PR #36, opened 2026-06-01
feat(cic): restore last focused window on cold load (#35)

### LINKED ISSUE #35: Restore last viewed channel on reload instead of server tab
On reload, cicchetto opens the server tab instead of the last channel the user was viewing before closing browser/app.

Expected: persist the last active channel and restore it on next session.

Reported by peluche on #sniffo.

— vjt-claude (acting on vjt account)

vjt:
## Summary

Closes #35. A registered identity's last focused channel/query/server window is now persisted to `localStorage` keyed by `user().id` and restored on cold start instead of always landing on `$home`.

- New module `cicchetto/src/lib/lastFocusedChannel.ts` — `load`/`save`/`clear`, scoped per identity, mirrors the existing `clientId.ts` / `theme.ts` storage discipline (single prefix, `try/catch` around every access, JSON-shape narrowing on read).
- `selection.ts` writes on every selection change for restorable kinds (`channel`/`query`/`server`).
- `Shell.tsx` cold-load arm tries restore before falling through to the existing `$home` default. Waits for `channelsBySlug()` to leave loading state so a fast reload doesn't miss the saved entry, then validates against `channelsBySlug` / `queryWindowsByNetwork` / `networkBySlug` and falls back to home if the saved window is no longer live (parted / closed / kicked).
- Visitor identities skip restore entirely (single-network ephemeral sessions get no useful payoff).
- `home` is the existing fallback target so it is not saved; `admin` is gated on `is_admin` and would redirect home on demote; `mentions` / `list` are ephemeral surfaces.

## Test plan

- [ ] Registered identity: open a channel, hard reload, lands back on that channel.
- [ ] Registered identity: open a query window, hard reload, lands back on the query.
- [ ] Registered identity: open the server window, hard reload, lands back on it.
- [ ] Part a channel that is currently persisted, hard reload, lands on `$home`.
- [ ] Visitor identity: hard reload always lands on `$home`.
- [ ] Multi-identity browser: each identity remembers its own last window.
- [ ] Safari Private Browsing: no crash, restore silently no-ops.

per chan request

### DIFF
--- a/cicchetto/src/Shell.tsx
+++ b/cicchetto/src/Shell.tsx
@@ -24,6 +24,7 @@ import { channelKey } from "./lib/channelKey";
 import { getDraft, setDraft, tabComplete } from "./lib/compose";
 import { loadUploadTtlSeconds } from "./lib/imageUploadOrchestrator";
 import { install, registerHandlers, uninstall } from "./lib/keybindings";
+import { loadLastFocused } from "./lib/lastFocusedChannel";
 import { mentionsBundleBySlug } from "./lib/mentionsWindow";
 import {
   openAdminPanel,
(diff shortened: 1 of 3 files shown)


vjt:
Superseded: the restore-last-focused-window feature (#35) already landed on main — `cicchetto/src/lib/lastFocusedChannel.ts`, `selection.ts`, `Shell.tsx` wiring and e2e `cp15-b6-archive-query-revival.spec.ts` are all present, and #35 is closed. This branch conflicts because main already carries the feature via another path. Closing as obsolete.

Issue #121, opened 2026-06-28
Visitor/session nick matching is case-sensitive → duplicate visitor accounts (same nick, different case)


vjt:
## Summary

grappa matches visitors/sessions by nick using a **case-sensitive** exact comparison. A reconnect with the same nick in a different case does not reattach to the existing session — it creates a second one. You end up with two visitor accounts/sessions for the same identity.

## Reproduction

1. Connect and identify via NickServ as `Mezmerize` → permanent (non-expiring) visitor session.
2. Disconnect the client WITHOUT closing the grappa-side session (e.g. the "X" in cicchetto's bottom bar): the client detaches but the grappa session stays alive (indefinitely).
3. Reconnect with the nick in a different case, `mezmerize`.
4. grappa does NOT recognise this as the same identity (case-sensitive compare) and creates a SECOND visitor session/account.
5. Result: two sessions with the same nick. A subsequent `/nick Mezmerize` returns "nickname already in use" because the orphaned session still holds the nick.

## Cause

The visitor/session nick lookups use a case-sensitive exact match:
- `Grappa.Visitors.get_by_nick_and_network/2` — `Repo.get_by(Visitor, nick: nick, ...)` (visitors.ex:539)
- `Grappa.Visitors.nick_in_use?/3` — same exact match (visitors.ex:552)
- `Grappa.Visitors.find_or_provision_anon/3` (visitors.ex:97), `lookup_visitor/2` (login.ex:203)

Meanwhile nicks ARE already compared case-insensitively in other places (`query_windows.ex:40`, `session.ex:914`) — but only **ASCII RFC 2812** folding (A–Z), and the visitor table lookup is the one place that doesn't fold at all. So the codebase is inconsistent and none of it handles the ircd's full casemapping.

## Fix (decided: unify ALL nick comparisons under rfc1459)

Azzurra runs bahamut (**rfc1459 casemapping**): besides A–Z, it folds `[ ] \ ~` → `{ } | ^`. Two nicks differing only by those are the SAME nick to the ircd.

1. Add a single **rfc1459 nick casemapper** (e.g. `Grappa.IRC.Identifier.canonical_nick/1`) — A–Z plus `[ ] \ ~` → `{ } | ^`. (Distinct from the existing `canonical_channel/1`, which is RFC 2812 ASCII for channels.)
[...]

Issue #114, opened 2026-06-27
docs: de-slop DESIGN_NOTES.md (moderate, careful verification)


vjt:
## Goal
`docs/DESIGN_NOTES.md` is ~13.4k lines / 720KB / 137 entries — the canonical decision log, loaded and grepped every session. The content is excellent (verified by a full audit), just heavily padded. De-slop to ~7-8k lines.

This is **polish, not a fix** — the doc works. The win is token cost + navigability, not correctness.

## Level: MODERATE (not aggressive)
- Tighten verbose prose ~40-50% (sample below).
- Drop stale ephemera: version/peer-warning notes, fully-superseded decisions restated at length, context now living in code/CLAUDE.
- Compress fully-superseded entries to a one-line "superseded by X".
- Merge obvious duplicate/overlapping entries.
- **KEEP every decision + its WHY + the rejected alternative + the invariant.** Those are the load-bearing bits.

Do **not** go aggressive (cutting entries to stubs + pointers). The risk is asymmetric — see below.

## Risk: asymmetric → "careful verification"
The danger is silently dropping a load-bearing *why* — the rejected alternative or non-obvious gotcha that stops someone re-introducing a bug. Live examples from the recent audit: the channel-directory `+s/+p` secret-leak rationale; "the `navigateFallbackDenylist` does NOT protect REST/WS". Cut those and nothing breaks today; it breaks months later when someone re-derives the bug. A dropped rationale is invisible in a 137-entry diff unless you read both versions.

### Verification requirement (the hard part)
For every entry trimmed/dropped, confirm its durable content is EITHER:
- already captured in a living doc (CLAUDE.md / OPERATIONS / TESTING) or a code moduledoc, OR
- genuinely dead (superseded, ephemeral).

If neither → keep it. Do this as a **reviewable pass**, not a blind rewrite. Coherence: the log is chronological + cross-referential ("see above", entries cite each other) — don't break internal refs or blur "what we knew when".

## Sample (target style — `2026-04-27 vite-plugin-pwa`, ~48 → ~22 lines)
Kept: the decision, the WHY, both non-obvious gotchas (denylist-doesn't-protect-REST, legacy-cache-intentional). Cut: prose padding + the stale `workbox-window@7.4.0` peer-warning note.

## Lower-risk alternative considered
[...]

PR #328, opened 2026-07-19
notify: /notify presence watch — server-side list, MONITOR/WATCH, re-arm on reconnect (#247)

### LINKED ISSUE #247: Implement /notify (presence watch list, persistent across reconnect)
## Summary

Implement a `/notify` presence-watch feature: a client command to track a list of nicks and get notified when they come online / go offline. The list is **server-side, per-account, per-network**, and survives both the client's own disconnect and any upstream reconnect.

## Why it's not a pure passthrough

`/notify` looks like a thin wrapper over the upstream IRC presence primitives, but grappa has to own real state because it's a persistent bouncer, not a transient connection:

- **Persistent notify-list per account/network** — the watch list must live in grappa and be re-armed automatically after an upstream reconnect (netsplit, server restart, leaf-spread rebalance per #271, the user's own client disconnecting and coming back). A raw client-side list would be lost on every reconnect, because `MONITOR`/`WATCH` registrations live on the *upstream connection* and die with it.
- **Upstream translation** — grappa maps the notify-list onto whatever the upstream network offers (see below), so the client speaks one command regardless of ircd.
- **Push presence events to cicchetto** — online/offline transitions get pushed to the connected client(s) over the existing event channel, so the UI can surface them regardless of which upstream mechanism produced them.

## Command UX (cicchetto client)

irssi-shaped, per-network (the active network context of the window the command is run in):

- `/notify` — list the watch list with each nick's current online/offline status.
- `/notify add <nick> [<nick> …]` — add one or more nicks to the list.
- `/notify del <nick> [<nick> …]` — remove one or more.
- `/notify list` — explicit alias of the bare form.
- `/notify clear` — wipe the list for the current network.

The command is thin sugar over a server-side list that the GUI also reads/writes: a **"Watched" panel** in cicchetto renders the same list with live status dots, and add/remove from the GUI hit the same server state. Command and GUI are two faces of one authoritative list.

## Server-side model (grappa)

- Storage keyed by **(account_id, network_id, nick)**. The same nick on two networks is two independent entries with independent state — presence on Azzurra ≠ presence on Libera.
- Nicks normalized per the network's `CASEMAPPING` (rfc1459 vs ascii); store canonical for matching + display form for the UI.
- The list is loaded into the session's presence-tracker on attach and **re-armed on every upstream (re)connect** as part of the post-registration replay phase (same lifecycle slot as channel-JOIN replay). This re-arm is the whole reason the feature is server-side.
- One authoritative list + state map per (account, network); all attached clients see the same list and the same events.

## Upstream translation

Decided per-network at connect from `ISUPPORT`/005:

- **MONITOR** where the token is advertised (Libera/solanum, OFTC). `MONITOR + nick,nick,…` respecting `MONITOR=<limit>`. Handle `RPL_MONONLINE (730)` / `RPL_MONOFFLINE (731)` / `RPL_MONLIST (732)` / `RPL_ENDOFMONLIST (733)` / `ERR_MONLISTFULL (734)`.
- **WATCH** where MONITOR is absent but WATCH exists (Azzurra / bahamut-inet6, legacy nets). `WATCH +nick`. Handle `RPL_LOGON (600)` / `RPL_LOGOFF (601)` / `RPL_NOWON (604)` / `RPL_NOWOFF (605)` / `RPL_WATCHOFF (602)` / `ERR_TOOMANYWATCH (512)`.
- **ISON polling** as fallback where neither exists. Batch nicks into `ISON` lines under the 512-byte line budget (incl. prefix), poll on a cadence (~15–30s, backing off as the list grows), and diff the returned present-set against the last-known set to synthesize online/offline transitions.

Both real networks are covered by push mechanisms (Azzurra→WATCH, Libera/OFTC→MONITOR); ISON is the graceful degradation for exotic upstreams.

## Presence-event push to cicchetto

- grappa keeps an authoritative online/offline map per (network, nick), fed by MONITOR replies, WATCH logon/logoff, or ISON diffs — normalized to one internal event.
- On a genuine transition, emit a presence event on the existing event channel to all attached clients: `{ type: "presence", network, nick, state: "online" | "offline", ts, source: "monitor" | "watch" | "ison" }`.
- **Not** a synthetic PRIVMSG — a first-class event so cicchetto can render a non-intrusive toast + update the watched-panel status dot.
- **Snapshot on attach:** when a client attaches, grappa sends the current watch list + each nick's known state, so the UI paints correct dots immediately without waiting for the next transition.

## Edge cases

- **Already-online on add** — MONITOR/WATCH reply immediately for already-present nicks (MONONLINE / NOWON), and the first ISON poll shows them. Treat the *initial* read as a **baseline snapshot** (paint the dot) — **not** a "came online" notification. Only transitions after the baseline fire toasts. This avoids a notification-storm when adding a large list at once.
- **List full** — respect `MONITOR=<limit>` / WATCH limit. Adding past the limit rejects with a clear error surfaced to the client (map `ERR_MONLISTFULL` / `ERR_TOOMANYWATCH`), never a silent drop. ISON has no hard server limit but obeys the line-length budget.
- **Duplicate add** — idempotent no-op.
- **Nick rename** — presence is nick-keyed, not account-keyed: a watched nick that changes nick reads as offline. Document as a known limitation; account-keyed tracking (extended-monitor / account-notify) is a future upgrade where the upstream supports it.
- **Casemapping** — `Foo` == `foo` per the network's CASEMAPPING; normalize on add and on match.
- **ISON cadence** — tune latency vs traffic; respect flood limits; pipeline multiple ISON lines for large lists.

## Scope

- **grappa (server):** notify-list storage (per account/network), re-arm on reconnect, upstream MONITOR/WATCH/ISON selection + polling loop, authoritative state map, presence-event push + snapshot-on-attach.
- **cicchetto (client):** `/notify` add/del/list/clear command, the "Watched" panel with live status dots, and a lightweight toast surface for online/offline transitions.

## Suggested v1

Command (add/del/list) + WATCH (Azzurra) + MONITOR (Libera/OFTC) + re-arm on reconnect + snapshot-on-attach + push to cicchetto. ISON fallback can be phase 2 if a network without MONITOR/WATCH ever matters.

---
Requested by @mannaggiaccristo on #grappa. Design expanded on user's behalf — design-first, not yet queued for build.

gmsrc:
Implements the `/notify` presence-watch design specified in #247, at the issue's own "Suggested v1" scope: command + Watched panel, MONITOR (solanum/Libera, OFTC) + WATCH (bahamut/Azzurra) upstream translation, re-arm on every (re)connect, snapshot-on-attach, and typed presence push to cicchetto. **ISON polling fallback is deferred to phase 2**, as the issue itself suggests — a network advertising neither mechanism logs honestly and keeps `unknown` dots.

Closes #247

## What's here (one commit per layer)

- **`isupport`** — capture `MONITOR=`/`WATCH=` 005 tokens; `presence_mechanism/1` picks MONITOR over WATCH, `:unlimited` on unparseable limits. No pre-005 mechanism seed — arming is advertisement-driven only.
- **`notify`** — durable per-(subject, network) watch list (`Grappa.Notify`, `notify_entries`): subject-XOR shape, rfc1459-folded partial unique expression indexes (#121), atomic batch add (`/notify add a b c` — one invalid nick rejects the whole batch), idempotent everything, `notify_list` full-snapshot broadcast on `Topic.user` (query_windows contract).
- **`session`** — pure `Grappa.Session.Presence` (chunked MONITOR/WATCH command building under the 512-byte budget + the authoritative presence map with baseline-vs-transition classification); arm at end-of-MOTD; EventRouter handling for 730/731/600/601/602/604/605 (delegated) and 734/512 (typed `presence_error`, raw numeric stays visible — never a silent drop); `presence_changed`/`presence_snapshot` wire events; `Session.notify_changed/4` live diff-sync.
- **`web`** — `GET/POST/DELETE /networks/:network_id/notify` (GET returns DB list AND live presence, `presence: null` when no session — two sources of truth, never fabricated); user-topic after-join pushes `notify_list` + per-network `presence_snapshot`. No nginx change needed (`/networks` prefix already allowlisted).
- **`cic`** — `/notify [list | add … | del … | clear]`, `notifyWatch` store, strict narrowers for the four new event kinds, Watched panel with status dots on the home-pane network cards, self-expiring transition toasts. Toasts fire only on genuine transitions (`initial: false`) — the post-arm baseline paints dots silently, so bulk adds never storm.

## Deviations from the issue text (and why)

- **Arm slot is 376/422 (end-of-MOTD), not the 001 JOIN-replay slot** the issue sketched: 005 arrives after 001, and the MONITOR-vs-WATCH pick needs the ISUPPORT tokens. A `presence_armed` latch keeps a mid-session `/motd` (which re-fires 376) from re-sending the burst. Reconnect re-arm needs no extra machinery — the crash-restart model replays 001→005→376 from scratch.
- **Storage is subject-XOR (`user_id` XOR `visitor_id`)**, not "account_id" — the repo's proven `query_windows` shape, visitor reaping CASCADEs included.
- **rfc1459 fold everywhere** rather than per-network CASEMAPPING: #121 made rfc1459 the server-wide fold invariant; a per-network fold would fork the identity rule. cic mirrors the fold for presence-map keys only (documented in `notifyWatch.ts` against the `normalizeNick` ascii tradeoff).
- **The Watched panel has no add-input.** The home pane is input-free by pinned design (`HomePane.test` "no compose / input affordance" KISS rule), so the panel is dots + remove; adding is `/notify add` from any compose box. If you'd rather have GUI-add somewhere, happy to move the panel or add an affordance wherever you prefer.

Durable rationale is in `docs/DESIGN_NOTES.md` (2026-07-18 entry). Deploy note: **COLD** — new migration + `Session.Server` state-shape change + cic bundle.

## Testing

- `scripts/check.sh` green (format, credo --strict, sobelow, audits, doctor, boundary, full ExUnit incl. new suites, wireTypes drift gate).
- New coverage: `Grappa.Notify` context (fold collisions, atomic batch, broadcasts, property tests), `Presence` (chunking properties incl. frame-budget, baseline/transition), ISupport tokens, EventRouter numerics (multi-target 730 hostmasks, gated 512), `Session.Server` end-to-end arm flows against the in-process fake IRC server (WATCH arm, MONITOR-wins, no-mechanism honesty, second-376 latch, live `notify_changed` sync, baseline-vs-transition broadcast), NotifyController request specs, cic vitest (`notifyWatch` store, `/notify` parser).
- cic `bun run check` (biome + tsc) green; full vitest green except a handful of pre-existing load-sensitive timeouts on my Windows host that pass in isolation and vary run-to-run — CI's Linux runners are the authoritative gate there.
- Full e2e integration suite: all 430 specs green locally, run as 6 fresh-stack Playwright shards + per-spec iso-reruns (my Windows host deterministically flakes the `issue211` multinet-reconnect spec when it runs deep in a long serial sequence, which then cascades — it and every cascade victim pass first-try on a fresh stack; the monolithic serial run is only viable on faster hardware, and CI's Linux run remains the canonical single-invocation gate).
- ~~e2e note: both testnet ircd families (bahamut → WATCH, solanum → MONITOR) are in place for a dedicated /notify integration spec; I didn't add one in this PR to keep scope contained — glad to follow up if you want e2e coverage before merge.~~ *Superseded — `issue247-notify-watch.spec.ts` added below.*

## Post-review hardening (2026-07-19)

A full codebase review of the repo (2026-07-19) flagged three MEDIUMs on this branch; four follow-up commits fix them before merge:

- **`notify: cap the watch list at 64 per (subject, network)`** — the list was unbounded at every layer (one POST could insert arbitrary rows; a list past the upstream MONITOR/WATCH limit re-earned 734/512 on every reconnect, its excess entries stuck at `unknown` forever). The cap is enforced on the post-state row count inside `add/4`'s transaction — counting after the inserts gets fold-dedup exactly right, so an idempotent re-add against a full list still succeeds — rolling back whole with 422 `list_full`. Static 64 rather than the ISUPPORT value: adds happen while parked/disconnected when no 005 exists, and 64 sits under every observed mechanism limit (solanum `MONITOR=100`, bahamut `WATCH=128`), so a within-cap arm burst can never be rejected upstream.
- **`cic: surface presence_error as an error toast`** — the typed `presence_error` event was routed only to `diagPush`, a ring buffer rendered solely behind the hidden `cic_diag` flag, i.e. invisible in production (while the comment claimed a toast). It now queues an error-styled toast through the shared presence-toast store; rides along: friendly copy for the new `list_full` token.
- **`cic: drop dead getNotify client + document sequential /notify del`** — `getNotify`/`NotifyIndexResponse` had zero call sites (the WS snapshots cover the panel's cold load); multi-nick `/notify del` stays sequential single-nick DELETEs with the trade-off written down at the loop.
- **`cic(e2e): issue247-notify-watch.spec.ts`** — the full loop against the testnet's WATCH mechanism: add-while-offline → 605 baseline dot with **no** toast; peer connect (600) → transition toast + flip; mid-session reload → dot repaints from `presence_snapshot` alone; peer quit (601) → offline toast; panel × removes via the server round-trip; REST cleanup in `finally`.

Post-review gates run locally: scoped ExUnit for the touched suites (29 tests), scoped vitest for the touched files (318 tests), `mix format`, `credo --strict`, `sobelow`, `dialyzer`, `bun run check` — all green. **The new e2e spec has NOT been executed locally** (e2e on my Windows host takes hours — bind-mount IO); please let the GH CI integration run be the gate for it, and shout if it flakes there.

### DIFF
--- a/cicchetto/e2e/tests/issue247-notify-watch.spec.ts
+++ b/cicchetto/e2e/tests/issue247-notify-watch.spec.ts
@@ -78,6 +78,10 @@ test("#247 — /notify add → dots + toasts + snapshot repaint + panel remove",
     //    dot must repaint ONLINE from presence_snapshot alone (no
     //    transition occurs during the reload window).
     await page.reload();
+    // Reload's last-focused restore lands on the seed CHANNEL (Home is
+    // not a persisted selection target — review 2026-07-19), so walk
+    // back to Home before asserting the panel.
+    await page.locator(".sidebar-channel-name").filter({ hasText: /^Home$/ }).click();
     await expect(page.locator(".home-pane-registered").first()).toBeVisible({ timeout: 15_000 });
     await expect(entryRow).toHaveCount(1, { timeout: 10_000 });
     await expect(dot).toHaveAttribute("data-state", "online", { timeout: 10_000 });
--- a/cicchetto/e2e/tests/issue247-notify-watch.spec.ts
+++ b/cicchetto/e2e/tests/issue247-notify-watch.spec.ts
@@ -1,0 +1,107 @@
+// #247 — /notify presence watch, end to end (review 2026-07-19 R3).
+//
+// The unit layers cover the store mechanics (notifyWatch vitest), the
+// slash-command parse, and the server context/controller contracts —
+// but none of them see the full loop: compose /notify add → REST →
+// Grappa.Notify → live WATCH + sync on the armed session → bahamut
+// 604/605 baseline + 600/601 transitions → typed wire events → Watched
+// panel dots + transition toasts. This spec drives that loop against
+// the real testnet (bahamut advertises WATCH=, so the WATCH mechanism
+// is the one exercised; MONITOR shares the same session plumbing and
+// is pinned at the unit layer in presence_test.exs).
+//
+// Assertions, in order:
+//   1. /notify add of an offline peer → Watched panel entry with an
+//      OFFLINE dot (605 RPL_NOWOFF baseline — dot painted, NO toast:
+//      the baseline-vs-transition rule).
+//   2. Peer connects → ONLINE transition toast + dot flips (600).
+//   3. Reload mid-session → dot repaints ONLINE from the
+//      presence_snapshot after-join push (snapshot-on-attach contract)
+//      with the peer still connected.
+//   4. Peer quits → OFFLINE transition toast + dot flips (601).
+//   5. Panel × removes the entry (server round-trip: DELETE → WATCH -
+//      sync → notify_list broadcast re-renders the panel empty).
+//
+// Cleanup is REST (DELETE /networks/:slug/notify) in finally so a
+// mid-spec failure can't strand the watch entry in vjt's durable list
+// for later specs.
+
+import { expect, test } from "../fixtures/test";
+import { composeSend, loginAs, selectChannel } from "../fixtures/cicchettoPage";
+import { GRAPPA_BASE_URL } from "../fixtures/grappaApi";
+import { IrcPeer } from "../fixtures/ircClient";
+import { AUTOJOIN_CHANNELS, getSeededVjt, NETWORK_NICK, NETWORK_SLUG } from "../fixtures/seedData";
+
+const PEER_NICK = "i247-watched";
+const SEED_CHANNEL = AUTOJOIN_CHANNELS[0];
+
+// Body is ~6 condition-waits (each fast in isolation) + one reload +
+// one peer register; 90s is the standard full-suite-load budget used
+// by the sibling session-lifecycle specs.
+test.setTimeout(90_000);
+
+test("#247 — /notify add → dots + toasts + snapshot repaint + panel remove", async ({ page }) => {
+  const vjt = getSeededVjt();
+  let peer: IrcPeer | null = null;
+
+  const panel = page.getByTestId(`watched-panel-${NETWORK_SLUG}`);
+  const entryRow = panel.locator(".watched-panel-item", { hasText: PEER_NICK });
+  const dot = entryRow.locator(".watched-panel-dot");
+
+  try {
+    await loginAs(page, vjt);
+
+    // Compose lives on channel windows (the home pane is input-free by
+    // design) — issue the add from the seed channel, then walk back to
+    // Home where the Watched panel renders.
+    await selectChannel(page, NETWORK_SLUG, SEED_CHANNEL, { ownNick: NETWORK_NICK });
+    await composeSend(page, `/notify add ${PEER_NICK}`);
+
+    await page.locator(".sidebar-channel-name").filter({ hasText: /^Home$/ }).click();
+
+    // 1. Entry present; dot settles OFFLINE via the 605 baseline (the
+    //    peer isn't connected yet). Baseline must NOT toast.
+    await expect(entryRow).toHaveCount(1, { timeout: 10_000 });
+    await expect(dot).toHaveAttribute("data-state", "offline", { timeout: 10_000 });
+    await expect(page.locator(".presence-toast")).toHaveCount(0);
+
+    // 2. Peer connects → 600 RPL_LOGON → genuine transition: toast +
+    //    dot flip. Arm the toast wait BEFORE connecting so the 6s
+    //    self-expiry can't outrace the first poll.
+    const onlineToast = page.locator(".presence-toast-online", { hasText: PEER_NICK });
+    const onlineToastSeen = expect(onlineToast).toBeVisible({ timeout: 15_000 });
+    peer = await IrcPeer.connect({ nick: PEER_NICK });
+    await onlineToastSeen;
+    await expect(dot).toHaveAttribute("data-state", "online", { timeout: 10_000 });
+
+    // 3. Snapshot-on-attach: reload with the peer still online — the
+    //    dot must repaint ONLINE from presence_snapshot alone (no
+    //    transition occurs during the reload window).
+    await page.reload();
+    await expect(page.locator(".home-pane-registered").first()).toBeVisible({ timeout: 15_000 });
+    await expect(entryRow).toHaveCount(1, { timeout: 10_000 });
+    await expect(dot).toHaveAttribute("data-state", "online", { timeout: 10_000 });
+
+    // 4. Peer quits → 601 RPL_LOGOFF → offline transition toast + flip.
+    const offlineToast = page.locator(".presence-toast-offline", { hasText: PEER_NICK });
+    const offlineToastSeen = expect(offlineToast).toBeVisible({ timeout: 15_000 });
+    await peer.disconnect("gone (#247 e2e)");
+    peer = null;
+    await offlineToastSeen;
+    await expect(dot).toHaveAttribute("data-state", "offline", { timeout: 10_000 });
+
+    // 5. Remove from the panel — server round-trip, panel re-renders
+    //    from the notify_list broadcast (cic never edits its own store).
+    await entryRow.getByRole("button", { name: `Stop watching ${PEER_NICK}` }).click();
+    await expect(entryRow).toHaveCount(0, { timeout: 10_000 });
+  } finally {
+    if (peer !== null) {
+      await peer.disconnect("cleanup (#247 e2e)").catch(() => {});
+    }
+    // Durable-list cleanup: idempotent whether step 5 ran or not.
+    await fetch(`${GRAPPA_BASE_URL}/networks/${NETWORK_SLUG}/notify`, {
+      method: "DELETE",
+      headers: { authorization: `Bearer ${vjt.token}` },
+    }).catch(() => {});
+  }
+});
--- a/cicchetto/src/HomePane.tsx
+++ b/cicchetto/src/HomePane.tsx
@@ -316,6 +316,10 @@ const DisconnectedRow: Component<{ row: HomeRow }> = (props) => {
           </Show>
         </div>
         <FeaturedLinks slug={props.row.slug} />
+        {/* #247 (review nit) — the watch list is durable server state,
+            so it stays visible while the network is parked/failed; dots
+            render ◌ unknown (no live session to report presence). */}
+        <WatchedPanel slug={props.row.slug} />
       </div>
     </li>
   );
--- a/cicchetto/src/HomePane.tsx
+++ b/cicchetto/src/HomePane.tsx
@@ -16,6 +16,7 @@ import { setSelectedChannel } from "./lib/selection";
 import { LIST_WINDOW_NAME, SERVER_WINDOW_NAME } from "./lib/windowKinds";
 import { windowStateByChannel } from "./lib/windowState";
 import NickText from "./NickText";
+import WatchedPanel from "./WatchedPanel";
 
 // #85 — operator-curated featured channels for a network, fetched on
 // home DISPLAY (component mount / slug change) so an operator config
@@ -247,6 +248,7 @@ const ConnectedRow: Component<{ row: HomeRow }> = (props) => {
         📇 Browse channels
       </button>
       <FeaturedLinks slug={props.row.slug} />
+      <WatchedPanel slug={props.row.slug} />
     </li>
   );
 };
--- a/cicchetto/src/PresenceToasts.tsx
+++ b/cicchetto/src/PresenceToasts.tsx
@@ -4,28 +4,49 @@ import NickText from "./NickText";
 
 // #247 — non-intrusive toasts for GENUINE /notify presence transitions
 // (`presence_changed` with `initial: false`; the post-arm baseline
-// never toasts — see notifyWatch.ts). Self-expiring, click-to-dismiss,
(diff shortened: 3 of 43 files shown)


gmsrc:
@vjt here's the documentation on the sharding implemented to work around the excessive time being observed on a Windows machine: https://gist.github.com/gmsrc/fbcbdb714508c53cca493aca5cc43712

gmsrc:
Further context about issues with CRLF on Windows: https://gist.github.com/gmsrc/ffeb5fcdd06d1ff00a52d17c76a7e932

vjt (commented):
## Review — server-side (Elixir) + cic

Both slices: **LGTM with nits, no code blockers.** Two things gate the merge, plus a deployment question and a handful of nits — collected here so the fix can go in one pass.

### Blockers before merge

**1. `integration` CI is red — but it's a test bug, not a feature bug.**
`test/e2e/issue247-notify-watch.spec.ts:81` asserts `.home-pane-registered` is visible after `page.reload()`. On reload, the last-focused restore lands on `#bofh` (selected earlier at line 57), so `HomePane` never mounts and the selector times out at 15s. The feature works; the spec just asserts against the wrong pane. Minimal fix: re-select Home after the reload, before the assertion. (The PR body already notes the spec was "not executed locally", consistent with this having never passed.)

**2. Rebase needed — PR is CONFLICTING.**
Same cause as #338: this branch and recently-landed work both append to the tail of `docs/DESIGN_NOTES.md`, and git won't auto-merge adjacent additions. A rebase on current `main` resolves it. (Aside: `docs/DESIGN_NOTES.md merge=union` in `.gitattributes` would make this whole class of conflict auto-resolve — tracked in #114.)

### Behavior mismatch to reconcile — 005 gating vs. intended 005-independence

The design intent is for `/notify` to work **regardless** of whether the upstream lists `WATCH=`/`MONITOR=` in RPL_ISUPPORT (005). The code as written does the opposite: the arm fires **only** on explicit advertisement (`isupport.ex:258-261`, to avoid per-reconnect `ERR_UNKNOWNCOMMAND`). Many bahamut forks support the WATCH *command* but omit the token from 005 — on such an ircd the feature silently no-ops (logs "presence arm skipped"), which is exactly the case the intent says should still work.

Reconcile one way or the other:
- **If it must work independent of 005** (the stated intent): drop the hard 005 gate and attempt `WATCH`/`MONITOR` unconditionally, handling `ERR_UNKNOWNCOMMAND` gracefully (or probe once and cache the capability), rather than refusing to arm when the token is absent.
- **If the 005 advertisement is a hard requirement**: document that, and confirm the target ircd actually lists `WATCH=`/`MONITOR=` in its 005 before relying on this in prod.

### Nits (non-blocking)

- **cic (LOW):** `WatchedPanel` isn't rendered inside `DisconnectedRow`, so the watch list is invisible while a network is parked/disconnected.
- `arm_presence/1` does a synchronous `Grappa.Notify.list/2` DB read inside the session GenServer at 376 (`server.ex:4509`); a Repo exception there crashes the session — mild tension with the "degrade, never crash" ethos, though consistent with the existing scrollback-persist calls already in the hot path.
- `NotifyController.create` passes the raw request batch as `added` (`notify_controller.ex:78`), so idempotent re-adds re-emit `MONITOR +`/`WATCH +` for already-armed nicks. Harmless (upstream no-op, `put_new` preserves state) but noisy.
- TOCTOU in `clear`: `removed` is snapshotted before the wipe (`notify_controller.ex:113-116`); a concurrent add in that window leaves one stale upstream registration until the next reconnect re-arm. Self-healing.
- `Notify.add/4` broadcasts `notify_list` even on a pure idempotent no-op — minor fan-out waste.
- No end-to-end test for re-arm on *reconnect* (drop → new 376 → burst re-sent). The mechanism is sound (reconnect = full session process restart → fresh `init` resets the latch), but it's the one PR-title claim without a dedicated test.
- **Fold-SQL triplication** (migration `fold/1`, `notify.ex @nick_fold_sql`, `Identifier.nick_fold/1`). The rfc1459 fold string is duplicated across three sites and must stay byte-identical — if it ever drifts, SQLite silently stops using the partial unique index and duplicate watch rows creep in. Verified identical today, but worth a test that pins all three equal (or a single shared source constant) so a future edit can't silently break dedup.

### Verified correct

Cap atomicity (`Notify.add` counts inside the transaction, ≤64; idempotent re-add at a full list stays at 64 → `:ok`), arm latch + re-arm, baseline-vs-transition (no notification storm on arm), chunk budget (400 < 512 with prefixes uncounted), numerics 730/731/600/601/604/605/734/512 (incl. 734/512 reaching EventRouter via the non-delegated routing branch), migration (XOR CHECK on user/visitor, CASCADE, reversible `down`).

Security specifically checked and cleared: **multi-tenant isolation** — every read/mutation scoped by subject + network, subject from the authn plug and network from `ResolveNetwork` (foreign/unknown → 404), `remove`/`clear` `delete_all` both scoped → no IDOR. **CRLF injection** on `DELETE /notify/:nick` — the raw nick reaches `Client.send_raw` directly, but that path screens the whole line via `Identifier.safe_line_token?` (`client.ex:443-445`), so an embedded CR/LF is dropped at the byte boundary, not smuggled (and a CRLF nick could never have been added — `Entry.changeset` requires a valid nick). DB deletes are parameterized fold queries.

gmsrc:
Review addressed, rebased on current main (`91408e0b`), all gates green.

**Blockers**
1. e2e spec reload assertion fixed as suggested — re-select Home after `page.reload()` before asserting the panel. The spec now runs green locally too: full monolithic suite **439 passed in 14.6m** (WSL2 clone; the PR-body caveat about local e2e is obsolete — turned out to be Docker Desktop bind-mount IO, ~157× slower than VM-native on small files; happy to share the write-up).
2. Rebased; DESIGN_NOTES conflict resolved keep-both. +1 on `merge=union` (#114).

**005 gating — reconciled toward 005-independence.** The advertisement is now a hint, not a gate: advertised pick wins (MONITOR over WATCH); with no token the arm probes `WATCH` optimistically and a 421 ERR_UNKNOWNCOMMAND downgrades it (WATCH → MONITOR → `:none`), cached per connection on `state.presence_mechanism`. Cost on a mechanism-less ircd: ≤2 probe lines per connect. The 512 list-full gate honours the resolved mechanism, so a probed-WATCH bahamut fork still surfaces `ERR_TOOMANYWATCH`. Covered by new Server tests (probe, fallback burst, double-421 → `:none`, no third attempt) and EventRouter tests (gated 421; unrelated 421s untouched).

**Nits taken**
- `WatchedPanel` now renders in `DisconnectedRow` too (durable list visible while parked; dots read unknown).
- `NotifyController.create` diffs the batch against the pre-add fold set — idempotent re-adds no longer re-emit `MONITOR +`/`WATCH +`.
- Fold-SQL triplication collapsed: `Identifier.nick_fold_sql/1` is the single source (consumed by Notify's conflict_target) and a drift-pin test asserts every folded-index migration embeds the canonical string byte-identically.

**Nits acknowledged, not changed:** clear-TOCTOU (self-healing on reconnect, as you noted), idempotent-no-op broadcast (fan-out cost negligible vs added bookkeeping), arm-time sync DB read (consistent with existing hot-path persist calls), reconnect re-arm e2e (mechanism covered by unit tests + fresh-process model; can add a spec leg as follow-up if you want it gated).

Issue #347, opened 2026-07-19
grappa bouncer: gate autojoin on upstream +r (identified) so +R channels join and ChanServ ops


vjt:
## Problem

For a per-user network credential using `auth_method: :nickserv_identify`, grappa's autojoin loop fires the `JOIN` commands **on `001 RPL_WELCOME`** (hardened in Phase 5 to also `376 RPL_ENDOFMOTD` / `422 ERR_NOMOTD` + a watchdog — `session/server.ex:45`, `:339`). But NickServ IDENTIFY is sent on 001 too (`irc/auth_fsm.ex:56-57`, `:559`) and processed **asynchronously** by the upstream: identification (self umode `+r`) lands *after* 001 / end-of-MOTD.

Consequence — the autojoin JOINs go out while the user is still unidentified, so on networks like Libera:
- `+R` (registered-only) channels reject the JOIN (`477 ERR_NEEDREGGEDNICK` — "you need to be identified to join that channel"), and
- ChanServ won't grant `+o` (it only ops identified users).

SASL (`auth_method: :sasl`) is unaffected — identification completes *before* 001, so autojoin on 001 already sees `+r`. This is specific to the `:nickserv_identify` path.

## Decision (vjt, 2026-07-20) — identify-first + short bounded wait

Chosen approach — the simplest, with **no NickServ NOTICE-text parsing**: at connect, send the **NickServ IDENTIFY first**, then **wait ~0.5s for it to propagate**, then fire the autojoin JOIN loop.

Concretely, for `:nickserv_identify` credentials, defer the autojoin JOINs until the **earlier** of:
- the self-`MODE +r` echo is observed — a standard MODE change the session already folds (`server.ex:536`, `221 RPL_UMODEIS`); this is the `+r` *umode*, **not** NickServ text-scraping, so it satisfies the "don't parse NickServ" constraint while being tighter than a blind timer; **or**
- a **~0.5s** timeout after IDENTIFY is sent (bounded also by the existing end-of-MOTD watchdog) — so a silent/slow NickServ never hangs autojoin: on timeout, JOIN best-effort (non-`+R` channels still work).

This also fixes the ChanServ `+o` miss (ops only granted post-identify).

[...]

PR #387, opened 2026-07-23
infra: native Linux (systemd) deploy path, with hot/cold Preflight support

Essency:
## Summary

Adds a third Docker-free production substrate alongside the existing
FreeBSD bastille jail: plain systemd + `mix release` on any modern
Linux host (tested on Debian 13). No container runtime involved —
`infra/linux/install.sh` provisions a `grappa` system user + the
asdf-managed Erlang/Elixir/bun toolchain, builds the release, and
installs a systemd unit + nginx reverse proxy; `infra/linux/deploy.sh`
does the day-2 update cycle.

Along the way this also extends `Grappa.Deploy.Preflight` with a
`:linux` substrate + a new COLD class for a changed systemd unit file,
so `deploy.sh` picks hot (rebuild + `POST /admin/reload`, sessions
preserved) vs cold (full restart) automatically instead of always
doing a full stop/start cycle — same mechanism `scripts/deploy.sh`
(Docker) and `infra/freebsd/deploy.sh` (jail) already use.

Two small unrelated bugs found and fixed along the way (both
pre-existing, not Linux-specific — see individual commits for the
live-repro details):

- `Mix.Tasks.Grappa.OptionParsing.parse_auth/1` could crash with "not
  an already existing atom" outside a full release boot.
- `GrappaWeb.Endpoint` was hardcoded into the supervision tree even
  for the one-shot `grappa.*` operator mix tasks, forcing
  `systemctl stop grappa` (or the Docker/jail equivalent) before every
  admin task even though those tasks never serve HTTP.

Plus a doc-drift fix: `AdminController`'s moduledoc still described
`/admin/reload` via `:code.modified_modules/0`, which `Grappa.HotReload`
moved off back on 2026-06-10.

## Commits (6, each self-contained — see individual messages for full rationale)

1. `infra: add native Linux (systemd) deploy path`
2. `fix(mix): OptionParsing.parse_auth crashes under bare mix task invocation`
3. `feat(deploy): extend Preflight classifier + deploy.sh with :linux hot/cold support`
4. `fix(app): make GrappaWeb.Endpoint opt-out, skip it for one-shot mix tasks`
5. `docs(admin): fix AdminController moduledoc drift on the reload mechanism`
6. `docs: drop stale LINUX.md pointers, fix cold-only description drift`

## Test plan

- [x] Full install + deploy cycle exercised live on a real Debian 13
      host (install.sh, deploy.sh hot path, deploy.sh cold path,
      systemd unit, nginx reverse proxy).
- [x] Hot-deploy path verified: `mix release --overwrite` + `POST
      /admin/reload` with the systemd service never stopped, same BEAM
      PID before/after, `/healthz` 200 throughout.
- [x] `GrappaWeb.Endpoint` opt-out verified live: ran a `grappa.*`
      admin mix task (theme seeding) against a running release without
      stopping the service first — no port conflict, same PID
      before/after.
- [ ] `mix test` / `mix credo` / `mix dialyzer` — not run in this
      environment (no local toolchain here); please run CI on this PR.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -578,8 +578,14 @@ is due. Don't just look at todo.md.
   change touches more than ~10 files unexpectedly, stop and confirm
   before continuing.
 - **Don't iterate through 10 wrong approaches.** Stop, think, ask.
-- **Don't propose split environments.** One Docker Compose stack. No
-  systemd, no bare `mix run`, no `infra/dev` vs `infra/prod`.
+- **Don't propose split dev environments.** Local dev is one Docker
+  Compose stack — no `infra/dev` vs `infra/prod` for development.
+  Production is substrate-plural by design: the FreeBSD bastille jail
+  (`infra/freebsd/`, prod since the m42 deploy) and the native Linux
+  systemd host (`infra/linux/`) are both supported, Docker-free
+  production paths — systemd and `mix release` for prod are not
+  off-limits, they're already how this ships. Don't propose a *third*
+  production substrate without discussing it first.
 - **NEVER run raw `docker compose`** — use `scripts/*.sh`. Always.
 - **NEVER `mix` on the host** — the container is the runtime.
 - **NEVER install hex packages on the host.** Add them to `mix.exs`,
--- a/docs/OPERATIONS.md
+++ b/docs/OPERATIONS.md
@@ -439,13 +439,11 @@ infra/linux/release.sh   # bin/grappa wrapper: version / eval / remote
 ```
 
 No SSH-to-a-separate-host indirection — run these directly on the
-target box as root. **Cold-only in v1 scope** (per `LINUX.md`'s
-original decision): `deploy.sh` always does a full stop → rebuild →
-migrate → start cycle, no hot/cold `Preflight` classification like the
-Docker/jail substrates. If that gap starts to hurt (frequent deploys,
-session-drop-sensitive usage), port `Grappa.Deploy.Preflight` here
-following the same pattern — not done yet because a first install
-should stay simple.
+target box as root. `deploy.sh` is preflight-driven, same as the
+Docker/jail substrates: `Grappa.Deploy.Preflight` classifies each
+deploy's diff and picks hot (rebuild + `POST /admin/reload`, sessions
+preserved) or cold (full stop → rebuild → migrate → start cycle)
+automatically — see `infra/linux/README.md` "Day-2 operations".
 
 Stop/start synchronization (the FreeBSD jail's `jail_beam_wait.sh`
 problem, defect #9) is solved differently here: the systemd unit runs
--- a/docs/OPERATIONS.md
+++ b/docs/OPERATIONS.md
@@ -425,6 +425,35 @@ at restart time — `--full-restart` does NOT touch it.** Never rehearsed
 against prod (it bounces the live jail + drops every session); bats-proven
 only — first real run is operator-driven.
 
+### Linux (systemd) — no host wrapper needed
+
+A third production substrate for a plain Ubuntu/Debian host, no
+Docker, no jail. Full setup + day-2 ops runbook lives in
+`infra/linux/README.md`; summary here for parity with the m42 section
+above:
+
+```
+infra/linux/install.sh   # first install, idempotent (see README.md for env)
+infra/linux/deploy.sh    # update: pull, rebuild, migrate, restart, healthcheck
+infra/linux/release.sh   # bin/grappa wrapper: version / eval / remote
+```
+
+No SSH-to-a-separate-host indirection — run these directly on the
+target box as root. **Cold-only in v1 scope** (per `LINUX.md`'s
+original decision): `deploy.sh` always does a full stop → rebuild →
+migrate → start cycle, no hot/cold `Preflight` classification like the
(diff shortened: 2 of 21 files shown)


vjt:
### Code review

Reviewed against current `main`. Good PR overall — the `Preflight` `:linux`/`:systemd_unit` classifier follows the existing `:jail`/`:docker` pattern cleanly, the `@auth_map` atom fix preserves all 5 modes, the `:start_endpoint` opt-out defaults to `true` and only flips for one-shot mix tasks, and the `Type=exec` + SIGTERM to clean `QUIT` claim on the unit holds. Two issues worth fixing before merge, plus a rebase heads-up.

**1. The secrets env file ends up world-readable (security)**

`install.sh` creates `/etc/grappa/grappa.env` as `0640 root:grappa`, but every `set_env_if_blank` / `force_set_env` write does `grep -v ... > "${ENV_FILE}.tmp" && mv "${ENV_FILE}.tmp" "${ENV_FILE}"`. The `.tmp` is a fresh inode created under root's umask (typically `0644`, `root:root`), and `mv` replaces the original with it — so the mode/group set moments earlier is lost. `force_set_env` runs unconditionally (`DATABASE_PATH`, `UPLOADS_STORAGE_ROOT`, `PHX_HOST`, `PORT`) *after* the secrets are written, so on the very first install the final file — holding `SECRET_KEY_BASE`, `SECRET_SIGNING_SALT`, `GRAPPA_ENCRYPTION_KEY` (at-rest key for network passwords), `RELEASE_COOKIE`, `VAPID_PRIVATE_KEY` — is left `0644 root:root`, readable by any local unprivileged user. Fix: re-`chmod 0640` + `chown root:grappa` after the last write, or wrap the writers in `umask 027`, or write through `install -m 0640`.

https://github.com/vjt/grappa-irc/blob/9888405cf723fc6bc6fbee4c93056d6f19371d36/infra/linux/install.sh#L96-L123

**2. `deploy.sh` hot-reload comment describes the wrong mechanism**

The new comment says the hot path tells the BEAM to walk `:code.modified_modules/0` and reload via `:code.load_file/1`. `Grappa.HotReload` actually walks the app's `ebin` by md5 and loads via `:code.load_abs/1` — its moduledoc explicitly says "never `:code.load_file/1`" and that `:code.modified_modules/0` was abandoned (blind to modules added post-boot). This is the same drift commit `533cc47d` corrects in `admin_controller.ex`; it looks copy-pasted from `infra/freebsd/deploy.sh` into the new file.

https://github.com/vjt/grappa-irc/blob/9888405cf723fc6bc6fbee4c93056d6f19371d36/infra/linux/deploy.sh#L150-L156

**Rebase heads-up (not a blocker):** the branch is ~2 commits behind `main` — the #347 `+r`-gated autojoin fix landed on `main` while this was being authored. The PR's own commits touch zero lines in `session/server.ex`, so GitHub's merge diff is clean and merging will not revert #347; but please rebase onto current `main` so the diff reflects only this work.

**Consistency note:** the `CLAUDE.md` edit blesses the systemd / `mix release` substrate, but the adjacent "NEVER `mix` on the host" and "NEVER install hex packages on the host" bullets are left intact — and this substrate necessarily does both. Worth reconciling those bullets so the policy stays self-consistent (author's/maintainer's call).

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

vjt:
Closing this because the work is **already on `main`** — it landed by cherry-pick on 2026-07-23, which is why GitHub reports the branch as conflicting (`add/add` on every `infra/linux/*` file): git has no record linking the two sets of commits, so it sees each file as added twice.

Every commit here has an exact counterpart on `main`. Verified with `git patch-id --stable` — the patches are byte-for-byte identical, not merely similarly titled:

| this PR | on `main` | subject |
| --- | --- | --- |
| `8a2ba2d4` | `a94059d5` | infra: add native Linux (systemd) deploy path |
| `10611d98` | `72bd2bbb` | fix(mix): OptionParsing.parse_auth crashes under bare mix task invocation |
| `f44ff814` | `c64e56ca` | feat(deploy): extend Preflight classifier + deploy.sh with :linux hot/cold support |
| `158f48f1` | `f939f2c2` | fix(app): make GrappaWeb.Endpoint opt-out, skip it for one-shot mix tasks |
| `533cc47d` | `431b5477` | docs(admin): fix AdminController moduledoc drift on the reload mechanism |
| `9888405c` | `294c894c` | docs: drop stale LINUX.md pointers, fix cold-only description drift |

`main` has also moved on top of them since, so merging this branch now would be a regression rather than a no-op — it would resurrect the pre-fix versions of these files. The follow-ups that would be undone include:

- `bdf7ba1b` — fix(linux): install.sh silently blanks both VAPID keys (#390)
- `8abeb8fe` — Fix native Linux installer bootstrap (#433, #434 — merged as #436)
- `cf3130ca` — Seed built-in themes during Linux install (#435 — merged as #437)
- `7a9c7cae` — packaging(#446/#419): 0.6.0 bump + libssl3 t64 dep
- `78afe487` — infra(linux): fix deploy.sh hot-reload comment to the real mechanism

So: closed as landed, not rejected. Thanks @Essency — the native systemd deploy path is in and has been built on since.

Issue #435, opened 2026-07-26
install.sh never seeds built-in themes — fresh install lands with an empty theme gallery


vjt:
## Summary

A fresh native-Linux install ends up with an **empty theme gallery**: `infra/linux/install.sh` never runs `mix grappa.seed_themes`, so none of the curated built-ins (solarized / gruvbox / nord / dracula / one-dark / …) exist in the DB until an operator discovers the task and runs it by hand.

Observed live on the first third-party self-host (2026-07-26, `irc.different-world.org`): install completed cleanly, gallery empty. Per chan request the requirement is explicit — theme seeding should be **part of setup, not a separate task to launch**.

## Evidence (verified on `origin/main`)

- The task exists: `lib/mix/tasks/grappa.seed_themes.ex`, idempotent upsert of the system-owned published built-ins (`lib/grappa/themes.ex:348`, `Grappa.Themes.Builtins`).
- It is referenced **only** by the e2e seeder (`cicchetto/e2e/compose.yaml:239`), `README.md:181` and `docs/DESIGN_NOTES.md` — `git grep seed_themes origin/main -- infra scripts` returns **nothing**. Neither `infra/linux/install.sh` nor the FreeBSD/packaging scripts call it.
- So the documented deploy ritual ("`mix grappa.seed_themes` must run on the target at ship") is operator tribal knowledge, not automation.

## Proposed fix

Run it in `install.sh` next to the first migration — step `6/10 first migration` runs plain `mix ecto.migrate` (not `release.sh eval`, for the documented `start_clean` boot reason); `mix grappa.seed_themes` belongs immediately after, in the same `MIX_ENV` context, **before** step `10/10 systemctl start grappa`.

Two properties make this safe:

- The task is idempotent (upsert; pinned by `test/mix/tasks/grappa/seed_themes_test.exs`), so it is harmless on re-runs and on upgrade paths.
- Since 2026-07-23 `Mix.Tasks.Grappa.Boot` suppresses `GrappaWeb.Endpoint`, so the task no longer binds port 4000 and no longer needs a `systemctl stop grappa` / crashes `:eaddrinuse`. Running it at install time (service not yet started) is doubly safe.

## Design questions for whoever implements

1. **Install-only or also upgrade?** New built-ins added later (the gallery grows) only reach existing installs if `deploy.sh` / the packaging upgrade path also seeds. Recommend seeding on both, given idempotency.
[...]

Issue #434, opened 2026-07-26
infra/linux: all *.sh committed 100644 but invoked by direct exec (fresh clone can't run install.sh)


vjt:
## Symptom

Every script under `infra/linux/` is committed non-executable, but they are
invoked by direct path, so a fresh clone cannot run the documented install
command.

Reported from a live install run (per chan report on #it-opers), same run as #433.

## Evidence (`git ls-tree origin/main`)

```
100644 infra/linux/cic_build.sh
100644 infra/linux/deploy.sh
100644 infra/linux/grappa_beam_wait.sh
100644 infra/linux/install.sh
100644 infra/linux/install_nginx.sh
100644 infra/linux/install_prereqs.sh
100644 infra/linux/install_systemd.sh
100644 infra/linux/install_toolchain.sh
100644 infra/linux/release.sh
```

against the sibling trees, which are correct:

```
100755 infra/freebsd/deploy.sh          (and every other jail_*.sh)
100755 infra/packaging/scripts/postinstall.sh   (and pre/post remove)
```

## Why the mode matters here

These are not `bash foo.sh` call sites — they are execs, which need the bit:

- `infra/linux/install.sh:54,66,257,260,263` — `"${SCRIPT_DIR}/install_prereqs.sh"`,
  `install_toolchain.sh`, `cic_build.sh`, `install_systemd.sh`, `install_nginx.sh`
- `infra/linux/deploy.sh:58` — `exec "${REPO_ROOT}/infra/linux/deploy.sh" "$@"`
  (self re-exec when deploy.sh itself changed)
- `infra/linux/deploy.sh:179,193` — `cic_build.sh`, `install_systemd.sh`
- `infra/linux/systemd/grappa.service:52` — `ExecStartPre=@REPO_ROOT@/infra/linux/grappa_beam_wait.sh …`
- `infra/linux/README.md:24` — documents `PHX_HOST=irc.example.org infra/linux/install.sh`

## The tell that this is an oversight, not a choice

`infra/linux/install_systemd.sh:23` already works around the symptom for one
file at runtime:

```sh
chmod 0755 "${REPO_ROOT}/infra/linux/grappa_beam_wait.sh"
```

That line exists because `ExecStartPre` would otherwise fail with `Permission
denied`. Fixing the mode in the repo makes the workaround redundant.

## Fix

```sh
git update-index --chmod=+x infra/linux/*.sh
```

[...]

Issue #433, opened 2026-07-26
infra/linux: install.sh creates REPO_ROOT parent as root, clone as grappa then fails (permission denied)


vjt:
## Symptom

On a fresh Linux install run, `infra/linux/install.sh` step 2 fails: the parent
directory of `REPO_ROOT` is created by root, then the clone runs as the
unprivileged `grappa` user and cannot write into it.

Reported from a live run (per chan report on #it-opers).

## Mechanism (verified against `origin/main`)

`infra/linux/install.sh:57-59`

```sh
if [ ! -d "${REPO_ROOT}/.git" ]; then
	mkdir -p "$(dirname "${REPO_ROOT}")"
	run_as_grappa "git clone '${GIT_REMOTE_URL}' '${REPO_ROOT}'"
```

`install.sh` runs as root, so `mkdir -p` creates the parent as `root:root 0755`
(root umask 022). `run_as_grappa` is `sudo -u grappa -H bash -c` — the clone
therefore needs write permission on that parent and does not have it.

The `chown -R "${GRAPPA_USER}:${GRAPPA_USER}" "${REPO_ROOT}"` on line 63 comes
*after* the clone and only covers `REPO_ROOT` itself, never its parent, so it
cannot rescue this.

`-H` also points `HOME` at `/home/grappa`; when that directory is root-owned,
git config/credential lookups write-fail there too.

## When the parent is missing

`infra/linux/install_prereqs.sh:53-58` only calls `useradd --create-home` when
the user does **not** already exist:

```sh
if ! id "${GRAPPA_USER}" >/dev/null 2>&1; then
	useradd --system --create-home --home-dir "/home/${GRAPPA_USER}" ...
else
	echo "[install_prereqs] user ${GRAPPA_USER} already exists"
fi
```

So the failure triggers whenever:

1. the `grappa` user pre-exists without a home directory (created by hand, by a
   package, or with its home since removed) — prereqs skips `useradd`, nothing
   creates `/home/grappa`; or
2. `REPO_ROOT` is overridden to a path whose parent does not exist
   (`REPO_ROOT=/srv/grappa/app`) — documented as supported at `install.sh:14`.

In the plain happy path (`grappa` absent, default `REPO_ROOT`) `useradd
--create-home` makes `/home/grappa` owned by `grappa` and the `mkdir -p` is a
no-op, which is why the default install passes.

## Suggested fix

Create the parent owned by the target user instead of by root, and generalise
[...]

PR #436, opened 2026-07-26
Fix native Linux installer bootstrap

abonforti:
## Summary

- mark native Linux operator scripts executable
- create a fresh checkout directory with the Grappa runtime user ownership before cloning

## Problem

A fresh Debian install failed twice:

```text
infra/linux/install.sh: line 54: .../install_prereqs.sh: Permission denied
fatal: could not create work tree dir '/home/grappa/grappa': Permission denied
```

The scripts were committed as `100644`, although the orchestrator executes them directly. The clone path was also created below a root-owned parent before `git clone` ran as the unprivileged `grappa` user.

## Validation

- `bash -n infra/linux/*.sh`
- `git diff --check`
- verified every `infra/linux/*.sh` is executable
- reproduced successful native Debian 13 install after applying both fixes


vjt:
Checked against the two issues this closes — #433 (root-owned parent) and #434 (missing `+x`): both are covered, all nine `infra/linux/*.sh` go 100644 → 100755 and the clone dir is now created with `install -d -o/-g` before the `sudo -u grappa` clone. Not a review (that goes through the usual worker path), just two notes from having filed those issues:

1. **Leftover tampon.** `install_systemd.sh:23` still does `chmod 0755 "${REPO_ROOT}/infra/linux/grappa_beam_wait.sh"` at runtime. That line existed only because the mode was wrong in the repo; with this PR it is dead weight and hides a regression (if the mode is ever lost again, that one file keeps working while the other eight break). Candidate for deletion in the same PR.

2. **Group name assumption.** `install -d -o "${GRAPPA_USER}" -g "${GRAPPA_USER}"` assumes a group named like the user. `install_prereqs.sh:52` skips `useradd` entirely when the user already exists — which is exactly the scenario that produced #433 — so a preexisting `grappa` with a different primary group makes `install -d` fail with `invalid group`. `-g "$(id -gn "${GRAPPA_USER}")"` would be immune. Non-blocking on a fresh Debian 13, where `useradd` creates the matching group.

Also worth knowing before this merges: #435, filed just now, is a third defect on the same fresh-install path (`mix grappa.seed_themes` is never called, so the theme gallery lands empty). Independent of this PR — no need to fold it in.

PR #437, opened 2026-07-26
Seed built-in themes during native Linux install

abonforti:
## Summary

- run the idempotent `grappa.seed_themes` task after the first migration
- update native Linux installer progress numbering and runbook

## Problem

A fresh native Debian 13 install completed healthy, but the theme gallery was empty. The built-in themes only appeared after manually running:

```sh
MIX_ENV=prod mix grappa.seed_themes
```

`Mix.Tasks.Grappa.SeedThemes` documents the operation as idempotent and safe on every cold deploy, so fresh installs should materialize the curated gallery automatically.

## Validation

- `bash -n infra/linux/install.sh`
- `git diff --check`
- verified all 11 progress markers are present
- manually confirmed the task populates the gallery on the fresh Debian 13 install

### DIFF
--- a/infra/linux/README.md
+++ b/infra/linux/README.md
@@ -44,13 +44,15 @@ directly and through the locally-installed nginx.
    `compile --warnings-as-errors`, `release --overwrite`.
 5. **Secrets bootstrap** — see below.
 6. **First migration** — `Grappa.Release.migrate()` via `release.sh`.
-7. **`cic_build.sh`** — builds the cicchetto PWA with bun into
+7. **Theme seed** — idempotently materialises the curated built-in theme
+   gallery in the database.
+8. **`cic_build.sh`** — builds the cicchetto PWA with bun into
    `runtime/cicchetto-dist/`.
-8. **`install_systemd.sh`** — installs + enables (doesn't start) the
+9. **`install_systemd.sh`** — installs + enables (doesn't start) the
    `grappa.service` unit.
-9. **`install_nginx.sh`** — installs nginx config + the shared
+10. **`install_nginx.sh`** — installs nginx config + the shared
    `infra/snippets/*.conf`, symlinks the PWA dist, starts/reloads nginx.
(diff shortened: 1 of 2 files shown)


vjt:
Read it against #435, which I filed — the fix is right and lands where the issue proposed: after `mix ecto.migrate` (now step 7/11), before `systemctl start grappa`, so `Mix.Tasks.Grappa.Boot`'s endpoint suppression is not even load-bearing here (nothing is listening yet). Progress renumbering is complete: 11 markers in `install.sh` and the `infra/linux/README.md` list matches. `set -euo pipefail` (install.sh:25) means a seeding failure aborts the install loudly, which answers design question 3 in the issue the way I would have argued for.

Three things still open, none of them a reason to hold this PR:

1. **The upgrade path still does not seed.** `infra/linux/deploy.sh` runs `mix ecto.migrate` (deploy.sh:189) and nothing else, so built-ins added to `Grappa.Themes.Builtins` after an install never reach an existing host — the operator is back to running the task by hand, which is what #435 was about. The task is idempotent by design, so the same block would be safe next to the deploy migrate. This is design question 1 in the issue, still unanswered.

2. **FreeBSD/packaging parity** (design question 2) — `infra/freebsd/` has its own `deploy.sh` and jail scripts and is untouched here, so the jail path keeps the gap.

3. **Merge order with #436.** Both PRs edit the same region of `install.sh` (#436 replaces the `mkdir -p` at line 56 and flips the file mode; this one rewrites the adjacent `say` line). GitHub reports both MERGEABLE against the current main, but that is pairwise — whichever lands second will most likely need a rebase, and #436 also carries the mode change that makes these scripts runnable at all. Merging #436 first and rebasing this one on top is the cheaper order.

Style nit, take it or leave it: the migrate block above uses a standalone `export MIX_ENV=prod`, this one inlines `MIX_ENV=prod` on the command. Same result, and the neighbouring blocks now read differently.

Issue #324, opened 2026-07-18
media-modal: same-host gate rejects deployment aliases → uploads open in-place, not modal


vjt:
## Problem

Upload links do not open in the in-app media viewer modal when the client is served from a **different hostname alias than the one the link carries**. They fall back to the plain anchor — and on iOS standalone PWA that triggers the navigate-in-place bug (raw media document, no chrome, return reloads cic).

### Repro (field-confirmed)

- cic opened from `https://irc.sindro.me`
- scrollback link: `📸 https://irc.sniffo.org/uploads/<slug>`
- both hostnames are aliases of the **same** grappa instance (shared `/uploads/<slug>` store)
- click → no modal, plain navigation.

## Root cause

`cicchetto/src/lib/mediaLink.ts` → `sameHostUrl()` gates on a **single** host equality:

```ts
if (url.host !== hostOf(origin)) return null;
```

`hostOf(origin)` is `window.location.origin`'s host. A deployment that answers on several hostname aliases (`irc.sindro.me`, `irc.sniffo.org`, …) mints upload URLs under one alias while a session may be loaded from another → cross-host → `classifyMediaLink` returns `null`.

## Fix

Replace the single page-origin host equality with **membership in the deployment's alias allowlist**. The existing re-root already does the heavy lifting:

```ts
rooted: `${origin}${url.pathname}${url.search}${url.hash}`
```

Because the aliases share one backend/upload store, a link `https://irc.sniffo.org/uploads/x` viewed from `irc.sindro.me` is re-rooted to `https://irc.sindro.me/uploads/x` → the modal's `<img>` stays **same-origin** → **CSP `img-src 'self'` is untouched (no loosening needed)**. The fix is one predicate change; the re-root machinery is unchanged.

## Alias-list source — server-provided, single source of truth

Per vjt: the alias set must be **server-provided**, with **no second hand-maintained copy** anywhere. The server already knows its own hostnames (nginx `server_name` / vhost pool / source addresses). Expose that set to cic at runtime (server-settings payload or an ISUPPORT token) and have the classifier read it. Adding a vhost then needs neither a cic redeploy nor a duplicated list.

Do **not** bake a static host array into the client — that is exactly the duplicated-list anti-pattern to avoid.

## Constraints to preserve

- **Keep the re-root.** The returned `href` must stay normalized to the page origin, never the alias host — otherwise mixed-content / cross-origin `<img>` and the CSP block reappear.
[...]

PR #451, opened 2026-07-26
Media management on shottinoo + bug fix chat echo

nextime:


vjt:
## Pre-merge security review + maintainer ruling

Apologies this lands after the fact, @nextime — the review below was done as a
pre-merge pass and should have been on the PR before we touched anything. It is
here now, together with the changes @vjt has approved to make **before** merge.
Posting it so the record is complete and nothing changes under your feet
silently. The encoders are genuinely solid: the pure functions are well tested
(`test_media.c` pins the framing, the DA1 near-miss parsing, cell fitting, and
NULL-safety), the decode/draw buffer math checks out under manual tracing, and
there is no shell injection (`run_cmd` = `execvp` with an argv array). Nice work.

### Findings (by severity)

**H1 — HIGH (threat-surface / design, not a memory bug): inline media auto-fetches
and auto-decodes arbitrary peer `http(s)` URLs through ffmpeg on scroll, default-on.**
`main()` sets `inline_media_enabled = true`, and the draw path promotes any row
carrying an image URL to `IM_FETCHING` the first time it is on screen
(`draw()` → `JOB_MEDIA` → `media_decode_job` → `run_cmd(ffmpeg -i <url>)`). So a
hostile channel peer who posts an image link causes the victim's client to
contact the attacker's host and hand untrusted bytes to ffmpeg with no action
beyond scrolling. Two costs: (a) a tracking-pixel — leaks IP + liveness/read
confirmation; (b) passive exposure of ffmpeg's substantial historical CVE
surface (malformed media; the HLS/concat class) running unsandboxed with the
user's privileges. `find_url` restricts to `http(s)://` (no direct `file://` —
good), and no `-protocol_whitelist` is set on ffmpeg. This is the meaningful
escalation over the old click-to-preview model.

**L1 — LOW (defensive, not reachable with current callers): `media_emit_sixel`
malloc without an overflow guard.** `malloc((size_t)w*(size_t)h*3*sizeof(int))`
and `malloc((size_t)w*(size_t)h)`. Today w/h come from terminal geometry
(`cols*6`, `rows*12`), so the products are bounded to tens of MB and not
attacker-controlled. But `media_emit_sixel` is a public, independently-tested
API — a future caller with large w/h would wrap on 64-bit.

**INFO — merge conflict.** `git merge-tree main <head>` reports exactly ONE
conflict, in `frontends/shottino/Makefile` (both sides add to the OBJS / TESTS /
header-dep lines and the coverage-caveat comment). `shottino.c` auto-merges
cleanly — the recent `decode_chunked` → `http.c/.h` extraction on main does not
touch the media path. Trivial union resolution: keep both `http.o`/`media.o`,
both test targets, both header deps.

### Verified and NOT a bug (coverage, so the depth is visible)
- `mime_for_path` now returns `NULL` for unsupported types: the only caller is
  `upload_command`, which null-checks it → no NULL-deref.
- `log_media` init: all three append sites (`log_line`/`log_line_mention`/
  `add_pending_echo`) set `-1`; `/clear` compaction now copies `log_ids` +
  `log_media` → no stale/OOB slot (draw also guards `mi < media_count`).
- `draw_inline_media` (rgb path): rows/cols clamped ≤ decode dims, stride =
  `m->cols` is stable (`if (m->rows <= 0)` never re-fits), resize-safe — no OOB.
- Bounds are correct in `copy_url_token`, `b64_write`, `da1_query`,
  `media_da1_has_sixel`, and the Floyd–Steinberg spread (`ny < 0` is impossible
  and correctly omitted).
- The decode/recycle race is handled: publish is guarded by
  `strcmp(url)==0 && state==IM_FETCHING`.

### Maintainer ruling (@vjt) — four changes before merge
1. **Auto-load only first-party grappa `/uploads/` links; click-to-load for every
   other peer `http(s)` URL — no automatic fetch on scroll.** (Addresses H1.)
2. **`-protocol_whitelist file,crypto,tcp,tls,http,https` on ALL ffmpeg argv.**
3. **Add the L1 defensive overflow guard in `media_emit_sixel`** even though it is
   not reachable today.
4. **Resolve the Makefile conflict** with the union above.

We will implement these as pre-merge follow-ups on top of your commits; happy to
walk through any of them. Thanks again — the inline-media work is a real feature.

vjt:
Merged into `main` as `merge #451: Media management on shottinoo + bug fix chat echo` (`ae780611`). Thank you, @nextime — the terminal-graphics work is genuinely solid and the pure encoders are well tested. Your six commits landed verbatim (original authorship preserved, nothing squashed or rewritten); we added three small pre-merge hardening commits on top, per the review + maintainer ruling above:

1. **H1 — first-party gate on auto-decode (ruling).** Inline media now auto-fetches/auto-decodes **only** for first-party `/uploads` URLs — the connect host unioned with the `#324` `http_host_aliases` set, at parity with cic's `mediaLink` classifier. Every other peer `http(s)` URL stays click-to-preview, so scrolling past a hostile link no longer hands untrusted bytes to ffmpeg on its own. Fallback is deliberately restrictive: if the alias list is unavailable at boot, only the connect host counts as first-party.
2. **H1 — ffmpeg `-protocol_whitelist file,crypto,tcp,tls,http,https`** inserted before `-i` on both ffmpeg argvs (kitty/iTerm2 PNG + sixel/art rawvideo).
3. **L1 — `media_emit_sixel` overflow guard.** A `size_t` overflow check on `w*h*3*sizeof(int)` / `w*h` before malloc. Not reachable with the terminal-bounded callers, but the encoder is a public, independently-tested API.

**Makefile:** the single merge conflict (both sides extend the OBJS / TESTS / header-dep lines) was resolved as a union — both `http.o`/`media.o`, both `test_http`/`test_media` targets, both header deps.

Full rationale in the review comment above. Closing as merged. 🙏

Issue #554, opened 2026-07-30
Session: an operator KILL/AKILL must be terminal, not a Backoff reconnect


vjt:
## Problem

An oper-initiated `KILL` (and therefore an AKILL, which the upstream delivers as a `KILL` plus a socket close) is **not modelled as a terminal condition**. `KILL` is parsed — `lib/grappa/irc/parser.ex:90` maps it to `:kill`, `lib/grappa/irc/message.ex:67` types it — but no `handle_info/2` clause in `lib/grappa/session/server.ex` or `lib/grappa/session/event_router.ex` matches it. Verified against `origin/main` (`3e6f10d4`):

```
$ git grep -nE ':kill' origin/main -- lib/grappa/session/server.ex lib/grappa/session/event_router.ex
(no match)
```

So the kill reaches us only as the subsequent `:tcp_closed`, which is exactly the shape `Grappa.Session.Backoff` is built to retry (see its moduledoc: "forced disconnect from upstream lands as `:tcp_closed`"). The session respawns and reconnects into a ban that was just applied.

The ban is eventually discovered, but only on the *next* connect, via numeric 465, which is already handled as terminal (`lib/grappa/session/server.ex:2590`, `handle_terminal_failure/2`). That costs one extra full connect per akill — and reconnect attempts against a banned address are precisely what upstream connection throttling punishes, so the recovery path makes the situation worse rather than better.

## Proposed behaviour

Treat a `KILL` whose target is our own current nick as terminal, on the same path as 465:

- mark the credential `:failed` with a reason carrying the kill's comment verbatim (same treatment 465 gives the trailing param),
- do **not** hand the disconnect to `Backoff`,
- surface it to the client with a distinct reason so the UI can say "killed by an operator" rather than showing a generic reconnect spinner.

A `KILL` targeting some *other* nick is ordinary traffic and must keep flowing to the client unchanged.

## Notes

- This is a separate concern from address mapping / ban granularity, which is tracked in #454. Filed separately on request.
- Worth deciding explicitly whether an operator kill should also stop the session from being restarted by `Grappa.Bootstrap` on the next deploy — 465 already does this by marking the credential `:failed`; the same should hold here, otherwise a deploy silently retries the ban.

vjt:
Shipped to production in the v0.8.0 batch (main `5f8f3616`, cold deploy 03:28–03:30 CEST). Verified live: `/api/config` reports `0.8.0-5f8f3616`, new bundle served.

Issue #537, opened 2026-07-29
Identifier folding: one fold at every boundary, plus per-network casemapping normalised at ingress


vjt:
Case-folding of identifiers is **sigil-gated**, so it applies to channel-shaped names only. Every other identifier — a DM peer nick, `$server`, any nick-shaped window key — is stored and compared case-sensitively, while the read paths that resolve those same targets are case-insensitive. Write-case-sensitive + read-case-insensitive is the shape that produced the permanent ghost badge in #532 (ask D): one window, N keys, and the keys nobody re-opens never clear.

Owner ruling on #grappa, 2026-07-29:

> il casefolding non deve esser limitato ai canali, ma a ogni identifier. così non mi rimangono i badge (1) fantasma

> potremmo anche usare canonical target ovunque eh, semplifichiamo

## Second axis — per-network casemapping, normalised at ingress

Owner ruling on #grappa, 2026-07-29 10:09:

> si aggiorna 537 e normalizza su reti rfc1459 i casi che porterebbero a dupe []{}^ etc, facciamolo in ingresso su grappa cosi il resto del codice puo tranquillamente considerare tutti ascii

This axis is live, not hypothetical. Measured today:

* The prod `networks` table holds three rows: `azzurra` (1), `libera` (3), `oftc` (4).
* Azzurra's 005 advertises `CASEMAPPING=ascii`; **Libera's 005 advertises `CASEMAPPING=rfc1459`** (measured from a live 005 line off `tungsten.libera.chat`). OFTC runs the same hybrid family and almost certainly advertises rfc1459 too — unconfirmed, no session was open to read its 005.
* **The ircd does not normalise the target it relays.** bahamut's `m_message` (`src/s_user.c`) puts `chan` on the outbound line — the raw `parv[1]` token exactly as the sender typed it — never `chptr->chname`: `sendto_channel_butone(cptr, sptr, chptr, ":%s %s %s :%s", parv[0], cmd, chan, parv[2])`. So on an rfc1459 network two members typing `#foo[1]` and `#foo{1}` are in ONE upstream channel and put TWO spellings on our socket.
* #525's defence — *"on a genuine rfc1459 network the lax case can't even arise (the ircd refuses the second nick)"* — holds for **nicks** only. Nobody refuses an alternate spelling of a channel you are already in, so for channels the ASCII-only fold splits one upstream channel into two windows, two scrollbacks, two cursors. That is the #532 family again, one door further out.

[...]

PR #563, opened 2026-07-30
fix(#554): operator KILL/AKILL on own nick is terminal, not a Backoff reconnect

vjt:
## What

An oper `KILL` (and an AKILL, delivered upstream as `KILL` + socket close) on our own nick is now **terminal**, on the same path as numeric 465 — instead of falling through to the generic delegate and letting the follow-on `tcp_closed` hand the drop to `Backoff`, which reconnected straight into the just-applied ban.

New `handle_info(:kill)` clause in `Session.Server` (next to 465/904):
- **Self-match** on `target` folded vs own nick (`fold_key/2`, network-aware `canonical_target`, #537/#121). **Sender-agnostic** — an AKILL from a server/service prefix still matches (fix the class, not the oper example).
- `handle_terminal_failure/2` → mark `:failed` with a **distinct `"killed:"` reason** (≠ `"k-line:"`/`"sasl:"`; Bootstrap skips it next deploy; cic shows a killed reason, not a reconnect spinner), no Backoff, `{:stop, :normal}`.
- **KILL on any other nick** delegates unchanged (still a `$server_event`, socket stays). Malformed KILL is non-terminal.
- **No user lockout** — `Networks.connect/2` accepts `:failed → :connected`, so a manual `/connect` still works. Terminal = no *auto* retry.

Server-only; cic's dedicated rendering of the `"killed:"` reason is a follow-up.

## Ordering (proven, not assumed)

`IRC.Client` is `packet: :line, active: :once` and forwards each parsed line to the Session **before** re-arming the socket, so `{:irc, kill}` is enqueued strictly before the `tcp_closed`-driven `{:EXIT}`; FIFO stops us in the KILL clause before the Backoff clause ever runs. ERROR is not terminal (no handler); only `tcp_closed` triggers Backoff, and it's strictly last.

## TDD evidence

**RED** (4 unit tests, before writing the handler):

```
4 tests, 2 failures (320 excluded)

  test terminal failure triggers — lenient :failed (S1.4) KILL targeting own nick ...
  Assertion failed, no matching message after 2000ms
  code: assert_receive {:DOWN, ^ref, :process, ^pid, :normal}
  stacktrace:
    test/grappa/session/server_test.exs:6253: (test)   # self-KILL
    test/grappa/session/server_test.exs:6297: (test)   # fold-case
```

**GREEN** after the handler: `4 tests, 0 failures`.

**Full `check.sh`**: `8 doctests, 47 properties, 4609 tests, 0 failures` · credo `found no issues` · dialyzer `Total errors: 0` · sobelow ok · bats `ok 152`.

**Real-ircd e2e** (scoped, live bahamut `bahamut(perimeter)-1.4(34)-azzurra`, `CASEMAPPING=ascii`): an opered peer (`381 :You are now an IRC Operator`) KILLs the seeded session's nick → credential `connection_state=failed` with a `killed:` reason, **stays failed** (no reconnect) → proves bahamut delivers the KILL to the victim and the terminal path end-to-end.

```
✓ issue554-operator-kill-terminal.spec.ts › #554 — operator KILL marks the network :failed with a killed reason and does not reconnect (9.2s)
1 passed (22.6s)
```

Full integration is the CI gate on this PR (base main).

Refs #554.

### DIFF
--- a/cicchetto/e2e/fixtures/ircClient.ts
+++ b/cicchetto/e2e/fixtures/ircClient.ts
@@ -362,6 +362,17 @@ export class IrcPeer {
     await kicked;
   }
 
+  // Issue a raw KILL <target> :<reason> (oper-only upstream) — #554 e2e:
+  // an opered peer KILLs the grappa session's upstream nick to prove the
+  // KILL-terminal path. Fire-and-forget: irc-framework doesn't echo own
+  // commands, and the victim's disconnect is observed SERVER-side (the
+  // credential goes connection_state :failed), not on this peer.
+  // `raw(array)` adds the trailing-param `:` itself, so the reason ships
+  // as one token. Requires a prior `oper(...)`.
+  kill(targetNick: string, reason: string): void {
+    this.client.raw(["KILL", targetNick, reason]);
+  }
+
   // /OPER up to ircop. Required to bypass bahamut's "no ops on new
   // channels in split-mode" gate that otherwise locks every freshly
   // created channel out of any kind of mode-setting (including +i, +k,
--- a/cicchetto/e2e/tests/issue554-operator-kill-terminal.spec.ts
+++ b/cicchetto/e2e/tests/issue554-operator-kill-terminal.spec.ts
@@ -1,0 +1,103 @@
+// #554 — an operator KILL / services AKILL is terminal, not a Backoff
+// reconnect.
+//
+// A peer OPERs up against the testnet O:line (testoper/testoperpass) and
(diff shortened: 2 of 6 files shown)


vjt:
Landed on main via rebase+ff (linear history is the house style), so the SHAs differ from this PR head `f52b4053` and GitHub cannot auto-mark it merged. The work IS on main: `d832dc30` (fix) + `f94b5bfb` (real-ircd e2e) + `3968f831` (docs). Closing as merged-by-rebase; #554 stays open until the batched deploy ships it.

Issue #543, opened 2026-07-29
Outbound addressing modes: fixed pool with reservations vs source-mapped pool with reservations (FreeBSD + Linux plumbing)


vjt:
Implementation issue for the configurable half of #454. #454 owns the *derivation* design and its rulings (keyed non-reversible hash, sizing, throttle constraint); this issue owns the **operator-facing addressing mode**, the **reservation semantics**, and the **per-platform kernel plumbing** — including the FreeBSD numbers measured tonight, which change one of #454's stated constraints.

Requested by vjt (2026-07-30): the vhost admin section should let the operator pick an addressing mode and define a prefix as an alternative to the curated pool, with per-user reservations taking precedence and everyone else mapped from their own source address. Explicitly: *"non c'è più il pool visitor a scelta random"*.

## Current state (verified on `origin/main`, not from memory)

- `vhosts` rows are curated addresses with two flags — `in_pool` (rotation membership) and `generally_available` (self-selectable by any subject) — `lib/grappa/vhosts/vhost.ex`.
- `vhost_grants` is already a per-subject grant of a `vhosts` row (`user_id` XOR `visitor_id`) — `lib/grappa/vhosts/grant.ex`. **This is the "reservations" table; it does not need inventing.**
- Resolution order today is `Grappa.Vhosts.effective_source/2` (#266): per-network `network_servers.source_address` wins absolutely → else the subject's selection (random among the selected) → else `nil`, and `Grappa.IRC.Client` falls through to `Grappa.OutboundV6Pool.pick/0` (random over `in_pool`), else kernel default.
- The pool is a `:persistent_term` set installed by `Grappa.Vhosts.resync_pool/1` → `OutboundV6Pool.apply_pool/1`, re-applied hot on any admin inventory edit (`lib/grappa_web/controllers/admin/vhosts_controller.ex`).
- The candidate universe offered by the admin UI is the host's **bound** addresses: `Grappa.Net.HostAddresses.list/0`, and `servers_controller.ex:245` gates a source on `HostAddresses.local_bindable?/2`. **This gate is the thing that has to change** for a prefix-derived source, which by definition is not in the bound list at config time.
- The actual bind already works and needs no change: `source_bind/2` in `lib/grappa/irc/client.ex` binds the literal as `ifaddr` after checking the upstream is reachable in that family.

## Proposal — one setting, two modes

An `addressing_mode` on the vhost section:

[...]

PR #564, opened 2026-07-30
feat(#543): deterministic /64→::cb source derivation (INC-2)

vjt:
## INC-2 of #543 — pure derivation module

Second increment of static-mapping outbound addressing. Adds `Grappa.Vhosts.SourceMapping`, a **pure** module that maps an untrusted subject's client prefix to ONE deterministic outbound address inside the configured `::cb::/80`. This is what replaces the random pool in mode 2 (wired up in INC-4/6).

Stacks on INC-1 (#562, now merged to main).

### Surface
- `client_key/1` — folds a v6 IP to its `/64` (interface id dropped, RFC 8981) or a v4 IP to its `/32`. This is the mapping key.
- `derive/2` — `sha256(@domain_tag <> key)` fills the host bits of the prefix; returns a canonical v6 literal strictly inside it. Deterministic + idempotent.
- `in_prefix?/2` — v6-only membership test (false on v4 / malformed input / malformed prefix). Feeds the INC-7 prefix-impact scan.

### Locked design decisions (#543)
- Key on the routable prefix, not the full address — a roaming client keeps ONE egress; NAT/CGNAT collapse is intended.
- Hash is **unkeyed** — reversibility is irrelevant, collision-freedom is the only property that matters; `@domain_tag` is namespace separation, not a secret.
- Reservations live OUTSIDE `::cb`, so a derived address never shadows a reserved one.

### Implementation note
`derive/2` extracts host bits via bitstring pattern (`<<host::size(host_bits), _::bitstring>>`) rather than a byte-aligned `:binary.part` slice, and re-slices `net‖host` straight into the 8×16 tuple with no integer round-trip — so it holds for EVERY prefix length (not just `/8`-aligned) and duplicates no bit helper from `IpLiteral`.

### Tests
Unit (determinism, interface-id insensitivity, IPv4 collapse, `/128` degenerate, invalid-prefix + out-of-prefix rejection) + a 100k distinct-`/64` injectivity check + StreamData properties over random v6 `/64`s, v4 `/32`s, and **arbitrary non-`/8`-aligned prefix lengths 0..128** (inside-prefix, interface-id-stable, idempotent).

### Gates
- Full `scripts/check.sh` — **exit 0** (`ok 152` bats tail), run twice (pre- and post-review-fix).
- Code review: `10x-engineer:code-reviewer`, synchronous — verdict *correctness CLEAN*; two nice-to-have items (arbitrary-length property gap + a duplicated bit helper) both applied before this push.

### DIFF
--- a/lib/grappa/vhosts/source_mapping.ex
+++ b/lib/grappa/vhosts/source_mapping.ex
@@ -1,0 +1,131 @@
+defmodule Grappa.Vhosts.SourceMapping do
+  @moduledoc """
+  Pure derivation of a deterministic outbound source address from a
+  client's network prefix — the core of the #543
+  `static_mapping_with_reservations` addressing mode.
+
+  Each untrusted subject egresses from ONE stable address inside the
+  configured `::cb::/80`, derived from the subject's OWN client `/64`
+  (v6) or `/32` (v4). This replaces the random pool for mode 2: an
+  address is `derive(client_key(client_ip), prefix)` — same client, same
+  address, forever (until the operator renumbers the prefix).
+
+  ## The mapping key (`client_key/1`)
+
+  The key is the client's routable prefix, NOT its full address:
+
+    * IPv6 → the first 64 bits (the `/64`). The interface id is IGNORED
+      because RFC 8981 rotates it (privacy extensions) — keying on it
+      would hand a single roaming laptop a new outbound address every
+      few hours, defeating accountability. Keying on the `/64` gives a
+      home/mobile subscriber ONE address for the life of their prefix.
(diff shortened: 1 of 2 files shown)


vjt:
Landed on main as `69c40cc1` via rebase+ff — the rebase onto #554 changed the SHA, so GitHub cannot auto-mark this PR merged (head stayed `2d152226`). Verified on main: `lib/grappa/vhosts/source_mapping.ex` + its test, 291 insertions. The rebased tree was re-gated with a full `check.sh` before the push. Closing as merged-by-rebase; #543 stays open for INC-3…INC-8.

PR #566, opened 2026-07-31
feat(#543): persist last client /64 per subject (INC-3 A+B)

vjt:
Part of #543 (static-mapping outbound addressing) — increment **INC-3**, Parts A+B only. Stacked on INC-2 (`SourceMapping`), which is now in `main`; this PR is INC-3-only.

## Why
The `static_mapping_with_reservations` mode derives an untrusted subject's outbound source from the subject's OWN client `/64`. That derivation runs at **upstream-connect**, where no client socket may be attached — so the subject's last-known client prefix must be persisted out-of-band and read back there. This lands the persistence machinery.

## What
- **Part A — `UserSettings`**: a dumb `"last_client_prefix64"` accessor pair (`get_last_client_prefix64/1`, `put_last_client_prefix64/2`), mirroring the `vhost_selection` split. A raw `client_key` binary is not JSON-safe, so the value is stored as an OPAQUE base16 string; `UserSettings` validates only "non-empty base16" and stays domain-agnostic. Works for user AND visitor subjects.
- **Part B — `Vhosts`** owns the domain logic: `record_client_source/2` normalises the remote IP to its `/64` (v6) or `/32` (v4) via `SourceMapping.client_key/1` (interface id dropped, RFC 8981), base16-encodes it, persists last-write-wins. `last_client_prefix64/1` decodes back to the raw binary, `nil` on never-seen OR malformed (`Base.decode16`, never the raising `!`). A persist failure returns `:ok` (a capture miss must never fail the connect) but is LOGGED, not silently swallowed (CLAUDE.md boundary rule).

## Part C split to its own increment
The plan claimed "RemoteIp already resolves the real client IP per RemoteIpFromProxy" at WS connect — that is FALSE. The WS socket declares `connect_info: [:auth_token]` only; `RemoteIpFromProxy` is an HTTP Plug that never runs for `connect/3`, so the client remote IP is not available at `UserSocket.connect/3` today. Wiring it needs a load-bearing `connect_info` edit (+ `:peer_data`/`:x_headers`) and REUSE of the (subtle, inverted-`:clients`) trust matrix via a shared helper — its own worktree, gate, and review (approved direction).

## Gates
- Local `scripts/check.sh` full: Credo strict (no issues), ExUnit (0 failures), Dialyzer (0 errors), Sobelow, bats — all green.
- Synchronous code review (clean prod; two boundary test-coverage gaps applied: decode-failure→nil, :ok+LOG on subject-gone).
- TDD: RED → GREEN.

### DIFF
--- a/lib/grappa/user_settings.ex
+++ b/lib/grappa/user_settings.ex
@@ -62,6 +62,9 @@ defmodule Grappa.UserSettings do
   |                        |                        | `put_display_prefs/2`,          |
   |                        |                        | `display_prefs_persisted?/1`    |
   |                        |                        | (#449)                          |
+  | `"last_client_prefix64"`| `String.t() \\| nil`  | `get_last_client_prefix64/1`,   |
+  |                        | (opaque base16)        | `put_last_client_prefix64/2`    |
+  |                        |                        | (#543)                          |
 
   ## Boundary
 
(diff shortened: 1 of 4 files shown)


vjt:
Landed on main as `e0faf535` via rebase+ff — the SHA differs from this PR head `fcfbf641`, so GitHub cannot auto-mark it merged. This PR was Parts A+B only (the persistence core); Part C (the WS connect-time capture) was split off, briefly landed as `78a1ada7`, and was reverted in `822e3b14` to keep it out of tonight's batch — it re-enters via its own PR from `543-addressing-inc3c-wsconnect`. #543 stays open for INC-4…INC-8.

PR #567, opened 2026-07-31
chore(cic): tooling group bump (biome 2.5.6 / TS 7) + source fixes — companion to #490 [HOLD]

vjt:
**Draft / on hold — do NOT merge yet.** Not part of tonight's batch. Companion to #490 (dependabot `tooling` group); vjt reconciles the two.

## Why
#490 bumps six `cicchetto/` dev-tooling deps but leaves the cic source failing the local `bun run check` gate under the new **biome 2.5.6** recommended rule set. (CI is green because `ci.yml` runs no cic biome/tsc/vitest step — the cic gates are local-only, so the breakage never surfaces in CI.) This branch is #490's six bumps **plus** the minimal source changes to make the gate green again.

## Contents
1. `dd692c66` — the exact #490 dep bumps (cherry-picked): biome 2.4.13→2.5.6, **typescript 6.0.3→7.0.2 (major)**, vite 8.0.10→8.1.5, vite-plugin-pwa 1.2.0→1.3.0, vite-plugin-solid 2.11.12→2.11.14, vitest 4.1.5→4.1.10. All `direct:development`, no CVE.
2. `83be97d5` — fix the 3 `lint/correctness/noUnsafeOptionalChaining` **errors**: `(putCall?.[1] as RequestInit).body` short-circuits to `undefined` then throws; each site already asserts `toBeDefined()` above, so `?.`→`!` (the codebase's established post-assert idiom). No behavior change.
3. `9b2e37bb` — `biome check --write` (safe fixes only) for the 4 `format` **errors** from biome 2.5.6's reformatter. The `default.css` change is **pure whitespace reflow**, verified `git diff --word-diff=porcelain --ignore-all-space` is empty → zero CSS cascade/semantic change, no e2e needed.

## Gates (local, lane-free)
- `bun run test` (vitest): **green** — 205 files, 3647 tests
- `tsc --noEmit` (TypeScript 7): **green** — exit 0, zero type errors (the TS 6→7 major is safe here)
- `bun run check` (biome + tsc): **green** — 0 errors (32 non-blocking warnings remain)

## Deliberately out of scope
The residual warnings (`noNonNullAssertion` ×~14, `noDescendingSpecificity` ×7, `noImportantStyles` ×2, `useOptionalChain` ×1) are **non-blocking** and left as-is. Fixing the CSS-specificity / `!important` ones would alter the cascade — a behavior change that would need an e2e pass — which is out of scope for a tooling bump. A follow-up (or a targeted biome rule config) can address them separately if desired.

### DIFF
--- a/cicchetto/bun.lock
+++ b/cicchetto/bun.lock
@@ -12,7 +12,7 @@
         "uqr": "^0.1.3",
       },
       "devDependencies": {
-        "@biomejs/biome": "^2.4.13",
+        "@biomejs/biome": "^2.5.6",
         "@fontsource/cascadia-code": "^5.3.0",
         "@fontsource/fira-code": "^5.3.0",
         "@fontsource/ibm-plex-mono": "^5.3.0",
@@ -24,12 +24,12 @@
         "@types/phoenix": "^1.6.7",
         "hack-font": "^3.3.0",
         "jsdom": "^30.0.0",
-        "typescript": "^6.0.3",
+        "typescript": "^7.0.2",
         "unicode-emoji-json": "^0.9.0",
-        "vite": "^8.0.10",
-        "vite-plugin-pwa": "^1.2.0",
-        "vite-plugin-solid": "^2.11.12",
-        "vitest": "^4.1.5",
+        "vite": "^8.1.5",
+        "vite-plugin-pwa": "^1.3.0",
+        "vite-plugin-solid": "^2.11.14",
+        "vitest": "^4.1.10",
         "workbox-precaching": "^7.4.1",
         "workbox-routing": "^7.4.1",
         "workbox-window": "^7.4.1",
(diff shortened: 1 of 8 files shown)


vjt:
Superseded: every bump this branch carried is already on `main` — `@biomejs/biome` 2.5.6, `typescript` ^7.0.2, `vite` ^8.2.0, `vite-plugin-pwa` ^1.3.0, `vite-plugin-solid` ^2.11.14, `vitest` ^4.1.10 — landed via the individual dependabot PRs, and the companion #490 has been closed since 2026-07-31. The source fixes that made the local gate green are moot on the current tree, and #716 now runs `bun run check` + `vitest` in CI unconditionally, so the local-only gate this branch worked around no longer exists. Closing; reopen if something regressed.

Issue #474, opened 2026-07-27
cicchetto: the rail shows server info on server windows (network, address dialled, connected-since, own nick)


vjt:
Ruling by @vjt (per chan request), from driving the post-#71 build:

> "aprire la sidebar su una finestra server dovrebbe mostrare info sul server che abbiamo già che ne so indirizzo da quanto siamo connessi nome rete etc."

## Context

The right rail (`.shell-members`) became a **permanent** surface in #71 INC-2 — it is mounted on every window kind, not only on channels. On a channel window it shows the members pane; on a server window it shows nothing but the button cluster. That empty rail is the slot this issue fills.

This is not a new idea, it is the deferred half of the rail's own design. `ActionCluster.tsx`'s header states it outright:

> *"R1's stated future is for the rail to ALSO carry per-tab-kind context (server → /lusers, query → /whois). That content grafts in as SIBLINGS of this cluster inside the rail container (Shell.tsx), NOT here."*

and `Shell.tsx:717` repeats it at the desktop mount: *"a future increment turns the rail into the per-tab-kind context surface — lusers/whois — grafting as a SIBLING of the cluster; NOT in scope now."*

So: same slot, and the scope @vjt is asking for is connection facts rather than `/lusers` output.

## What to show on a server window

Facts the client **already has** — this should need no new IRC traffic:

- **network name** (and its slug);
- **the server actually connected to** — the address the session dialled, not the network label. On a round-robin network these differ, and which box you landed on is exactly what you want to know;
- **connected since** — as a duration (*"connected for 4h 12m"*), which is what a bouncer user actually reads for;
- **your nick on this network**, and whether the identity is registered/identified.

Reasonable extras if they are already in the store, not worth new plumbing on their own: TLS yes/no, the server's advertised name/version from `RPL_MYINFO`/`RPL_ISUPPORT`, lag/ping.

## Constraints

- **Grafts as a sibling** of the button drawer inside the rail container, per the contract quoted above. Do not fold it into the cluster component.
- Whatever the button drawer settles into (see the companion rail issue), this content sits above it and does not push it off-screen on a phone.
- Facts only, sourced from existing session/network state. If something needs a fresh numeric to be accurate, prefer leaving it out over showing a stale value confidently.

## Follow-on, not scope here

Query windows getting `/whois` context in the same slot is the other half of the deferred design. Keep it in mind so the container is built to hold *some* per-kind content rather than a hardcoded server card, but ship the server one first.

[...]

Issue #71, opened 2026-06-22
cicchetto: left bar (server/channel sidebar) rework — desktop + mobile consolidation


vjt:
## Summary

Rework the **left bar** (the server / channel / nick sidebar) for better visual hierarchy. Today the server row, channels and nicks render as a near-flat list, the server name reads as *reverse-indented* because of the leading cog, and the user's own nick is shown nowhere. Take visual inspiration from Discord's sidebar information hierarchy — **but this stays an IRC client**: no avatars, no voice, no Discord-isms. IRC-native, just better organised.

Started as desktop-only; **mobile is now in scope too** — the mobile bottom bar is being killed in favour of this consolidated, richer left bar. See the Mobile section below.

### Current state

The cog sits to the left of the (lowercase, config-supplied) network label, so the server line looks indented *backwards* relative to the channels under it. Everything is flat — no grouping, no indentation, channels/queries/nicks all at the same level.

![current cicchetto left bar](https://irc.sniffo.org/uploads/hqh7fzgpea45yzfurh7dspjh5m)

### Reference (information hierarchy only — NOT the chrome)

Discord's sidebar, for the grouping/indentation/own-identity ideas only. We do **not** want avatars, server-icon rail, or voice UI.

![discord ref 1](https://irc.sniffo.org/uploads/a65fkdf6lgd473wh7dguj63to4)
![discord ref 2](https://irc.sniffo.org/uploads/wemgc5hqylhcraiq3mn3tm7f7e)

## Requested changes

- [ ] **Server row affordance** — replace the leading cog next to the network name with something that doesn't make the server line read as reverse-indented. A settings/config affordance can live on hover or at the row's trailing edge instead.
- [ ] **Distinguish the server row** — give the network header a different background / weight so it's clearly the parent, not a sibling of the channels.
- [ ] **Indent channels & nicks** under their network so the parent/child relationship is visible.
- [ ] **Per-network grouping rail** — a vertical line/bracket spanning all of a network's channels and queries, visually tying them to their server header.
- [ ] **Collapsible network groups** — a disclosure chevron (archive-style) to fold/unfold a whole network. Most useful once multiple networks are connected. Note: multi-network is currently **totally untested**, so treat this item as lower priority / verify it works at all first.
[...]

Issue #294, opened 2026-07-17
theme picker: curated built-in background image SET + background picker


vjt:
Follow-up to #75 (theme picker v1). vjt directive 2026-07-17: a curated set of
built-in background images for the theme picker. **Added to the active pack
2026-07-17 with finalized answers below.**

## Scope
- A curated set of **system-owned, read-only** built-in backgrounds selectable
  in the theme editor — same ownership model as the built-in themes / `sux.theme`.
- Implies a **background-image PICKER UI** in the editor, alongside the existing
  custom-upload path (#75 Piece C ships the wallpaper layer + `theme-has-bg`
  class + opacity + upload UI — this adds a gallery of built-ins to pick from).
- Images **generated + cropped by Claude** (generate-image skill → Gemini),
  compressed. Content: mostly abstract patterns + a few funny IRC-themed.

## Finalized spec (vjt, 2026-07-17)
- **Count: 12 images.**
- **Seamless tileable** — CSS `background-repeat: repeat`, NOT full-bleed cover.
  Tile size modest (seamless tiles don't need to be full-res); hard ceiling
  2560×1440 but **prefer small — they must not be too big** (filesize).
- **Storage: static assets.** Set up a new **`assets/` directory** for static
  assets (so we can leverage future CDN caching). Served with **long-lived
  client caching** — `Cache-Control: max-age` large + `Expires` max — so clients
  don't re-download. This is the CDN-friendly posture.
- Content mix: mostly seamless **patterns** + a few **funny IRC-themed** tileable
  motifs (Claude's creative call).

## Split of work
- **Claude (orchestrator) track:** generate the 12 seamless tiles + compress.
- **Worker track (after tiles exist):** create `assets/` static dir + wire the
  caching headers (Phoenix static / nginx) + the background PICKER UI consuming
  the built-in tiles as selectable wallpapers. Mechanical once tiles land.

## Open / decide at plumbing time
- Exact `assets/` mount path + which layer sets the cache headers (Phoenix
  `Plug.Static` vs nginx `location`) — worker proposes.
- Tile pixel size (tune for seamless repeat + small filesize during generation).

Related: #293 (runtime live `.theme` import). Follows #75.

vjt:
**v1 dispatch (2026-07-17): ship these 8 backgrounds now.** Gemini rate-limited mid-set; the ~10 small tileable patterns come as a next-session follow-up. v1 = 8 selectable **cover** (full-bleed) backgrounds.

[...]

PR #572, opened 2026-07-31
cicchetto: server-info context on the rail for server windows (#474)

vjt:
Fills the empty right-rail slot on a **server window** with per-window-kind
context: connection facts the client already holds — network slug, connection
state, connected-since duration, own nick, services flavor.

Refs #474 (vjt closes post-deploy).

## What

`.shell-members` became a **permanent** desktop column in #71 INC-2 — on a
channel it shows the members pane, on a server window it showed only the
RailActions button drawer. This grafts a `RailContext` container as a
**sibling** of that drawer (both desktop + mobile rail branches), dispatching
on the active window's `kind`:

- **server** → `ServerInfoCard` (this PR)
- **query** → a `/whois` card is the deferred follow-on — one `<Match>` away,
  no second Shell edit (built as a generic container, not a hardcoded card).

`ServerInfoCard` is pure-presentational, fed only from the existing `Network`
store. `connected-since` renders **only while connected** (honesty — a duration
next to a parked/failed state would lie). Duration formatting is **lifted** from
`WhoisCard`'s private `formatIdle` into `lib/duration.ts` and shared by both
call sites (DRY — WhoisCard's only change is the import).

## Deliberately out of scope

The dialled server address / TLS / registered-identified state are **not shown**
— they are admin-only wire fields absent from the user store. Per #474's
"prefer leaving it out over showing a stale value confidently", they graft in
later behind an additive wire field and this card inherits them for free.

## Facts only

No new IRC traffic, no server-side work, no new wire field — pure cic, sourced
from existing session/network state.

## Gates

- `bun run check` (biome + tsc): exit 0, no new warnings
- vitest **full**: 208 files / 3676 tests green (duration 9, ServerInfoCard 5,
  RailContext 5, WhoisCard 20 unchanged after the DRY refactor)
- e2e (chromium, `scripts/integration.sh --grep "#474"`): **1 passed** — real
  login → server window → asserts `[data-testid=rail-server-info]` shows the
  slug, own nick, connection state, and a non-empty connected-since row from the
  live-hydrated store
- code review (sync `10x-engineer:code-reviewer`): clean, all six design
  constraints satisfied

10 files, all under `cicchetto/`.

### DIFF
--- a/cicchetto/e2e/tests/issue474-server-info-rail.spec.ts
+++ b/cicchetto/e2e/tests/issue474-server-info-rail.spec.ts
@@ -7,6 +7,18 @@
 // already in the client store (network slug, own nick, connection state,
 // connected-since) rendered by `RailContext` → `ServerInfoCard`.
 //
+// Scope B adds the LIVE upstream facts under `network.connection`, resolved
+// from `Session.connection_info/2` at the `GET /networks` boundary (the
+// additive #447 wire field): the dialled `server:port` (resolved peer IP,
+// #550), a 🔒 when TLS, and whether the nick is identified to services
+// (`registered`, from the +r umode — the #561 signal). `connection` is null
+// whenever there is no live connected session, so these rows render ONLY on
+// a real socket. This spec asserts them against what the e2e seeder actually
+// dials: `bahamut-test:6667 --no-tls --auth none` (compose.yaml) — so the
+// server row shows `:6667`, there is NO lock (plaintext), and `identified`
+// reads "no" (no SASL/NickServ → the nick never gains +r). Honesty rule:
+// never assert a 🔒 that the plaintext transport does not render.
+//
 // vitest (RailContext/ServerInfoCard/duration) proves the pure dispatch +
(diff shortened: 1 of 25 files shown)


vjt:
**The red `integration` on this branch is NOT from this PR.**

Run [`30633749708`](https://github.com/vjt/grappa-irc/actions/runs/30633749708) @ `190af1a4`: **563 passed, 1 failed**.
The single failure is `tests/issue294-builtin-bg-picker.spec.ts:38` — the stale `toHaveCount(8)` assertion left
behind when #334 grew the built-in background catalog from 8 to 14. It is inherited from this branch's base
(`6aa9c877`), not caused by anything here.

Fixed on `main` by `34bce335` (`8 → 14`), proven green by a scoped run before it landed:

    ✓ 1 [chromium] #294 picker renders the catalog + applies wallpaper live
    ✓ 2 [chromium] #294 built-in assets long-lived cache headers
      2 passed

**Rebasing this branch onto current `main` clears it.** The 563 passing includes this PR's own new spec
(`issue474-server-info-rail.spec.ts`), which was also verified green in a scoped run.

Issue #576, opened 2026-07-31
cicchetto: unread badge counts the operator's own messages (no own-sender exclusion)


vjt:
The unread badge counts the operator's **own** content rows. Nothing filters by sender: a message you just typed is "unread" until some later event happens to move the read cursor past it. Today that is masked by an optimistic cursor advance on send, but that advance has gaps, so the badge is intermittently wrong in exactly the way a user cannot explain.

Reported on IRC by an operator who saw a DM badge whose count was **only his own lines**, with zero peer lines after them; `F5` cleared it (server cursor was already ahead, the client copy had fallen behind). No reproducible trigger, so this issue is not a repro report — it is the hardening that makes the trigger irrelevant.

**Source (origin/main)**

- `cicchetto/src/lib/selection.ts` — `perChannelUnread` counts every scrollback row with `row.id > cursor`, split by `isContentKind` and filtered only by `presenceRowVisible`. There is no own-sender predicate.
- `cicchetto/src/lib/scrollback.ts` — send advances the read cursor optimistically to the new `row.id`, which is the *only* reason your own messages normally don't show up in the badge.
- `cicchetto/src/lib/readCursor.ts` — `applyReadCursorSet` adopts an incoming server cursor last-write-wins, with no forward-only gate (deliberate, so a cross-device backwards move is honoured). A cursor that lands lower than your own last message re-exposes it as unread.
- `lib/grappa/scrollback.ex` — the cold-load seed (`count_after_split/5`) has the same shape: `exclude_own_presence/2` (#532 A) strips the subject's own **presence** rows, and its comment states explicitly that "CONTENT rows are untouched (own content still counts toward `messages`)". Twin in `ReadCursor.bulk_unread_split/2`.

**Why it matters**

A row you authored is read by definition. Leaving it countable makes the badge depend on cursor timing, and the two known ways that timing slips both produce a badge the operator cannot clear by reading:

1. the send-time advance is gated on the pane already having rendered rows (`hasRenderedRow`, issue #50 / m6) — a query opened with `/msg` and written into immediately never advances;
2. a server cursor arriving from another session (or an out-of-order write) can pull the local cursor *below* your own messages.

Both are invisible to the user, who sees a badge on a window where the last lines are theirs.

**Ask**

Exclude own-authored **content** rows from the unread count — the content-row twin of #532 A — so the badge stops depending on cursor timing:

[...]

Issue #532, opened 2026-07-29
Stale unread badges: self-PART cursor, per-casing duplicate DM cursors, outbound DMs counted as notify-worthy, no unread state in Archive


vjt:
Unread counters hold state the user cannot see or clear. Four defects, one theme, filed together because they were hunted down in one session and the user hits them as a single symptom ("there is a 1 stuck in my PWA and nothing clears it"):

* **A** — leaving a channel leaves a permanent unread behind (server).
* **B** — the Archive modal gives no way to find which window is holding it (client).
* **C** — the PWA icon badge counts the user's **own outgoing DMs** as notify-worthy (server).
* **D** — a DM window accumulates one read cursor **per casing**; opening it advances only one of them (server).

C and D are the ones that survive everything: reinstalling the app does not help (the server genuinely returns a non-zero badge), and neither does opening the offending window (it advances a different cursor row than the one holding the count). D is the root cause of the reported symptom; C is why that particular row was counted as notify-worthy in the first place.

---

### What happens

1. You PART a channel. `EventRouter`'s self-PART arm persists a `:part` audit row for the subject; `Session.PartCleanup.cleanup_local/2` evicts members/topics/modes/window_state — **nothing touches the read cursor**.
2. The window leaves the sidebar (parted) and reappears in Archive (`Scrollback.list_archive/3` derives archive as "targets with rows" minus the active keyset).
3. The read cursor is still parked *before* that own PART row, so `WindowCounts.snapshot/6` keeps returning `%{messages: 0, events: 1, severity: :event}` for that window — and `MeController.build_unread_counts/2` keys the envelope off **read cursors**, not off active windows, so the archived window keeps contributing to the unread envelope forever.

Net effect: a stale "1 unread" that the user can neither locate nor clear from the UI, produced by an event **the user themselves generated**.

### Evidence (measured on prod, read-only)

A real case on the reporting user's account: archived channel with read cursor at message `1052543`; the only row after it is `1052676`, `kind = part`, `sender` = the user themselves, i.e. their own leave. Content-kind unread (`privmsg | notice | action`) across **every** window on all three networks of that account: zero. The only unread anywhere was that self-generated presence row.

### Ask

[...]

Issue #396, opened 2026-07-25
Cold-load logon issues ~2 queries per window (~100 round trips) — collapse to a single GROUP BY


vjt:
## Problem

The cold-load path builds unread counts with a **query per window**, and there are two of them per window.

`MeController.build_unread_counts/2` (`me_controller.ex:248-266`) loops `for {slug, per_channel} <- cursor_envelope` → `for {channel, cursor} <- per_channel` and calls `WindowCounts.snapshot/6` for each. Each `snapshot/6` (`window_counts.ex:119-133`) issues:

1. `Scrollback.count_after_split/5` — the content/events aggregate;
2. `count_mentions/6` → `Scrollback.unread_content_tail/6` — fetches up to `@mention_scan_cap` rows, then folds `Mentions.mentioned?/3` over them **in Elixir**.

So a subject with ~50 windows pays **~100 round trips** on every logon, plus ~50 regex folds. Each individual query is now fast (#393: 5-6ms channel, sub-ms DM) — the remaining cost is the **fan-out itself**.

## Goal

Collapse the fan-out to a **constant number of queries**, independent of window count. Target: **2 queries total** instead of 2N.

## Why this is expressible now

#393 rewrote `where_dm_peer/2` to a single sargable predicate, `nick_fold(COALESCE(dm_with, channel)) == ^folded_peer`. That unifies channel and DM windows under **one** join condition:

- channel window → `dm_with IS NULL` → `COALESCE` yields `channel`;
- DM window → `COALESCE` yields `dm_with`, and the cursor's `channel` **is** the peer nick.

So `nick_fold(COALESCE(m.dm_with, m.channel)) = nick_fold(rc.channel)` covers both window shapes in a single predicate — and it is served by the indexes already live on prod (`messages_{user,visitor}_id_network_id_dm_coalesce_fold_id_kind_index`). With the old `OR` form this query would not have planned.

## Shape

**Query 1 — all window counts in one grouped statement.** Drive from the cursors (either a join on `read_cursors`, or a `VALUES` list of `(channel, cursor)` pairs, since `build_unread_counts/2` already holds the envelope in memory), then:

```
… WHERE m.id > rc.last_read_message_id
GROUP BY m.channel, <content-vs-events bucket>
```

Per group this is the same index seek + range walk the per-window query does today, so the total work is unchanged — it is the round trips and query planning that collapse.

[...]

Issue #50, opened 2026-06-09
cic: own /msg to a brand-new nick can vanish (refreshScrollback recovery defeated by own-send read-cursor advance)


vjt:
## Symptom

Sending `/msg <new-nick> <body>` to a nick with **no existing query window** can leave the freshly-opened query window showing **"no messages yet"** — your own outbound message never renders — until a page reload. Intermittent, timing/load-dependent.

Surfaced as a recurring e2e flake in `m6-cicchetto-to-priv-opens-query.spec.ts` (and originally `cp15-b6-archive-query-revival.spec.ts`): under full-suite load on the Raspberry Pi dev box, m6 fails at the own-row render assertion (~17s, row absent) even on a settled host. `assertMessagePersisted` passes first, so the message **is** persisted server-side — cic just never displays it. Faster CI (ubuntu) usually wins the race, so this is mostly invisible there.

## Root cause (verified, not a timeout)

Three delivery paths for the own outbound row, all of which can miss it for a brand-new window:

1. **`loadInitialScrollback`** (`scrollback.ts:146-158`) is fired by `setSelectedChannel` in the `/msg` compose handler **before** the message POST (`compose.ts:327` then `:328`). It ships before the row persists → returns empty → marks the channel load-once (won't re-fetch).
2. **Live WS append** needs the channel-topic subscription, which is **reactive** (the `subscribe.ts` query-windows loop joins after the window appears in `queryWindowsByNetwork`). If the server broadcasts the row before that join completes, Phoenix drops it — no replay to late subscribers.
3. **`refreshScrollback`** (the CP29 R-5 recovery, on join-ok, `scrollback.ts:316-354`) is *supposed* to backfill — but it is **defeated by a cursor-poisoning ordering bug**:
   - `sendMessage` advances the **server read cursor to the just-sent message's id** immediately after the POST (`scrollback.ts:261-264`), optimistically, before the row is rendered.
   - `refreshScrollback` resolves its resume cursor via `getResumeCursor` (`scrollback.ts:321`). Because **no row was ever rendered**, `lastSeenIdByKey` is empty, so `getResumeCursor` **falls back to the read cursor** = the message's own id (`reconnectBackfill.ts:71-75`).
   - It fetches `?after=<messageId>` → **empty** → the row is never recovered.

So when path 2 loses the subscription race, paths 1 and 3 both miss the row and it is **unrecoverable** without a reload.

[...]

PR #583, opened 2026-07-31
fix(#576): exclude own-authored content from the unread count

vjt:
Refs #576. (No auto-close keyword — merge ≠ ship; vjt closes the issue post-deploy.)

## Problem
The cic unread badge counted the operator's OWN content rows. Nothing filtered
the count by sender, so a line you typed stayed "unread" until some later event
moved the read cursor past it. The optimistic send-time cursor advance
(`cicchetto/scrollback.ts`) normally masks this, but it has gaps — the #50
empty-pane gate skips the advance for a `/msg` window written into immediately,
and a cross-device `read_cursor_set` (last-write-wins, no forward-only gate) can
pull the local cursor BELOW your own last line. Both leave a DM/channel badge
made only of your own lines, with no peer line after them, that the operator
cannot clear by reading. Reported on IRC.

## Fix — a row you authored is read BY DEFINITION
Exclude own-authored CONTENT from the count, the content-row twin of #532 A
(which already excludes own PRESENCE), on BOTH count doors ("one feature every
door"):

- **client** — `perChannelUnread` (`selection.ts`): a content row whose `sender`
  folds to the per-network own nick no longer increments `messages`.
- **server** — the cold-load seed twins that stripped own presence now strip own
  content too: `Scrollback.exclude_own_presence/2` → `exclude_own_authored/3`
  (`count_after_split/5`), `ReadCursor.own_presence_dynamic/1` →
  `own_authored_dynamic/2` (`bulk_unread_split/2`).

**NICK-FOLD MATCH:** `sender` stays RAW (display), only the compare folds
(`Identifier.nick_fold/1` / `canonical_target/1` server, `nickEquals` client) —
never a bare `downcase`/`==`.

## Self-window carve-out (#576 × #396)
The own-nick SELF window (`channel == own_nick`) is the ONE window where own
content is legitimate payload — a note-to-self (`/msg <ownnick>`) — which #396
deliberately counts. The **predicate** is restricted, not the test:
`exclude_own_authored/3` takes a `self_window?` flag and there strips own
PRESENCE ONLY (byte-for-byte the pre-#576 #532 A predicate); own content still
counts. `bulk_unread_split/2` spans all windows, so its per-row twin lives in the
JOIN on-clause (`nick_fold(rc.channel) != folded`). Own content is dropped in
peer/channel windows (the reported bug) and KEPT in the self-window (#396
unchanged, its asserts untouched). #532 A presence exclusion applies in BOTH.
The optimistic cursor advance STAYS (belt-and-braces).

## Tests
- server: `count_after_split` + `bulk_unread_split` (own content excluded in
  peer/channel + ASCII fold + `own_nick=nil` boundary + COUNTS in self-window),
  `window_counts` (channel own-msg excluded; #396 self-window counts unchanged).
- client: `selection.test.ts` (peer-window exclusion + ASCII fold + self-window
  carve-out — the client twin of the server unit tests), `subscribe.test.ts`
  (explicit peer sender).
- e2e: `issue576-own-msg-unread-badge.spec.ts` — send own lines, force the read
  cursor back below them (the cross-device regress), badge shows nothing; a peer
  reply → "1".

## Gates (local)
`scripts/check.sh` green (4766 tests / 0 failures, credo/dialyzer/sobelow/bats);
cic `bun run check` exit 0; cic vitest green; `--grep "#576"` e2e passed.
DESIGN_NOTES entry appended (2026-07-31). No wire change → no CLIENT_PROTOCOL bump.

### DIFF
--- a/cicchetto/e2e/tests/issue576-own-msg-unread-badge.spec.ts
+++ b/cicchetto/e2e/tests/issue576-own-msg-unread-badge.spec.ts
@@ -1,0 +1,145 @@
+// GH #576 — the unread badge counted the operator's OWN content rows.
+//
+// Nothing filtered the unread count by sender: a line you typed was
+// "unread" until some later event moved the read cursor past it. The
+// optimistic send-time cursor advance (scrollback.ts) normally masks it,
+// but that advance has gaps — and a server `read_cursor_set` arriving from
+// another session (last-write-wins, no forward-only gate) can pull the
+// local cursor BELOW your own last message, re-exposing it as unread with
+// no peer line after it. The operator sees a badge on a window whose last
+// lines are all theirs and cannot clear it by reading (F5 fixed it: the
+// server cursor was already ahead, the client copy had fallen behind).
+//
+// The fix excludes own-authored CONTENT from the count — the content-row
+// twin of #532 A (own PRESENCE), on BOTH count doors: the client
+// `perChannelUnread` memo (selection.ts) AND the server cold-load seed
+// (`Scrollback.exclude_own_authored/3` + `ReadCursor` twin). This spec
+// pins the CLIENT surface in a real browser (per feedback_ux_e2e_mandatory
+// — vitest can't exercise the live WS echo + cross-device cursor regress +
+// Solid render path the badge depends on); the server exclusion is pinned
+// by the Elixir unit tests (scrollback_test / read_cursor_test /
+// window_counts_test).
+//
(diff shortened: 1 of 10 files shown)


vjt:
Superseded: landed on main as 3821565f (rebased head). Shipped in the 2026-08-01 cold deploy; #576 closed. Closing to keep the stale pre-rebase head from being merged again.

Issue #585, opened 2026-07-31
flake: SourceAliasManagerTest races its own boot reconcile (Mox expect consumed by handle_continue)


vjt:
## Symptom

`test + lint + audit` went red on PR #584 with **1 failure out of 4777** — and #584 touches
`cicchetto/e2e/**` only, so the failure is unrelated to the PR. The same job was **green on `main`
at `3821565f`**, i.e. the same code passed minutes earlier.

```
1) test reconcile reconcile keeps a live-held alias even when the refcount table is empty
   (Grappa.Net.SourceAliasManagerTest)
   ** (exit) exited in: GenServer.call(Grappa.Net.SourceAliasManager, :reconcile, 5000)
       ** (EXIT) an exception was raised:
           ** (Mox.UnexpectedCallError) expected Grappa.Net.SourceAliasMock.list_aliases/1
              to be called once but it has been called 2 times
   test/grappa/net/source_alias_manager_test.exs:96: (test)
```

Job: https://github.com/vjt/grappa-irc/actions/runs/30650139894/job/91221092058

## Root cause — a setup race, not a flaky assertion

`Manager.init/1` returns `{:ok, state, {:continue, :reconcile}}`, and `handle_continue(:reconcile, …)`
calls `adapter.list_aliases/1` (`lib/grappa/net/source_alias_manager.ex:127,131,214`).

`start_supervised!/1` returns as soon as `init/1` returns — **the boot reconcile then runs concurrently
with the test process.** The test's ordering assumption is:

1. `stub(list_aliases, …{:ok, []})` — meant to absorb the boot reconcile
2. `start_supervised!(...)`
3. `expect(list_aliases, …{:ok, [@addr, @addr2]})` — meant for the explicit reconcile
4. `Manager.reconcile()`

If the boot reconcile's call lands **after** step 3 (which is exactly what CI load makes likely), it
consumes the **expect** instead of the stub, and the call at step 4 becomes the second call against a
`once` expectation. Nothing about the production code is wrong — the test simply has no barrier
between "manager started" and "expectations declared".

## Fix (setup determinism — do NOT relax the assertion)

Put a synchronisation point between `start_supervised!` and the `expect` declarations, so the boot
reconcile is provably done before any expectation exists. A single synchronous call is enough
(`handle_continue` is guaranteed to run before any queued message), e.g.:

```elixir
pid = start_supervised!({Manager, adapter: …, prefix: @prefix, held_source_fn: fn -> [@addr] end})
[...]

Issue #552, opened 2026-07-30
e2e flake: issue168-scroll-authority:165 @chromium — send-snap fails under full-gate load (3rd sighting)


vjt:
## What

`cicchetto/e2e/tests/issue168-scroll-authority.spec.ts:165:3` — *"issue #168 — send pins to bottom, never jumps to the unread marker › operator paged UP → send snaps back to the bottom (unconditional)"* — has now failed **three separate times** under a full `scripts/integration.sh` run, always at the **same file and line**, and has **never** reproduced on GitHub CI.

Filed per the standing rule: a **third** recurrence stops being "a flake we note and proceed past" and becomes something that must be tracked and fixed deterministically. The two earlier sightings were each individually written off; that is exactly the retry-mask pattern #268 was filed about.

## Sightings

1. A rebased local full gate — 1 red, 554 passed. Did **not** reproduce on GitHub CI. (Also surfaced as #485's old `idx99`.)
2. Second sighting, same spec, same line.
3. **2026-07-30**, `#535`'s ship gate, `chromium`: `559 tests using 1 worker` → **557 passed, 2 failed** in 20.9m. The other failure was #519 (`ux-2-mobile-archive @webkit`), already tracked.
   ```
   ✘  99 [chromium] › tests/issue168-scroll-authority.spec.ts:165:3 › issue #168 — send pins to bottom,
      never jumps to the unread marker › operator paged UP → send snaps back to the bottom (unconditional) (6.3s)
   ```

## Shape

Same signature as #519 and #531: **green in isolation, red under full-gate load with a single worker**. The failing assertion is the send-path snap-to-bottom; the theory that best fits the evidence is that the send-snap races something under load rather than being logically wrong.

Established during #535's triage, and worth recording because it narrows the search: the `#535` visibility-return change is **inert** for this spec — `issue168` never enters the visibility-return path (the new mode is reachable only via visibility-return with `atBottom=false`), and the spec was green on a run with byte-identical product code. So this is **not** a #535 regression; it predates that branch.

## What a fix must and must not do

- **Make the SETUP deterministic.** Find what the send-snap is racing under load and wait on it explicitly.
- 🔴 **Do NOT weaken the assertion to get green.** The `#535` bucket already proved the point in the other direction: three specs that looked wrong were right, and went green untouched once the actual bug was reverted.
- An isolated `--repeat-each` pass does **not** clear this. Only a full-suite green does — that is the whole reason a full-gate-only failure is dangerous.

[...]

Issue #531, opened 2026-07-28
e2e flake: ux-6-a-mobile-members-scroll:108 @webkit — overlay touch-action suspend fails under full-gate load (green 5/5 iso)


vjt:
`cicchetto/e2e/tests/ux-6-a-mobile-members-scroll.spec.ts:108` — *@webkit ux-6-a — html.overlay-open suspends root touch-action so gesture-escalation can't reach installScrollPin* failed at ~1.7s in the full `integration.sh` gate for #529 (HomePane-only change; no overlap with members-scroll or touch-action).

**Isolation result: 5/5 green** on a fresh stack ⇒ load-dependent, not a regression.

Same webkit-under-full-gate-load family as #519 (`ux-2-mobile-archive` detaching mid-click), but a **different spec**, so it needs its own registration — treating them as one masks whichever is actually flaking on a given run.

Registered so the next gate can distinguish it from a real red without spending an isolation run to find out. **Do not weaken the assertion**: the spec pins the touch-action suspension that keeps gesture-escalation from reaching `installScrollPin`, which is load-bearing on real iOS where Playwright webkit already under-represents the behaviour.

vjt:
Consolidated into #653 — "the test suite is not load-stable under the full gate". Triaged 2026-08-02: this report and seven others all describe the same signature (green in isolation, red under full-gate load), so they are being worked as one phenomenon rather than eight spec-level patches. The repro detail here stays readable and #653 links back to it; reopen if the load explanation turns out not to cover this one.

Issue #519, opened 2026-07-28
e2e flake: ux-2-mobile-archive @webkit — "open members sidebar" detaches mid-click under full-gate load (green 3/3 iso)


vjt:
## Summary

`cicchetto/e2e/tests/ux-2-mobile-archive.spec.ts:54` — `@webkit UX-2 — rail archive opens the grouped modal + delete drops entry` (project `webkit-iphone-15`) fails in the full `scripts/integration.sh` gate but passes **3/3 in isolation**. Same green-3/3-iso / red-full-gate profile as #507, but a **distinct break point** (see below). Never filed.

## Observed (full gate, 2026-07-28)

Test timeout of 30000ms on the members-sidebar **open** step:

```
locator.click: Test timeout of 30000ms exceeded.
Call log:
  - waiting for getByLabel(/open members sidebar/i).first()
    - locator resolved to <button ... aria-label="open members sidebar"
        class="topic-bar-hamburger shell-chrome-btn">☰</button>
  - attempting click action
    - waiting for element to be visible, enabled and stable
    - element was detached from the DOM, retrying
```

The hamburger button resolves, but is **detached from the DOM mid-click** and never re-stabilizes — a re-render race on the topic-bar chrome under full-suite load. WebKit-only (webkit-iphone-15).

## Isolation

3/3 green:

```
scripts/integration.sh --project webkit-iphone-15 --grep "UX-2" --repeat-each 3
```

Deterministic pass in isolation → not a logic bug in the spec's assertions; a full-gate ordering / re-render timing hazard.

## Distinct from #507 (same area, different break point)

#507 (`cp15-b6`, chromium/desktop) fails on the members-pane **content**: `.members-pane li` filtered by the nick is never *visible* ("element(s) not found", 5s) — the list under-populates (candidate root: #506 db-lock seed race).

This issue (ux-2, webkit-iphone-15/mobile) fails one step **earlier** and on a **different element**: the toggle button that *opens* the pane (`button.topic-bar-hamburger[aria-label="open members sidebar"]`) is detached from the DOM mid-click ("element was detached from the DOM, retrying", 30s) — a topic-bar re-render race, before the pane content is ever queried.

Same green-3/3-iso / red-full-gate profile; different surface, element, and mechanism.

## Related

- Sibling members-pane-area flake, distinct break point: #507.
- WebKit-iphone-15 flake precedent: #339.

## Likely direction

[...]

PR #586, opened 2026-07-31
test(#585): barrier the boot reconcile before Mox expects

vjt:
Refs #585

## Problem
`SourceAliasManagerTest` raced its own boot reconcile. `Manager.init/1`
returns `{:ok, state, {:continue, :reconcile}}`, so
`handle_continue(:reconcile, …)` — and its `adapter.list_aliases/1` boot
call — runs **async** after `start_supervised!` returns. Under CI load the
boot `list_aliases` could land **after** a test's `expect/3` and consume
it, turning the explicit `Manager.reconcile()` into a second call against a
`once` expectation → `Mox.UnexpectedCallError`. Green on `main` minutes
earlier, red on an unrelated (cic-only) PR — a load-sensitive **setup**
race, not a flaky assertion. Same family as #519 / #531 / #552.

## Fix (at the setup, not the assertion)
A private `drain_boot_reconcile/1` helper does one synchronous
`:sys.get_state/1`. A gen_server answers that system message only from its
receive loop, which it reaches strictly **after** `handle_continue`
finishes — so returning proves the boot reconcile is done before any
expectation exists. Applied after **every** `start_supervised!` in the file
(5 sites, not just the one that fell today), so no test in the class can
race and no future copy inherits the gap.

- **No production code touched. No assertion touched.** Diff is +29/-9 in
  the test file only.
- `arm_check` runs synchronously in `init/1`, so its pre-start `expect`
  never raced; the barrier only settles the async `list_aliases` reconcile.

## Evidence
- Scoped stress: `test.sh source_alias_manager_test.exs --repeat-until-failure 50`
  → 50/50, 8 tests each, 0 failures.
- Full gate `scripts/check.sh` → exit 0; **4777 tests, 0 failures** (8
  doctests, 50 properties) + bats 171 ok.
- Sync code review: clean, no findings (barrier soundness confirmed from
  the production `init` continue + `handle_continue`; all 5 sites covered;
  sibling `server_test.exs` confirmed correctly out of scope).

Merge ≠ ship: no `Closes` keyword; the issue is closed post-deploy.

### DIFF
--- a/test/grappa/net/source_alias_manager_test.exs
+++ b/test/grappa/net/source_alias_manager_test.exs
@@ -27,8 +27,20 @@ defmodule Grappa.Net.SourceAliasManagerTest do
     stub(Grappa.Net.SourceAliasMock, :arm_check, fn @prefix -> :ok end)
     stub(Grappa.Net.SourceAliasMock, :list_aliases, fn @prefix -> {:ok, []} end)
 
-    start_supervised!({Manager, adapter: Grappa.Net.SourceAliasMock, prefix: @prefix})
+    pid = start_supervised!({Manager, adapter: Grappa.Net.SourceAliasMock, prefix: @prefix})
+    drain_boot_reconcile(pid)
+    pid
   end
+
+  # Barrier between "manager started" and "expectations declared". `init/1`
+  # returns `{:continue, :reconcile}`, so `handle_continue(:reconcile, …)` — and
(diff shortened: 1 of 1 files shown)


vjt:
Superseded: landed on main as 8a7650d2 (rebased head). Shipped in the 2026-08-01 cold deploy; #585 closed. Closing to keep the stale pre-rebase head from being merged again.

Issue #591, opened 2026-07-31
cicchetto: implement /ctcp and /ping (send side) — both return "unknown command"


vjt:
Per channel request (#sniffo), maintainer priority call: P0.

## Ask

cicchetto has no `/ctcp` and no `/ping`. Typing either returns `unknown command: /ctcp` from the parser's fallback (`cicchetto/src/lib/slashCommands.ts:898`). Both are standard client commands in the irssi grammar cicchetto otherwise follows.

## Current state (verified against `origin/main`, not assumed)

**Missing from the DISPATCH table** in `cicchetto/src/lib/slashCommands.ts` — no `ctcp` key, no `ping` key. Present today: `me join part topic nick msg query quit disconnect connect away op deop voice devoice kick kb kill ban unban banlist invite umode mode who names list links lusers info version motd stats rehash whois whowas notify watch hilight dehilight highlight cs ns ms os hs rs alias unalias quote oper`.

**The send-side plumbing already exists.** `/me` sends an ACTION wrapped in CTCP framing (`\x01ACTION <text>\x01`) through `scrollback.sendMessage` — see `cicchetto/src/__tests__/compose.test.ts:350`. So this is generalising an existing seam from the single hardcoded ACTION verb to an arbitrary CTCP verb, not building a new path.

**The server side is already CTCP-aware.** `Grappa.IRC.CTCP` is the single source of truth for CTCP framing classification (`lib/grappa/irc/ctcp.ex`); the parser deliberately preserves `\x01` verbatim in trailing params so CTCP round-trips through the scrollback `body` column (`lib/grappa/irc/parser.ex:29-33`); `LineSplit` knows non-ACTION CTCP verbs are single-line by convention (`lib/grappa/irc/line_split.ex:82`). No server change is expected for the send path — worth confirming rather than assuming.

**`/quote` is not a usable workaround.** The escape hatch exists (`/quote <raw irc line>` → `GrappaChannel.handle_in("raw", ...)` → `Session.send_raw`), but a CTCP frame requires literal `\x01` bytes, which cannot be typed into the compose textarea. This is precisely why the dedicated command is needed.

## Scope

**`/ctcp <target> <VERB> [args]`**
- Wrap as `\x01VERB args\x01` and send as PRIVMSG to target. Verb uppercased per convention.
- Reuse the `/me` framing seam rather than adding a second CTCP writer — one place that builds `\x01` frames.
[...]

Issue #581, opened 2026-07-31
cicchetto+grappa: "recover my identity" — /recover alias + button on one action path, with a progress modal (split from #561 point 4)


vjt:
Split out of #561 point 4. #561 is `status:soon` — built and awaiting verify — so the recover affordance lands here rather than re-opening scope on an issue sitting at the ship gate. It depends on #561: without "stop persisting the observed nick", the button fights its own database.

## Channel ruling (2026-07-31, vjt)

> "si /recover e bottone, entrambi scatenano le stesse azioni. e mostrano una modale col progress"

Three things, in this order:

1. **One action path.** The sequence — `IDENTIFY` → `RECOVER`/`RELEASE` (verb picked from the numeric, see the product-answers comment) → `NICK <credential nick>` — is implemented **once**. `/recover` and the button are two entry points onto the same code; neither re-implements the steps.
2. **Both entry points.** The `/recover` alias (asked for in the #561 thread) and a button in the client. Same action, same outcome.
3. **A progress modal.** The sequence is multi-step and each step can fail on its own, so it is not a fire-and-forget: the modal shows which step is running, which succeeded, and where it stopped.

## Constraints carried over from #561

- **Do not parse NickServ notices.** Their text is per-network folklore. The signal grappa trusts is the `+r` user mode (`{:visitor_r_observed, _}`, `lib/grappa/session/event_router.ex`). Any "did it work?" check in the progress modal must be built on mode state, not on service message text.
- **Visitor subjects only** — the same scope boundary #561 draws. User-subject nicks are operator-driven.

## Progress states worth distinguishing

Because each step fails differently, and the modal is the only place the operator finds out:

- `IDENTIFY` sent, `+r` never observed → wrong password (the common case; this is the one that guested them in the first place).
- `+r` observed, `RECOVER`/`RELEASE` refused → the nick is not theirs, or services declined.
- `NICK` refused → **failure** (vjt's ruling, 2026-07-31 — this reverses the earlier "retry, not an error" wording). An empty retry never wins the nick, which is why the verb above is `RECOVER` rather than `GHOST`. The modal says it failed and stops there.
- All three fine → close on success, and the credential nick is now the one they identified with.

vjt:
## Product answers (vjt, #grappa 2026-07-31 21:38)

The three open questions on this issue are answered. Verbatim:

[...]

Issue #14, opened 2026-05-18
cicchetto: CTCP frames (incl. /me ACTION) surface as raw PRIVMSG text instead of being parsed


vjt:
**Reported by:** Mezmerize on #it-opers (2026-05-18)
**Screenshot:** https://litter.catbox.moe/0unt4z.jpg

## Behavior
All CTCP types except `VERSION` reach cicchetto as plain `PRIVMSG`-style lines from the sender, opening (or routing into) a normal query window.

Includes `/dcc *`, `/ctcp *`, and `ACTION` (i.e. `/me`):

- `/me prova` in a **query** renders as: `timestamp <nick> ACTION prova`
- `/me prova` in a **channel**: same — raw `ACTION prova` text.
- `/dcc` and `/ctcp` (other than `VERSION`) similarly leak into query windows as if they were normal privmsgs.

## Expected
- `ACTION` → standard action rendering (`* nick prova` / italic).
- Other CTCPs → either handled (VERSION/PING/TIME with auto-reply) or surfaced separately (status / dedicated CTCP indicator), not dumped into a normal query as plain text.

## Notes
Common root cause: the client isn't unwrapping `\x01<TYPE> ...\x01` framing before display. `VERSION` works, so the parsing path exists but isn't generalised.

vjt:
**Triage 2026-06-08 — still open.**

CTCP ACTION stripping landed in `eb4a324` (2026-05-06, M10 ACTION), but this was filed 2026-05-18 — *after* that commit — and still reports raw PRIVMSG text for `/me`. So the strip path is incomplete or a regression slipped past. Fix: re-check the `\x01ACTION` detection in the cic display layer (ScrollbackPane) against the actual wire bytes. The server preserves `\x01` (per CLAUDE.md, round-trip fidelity), so the parse must happen client-side and is likely missing a framing case.

— automated triage (claude, on vjt's account)

PR #600, opened 2026-08-01
cicchetto CTCP send side: /ctcp + /ping (#591)

vjt:
## What

cicchetto's CTCP **send** side — `/ctcp <target> <VERB> [args]` and `/ping <nick>` sugar. Completes #591 (the `/me` ACTION **receive** side shipped in #14; inbound CTCP PING **answer** shipped upstream in 86416a21/96bedfdd).

## Design (rulings)

- **Server SSOT (Option 2):** `Grappa.IRC.CTCP.verb_args/1` classifies a CTCP frame into typed FLAT meta (`ctcp_verb` + `ctcp_args`). cic consumes the TYPED meta and **never parses `\x01`**.
- **Self-echo:** own `/ctcp` + `/ping` render as `→ CTCP VERB args to <target>`, never the raw `\x01` envelope (pre-#591 that leaked as a bare chat row).
- **RTT synthesised LOCALLY** in the `/ping` source window (irssi behaviour), decoupled from routing. A peer's CTCP PING **reply** NOTICE routes to `$server` (CTCP-framed — a reply is protocol, not a DM: 86416a21/96bedfdd); cic is always subscribed to `$server`, so the correlation gate consumes it with no query-window subscribe race.
- **Consume-on-match only:** an unmatched reply is left to normal routing (additive).

## Commits (6)

1. `feat(#591)` classify CTCP frames server-side into typed meta (`CTCP.verb_args/1` + `strip_trailing_delim/1` coexist with main's `framed?/1`; event_router `ctcp_meta/1` + NOTICE arm; `ctcp_self_echo_meta/1`).
2. `feat(#591)` cic send — `/ctcp`+`/ping` slash commands, `ctcpFrame` compose, `pingCorrelation` (register/resolve, RTT pure, ascii-fold nick).
3. `feat(#591)` cic receive — `subscribe.ts` `maybeConsumePingReply` gate → RTT into source window, no beep, no raw `\x01`.
4. `feat(#591)` cic render — ScrollbackPane renders the `→ CTCP …` line off `meta.ctcp_verb`.
5. `refactor` route `ctcp_verb/1` + `ctcp_payload/1` through the `CTCP.verb_args/1` SSOT. **Separate, revertible-alone, zero-behaviour-change** commit (it touches just-landed inbound-PING code 86416a21/96bedfdd).
6. `test(#591)` e2e — `/ctcp` self-echo render + `/ping` RTT (2 tests, full-stack round-trips jsdom cannot see).

## Gate

- `scripts/check.sh` green (compile `--warnings-as-errors`, credo `--strict`, doctor, ExUnit, dialyzer, wire-type drift, bats).
- cic `bun run check` / `test` / `build` green.
- **e2e (`scripts/integration.sh --grep "#591"`): 2 passed.** The **full** `scripts/integration.sh` on this PR is the ship gate (scoped `--grep` proves only the new spec).

Rebased clean onto `main` (`d53eeb9b`); the intervening #581/shottino commits touch no CTCP routing.

Refs #591

### DIFF
--- a/cicchetto/e2e/fixtures/ircClient.ts
+++ b/cicchetto/e2e/fixtures/ircClient.ts
@@ -154,6 +154,38 @@ export class IrcPeer {
     this.client.raw(["NOTICE", target, body]);
   }
 
+  // #591 — answer an inbound CTCP PING by echoing its token straight back as a
+  // CTCP PING NOTICE, exactly as a real client (or shottino) does. Parsed off
+  // the RAW wire line rather than irc-framework's CTCP middleware, so it works
+  // regardless of the library's own CTCP handling; a `done` latch makes it
+  // one-shot, so a stray library auto-response can't double-fire from here.
+  // Call BEFORE the operator's `/ping` so the listener is armed when the query
+  // arrives. Grappa routes the reply to `$server` (CTCP-framed, 96bedfdd) and
+  // cic's correlation gate synthesises the RTT in the window `/ping` was typed.
+  answerCtcpPing(): void {
+    const delim = String.fromCharCode(1);
+    const marker = `:${delim}PING`;
+    let done = false;
+    this.client.on("raw", (event: { line: string; from_server: boolean }) => {
+      if (done || !event.from_server || event.line[0] !== ":") return;
+      const line = event.line;
+      // ":asker!user@host PRIVMSG <me> :\x01PING <token>\x01"
+      if (!line.includes(" PRIVMSG ")) return;
+      const at = line.indexOf(marker);
+      if (at < 0) return;
+      done = true;
+      const asker = line.slice(1, line.search(/[! ]/));
+      // Token = the bytes after ":\x01PING " up to the closing \x01 (optional),
+      // echoed verbatim — the whole protocol is that it returns byte for byte.
+      let rest = line.slice(at + marker.length);
+      if (rest.startsWith(" ")) rest = rest.slice(1);
+      const end = rest.indexOf(delim);
+      const token = end >= 0 ? rest.slice(0, end) : rest;
+      const body = token === "" ? `${delim}PING${delim}` : `${delim}PING ${token}${delim}`;
+      this.client.raw(["NOTICE", asker, body]);
+    });
+  }
+
   // Register a nick with NickServ. Used by P-0a e2es to put a peer
   // into +r (registered) state so a subsequent /whois returns 307
   // RPL_WHOISREGNICK. With EMAIL:1 (the config since GH #349 wired the
--- a/cicchetto/e2e/tests/issue591-ctcp-ping.spec.ts
+++ b/cicchetto/e2e/tests/issue591-ctcp-ping.spec.ts
@@ -1,0 +1,91 @@
+// Issue #591 — cicchetto's CTCP send side: `/ctcp` + `/ping`.
+//
+// Two full-stack round-trips jsdom cannot see (per feedback_ux_e2e_mandatory
+// + feedback_cicchetto_browser_smoke — the compose → REST → self-echo/reply →
+// WS → render path only exists against a real bahamut + a real peer):
+//
+//   1. /ctcp <target> <VERB> [args] — the operator's OWN outbound CTCP query
+//      self-echoes as a :privmsg carrying the raw `\x01VERB args\x01` body.
+//      The server tags it with typed meta (Grappa.IRC.CTCP.verb_args/1) and
+//      ScrollbackPane renders "→ CTCP VERB args to <target>" — NOT the raw
+//      \x01, which pre-#591 surfaced as a bare "<nick> VERB" chat row. cic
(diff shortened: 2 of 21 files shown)


vjt:
## #600 CI red on the /ping RTT e2e — falsified as a flake, hardened at the root

**Falsification (rebased vs pre-rebase, `--repeat-each 3`, chromium):**

| commit | label | result |
|---|---|---|
| `5927d5c9` | rebased (== what CI ran red) | **3/3 green** |
| `3fc985cf` | pre-rebase tip carrying the spec | **3/3 green** |

The CTCP-ping path is **byte-identical** across the rebase:
- `ctcp.ex` / `subscribe.ts` / `pingCorrelation.ts` → 0-diff
- `event_router.ex` diff = only #581 `strip_r_on_self_rename` (no CTCP/notice/$server lines)
- `compose.ts` diff = only #581 `/recover`

⇒ the rebase did **not** change `/ping` behaviour. The CI red is a **flake**, not a regression.

**Root cause:** the spec used a fixed peer nick `m591peer`. Under CI rerun + bahamut per-IP clone/flood, a leftover ghost (or a QUIT not yet propagated) makes `IrcPeer.connect` collide (433 `ERR_NICKNAMEINUSE`); irc-framework then registers under an alternate nick, but the fixture's `peer.nick` field stays the requested value (`connect()` never reconciles it with the `registered` event). So `/ping <peer.nick>` DMs a nick that isn't on the network → no CTCP echo → 15s timeout.

**Fix (`5ab2d5d7`, cic-only, e2e-spec-only):** per-run unique nick `m591-<uuid5>`, mirroring the entrenched sibling idiom (`k16peer` / `e2e263` → `crypto.randomUUID().slice(...)`). A fresh nick per run/retry can't collide with a ghost. The assert is unchanged (`/CTCP PING reply from <nick>: \d+ ms/`, kind `notice`, source window) — `peer.nick` now equals the unique nick and auto-follows. Peer readiness stays deterministic: `connect()` awaits `registered` and `answerCtcpPing()` arms synchronously before `/ping`.

**Verification:** `bun run check` green; full `issue591-ctcp-ping.spec.ts` **6/6 green** (both tests × 3, chromium) on the lane post-fix.

_Optional follow-up (out of scope, non-blocking): reconcile `peer.nick` from the `registered` payload inside `IrcPeer.connect` — that would harden the whole suite against alternate-nick divergence in one place, not just this spec._

Refs #600
Refs #591

vjt:
Closing and reopening to force a CI run: the push of 5ab2d5d7 (the flake fix) created no workflow run at all — zero check-runs for that SHA — even though its only changed file (cicchetto/e2e/tests/issue591-ctcp-ping.spec.ts) matches the cicchetto/e2e/** path filter on both workflows. No merge implied.

vjt:
## Correction: #600 is NOT a flake — real client-side ordering race (fixed in `346995f3`)

My earlier "flake" comment was **wrong**. `/ping` failed CI **deterministically** — two consecutive runs (`5927d5c9`, `7bc34109`), always the same spec — while passing in isolation, in a 3-test slice, and in a **local full-suite run (460 tests, same first-391 ordering as CI)**. Deterministic-on-CI + invisible-locally = a load-sensitive race, not a flake.

**Root cause** (`cicchetto/src/lib/compose.ts`, `/ping` handler):

```
await sendPrivmsg(...)   // scrollback.sendMessage → apiSendMessage: a REST POST
registerPing(...)        // pending correlation registered AFTER the await
```

The peer's CTCP PING reply returns over the **separate, already-open WebSocket**. On a slow/loaded CI runner the REST POST's ack resolves **after** the reply is dispatched on the WS: `maybeConsumePingReply → resolvePing` runs with **no pending entry yet** → returns null → drops the reply; `registerPing` then stores an orphan. RTT line never renders → 15s timeout. On a fast local box the POST resolves first, so registerPing wins → passes. The old comment's "the reply cannot precede the send" held only for the **wire**, not for JS event-loop interleaving during the `await`.

**Fix:** register the pending **before** `await sendPrivmsg`. `registerPing` is a synchronous `Map.set` with no `await` before it in `submit()`, so the entry is present before control can yield to the WS-dispatching loop — the reply always finds it. An orphaned entry (if the send throws) is inert (one-shot, read only by a matching reply, cleared on identity change).

**TDD:** new `compose.test.ts` case models the slow POST with a deferred send and asserts `registerPing` fired **while the send is still pending** — RED before (0 calls), GREEN after. Full cic gates green (`bun check` rc0, vitest **3745/3745**). e2e assert + 15s timeout **untouched**.

**Not the SSOT refactor** (`24c7ba2e`): that commit is `:privmsg`-arm only and cannot touch this path; falsified by code + experiment. No server change needed.

**recover-identity-real:193** failed only in the `7bc34109` run (passed on the prior branch CI, on main, and in the local full-suite) — a single-occurrence flake, likely collateral to the /ping 15s stall that run. This fix removes the stall; CI will confirm.

Refs #600
Refs #591

Issue #588, opened 2026-07-31
cicchetto: rail "actions" menu cannot scroll when its rows overflow the space above the launcher


vjt:
Reported by vjt on #grappa (2026-07-31 22:06):

> "se apro il menu 'actions' e i bottoncioni non entrano nello spazio verticale non posso scrollare"

Open the rail `actions` launcher; if the menu rows do not fit in the vertical space above the launcher, the overflowing rows sit **off-screen above the viewport top** and **nothing scrolls** — neither the menu nor the page. Those actions are unreachable.

## Not reproduced by me — triaged from source

I did not reproduce this in a browser. What follows is read off `origin/main`, and the mechanism is specific enough to be worth the issue as-is; if the cause below is wrong, the repro will say so.

## Mechanism

`.rail-actions-menu` (`cicchetto/src/themes/default.css`, the `#500` block) is:

```css
.rail-actions-menu {
  position: absolute;
  bottom: 100%;           /* opens UPWARD from the launcher */
  max-height: var(--viewport-height, 80vh);
  overflow-y: auto;
}
```

`--viewport-height` is written by `cicchetto/src/lib/viewportHeight.ts` as **the full visual-viewport height in px** (`vv.height`, e.g. `852px`).

So the cap is the height of the *whole viewport*, while the space the menu actually has is only what lies **above the launcher** — and the launcher is pinned at the *bottom* of the rail (`.rail-actions { margin-top: auto }`). The two quantities differ by roughly the launcher row plus its padding, and by much more once the visual viewport shrinks (mobile keyboard up).

The consequence is not a clipped scrollable box, it is the opposite:

1. The menu's own content is **shorter than `max-height`**, so it is not overflowing *itself* → `overflow-y: auto` produces **no scrollbar**. The scroll affordance is there in the CSS and never engages.
2. `.shell-members` is deliberately `overflow-y: visible` (same `#500` block: *"so the absolute launcher menu overlays unclipped"*), so no ancestor clips or scrolls it either.
3. The menu therefore grows straight past `y = 0` and off the top of the screen. The rows up there cannot be reached by any gesture.

The two decisions are individually correct and only bite in combination: unclipped overlay (so the menu can cover the nick list) + a cap measured against the wrong box (so the menu never learns it ran out of room).

## Why the sibling menu does not have this

[...]

PR #602, opened 2026-08-01
cicchetto: cap rail actions menu to space above its launcher

vjt:
## What

The rail **actions** menu opens UPWARD from a launcher pinned at the bottom of the rail (`.rail-actions-menu { position:absolute; bottom:100% }`), but its CSS capped `max-height: var(--viewport-height)` — the **whole viewport**, not the space that actually lies **above the launcher**. On a short viewport (a handful of rows already suffice; mobile keyboard-up makes it worse) the overflowing rows grew straight off the top of the screen: the menu was shorter than its own oversized cap, so the `overflow-y: auto` already present never engaged, and neither the menu nor the page scrolled. The topmost actions sat above `y=0`, unreachable by any gesture.

The sibling `UserContextMenu` never had this because it clamps through the pure `lib/menuPosition` seam. `RailActions` was pure-CSS placement and was the one menu never wired to it.

## Fix

Reuse the seam instead of a second, differently-buggy guess:

- `lib/menuPosition.ts`: add pure primitive `spaceAbove(anchorTop, gap) => Math.max(0, anchorTop - gap)` — px available above the anchor, **clamped at 0** so a launcher near `y=0` never yields a negative `max-height` (invalid CSS → ignored → the bug returns).
- `RailActions.tsx`: measure `rootRef` (the `.rail-actions` container the absolute menu anchors to) on open, and on `resize` / `visualViewport` resize (the mobile-keyboard case), capping the menu's inline `max-height` via `spaceAbove`. Listeners are registered only while open and removed on close + unmount.
- CSS `var(--viewport-height)` stays as the **pre-measure fallback**. Once the cap is right, the existing `overflow-y: auto` starts doing its job.

## Proof (TDD)

- **Pure unit** (`menuPosition.test.ts`): `spaceAbove` normal + clamp-at-0 boundaries (jsdom-safe, like `placeAxis`).
- **Real e2e** (`issue500-rail-launcher-overflow.spec.ts`, adjacent guarantee): short viewport, open the menu, assert the topmost row (`home`) sits inside the viewport — **RED at `y=-54` before the fix, green after**. Precondition asserts the menu genuinely overflows, guarding the tall-viewport false-pass the issue calls out.

## Gates

- `bun run check` (biome + tsc): clean.
- `bun run test` (vitest): 3726 passed.
- `scripts/integration.sh` full suite: **570 passed / 0 failed** (chromium + webkit-iphone-15). The new e2e is chromium-only (untagged; webkit project greps `@webkit`), so it adds exactly 1 test.

Refs #588

### DIFF
--- a/cicchetto/e2e/tests/issue500-rail-launcher-overflow.spec.ts
+++ b/cicchetto/e2e/tests/issue500-rail-launcher-overflow.spec.ts
@@ -92,3 +92,57 @@ test("#500 — an overflowing member list keeps the rail launcher reachable and 
   await expect(page.locator(".rail-actions-menu")).toBeVisible();
   await expect(page.locator(".rail-actions-menu [data-testid='action-cluster-cog']")).toBeVisible();
 });
+
+// #588 — sibling guarantee to #500. #500 keeps the LAUNCHER reachable; this
+// covers what the launcher OPENS. The menu opens UPWARD from the bottom-pinned
+// launcher, but pre-fix it capped `max-height` at the WHOLE viewport height,
+// not the space that actually lies above the launcher. On a short viewport the
+// overflowing rows therefore grew straight off the top of the screen with NO
+// scroll (the menu was shorter than its own oversized cap, so `overflow-y:
+// auto` never engaged) — the topmost actions unreachable by any gesture.
+//
+// The trap the issue calls out: a menu that fits (tall viewport, few rows)
+// false-passes. So shrink the viewport until the menu GENUINELY overflows the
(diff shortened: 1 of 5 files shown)


vjt:
Già su main come ad284c5f (stesso patch-id, byte per byte): mergiata via rebase, quindi il commit atterrato ha uno sha diverso e questa PR resta aperta con la head pre-rebase. Chiudo per evitare di rishippare lo stesso lavoro. Il contenuto è nel batch 0.9.0.

Issue #589, opened 2026-07-31
iOS: no text selectable anywhere in the admin pane (keyboard up or down)


vjt:
On iOS, no text inside the admin pane can be selected — not with the keyboard up, not with it down. Reported by @vjt (2026-07-31). The admin tables are exactly the surfaces whose text you want to copy out (IPs, session/visitor ids, vhosts, UA strings, error text), so today the only way off the device is retyping by hand.

**Not reproduced on device by me** — triage below is read from `origin/main` sources, not from a live iPhone. The keyboard-down half is certain (it is pure CSS, no JS involved); the keyboard-up half depends on whether compose keeps focus behind the pane, see "Open question".

## Cause (keyboard DOWN — certain)

`themes/default.css` applies a blanket selection kill on iOS:

```css
html.is-ios {
  position: fixed;
  inset: 0;
  -webkit-user-select: none;
  -webkit-touch-callout: none;
}
```

`user-select` inherits, so it lands on every descendant. The policy is deliberate (Telegram Web K's pattern) and is paired with an explicit re-enable allowlist right below it:

```css
html.is-ios .scrollback,
html.is-ios .topic-modal-text,
html.is-ios input,
html.is-ios textarea { -webkit-user-select: text; user-select: text; }
```

plus `html.is-ios select` (#508). **The admin pane is not in that list.** `.admin-pane` / `.admin-tab-panel` / all the `.admin-*-table` cells inherit `none`, so there is nothing to select. Form fields inside admin already work (they are `input`/`textarea`), which matches the report: the *fields* are fine, the *tables and labels* are dead.

This half is unconditional — it does not care about the keyboard — which is why it fails "mai".

## Cause (keyboard UP — conditional)

`lib/keepKeyboard.ts` `handleMouseDown` bails immediately unless compose is focused:

```ts
if (!isTextEntry(document.activeElement)) return;
```

So with the keyboard down the JS contributes nothing (confirming the CSS is the whole story there). With compose focused, any target that is **not** in `SELECTABLE_TEXT_SURFACES` (`".scrollback, .topic-modal-text"`) falls through to the generic-chrome `e.preventDefault()` at the end of the handler, which cancels the press that would start a selection. Admin is not in that list either.

The CSS comment already states the invariant: the two lists must stay in sync, "a surface added here but not there stays unselectable while compose is focused".

## Direction

[...]

PR #603, opened 2026-08-01
cic(#589): make admin pane text selectable on iOS

vjt:
Refs #589.

## Problem
On iOS NO text inside the admin pane could be selected, keyboard up or down — yet the admin tables are exactly the surfaces an operator needs to copy off the device (IPs, session/visitor ids, vhosts, UA strings, error text). The only escape was retyping by hand.

## Root cause (pure CSS)
`html.is-ios` applies a blanket `-webkit-user-select: none` (Telegram Web K pattern) that inherits to every descendant, paired with an explicit re-enable allowlist (`.scrollback`, `.topic-modal-text`, `input`, `textarea`, `select`). The admin pane was never in that list, so `.admin-tab-panel` and every table cell inherited `none`. Admin form fields already worked (they are `input`/`textarea`) — matching the report: the fields were fine, the tables and labels dead.

## Fix
- Re-enable `.admin-tab-panel` — the single wrapper around every tab's content, so ONE selector covers all ten tabs.
- Re-exclude the controls inside it (`button`) so a long-press on a refresh/action button doesn't pop the selection magnifier over a control — same carve-out `.scrollback-invite-join` makes for the `[Join]` CTA. Buttons only: `input`/`select` stay selectable via the global re-enables, anchors keep their URL text copyable.

## Open question — resolved
The spec asked whether compose can keep focus behind the pane (which would need a `keepKeyboard.ts` `SELECTABLE_TEXT_SURFACES` change too). It cannot: the AdminPane `<Match>` arm **unmounts** ComposeBox (`Shell.tsx` `<Switch>`), so no text input holds focus behind the pane and `handleMouseDown` bails on `document.activeElement`. On real iOS a long-press synthesizes no mousedown anyway. **CSS-only; `keepKeyboard.ts` deliberately untouched.**

## Test
e2e (`@webkit` iPhone-15), not jsdom — jsdom is blind to CSS and has a no-op `Selection`, so a unit test passes against the broken CSS. Asserts the computed `user-select` on a real `.admin-visitors-table` cell is `text` and on a control inside the panel is `none`, under the active is-ios global kill (same posture as `text-selection-restored.spec.ts`).

## Verification
- `bun run check` (biome + tsc): green
- `bun run test` (vitest): 3722 passed
- e2e RED→GREEN (scoped, `--project webkit-iphone-15`): without the fix `panelUserSelect === "none"` → **FAIL**; with the fix → **1 passed**
- Full `scripts/integration.sh` suite: NOT run here (ship gate — deferred to reviewer/merge)

Real-device dogfood remains the final iOS check.

### DIFF
--- a/cicchetto/e2e/tests/issue589-ios-admin-select.spec.ts
+++ b/cicchetto/e2e/tests/issue589-ios-admin-select.spec.ts
@@ -1,0 +1,105 @@
+// #589 (2026-08-01) — on iOS NO text inside the admin pane was selectable,
+// keyboard up or down. The admin tables are exactly the surfaces an operator
+// needs to copy off the device (IPs, session/visitor ids, vhosts, UA strings,
+// error text), so the bug forced retyping by hand.
+//
+// ROOT CAUSE (pure CSS): `html.is-ios` applies a blanket
+// `-webkit-user-select: none` (Telegram Web K pattern, themes/default.css) that
+// inherits to every descendant, paired with an explicit re-enable allowlist
+// (.scrollback, .topic-modal-text, input, textarea, select). The admin pane was
+// NOT in that list, so `.admin-tab-panel` and every table cell inherited `none`.
+// Fix: re-enable `.admin-tab-panel` (one wrapper covers all ten tabs) and
+// re-exclude the controls inside it (`button`) so a long-press on a control
+// doesn't pop the selection magnifier — the same carve-out `.scrollback-invite-join`
+// makes for the [Join] CTA.
+//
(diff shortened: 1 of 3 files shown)


vjt:
Già su main come 95ffb09c: verificato per contenuto (spec issue589-ios-admin-select.spec.ts e regola CSS presenti su main; l'unica differenza è il terminatore di commento che #588 ha esteso). Head pre-rebase, quindi non è antenato di main. Chiudo per evitare il doppio ship. Il contenuto è nel batch 0.9.0.

Issue #606, opened 2026-08-01
cicchetto: query windows get a rail context — heading + whois card, fetched on select (the deferred half of #474)


vjt:
Requested by the maintainer on IRC (`#grappa`, 2026-08-01 10:24–10:34), as the follow-on to
#474's explicitly deferred half ("query → a /whois card is the deferred other half of this
design", `cicchetto/src/RailContext.tsx:11-13`).

Today `RailContext` has a single `<Match>` for `kind === "server"`. On a query window the rail
renders nothing at all. The ask: give query windows their own rail context — a heading and a
whois card, fetched on select.

## Scope

**1. Heading, mirroring the channel rail.** Channels render `<h3>members ({list().length})</h3>`
(`cicchetto/src/MembersPane.tsx:170`, uppercased by CSS). Query windows get the same heading
slot with `private conversation with <NICK>`. The nick must stay live: hook the existing
nick-change handling rather than snapshotting the name at open time, so a `NICK` while the query
is open re-labels the heading.

**2. Whois card in the rail, auto-fetched on select.** Reuse the existing `WhoisCard` presentation
rather than a second card component. Fetch policy:

* Fire the WHOIS when a query window is selected.
* **Per-nick TTL cache** — re-selecting the same query inside the TTL must NOT re-issue the WHOIS.
  Every selection being a round trip is how you get rate-limited by the ircd.
* **De-dupe in-flight requests** per nick, so fast window switching cannot stack them.

**3. Do not disturb the explicit `/whois`.** `/whois` today renders its own card
(`ScrollbackPane.tsx:2909`), not a scrollback line, and the maintainer wants that unchanged:
running `/whois` on an already-open query must still show that card even though the rail already
shows one.

This is the part that needs real care. `cicchetto/src/lib/whoisCard.ts` holds **one bundle per
network slug** (`whoisCardBySlug`, single-slot, `dismissWhoisCard` clears it) and the server pushes
`whois_bundle` on the **user** topic, landing on whatever window is active at arrival time. So the
rail cannot simply read/write that store: opening two queries in a row would clobber the card, and
an auto-fetch would forge a card the user never asked for. The rail needs **its own per-nick state**,
with the existing single-slot store left to the user-issued command.

## Out of scope (maintainer, same thread)

* **No message counters / conversation stats.** For the record on the cost, since it came up:
  `messages` carries `(user_id, network_id, channel, server_time)`
[...]

Issue #605, opened 2026-08-01
cicchetto: cap rail width so the server-window status card can't starve the message area on small desktop viewports


vjt:
Reported on IRC by morph (`#grappa`, 2026-08-01): on the desktop layout the network-window
status card eats too much horizontal space on small desktop-class screens (iPad Pro 11"),
squeezing the message area.

Maintainer scoping in-channel, same session: **do not add a hide/collapse control** —
"nasconderlo magari no, ma un max-width sì". So the ask is a width cap, not a new toggle.

## What was measured (`origin/main`)

* The status card is `ServerInfoCard`, mounted by `RailContext` when the selected window
  kind is `server` (`cicchetto/src/RailContext.tsx:39`). It renders inside `.shell-members`,
  i.e. the right rail — not inside the scrollback.
* On desktop that rail is a **permanent column** (#71 INC-2 / #473): `Shell.tsx:396-402`
  states `membersOpen` / `.open` have no desktop CSS and the TopicBar hamburger is
  CSS-hidden there. So today there is genuinely no way to get that width back.
* Grid: `.shell { grid-template-columns: var(--sidebar-width, 16rem) 1fr var(--members-width, 14rem) }`
  (`cicchetto/src/themes/default.css:955`). Default rails = 16rem + 14rem = **30rem ≈ 480px**.
* The rail *is* drag-resizable (`ResizeHandle side="right"`, `lib/sidebarWidths.ts`):
  min `160px`, max `50%` of `window.innerWidth`, persisted to `cicchetto.membersWidth`.
  The 50% clamp is **per side**, so left + right can jointly claim the whole viewport.
* The existing compact tier does not help here. It is gated
  `@media (orientation: landscape) and (max-height: 500px) and (min-width: 769px)`
  (`default.css:1050`), and the comment deliberately excludes iPad
  ("an iPad in landscape is 768px TALL, well above 500px … do NOT regress iPad landscape").
  An iPad Pro 11" is 1194x834, so it never enters that tier and always gets full-width rails:
  landscape leaves 714px of centre out of 1194; portrait leaves **354px out of 834**.

## Suggested direction (not prescriptive)

Cap the rail width against the viewport rather than adding a control — e.g. make the grid
track `min(var(--members-width, 14rem), <fraction>vw)`, and/or make `clampWidth` in
`lib/sidebarWidths.ts` enforce a **joint** budget across both rails instead of 50% each.
Either keeps the drag affordance and the persisted preference intact while making the centre
[...]

Issue #373, opened 2026-07-21
cicchetto: query window does not follow peer NICK change → keeps old nick, sends 401 (no such nick)


vjt:
## Summary

A query window does not follow the peer's NICK change. When the other party in an open query renames, the query window keeps showing the **old** nick, and outgoing messages keep routing to the old nick — which no longer exists — producing `no such nick/channel` (401).

## Repro (reported on #grappa, testnet)

1. Open a query with a peer (here: `Guest87449`).
2. The peer changes nick (`Guest87449` → `NickTemporaneo`).
3. The query window still shows `Guest87449`, and sending in that window routes to `Guest87449` → server replies `no such nick/channel`.

## Why it happens (source)

Query windows are keyed by `targetNick` and are server-backed (`cicchetto/src/lib/queryWindows.ts` — `openQueryWindowState` / `closeQueryWindowState` / `setQueryWindowsByNetwork`; the `Grappa.QueryWindows` row is unique on `(subject, network_id, …)`). There is **no rename-on-NICK path**: nothing migrates a query window's `targetNick` from old → new when a NICK is observed.

By contrast the members store already handles this — `cicchetto/src/lib/members.ts` renames a member on NICK (`nickEquals(m.nick, msg.sender) ? { ...m, nick: newNick } : m`). Query windows were simply not given the same treatment.

## Fix direction

- On observing a NICK change for a nick that has an open query window, migrate that window's `targetNick` old → new: the client signal (`queryWindowsByNetwork`) **and** the server-side `Grappa.QueryWindows` row, so routing follows the peer.
- Preserve the existing scrollback/history under the migrated window.
- Add an e2e: open a query, peer sends NICK, assert the window relabels to the new nick and a subsequent send routes correctly (no 401).

## Relation to #371 / #372

Distinct root cause. All three touch query-window identity: #371 = services allowlist (services shouldn't open query windows at all), #372 = case-sensitive incoming routing (casing splits windows), this one = query windows don't track NICK renames. Fix independently.

vjt:
Shipped in grappa 0.4.0 — cold-deployed to m42 prod 2026-07-25 (healthz 200, 33 sessions reconnected, migrations applied). Release: https://github.com/vjt/grappa-irc/releases/tag/v0.4.0

PR #613, opened 2026-08-01
feat(#606): query windows get a rail context — heading + WHOIS card

vjt:
Refs #606 — the deferred half of #474 (server-info rail).

## What

A query (DM) window's right rail now renders a context, mirroring `MembersPane` for channels:

1. **Heading** `private conversation with <NICK>` — reads live off `selectedChannel`, so a #373 peer-nick rename re-labels it (`followQueryNick` already swaps the selection).
2. **A persistent WHOIS card** for the partner, **auto-fetched on select**. `WhoisCard` is refactored **prop-driven** (`bundle` + optional `onDismiss`): the scrollback overlay keeps its `×` dismiss; the rail card omits `onDismiss` → no `×`, persistent, like `ServerInfoCard`.

## How — Option 2: server marks the WHOIS origin

Kills the `whois_pending` collapse race **by construction** rather than papering over it client-side.

- Add-only wire field `source: :user | :rail` on `whois_bundle` (atom-union → `"user"`/`"rail"`; **no `protocol_version` bump**, additive-only; absent → `"user"`).
- Flow: `pushWhois(…, origin)` (only the rail passes `"rail"`) → channel `handle_in("whois")` `normalize_whois_origin/1` (**untrusted**: only `"rail"` opts in, everything else → `:user`) → `Session.send_whois/5` primes `:source` in the per-target `whois_pending` accumulator → `Wire.whois_bundle/3` projects `Map.get(accum, :source, :user)`.
- **Consumer rule** (the collision point — `whois_pending` is keyed per nick; concurrent requests share one accumulator, last `:source` wins, since ircd numerics carry no request id): the single-slot store (`whoisCard.ts`) consumes **only** `source:user`; the per-nick rail cache (`railWhois.ts`) consumes `source:rail` **and** a `source:user` bundle for the nick the rail is currently showing (free refresh). A collapse becomes a cache hit, not a race — `/whois` always gets its card, the rail gets a refresh.
- Rail cache: per-nick TTL 60s + in-flight de-dupe; fetch is fire-and-forget (`.catch(()=>{})`, pending-TTL 30s covers retry), unlike compose's awaited `/whois`.

## #605 coupling

The query rail lives in the `fit-content(14rem)` `.shell-no-members` track (landed with #605). The cap binds because both the reused `.whois-card-fields dd` and the new `.rail-query-heading` `word-break` — the exact #605 gotcha.

## Out of scope

No message counters / stats (maintainer's call, per issue #606).

## Tests

- server: `wire_test`, `server_test`, `grappa_channel_test` (origin normalize + `:source` projection).
- cic: `railWhois.test.ts` (new), `whoisCard.test.ts` (new), `WhoisCard.test.tsx`, `RailContext.test.tsx`, `userTopic.test.ts`, `ScrollbackPane.test.tsx`, `Shell.test.tsx`.
- e2e: `issue606-query-rail-whois.spec.ts` (new; chromium; live stack).

## Gates (local, this branch @ c0e6e3ba)

- cic `bun run check` → exit 0 (biome + tsc).
- cic `bun run test` → **3775/3775** (212 files).
- Server `check.sh` verified green pre-rebase (tree byte-identical after rebase; full suite runs in PR CI).

### DIFF
--- a/cicchetto/e2e/tests/issue606-query-rail-whois.spec.ts
+++ b/cicchetto/e2e/tests/issue606-query-rail-whois.spec.ts
@@ -1,0 +1,102 @@
+// #606 — a query window gets its own rail context (the deferred half of
+// #474): a heading + a WHOIS card for the conversation partner, auto-fetched
+// on select. This e2e pins the LIVE end-to-end behaviour the vitest units
+// can't reach over the real socket + a real upstream WHOIS round-trip:
+//
+//   * selecting a query shows the heading `private conversation with <NICK>`
+//     AND a WHOIS card in the RAIL (NOT the scrollback overlay), auto-fetched;
+//   * the rail card is persistent — no × dismiss affordance (unlike the
+//     scrollback /whois card);
+//   * a peer NICK while the query is open re-labels the heading (#373 swaps
+//     selectedChannel in place, which the heading reads live);
+//   * an explicit `/whois` still renders its OWN card in the scrollback
+//     overlay, even though the rail already shows one — the two stores are
+//     disjoint by the server-marked `source` (#606 option 2), so `/whois` is
+//     undisturbed.
+//
+// The #605 rail-width cap on this same track is proven by
+// `issue605-rail-width-cap.spec.ts`; it binds here too because the reused
+// `.whois-card-fields dd` and the `.rail-query-heading` both `word-break`
+// (the coupling gotcha #605 flagged). Per `feedback_ux_e2e_mandatory` a
+// UX-behaviour change ships a Playwright e2e; the surface is subject-agnostic
+// (`feedback_e2e_user_class_parity_matrix`) so one user-class spec suffices.
+// Feature logic is engine-agnostic → chromium only (no `@webkit`), mirroring
+// nick-follow-query.spec.ts.
+
+import {
+  composeSend,
+  loginAs,
+  selectChannel,
+  waitForQueryWindowReady,
+} from "../fixtures/cicchettoPage";
+import { IrcPeer } from "../fixtures/ircClient";
+import { AUTOJOIN_CHANNELS, getSeededVjt, NETWORK_NICK, NETWORK_SLUG } from "../fixtures/seedData";
+import { expect, test } from "../fixtures/test";
+
+// Unique suffixes so retries / sibling specs don't collide on a nick already
+// in use upstream (same rule as nick-follow-query.spec.ts).
+const RUN_ID = crypto.randomUUID().slice(0, 8);
+const PEER = `Rail${RUN_ID}`;
+const PEER_RENAMED = `Rail2${RUN_ID}`;
+const CHANNEL = AUTOJOIN_CHANNELS[0];
+
+test("query rail shows heading + auto-fetched WHOIS card, follows NICK, coexists with /whois", async ({
+  page,
+}) => {
+  const vjt = getSeededVjt();
+  await loginAs(page, vjt);
+  await selectChannel(page, NETWORK_SLUG, CHANNEL, { ownNick: NETWORK_NICK });
+
+  const peer = await IrcPeer.connect({ nick: PEER });
+  try {
+    // Share a channel so the peer is reachable AND grappa observes a later
+    // NICK (IRC only relays a NICK to channel-sharing users).
+    await peer.join(CHANNEL);
+
+    // Open + focus the query — this is what fires the rail's fetch-on-select.
+    await composeSend(page, `/q ${PEER}`);
+    await waitForQueryWindowReady(page, NETWORK_SLUG, PEER);
+
+    // Heading in the RAIL (scope to `.shell-members` so a regression that
+    // renders it elsewhere fails here).
+    const rail = page.locator(".shell-members");
+    const ctx = rail.getByTestId("rail-query-context");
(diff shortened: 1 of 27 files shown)


vjt:
Rebased onto `main` on vjt's authorization (was 214 commits behind, 1 ahead): `13ebd577` → `065867cd`, force-pushed with `--force-with-lease`. Old tip kept locally as `backup/613-pre-rebase-13ebd577`.

**One conflict, purely structural** — `lib/grappa_web/channels/grappa_channel.ex`: `main` has since moved the verb clauses behind a dispatcher (`def handle_in/3` at :400 → `defp do_handle_in/3`), while this branch changed the two `"whois"` clauses to bind `= params` and derive `origin = normalize_whois_origin(params)`. Resolution keeps `main`'s `defp do_handle_in` shape and layers this branch's two additions on top — no behavior taken from either side, no clause dropped. `normalize_whois_origin/1` (:1575) is untouched and both call sites resolve.

Everything else auto-merged clean (24 files).

**Not verified locally**: this host is aarch64 with no elixir/bun toolchain, so neither `mix` nor `bun run check` ran here — CI is the gate, deliberately not claiming a green I didn't observe.

Note on the prior failure: the e2e red on the old base was `nick-follow-query.spec.ts:48` (`IrcPeer: timeout waiting for privmsg ... (5000ms)`), reproduced 2/2 — but on a base 214 commits stale, so that verdict is void. Re-judging on this run.

vjt:
Red on `065867cd`, two checks, **two unrelated causes** — one mine, one not.

**1. `cicchetto (types + lint + unit)` — my rebase, fixed in `4a969e76`.** The merge was textually clean and semantically wrong: the `#673 extra_lines` describe in `WhoisCard.test.tsx` came from `main`, written against the store-driven card (`setWhoisBundle` / `dismissWhoisCard` / `<WhoisCard networkSlug=…>`), while this branch converted `WhoisCard` to take `bundle` + `onDismiss` as props. Neither side touched the same lines, so git had nothing to flag — `tsc` did: TS2304 on both helpers, TS2322 on `networkSlug`. Ported the six specs onto `renderCard` (the helper every sibling describe here already uses) and dropped the `afterEach` store teardown, which has no slot to dismiss once the bundle is a prop. Assertions unchanged. The store is untouched and still covered by `whoisCard.test.ts`.

Worth noting for the reviewer: this class of breakage is exactly what a stale branch buys you, and it is invisible to `git` — only the typecheck catches it. It is also the second time today that #716's unconditional JS job earned its keep.

**2. `cicchetto + grappa + azzurra-testnet` — not this diff.** The run died 3s into container startup: `Container azzurra-hub Started` at 14:17:53 → `Container azzurra-hub Error dependency hub failed to start` at 14:17:56. No Playwright test ever ran, so there is no report or trace. Nothing on this branch touches the testnet compose or the hub image. Re-running now on `4a969e76`; if it reproduces it is a testnet/infra issue and gets its own issue, not a fix here.

Still not claiming a local green — no elixir/bun toolchain on this host, CI remains the only judge.

vjt:
CI on `4a969e76`: the JS job is now green (the rebase breakage is fixed). One check remains red, and it is a **real regression in this PR** — not infra, not a stale base.

**Correction to what I said earlier.** I wrote that the pre-rebase red was "void" because the base was 214 commits stale. That was wrong, and the correction matters because it flips the conclusion: the same spec fails on the current base too, so the old signal was informative all along.

**Evidence:**

| | |
|---|---|
| Failing spec | `nick-follow-query.spec.ts:49` — *query window follows a peer NICK change — relabels, keeps history, routes with no 401* |
| Result on this PR | `1 failed, 590 passed` (24.6m) — red 3/3 now, across two different bases |
| Result on `main` | run `30822009592` @ `c1a0e535` — **success** |
| Does this PR touch the spec? | No — `git diff origin/main...4a969e76 -- cicchetto/e2e/tests/nick-follow-query.spec.ts` is empty |

Main green + spec untouched + reproducible here ⇒ the change set causes it.

**Where it fails.** Not on the relabel and not on history — the sidebar swap and the scrollback assertions all pass. It dies at STEP 4:

```
Error: IrcPeer: timeout waiting for privmsg from vjt-grappa containing "#373 followup f995933e" (5000ms)
```

That is the send **after** the rename never reaching the renamed peer. The spec's own comment names the failure mode it was written to lock: *"Pre-fix it routed to the vanished old nick → 401 and never arrived."* So this is the #373 fix coming undone — the window relabels correctly but the outbound still targets the old nick. Note the spec gates on `waitForQueryWindowReady(page, …, NEW_NICK)` before sending, so a re-subscribe race is already excluded by construction; the assertion that fails is routing, not rendering.

**Where I would look** (pointer, not a diagnosis I have confirmed): the rename → re-target path, i.e. this PR's changes to the query-window handling in `cicchetto/src/lib/userTopic.ts` and the server-side `session.ex` / `server.ex` / `wire.ex` / `grappa_channel.ex` edits. The `socket.ts` change is only the `WhoisOrigin` add-only field and is very unlikely to be involved.

**I am not writing the fix** — flagging and handing over, per the standing split. The hub failure from the previous run did **not** recur (`azzurra-hub Healthy`), so that one was transient and needs no issue.

Artifacts for whoever picks this up: the run has `trace.zip`, a screenshot, and `error-context.md` under `test-results/nick-follow-query-query-wi-a12a8--history-routes-with-no-401-chromium/`.

Issue #592, opened 2026-07-31
testnet.sh submodule auto-init always fails in a worktree (missing protocol.file.allow)


vjt:
## What

`scripts/testnet.sh` auto-inits the `cicchetto/e2e/infra` submodule when a fresh
worktree lands without it (testnet.sh:41-48). The intent is stated in the comment:

> Auto-init instead of dying on a manual step everyone forgets — idempotent and a no-op once present.

**In a worktree that auto-init can never succeed.** The superproject is reached over a
local path (`/…/grappa/.git/modules/cicchetto/e2e/infra`), and git has refused the
`file://` transport for submodule clones since the CVE-2022-39253 mitigation, so the
clone dies with `fatal: transport 'file' not allowed`, retries once, and `die`s:

```
testnet: azzurra-testnet submodule empty (fresh worktree?) — initialising…
Cloning into '…/.worktrees/581-recover-identity/cicchetto/e2e/infra'...
fatal: transport 'file' not allowed
fatal: clone of '…/grappa/.git/modules/cicchetto/e2e/infra' into submodule path '…' failed
Failed to clone 'cicchetto/e2e/infra'. Retry scheduled
… second attempt, same failure …
Failed to clone 'cicchetto/e2e/infra' a second time, aborting
submodule auto-init failed — run: git -C '…' submodule update --init cicchetto/e2e/infra
```

The manual heal everyone ends up running is the same command **plus** the allow flag:

```sh
git -C "$SRC_ROOT" -c protocol.file.allow=always submodule update --init cicchetto/e2e/infra
```

## Why it matters

Every e2e run started from a fresh worktree hits this, and the auto-init that exists
precisely to spare the manual step does not spare it — it just fails more verbosely.
This has now recurred across several sessions.

## Note on the exit code (NOT a bug — recorded so nobody re-chases it)

This surfaced as an `integration.sh` run that "exited 0" while nothing ran. That was a
**reporting artifact of the caller**, not a masked failure:

- `die` exits 1, `integration.sh` runs `set -euo pipefail`, and a bash `EXIT` trap that
  does not call `exit` does NOT override the exit status (verified on bash 5.2:
  a trap whose last command succeeds still leaves `$?` = 1).
- The observed 0 came from the invoking command ending in `…; echo "SPIKE_EXIT=$?"` —
  the reported status was the `echo`'s.

`integration.sh`'s propagation is fine. The second copy of the failure in the log is
[...]

Issue #277, opened 2026-07-16
e2e flake: testnet nick-release race — issue195 spec 500s in _vjtReset beforeEach (nick_rejected 433)


vjt:
## What
CI integration run [29539788449](https://github.com/vjt/grappa-irc/actions/runs/29539788449) (on batch tip `a97569ac`) went **1 failed / 393 passed**. The single failure was NOT in any product spec:

- **`issue195-leave-confirm-modal.spec.ts:71`** failed inside the **`_vjtReset.auto` beforeEach** (`fixtures/test.ts:119` → `resetSubject`), with error **500 `{:client_exit, {:nick_rejected, 433, "vjt-grappa"}}`** on `bahamut-test`.
- The **next test in the same file passed immediately** (nick had been freed by then).

## Classification: CASCADE (not a defect, not a regression)
- **iso `integration.sh --grep "#195" --repeat-each 5` = 10/10 passed** → TESTING.md-definitive **cascade**, not a spec defect.
- Local FULL `integration.sh` = **394/0/0** on the identical tree.
- All 3 batch specs (#257/#264/#234) were GREEN in the CI run itself.

## Root cause (hypothesis — verify with data)
A **testnet nick-release / ghost race**: an earlier spec's `vjt-grappa` session isn't fully released upstream (bahamut testnet) before `_vjtReset`'s beforeEach re-registers, so the re-register hits **433 nick in use** and the reset 500s. This is the **chronic testnet flake tail** the #268 campaign was systematically greening — same class.

## Fix direction
Make `resetSubject` robust to a not-yet-released nick: wait-for-nick-free / retry-on-433 with a bounded backoff in the reset path (or force-release the prior session deterministically before re-register). Do NOT paper over by rerunning CI — first run is truth.

_Filed by orchestrator: batch #257/#264/#234 shipped healthy despite this CI red because it was proven unrelated; tracking the flake here so it isn't buried._

vjt:
Another surface of this nick-release race: on the #281 integration CI run, `bughunt-1-b-archive-mobile-seed` failed in its `resetSubject(vjt)` setup with `500 session_reconnect_failed :nick_rejected 433 vjt-grappa` (bahamut still holding `vjt-grappa` when the per-spec reset reconnected). Iso 5/5 green, so a transient testnet reconnect flake — logging it here as another data point for the #277 nick-release-race family.

vjt:
[...]

PR #616, opened 2026-08-01
fix(#592): pass protocol.file.allow=always on submodule auto-init

vjt:
## What

The git-submodule auto-init in **both** `scripts/testnet.sh` and `scripts/bats.sh` now passes `-c protocol.file.allow=always`.

Refs #592.

## Why

In a git worktree, git clones the submodule from the superproject's **local** module store (`$REPO/.git/modules/…`) over the `file://` transport — which the CVE-2022-39253 mitigation blocks by default. So each auto-init died with `fatal: transport 'file' not allowed`, retried once, and `die`d. The auto-init that exists precisely to spare the manual `git submodule update --init` never spared it — it just failed more verbosely, and everyone ended up hand-running the same command **plus** the flag. The flag is **required, not cosmetic**: without it every fresh worktree's first e2e / bats run fails.

## It's a class, not a case

`grep -rn "submodule update" scripts/` surfaced **two** sites with the identical pattern (bats.sh's comment even says it "mirrors the testnet.sh submodule auto-init pattern"). Both are fixed in one pass:

- `scripts/testnet.sh:45` — `cicchetto/e2e/infra`
- `scripts/bats.sh:31` — `vendor/bats-core`

Each `die` message's manual-heal command gets the flag too — the fallback advice must not reproduce the bug it's recovering from.

## Test — TDD, asserts on the constructed command

New bats spec `test/scripts/submodule_init_flag_test.bats` (2 tests, one per script). Reproducing the real `file://` clone failure under bats is impractical (it needs a populated superproject module store hit under the exact CVE conditions), so — per the issue's explicit steer — it **asserts on the constructed command** instead of faking a clone: a `git` stub on PATH records every invocation, forwards everything except `submodule update` to real git, and the test runs each script from a **real** throwaway worktree and asserts the recorded `submodule update` line carries `-c protocol.file.allow=always`. An honest, tight assert beats a test that pretends to clone.

- **RED** (verified by reverting the two scripts): both tests fail exactly at the flag grep.
- **GREEN** (with the fix): both pass.

## Gates

- **bats**: full suite green — `171 → 173` tests, exit 0 (the +2 are this spec; no regressions).
- **shellcheck**: `shellcheck -x scripts/testnet.sh scripts/bats.sh` clean. Both files gained the `# shellcheck source=scripts/_lib.sh` directive (the `deploy.sh` convention) so shellcheck follows the include, sees `set -euo pipefail`, and lints them clean — neither file was in CI's shellcheck list before.

## Scope

Fix + directive + test only. Did **not** take the issue's *optional* "move the submodule check below the verb dispatch so `down` doesn't require it" — that touches testnet flow and the orchestrator scoped this to the flag + the class hunt.

### DIFF
--- a/scripts/bats.sh
+++ b/scripts/bats.sh
@@ -14,6 +14,7 @@
 # container is only invoked transitively when a test exercises a verb
 # that shells out to docker (those tests stub `docker` via PATH).
 
+# shellcheck source=scripts/_lib.sh
 . "$(dirname "$0")/_lib.sh"
 
 cd "$SRC_ROOT"
@@ -27,9 +28,14 @@ if [ ! -x "$bats_bin" ]; then
     # tree). Init it rather than making the operator run the incantation
     # by hand — mirrors the testnet.sh submodule auto-init pattern so
     # check.sh works first-try from any worktree.
+    # `-c protocol.file.allow=always` is REQUIRED, not cosmetic (#592): in a
+    # worktree git clones the submodule from the superproject's LOCAL module
+    # store over the file:// transport, which the CVE-2022-39253 mitigation
+    # blocks by default — without it a fresh worktree dies with
+    # `fatal: transport 'file' not allowed`.
     printf 'scripts/bats.sh: vendor/bats-core missing — initialising submodule...\n' >&2
-    git -C "$SRC_ROOT" submodule update --init vendor/bats-core >&2 \
-        || die "vendor/bats-core init failed. Run: git -C \"$SRC_ROOT\" submodule update --init vendor/bats-core"
+    git -C "$SRC_ROOT" -c protocol.file.allow=always submodule update --init vendor/bats-core >&2 \
+        || die "vendor/bats-core init failed. Run: git -C \"$SRC_ROOT\" -c protocol.file.allow=always submodule update --init vendor/bats-core"
 fi
 
 if [ ! -x "$bats_bin" ]; then
--- a/scripts/testnet.sh
+++ b/scripts/testnet.sh
@@ -23,6 +23,7 @@
 
 set -euo pipefail
 
+# shellcheck source=scripts/_lib.sh
 . "$(dirname "$0")/_lib.sh"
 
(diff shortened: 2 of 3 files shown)


vjt:
## Integration red triaged — **none of the three failures is this PR's**

Run [30706528904](https://github.com/vjt/grappa-irc/actions/runs/30706528904): **3 failed / 572 passed**, both projects ran (chromium + webkit-iphone-15).

| # | spec | signature | verdict |
|---|---|---|---|
| 1 | `issue605-rail-width-cap.spec.ts:120` (chromium) | `route.fetch: Target page, context or browser has been closed` — the intercepted `GET /networks` returned **200**, no assertion failed | **route handler outliving the test** → filed as #619 |
| 2 | `push-trigger-channel-mention.spec.ts:55` | `resetSubject(vjt) failed: 500 {:client_exit, {:nick_rejected, 433, "vjt-grappa"}}` at `fixtures/test.ts:120` | **#277** |
| 3 | `recover-identity-real.spec.ts:193` | same `_vjtReset` 433 signature | **#277** (the persistent one this week) |

2 and 3 are keyed off the **error signature + the `_vjtReset.auto` fixture frame**, not the spec name — an auto-fixture's race surfaces in whatever spec happens to be running, so the spec name is never the discriminator.

1 is a genuine defect but a **latent, suite-wide** one, not a #592 regression: 13 of the 14 specs that call `page.route(` never unroute, so a late `connection_state_changed`-driven refetch can still be in the callback when the context closes. `issue605-rail-width-cap`'s `@webkit` twin **passed in this very run**, and the spec is green on `main` at `d536ecba`. Full analysis + the one-seam fix in #619.

**This branch touches `testnet.sh` + `bats.sh` submodule init only** — there is no causal path from that to any of the three. Nothing to fix here.

**No rerun now, deliberately.** This PR is held behind the `v0.9.0` tag and will have to rebase past it anyway, so a green at this base proves nothing at merge time. Re-gate after the rebase.

Issue #503, opened 2026-07-27
Docker becomes a first-class deploy substrate: one idempotent infra/docker/deploy.sh, shared hot/cold with linux+freebsd, ghcr + curl|bash install/update


vjt:
> **Consolidated issue.** Absorbs #51 (shared deploy decision lib + docker parity) and #439 (grappa image on ghcr.io), both closed as duplicates of this one. It was one job split across three tickets; the split cost more than it bought.

## Ask

Filed per channel request: the Docker/self-host path must become a first-class deploy substrate, on the same footing as the jail and native-Linux ones — same layout, same hot-vs-cold decision code, one idempotent entry point — and it must extend to a published container image and to `curl | bash` install/update one-liners.

## Decided up front (channel ruling, vjt)

- **The published image is a self-contained release image**, not the current dev/toolchain image. This settles the A/B that #439 carried.
- **Updates are hot when the preflight verdict allows it, cold otherwise** — the same rule the other two substrates already follow.

## Where we actually are (measured on `main`, not from memory)

**Layout is asymmetric.** `infra/freebsd/deploy.sh` and `infra/linux/deploy.sh` exist; **there is no `infra/docker/`**. The Docker path lives in `scripts/quickstart.sh`, `scripts/quickstart-update.sh`, `scripts/quickstart-stop.sh`, next to the operator-only `scripts/deploy.sh` (which targets a specific production host).

**Hot/cold decision is shared by two substrates out of three.** `Grappa.Deploy.Preflight` is the source of truth and both `infra/linux/deploy.sh` and `infra/freebsd/deploy.sh` are thin invokers of it (`infra/linux/deploy.sh:5,67`). `scripts/quickstart-update.sh` does **not** use it: it re-derives its own change classification with path regexes (`scripts/quickstart-update.sh:124-146`) and then **always** does `compose up -d --force-recreate` (line 153). Its header says so explicitly: *"This is NOT the hot-reload path … here we always recreate"* (lines 24-27). The self-hoster gets every session dropped for changes the other two substrates hot-reload.

**The install/update split is not obviously load-bearing.** `scripts/quickstart.sh` already documents itself as re-runnable ("existing secrets in `.env` are never regenerated, and an already-up stack is just reconciled").

[...]

Issue #468, opened 2026-07-27
`.env.example` ships PHX_HOST=grappa.bad.ass — third-party installs mint upload links pointing at someone else's host


vjt:
## What happens

A self-hosted install that copies `.env.example` without editing `PHX_HOST` mints upload links on
**`grappa.bad.ass`** — the maintainer's home-LAN deployment — instead of its own hostname.

Observed in the wild: a third-party instance (Docker, 0.5.0, installed 2026-07-26) posted

```
📸 https://grappa.bad.ass/uploads/<slug>.png
```

to an IRC channel. The link is dead for every reader, and it advertises a host that has nothing to do
with that deployment.

## Why

`.env.example` (lines ~118-122) does not ship a placeholder, it ships a real host as the value:

```
# Public hostname the bouncer is reached at via nginx. Defaults to
# grappa.bad.ass (the home-LAN deployment). Override here if your
# deployment lives elsewhere — Phoenix's `check_origin` rejects
# WebSocket handshakes whose Origin header doesn't match this.
PHX_HOST=grappa.bad.ass
```

`compose.override.yaml.example:36` does the same with `voygrappa.bad.ass`.

`PHX_HOST` is not only the `check_origin` gate: with `EXTRA_CHECK_ORIGINS` it is the source of the
deployment's HTTP host-alias set (`lib/grappa/http_hosts.ex`, #324), which is what upload links are
built from. So the wrong value does not fail loudly — it silently produces links to the wrong host.

Note the value is **not** hardcoded in the upload path: the `|| "grappa.bad.ass"` runtime fallback was
already removed (`config/runtime.exs`, prod raises when `PHX_HOST` is unset), and the uploader host is
already switchable `embedded` / `litterbox` via `upload.active_host`. The remaining source is the
example file being copied verbatim.

## Suggested fix

1. `.env.example` and `compose.override.yaml.example`: use a placeholder (`grappa.example.com`) and
   reword the comment so it reads as "you must set this", not "defaults to".
2. Optionally, refuse to boot in prod when `PHX_HOST` matches the shipped placeholder or a known
   upstream host — same shape as the existing unset-in-prod raise, so a copy-paste install fails
   loudly at boot instead of quietly minting broken links.
3. Consistent with the DESIGN_NOTES entry on replacing the `grappa.bad.ass` occurrences with a
   placeholder plus a generic comment.

Reported on IRC by a third-party operator running a self-hosted 0.5.0 instance.

vjt:
[...]

PR #617, opened 2026-08-01
feat(#503 D): release-image docker run + curl|bash bootstrap

vjt:
## What

Unit D of #503 — the run-time env for the self-contained ghcr release image
unit C published. Grows a **second substrate** onto `infra/docker/deploy.sh`
(the checkout-less host that runs the published image with plain `docker` — no
compose, no source, no `mix`) plus the `curl | bash` bootstrap
`infra/docker/get.sh`. Closes unit C's forward-refs (`bring-up + one-liners
are unit D`).

Three commits:
1. `feat(#503 D)` release-image mode in `deploy.sh` (mode-detect, release
   `install`/`update`/`stop`/bare verbs against `docker run`, secret gen,
   PHX_HOST tty prompt) + 20-test bats suite.
2. `feat(#503 D)` `get.sh` POSIX-sh `curl|bash` bootstrap (mirrors the two
   shell files into `$GRAPPA_HOME`, execs deploy.sh in forced release mode) +
   8-test bats suite + wired into ci.yml's shellcheck/dash gate.
3. `docs(#503 D)` INSTALL.md / README.md / OPERATIONS.md runbook /
   DESIGN_NOTES entry filling the placeholders.

## Design (all decided with vjt/orchestrator — not re-opened here)

- **COLD-only, as the finding the issue sanctions.** The release image ships
  no `Phoenix.CodeReloader`, so release-mode `update` skips preflight and
  force-colds through the shared `deploy_common.sh` cold path (pull → migrate →
  `rm -f` + `run -d` recreate). Hot-on-image is **#503 unit E** — documented in
  every operator doc, not silently dropped.
- **Secret model = fork C.** Four random secrets minted on the host with
  `openssl`; the VAPID keypair minted by the image's OWN `:crypto`
  (`docker run … eval`, mirroring `mix grappa.gen_vapid` — the release image has
  no `mix`). Kept shell-only deliberately to avoid dragging a COMPILE lane into
  a shell-only unit.

## Security invariants (all tested)

- `umask 077` before any create + `chmod 600` — the env file is never
  world-readable, not even for an instant.
- No secret on argv (`ps`-visible) or stdout — generated into shell vars / a
  0600 `--env-file`.
- **Idempotent:** an existing `$GRAPPA_HOME/grappa.env` is NEVER regenerated
  (rotating SECRET_KEY_BASE/GRAPPA_ENCRYPTION_KEY under a live box invalidates
  sessions + makes Cloak creds undecryptable).
- `PHX_HOST` asked on `/dev/tty` (a piped one-liner reads the answer from the
  terminal, not the pipe), env fallback, HARD-FAIL if neither — never a silent
  `localhost` (a wrong PHX_HOST mints dead upload links, #468).

## ⚠️ Published-image verification is PENDING the v0.9.0 tag

The `docker` release job is **tag-driven with ZERO prior real runs**, and ghcr
carries **no grappa image yet**. Unit D was built + **shell-tested against the
image's SHAPE only** (bats stub the docker daemon; shellcheck/dash gate the
shell). A real `docker run` of the PUBLISHED image is verified **after the tag
cuts**. The local shell tests below are NOT published-image verification.

## Gate evidence

- Full bats set green: `bats test/bin/ test/infra/ test/scripts/` → **199 ok,
  0 not-ok** (exit 0). New: 8 `get.sh` tests + 20 release-image tests.
- CI shellcheck/dash gate replicated locally, clean: `shellcheck -s sh …
  get.sh`, `shellcheck -x … deploy.sh`, `dash -n … get.sh`.
- Shell-only unit → COMPILE/STACK lanes not run locally by design; PR CI is the
  ship gate.

Refs #503.


vjt:
## CI red diagnosed — it is the TEST, not the installer

`test + lint + audit` fails on exactly one bats case (run [30706816699](https://github.com/vjt/grappa-irc/actions/runs/30706816699)):

```
not ok 45 install: writes a 0600 env file carrying PHX_HOST + generated secrets
# (in test file test/infra/deploy_docker_release_image_test.bats, line 93)
#   `[ "$(stat -f '%Lp' "$ENV_FILE" 2>/dev/null || stat -c '%a' "$ENV_FILE")" = "600" ]' failed
```

**The production side is correct** — `infra/docker/deploy.sh:826` sets `umask 077` *before* any create (so the file is never world-readable even for an instant) and `:883` re-asserts `chmod 600 "$ENV_FILE"`. Nothing to fix there.

**The assertion is BSD-first and that ordering is the trap on GNU coreutils.** `stat -f` on GNU means `--file-system`, so `stat -f '%Lp' FILE` parses `%Lp` as a *second file operand*: it errors on that operand (stderr, swallowed by `2>/dev/null`), **but still writes the filesystem dump of `FILE` to stdout**, and exits non-zero — so the `||` fallback fires too and appends `600` to output already containing the dump. Reproduced on Linux:

```console
$ echo "[$(stat -f '%Lp' t600 2>/dev/null || stat -c '%a' t600)]"
[  File: "t600"
    ID: 5a701d5b71693b22 Namelen: 255     Type: tmpfs
Block size: 16384      Fundamental block size: 16384
Blocks: Total: 519006     Free: 482427     Available: 482427
Inodes: Total: 1048576    Free: 1030232
600]
```

Green on macOS (where `stat -f` is the mode format), red on the Linux runner. Nothing about the branch's behaviour changed between the two.

**Fix: put the GNU form first**, since BSD `stat -c` is an illegal option that exits non-zero with empty stdout, making the fallback clean in that direction:

```sh
[ "$(stat -c '%a' "$ENV_FILE" 2>/dev/null || stat -f '%Lp' "$ENV_FILE")" = "600" ]
```

Scanned the rest of the tree for the same shape: `test/infra/deploy_docker_release_image_test.bats:93` is the **only** occurrence, on this branch or on main. One-line change, no other site to sweep.

vjt:
## Update — the bats fix worked; it uncovered a **second, independent** defect

Run [30708722133](https://github.com/vjt/grappa-irc/actions/runs/30708722133) at `25bf4184`: **no `not ok` anywhere** — case 45 passes now, so the `stat` GNU-first swap did what it was diagnosed to do.

The job is still red, but at a **later step** that had never executed before:

```
Shellcheck + POSIX-lint deploy scripts (#503 anti-drift)
In infra/docker/deploy.sh line 84:
CANDIDATE_ROOT="$(cd "$SELF_DIR/../.." 2>/dev/null && pwd || true)"
                                                   ^-- SC2015: Note that A && B || C is not if-then-else
```

The bats stage aborted the job before shellcheck could run, so this was masked. Worth stating plainly: *a gate going red does not mean only the thing you can see is broken* — the stage that fails first hides every stage behind it.

**This one is the branch's own.** `CANDIDATE_ROOT` does not exist in `infra/docker/deploy.sh` on `main`, so the line is new here. And the lint invocation is unchanged: `infra/docker/deploy.sh` was already on the `shellcheck -x` line on main; this PR only adds `infra/docker/get.sh` to the `-s sh` line. So nothing about the CI config caused it.

**Fix by restructuring, not by suppressing.** The intent is "empty if the `cd` fails", which reads directly as:

```sh
CANDIDATE_ROOT="$(cd "$SELF_DIR/../.." 2>/dev/null && pwd)" || CANDIDATE_ROOT=""
```

Same semantics — the substitution simply fails instead of being laundered through `|| true` — and it drops the `A && B || C` shape that SC2015 is warning about. A `# shellcheck disable=SC2015` would silence a warning that is pointing at something real.

Issue #621, opened 2026-08-01
doctor (dev env) scores every `Mix.env() == :test`-gated function as undocumented


vjt:
## The artifact

`mix doctor` derives a module's function **count from the source AST** but reads
each function's **doc/spec presence from the compiled beam of the env it runs
in**. In `MIX_ENV=dev` (how `ci.check` runs it), every function defined inside an
`if Mix.env() == :test do … end` block is compiled OUT, so it has no beam entry
— doctor still counts it (from source) but scores it as **"No Docs" AND "No
Specs"**, even when the source gives it `@doc false` + a full `@spec`.

Net: a module's dev-env doc-coverage is
`(always-compiled documented fns) / (all source fns incl. test-gated)`, which
sinks as its test-only seam grows. The **test-env** doctor (`check.sh` also runs
one, for `test/support` parity) reads the same functions from the test beam and
correctly scores them 100%.

## Why it's repo-wide, not a one-module problem

Several modules gate test-only helpers this way:
`Grappa.Repo.BusyRetry`, `Grappa.WSPresence`, `Grappa.Net.SourceAliasManager`,
`Grappa.Net.SourceAlias.Config`, `Grappa.Admission.Config`,
`Grappa.Push.BadgeSource`, `Grappa.WindowCounts.PushSource`.

They pass today only because their **real** functions outnumber their seams
(BadgeSource / PushSource sit at 75%, above doctor's 40% default floor).
`BusyRetry` (3 always-compiled + a large fault-injection seam) is simply the
**first to cross the floor** — a #594 change that added one more test-gated
helper dropped it from 43% to 38% and reddened the gate, despite the seam being
fully `@doc false` + `@spec`'d (test-env doctor: 100%). The next module to grow a
seam will rediscover this from a red gate.

## Options to consider (not decided here)

- Run the **gate's** doctor only in `MIX_ENV=test` (where the count is honest),
  keeping the dev run informational — but the dev run was added deliberately for
  a reason (`check.sh` comment), so this needs care.
- A `.doctor.exs` that excludes `test/support`-style seams or `@doc false`
  test-gated functions from the *dev* count.
- A convention: keep test-only fault/injection seams in `test/support` modules
  so prod modules never carry them in the source count.

## Immediate context

Surfaced by #594 (BusyRetry `fire_on:` seam). That PR works around it narrowly
by collapsing `arm_faults/2` into a single explicit-`fire_on` arity (a good
[...]

Issue #594, opened 2026-07-31
Cross-process terminal coverage for the BusyRetry fault seam (do it in the #590 pass)


vjt:
## What

The `Repo.BusyRetry` fault-injection seam (`busy_retry.ex`, `Mix.env() == :test`-gated)
stores its armed fault in the **process dictionary**. That is a deliberate anti-leak
choice: a per-process fault cannot bleed into a concurrent async test.

The cost is reach. A test can only inject a busy into work that runs **in the test
process**, so the ROOT A / B1 terminal policy is provable for some terminals and not
others:

| terminal | proven by a test? |
|---|---|
| REST 503 (`FallbackController`) | ✅ `themes_controller_test` (core) |
| auth door 503 (`POST /auth/login`, explicit clause above the catch-all) | ✅ added in B1 |
| **WS `close_failed`, socket stays open** | ❌ channel runs in its own process |
| **#590 drop: `maybe_open_query_window` logs and the session survives** | ❌ `Session.Server` is its own process |

The two unproven ones are structurally correct today (threaded through `with` /
`FallbackController`, no hard match left), and were verified by a caller read-review —
but nothing pins them, so a future refactor can reintroduce a crash unseen. That is
exactly how the `maybe_open_query_window` `CaseClauseError` got in: **Dialyzer does not
flag a non-exhaustive `case`.**

## Why this is deferred rather than done

Reaching another process needs the seam to become pid-keyed (e.g. ETS keyed by pid,
armed on the channel/session pid, cleaned in `on_exit`). Bolting that onto the core to
serve two tests makes the mechanism heavier than the problem it solves.

**The #590 pass wants the same capability anyway** — it hardens background writers
(`destroy_visitor/1` and friends), whose whole terminal *is* "the sweeper process
survives and drops". Building the cross-process arming there pays for both, so it
belongs there.

## Acceptance

With the pid-keyed seam in place, add:

1. a channel test: a busy on `close_query_window` replies `close_failed` **and the
   socket stays open** (never an error that tears the connection down);
2. a session test: a busy on the query-window auto-open **logs and the session keeps
   running** — no crash, matching the function's own "session continues" moduledoc.

Related: #590 (background-writer best-effort DROP), #523, #518.

vjt:
Shipped to production in the 2026-08-02 jail cutover — prod runs `b41b8d09`, released as [v0.9.0](https://github.com/vjt/grappa-irc/releases/tag/v0.9.0).

Issue #590, opened 2026-07-31
web: extend BusyRetry to background writers (best-effort drop, NOT 503)


vjt:
## Context

Follow-up carved out of the #523/#518 ROOT-A fix. That change introduces the
shared SQLite busy-retry engine `Grappa.Repo.BusyRetry` and applies it to the
**web-reachable** write paths (Themes, Visitors, QueryWindows, ReadCursor,
User/ServerSettings), where a transient `SQLITE_BUSY` that outlives the retry
budget degrades to `{:error, :db_unavailable}` → a clean **503** via
`FallbackController` (instead of a 500 raise).

## What was deliberately left OUT of #523's commit, and WHY

The **background** writers were NOT wrapped in that change:

- `Grappa.Visitors.Reaper`, `Grappa.Uploads.Reaper`, `Grappa.Accounts.Reaper`
- `Grappa.SessionLog`, `Grappa.AdminEvents` (+ their ring-buffer trims)
- `Grappa.Notify` / `Grappa.Push` `delete_all` sweeps

These have a **different terminal policy**: there is no HTTP client on the other
end, so a sustained-busy outcome must be **best-effort drop + log** (exactly like
`Grappa.Scrollback`'s `:persist_unavailable` #336 contract), **never a 503**.

Folding them into the #523 commit would have mixed two terminal semantics
(client-facing 503 vs background best-effort-drop) into a single change — a
boundary violation. The retry ENGINE is shared; the terminal handling is the
20% that differs per caller (CLAUDE.md "reuse the verbs, not the nouns").

## Scope of this issue

Apply `Grappa.Repo.BusyRetry.run/1` to the background writers above, with a
best-effort-drop terminal (log + continue on `{:error, :db_unavailable}`), so a
transient contention during a sweep/log-write is ridden out rather than crashing
the reaper/sink or escaping. No 503 (no client). Verify the classifier + budget
are the shared engine's, not a forked copy.

Refs #523, #518.

vjt:
Shipped to production in the 2026-08-02 jail cutover — prod runs `b41b8d09`, released as [v0.9.0](https://github.com/vjt/grappa-irc/releases/tag/v0.9.0).

Issue #523, opened 2026-07-28
web: DB writes outside Scrollback lack busy-retry — a transient SQLITE_BUSY raises a 500 instead of retrying


vjt:
## Summary

Every DB write path in the app **except** `Scrollback.persist_event/1` performs a
bare `Repo.insert!` / `Repo.insert` / `Repo.update` / `Repo.transaction` with no
handling for a transient SQLite `Exqlite.Error: Database busy` (SQLITE_BUSY).
Under write concurrency (`pool_size > 1` + WAL) such a write **raises**, the
controller action fails to return a `%Plug.Conn{}`, and the client gets a **500**
— where a bounded busy-retry (and, failing that, a clean 4xx/503) is the correct
behaviour.

The mitigation pattern **already exists in-house**:
`Grappa.Scrollback.with_pool_retry/1` (`lib/grappa/scrollback.ex:270`) rescues
`Exqlite.Error` / `DBConnection.ConnectionError`, classifies `transient_fault?`,
and retries with backoff on a wall-clock budget before degrading (never crashes,
per #336). It is wired into the hot message-insert path **only**. This issue is
to extend the same discipline (a shared helper reused across contexts — not
copy-paste) to the other write paths. The pattern is in the codebase; it just
isn't applied beyond Scrollback.

## Evidence

Reproduced in an e2e run (grappa-test = `MIX_ENV: dev` → `pool_size: 5`, WAL,
`busy_timeout: 30_000`). `POST /themes/:id/copy` raised:

```
** (Exqlite.Error) Database busy
    INSERT INTO "themes" ("name","visitor_id","published","payload","apply_count","inserted_at","updated_at") VALUES (?1,?2,?3,?4,?5,?6,?7) RETURNING "id"
    (ecto_sql) lib/ecto/adapters/sql.ex:1121: Ecto.Adapters.SQL.raise_sql_call_error/1
    (grappa)   lib/grappa/themes.ex:258: anonymous fn/2 in Grappa.Themes.do_copy/3
    (grappa)   lib/grappa/repo.ex:14: anonymous fn/1 in Grappa.Repo.transaction/2
```

The plug then crashed (`Expected .call/2 to return %Plug.Conn{} but got: ...`)
→ **500**. `Grappa.Themes.do_copy/3` is a `Repo.transaction` whose insert is a
bare `Repo.insert!` — no busy-retry, unlike the `Scrollback` insert running
concurrently, which rode the contention out and survived.

[...]

PR #622, opened 2026-08-01
BusyRetry into background writers (#590) + pin the two out-of-process DB-busy terminals (#594)

vjt:
Refs #590
Refs #594

## #590 — BusyRetry into the background writers (B2)

B1 wired `Grappa.Repo.BusyRetry` into every web-reachable write (terminal →
503). B2 wires the SAME shared engine — classifier + budget reused verbatim —
into the **background writers**, the ones with no client waiting. The terminal
differs per the three-doors split: a background write that outlasts the retry
budget **DROPS best-effort (log + continue)**, never a 503, never a crash (the
`:persist_unavailable` posture #336 gave the scrollback hot path, generalised).

- **Uniform contract, per-caller terminal.** Every context write fn returns
  `{:error, :db_unavailable}`; each CALLER maps its own terminal.
  `Visitors.delete/1` PROPAGATES (operator/admin want the 503 / a raised CLI
  error); `Visitors.purge_if_anon/1` ABSORBS to `:ok` + a log (a
  failed/preempted login has no client; the anon row waits for the reaper).
- **Writers migrated:** `Accounts.delete_expired_sessions`,
  `Visitors.destroy_visitor`, `Uploads.soft_delete/2`, `SessionLog`
  persist+prune, `AdminEvents` persist+prune+load_recent (a degraded boot
  reload yields an empty ring, not a crash-loop), `Push.delete_dead`. Each
  reaper's `sweep/0` maps the drop back to `{:ok, 0}` + a log so its
  `handle_info` stays total.
- **Small web fan-out:** only `admin/uploads_controller.delete` changed (its
  `{:ok, _} =` hard match became `with <-`); every other `with`-based action
  already flows `:db_unavailable → 503` via the shared `action_fallback`.
- **Notify is OUT (deliberate):** no background sweep — every caller is a web
  controller, so it belongs to B1's web-503 door, not B2. Its
  `notify_controller` `:ok =` hard matches are filed as a #523 follow-up, not
  smuggled in here. `Push.touch_last_used` (a post-delivery UPDATE) likewise.

## #594 — pin the two out-of-process DB-busy terminals

Two terminals live out of process from any test, and **Dialyzer does NOT flag a
non-exhaustive `case`** — so nothing but a test pins them; a refactor dropping
either `{:error, :db_unavailable}` arm would crash on every DB-saturated path,
unseen (exactly how the original `maybe_open` arm got in).

- **(a) WS `close_query_window`** (runs in the Channel's pid): a sustained busy
  replies `close_failed` and leaves the socket open.
- **(b) Session query-window auto-open** (`maybe_open_query_window/2`, runs in
  the `Session.Server`'s pid): a sustained busy logs + the session survives.

Both are armed via a new **cross-process ETS fault seam** keyed by the target
pid. (b) can't be isolated by a plain per-pid counter (the open runs only after
the persist succeeds ⟹ with one counter it never faults), so the seam grew a
**positional `fire_on:` selector**: it rides out the first `k-1` fault-CHECKS in
the pid, then fires from the `k`-th. For the inbound-DM flow `fire_on: 3`
(persist insert + preload, then the open — the push obligations run in their own
Tasks, adding no session-pid checks); a wrong value fails LOUD (too low faults
the persist, too high lets the open succeed), so `3` is empirically confirmed by
the triangulating assertions, not hardcoded.

`arm_faults` is a **single arity with a required `fire_on:`** — no immediate-mode
2-arity. Two arities where the shorter means "fire on the next check" is a
default argument in disguise (`arm_faults(pid, n)` never says WHEN it fires),
the silent-degradation path CLAUDE.md bans; `fire_on: 1` is the explicit
immediate case at every call site.

### Note on the doctor gate

The extra test-gated arity briefly pushed `Grappa.Repo.BusyRetry` under `mix
doctor`'s 40% dev-env doc floor. The single-arity refactor restores it as a
**side effect**, not the reason. The floor itself is a general artifact: doctor
counts functions from source but reads doc/spec presence from the **dev beam**,
where `Mix.env() == :test` helpers are compiled out and scored "missed" despite
`@doc false` + `@spec`. The **test-env doctor scores this module 100%** — the
seam is fully documented; this was never a docs-quality problem. The repo-wide
artifact is tracked as #621.

## Gate

`scripts/check.sh` → EXIT 0: 4867 tests / 0 failures, Dialyzer 0 errors, Credo
+ Sobelow + doctor (dev + test) + wire-types + bats 171/171 all green.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -25619,19 +25619,31 @@ from any test: (a) the WS `close_query_window` `close_failed` reply runs in the
 Phoenix Channel's own pid; (b) the session query-window auto-open
 (`maybe_open_query_window/2`) runs in the `Session.Server`'s pid. The existing
 process-dictionary fault seam only reaches work in the TEST process, so #594
-added a cross-process ETS seam keyed by the TARGET pid (`arm_faults(pid, n)`).
-That pins (a) directly. But (b) can't be isolated by a plain per-pid counter:
-the auto-open runs ONLY after the persist returns `{:ok, _}`, and with a single
-counter "persist succeeds ⟺ fault count is 0 ⟹ the open never faults." So the
-seam gained a POSITIONAL selector: `arm_faults(pid, n, fire_on: k)` rides out
-the first `k-1` fault-CHECKS in the pid (a "check" = one `maybe_inject_fault/0`
-at the top of a `BusyRetry.run/1` attempt) and fires from the `k`-th. For the
-inbound-DM flow the pre-open checks are persist insert (1) + persist preload (2)
-— the `push: true` obligations (`Push.Triggers`, `WindowCounts`) each run in
-their OWN Task, so they add ZERO checks to the session pid — making the open's
-insert check **3**. `fire_on:` counts checks, NOT operations, and was named for
-exactly that; it is a separate explicit arity (NO default-arg drift), so every
-existing channel/reaper caller of `arm_faults/2` is untouched.
+added a cross-process ETS seam keyed by the TARGET pid. That pins (a) directly.
+But (b) can't be isolated by a plain per-pid counter: the auto-open runs ONLY
+after the persist returns `{:ok, _}`, and with a single counter "persist
+succeeds ⟺ fault count is 0 ⟹ the open never faults." So the seam gained a
+POSITIONAL selector: `arm_faults(pid, n, fire_on: k)` rides out the first `k-1`
+fault-CHECKS in the pid (a "check" = one `maybe_inject_fault/0` at the top of a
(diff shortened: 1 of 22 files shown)


vjt:
Merged into main via the orchestrator-directed merge train (rebased onto f85cea99 — it was two merges behind its f3119a73 cut; fully content-disjoint from #592/#604, clean rebase, ff-merge).

**Tolerated red — #623.** The integration run was RED with exactly ONE failure: `recover-identity-real.spec.ts:193` — the #623 mechanism (RECOVER ok, post-RECOVER IDENTIFY hangs, server pushes `outcome=failed`), surfacing as a `toBeVisible` element-not-found at 31.7s. That is a diagnosed-and-filed red, not a masked one; every other check is green. Merged by explicit orchestrator ruling; CI intentionally not re-run.

Issue #642, opened 2026-08-02
per-IP session cap counts the reverse proxy address, and a rejected spawn leaves the credential 'connected' (user wedged)


vjt:
Two defects, found together while setting up a second test account on staging. They compound: the first makes the per-IP cap fire when it should not, the second turns that rejection into a permanently unusable credential.

## 1. The per-IP cap counts the proxy's address

Every browser session on staging registers `sessions.ip = 192.168.144.1` — the Docker gateway, i.e. the reverse proxy — not the client address. With `max_per_ip` at its default of 1, the second distinct human to open the app is refused because the first one "already" holds the budget for that IP. The cap is doing exactly what it was asked to do; the address it was handed is wrong.

Worked around on staging by raising `azzurra.max_per_ip` to 5 via `Networks.update_network_settings`. That is a workaround, not a fix — with the proxy address recorded, the cap cannot distinguish anyone from anyone.

**To check on prod**: if the address recorded there is the proxy's too, the cap is either inert (high limit) or a shared budget across all users (low limit), and no per-IP policy is actually being enforced. The client address needs to come from the forwarded-for chain, from a trusted proxy list, and nothing else should be accepted.

## 2. A rejected spawn leaves the credential `connected` — the user is wedged

`SessionController.add_user_network/3` calls `bind_user_credential` first, which writes `connection_state: :connected` (schema default, `credential.ex:237`), and only then calls `NetworkSpawn.orchestrate`. When admission refuses — `:ip_cap_exceeded` in my case — there is no rollback. The credential stays `connected` while no session exists.

Measured on staging:

- the UI shows the network as CONNECTED, with a nick and a climbing uptime;
- `LiveIntrospection.list_sessions` shows **no** `Session.Server` for it;
- connecting again → **409** `already_attached` ("You're already connected to that network");
- sending anything → **404** `not_found` ("That network or resource doesn't exist"), because `Session.send_privmsg → call_session` finds no registry entry.

Server log:

```
07:29:38  error=:ip_cap_exceeded ... session spawn rejected
07:32:33  POST /session/networks → 409
```

The only way out is Disconnect (credential → `PARKED`) followed by Reconnect. A user who does not know that sees an account that claims to be online and silently swallows every message.

### Expected

[...]

Issue #632, opened 2026-08-02
Verify the #171 per-IP admission cap is still enforcing in prod — more clients from one IP than the cap allows


vjt:
## Report

vjt, from prod (2026-08-02, post-cutover): **more clients connected from the same IP than the per-IP cap should
allow.** Ask: verify the limit still works, fix it if it does not. **Already-connected users may stay — the
requirement is that NEW connections from an over-cap IP are rejected.**

That last part matches the existing design: `Admission.check_ip_cap/2` runs at **admission** only, so nothing
re-evaluates a live session. No change needed there; it is the admission decision itself that must be trusted.

## What the cap actually does today (`lib/grappa/admission.ex:331-369`)

Counts **DISTINCT SUBJECTS** for `(source_ip, network)` — network-bound, non-revoked, self-excluding the
requesting subject, disjoint by subject kind — keyed on the persisted `accounts_sessions.ip`, compared against
`Network.max_per_ip` (falling back to `@default_max_per_ip_per_network`).

## Three hypotheses, each with a decisive check. Run them in this order — do NOT start coding.

**H1 — it is working, and the expectation is wrong (check this FIRST).**
The cap counts distinct SUBJECTS, not connections. One subject with many browser tabs / devices / sockets is
**allowed by design**, and several such clients from one IP is exactly what you would see.
🔎 Decisive: for the offending IP, count `DISTINCT subject` vs `COUNT(*) sessions` on that network. If subjects
≤ cap and only sessions exceed it, there is no bug — and then the real question for vjt is whether the cap
should be per-subject (today) or per-session/per-connection (a deliberate product change, NOT a silent fix).

**H2 — the INPUT is wrong: `accounts_sessions.ip` no longer records the real client IP.**
The IP comes from `GrappaWeb.RemoteIP.format(conn)` (`network_spawn.ex:55`), which depends on
`Plugs.RemoteIpFromProxy` (`endpoint.ex:136`) trusting `x-forwarded-for` / `x-real-ip` from the fronting nginx.
**Tonight's cutover moved the app into a new jail with new addresses and re-ran `jail_install_nginx.sh`** — if the
proxy hop is no longer trusted, or the header is no longer set, every session records the proxy address (or a
constant), and the cap is then counting the wrong bucket entirely.
🔎 Decisive: distribution of `accounts_sessions.ip` for rows created AFTER the 00:22 cutover vs before. Look for
[...]

Issue #623, opened 2026-08-01
e2e mechanism: #581 recover-identity-real fails when the post-RECOVER IDENTIFY never completes (+r never lands, server pushes outcome=failed)


vjt:
## Summary (a mechanism, not a proven root cause)

`cicchetto/e2e/tests/recover-identity-real.spec.ts:193` (the #581 real-services
positive: *"a returning visitor on a Guest nick recovers a held registered nick
via RECOVER → +r"*) intermittently fails at its final ship-gate
(`spec.ts:268` — `getByTestId('recover-modal-success')` `.toBeVisible({ timeout: 30_000 })`).

Trace evidence (below) shows the recover modal reaches a **terminal
`data-outcome="failed"`** — the server pushes a recovery *failure*, `+r` is
never acquired. It is therefore:

- **NOT a cic render defect** — cic renders the failed outcome faithfully;
  `recover-modal-success` legitimately never exists because recovery genuinely
  failed. (The "server reached +r but the testid didn't render" hypothesis is
  falsified by the trace.)
- **NOT the #277 fixture family** — the `resetSubject(vjt) 500
  {:nick_rejected, 433, "vjt-grappa"}` line appears in only one run, at
  `grappaApi.ts:678`, logged in the report/teardown phase ~9 min AFTER the
  assertion failed. It is a downstream teardown artifact of the already-stalled
  session, not the cause.

## The mechanism (from the trace step-machine)

The recover modal drives three steps: `nick` ("Reclaiming your nick") →
`recover` ("Recovering your nick") → `identify` ("Identifying to services").
Across the 30s wait, two independent failing traces show the SAME shape:

| step | label | status at give-up |
|------|-------|-------------------|
| `nick` | Reclaiming your nick | **failed** (433 — holder holds the nick) |
| `recover` | Recovering your nick | **ok** (NickServ RECOVER succeeded; hold freed) |
| `identify` | Identifying to services | **running — stuck the entire 30s** |
| — | modal `data-outcome` | **failed** |

Reading: reclaim `NICK rec-id-nick` → **433** (the holder legitimately holds it;
bahamut emits verbatim *"This nick is owned by someone else… identify yourself
with /ns IDENTIFY password. Your nick will be changed in 60 seconds if you do
not comply."*) → NickServ **RECOVER succeeds** and frees the hold → but the
**follow-on IDENTIFY to re-acquire `+r` never completes** (`identify` hangs) →
`+r` never lands → the server pushes `outcome=failed`.

[...]

PR #643, opened 2026-08-02
fix(#642 defect 2): roll back the accreted credential when its spawn is refused

vjt:
Refs #642 (defect 2 only).

## Problem
`POST /session/networks` user accretion (`SessionController.add_user_network/3`)
bound the credential — `connection_state: :connected`, the `Credential` schema
default — **before** `NetworkSpawn.orchestrate`, with no rollback. When admission
refused the spawn (`:user_cap_exceeded` on staging) the row stayed `:connected`
while no `Session.Server` existed: the UI showed the network CONNECTED with a
climbing uptime, `POST /messages` 404'd (no registry entry), and a reconnect
409'd `already_attached`. The only escape was Disconnect+Reconnect.

## Fix
Accretion is now **atomic**: bind + spawn succeed or fail together. A new
`spawn_or_rollback/4` runs plan-resolve + orchestrate in a `with/else`; on **any**
`{:error, _}` after the bind it rolls the just-bound credential back via the
canonical `Credentials.unbind_credential/2` and returns the original typed error
(FallbackController maps it as before). Nothing stays attached, the refusal
surfaces honestly (503), and a later attempt starts clean — no
Disconnect+Reconnect.

**General rule (not the one incident):** the catch-all covers every post-bind
failure path — `:resolve_failed`, `:ip_cap_exceeded` / `:user_cap_exceeded`, the
`{:network_circuit_open, _}` / `{:start_failed, _}` tuples, and the
subject-row-gone `:not_found`.

This is the accretion twin of the PATCH /connect **U-0** stop-swallow fix
(`NetworksController.apply_transition/5`): both uphold the invariant *no observer
sees `:connected` without a live session*. PATCH transitions a pre-existing
parked row (commit-on-success); accretion **creates** the row, so restoring the
pre-operation state is a compensating **delete**, not a park (which would strand
a binding the user never established and re-409 the retry). See DESIGN_NOTES
2026-08-02 for the full rationale (incl. why not bind-parked-then-connect).

## Scope
**#642 defect 1** (the per-IP cap counting the reverse-proxy address) is
**deliberately out of scope** — it turns on whether prod records the proxy
address and whether the cap should be per-connection or per-subject (same axis as
#632), an open product question owned by @vjt.

## Tests / gates
- New regression test: drives `:user_cap_exceeded`, asserts the credential is
  gone (`{:error, :not_found}`), no live session, a retry succeeds once the cap
  clears, and exactly one `:connected` credential remains. Fails if the rollback
  is removed.
- Full `scripts/check.sh` green: 4892 tests / 0 failures, dialyzer 0 errors,
  credo --strict / sobelow / format / doctor / wire-types / bats all clean.
- Code-reviewed (mandated gate): fix judged sound; the one finding (comment
  atom-name accuracy) is folded in.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -26098,3 +26098,66 @@ triad: `ServerSettings.put_addressing_mode/1` + `put_static_mapping_prefix/1`
 no bind) with a mode-2 `base_plan` resolution afterward; it deliberately does
 NOT isolate either commit (either fix alone still admits), so its falsification
 is the DOUBLE revert of both commits ⇒ `{:hold, :no_client_source}` ⇒ RED.
+---
+
+## 2026-08-02 — #642 defect 2: a refused accretion spawn must not strand a `:connected` credential (the wedged user)
+
+`POST /session/networks` user accretion (`SessionController.add_user_network/3`)
+bound the credential — `connection_state: :connected`, the `Credential` schema
+default — BEFORE `NetworkSpawn.orchestrate`, with no rollback on refusal. When
+admission rejected the spawn (`:user_cap_exceeded` on staging, but the class is
+EVERY rejection out of orchestrate) the row stayed `:connected` while no
+`Session.Server` existed: the UI showed the network CONNECTED with a climbing
+uptime, `POST /messages` 404'd (no registry entry), and a reconnect 409'd
+`already_attached`. The only escape was Disconnect (→ `:parked`) + Reconnect — a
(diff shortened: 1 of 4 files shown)


vjt:
Merged to main via fast-forward at `3d986666` (rebased onto origin/main `b42b3c2e`). Sole red on the pre-merge e2e was `recover-identity-real.spec.ts` (assert L268), 1 failed / 579 passed — the known flake #623; per the pre-registered posture a PR whose only red is #623 is authorized to merge. Deploy remains held for vjt.

Issue #639, opened 2026-08-02
e2e flake: #360 scroll-to-bottom badge — tap 3 races a scrollback re-render (element detached from the DOM)


vjt:
## What happened

`integration` on PR #638 (attempt 1, run `30734371652`) failed with a single red:

```
1) [chromium] › tests/issue360-scroll-to-bottom-mention-badge.spec.ts:120:3
   TimeoutError: locator.click: Timeout 10000ms exceeded.
   - waiting for locator('[data-testid="scroll-to-bottom"]')
     - locator resolved to <button ... data-testid="scroll-to-bottom">↓</button>
     - attempting click action
       - element is visible, enabled and stable
       - scrolling into view if needed
     - element was detached from the DOM, retrying
   > 221 | await page.locator(SCROLL_TO_BOTTOM).click({ timeout: 10_000 });
```

**Rerun of the same job: GREEN — 579 passed / 0 failed (22.8m), `✓223 issue360` in 32.1s.**
So it does not reproduce, and it is a flake rather than a regression.

## Why it is not PR #638's doing

Recorded so the next person does not re-derive it:

1. #638 changes four cic files only — `pingCorrelation.ts` plus its tests and a new spec. #360 never
   enters that path: its six `ping` occurrences are **message text** (`first ping i360 mention one`),
   not CTCP PING correlation.
2. Timing rules out cross-talk from the new spec — #360 finished failing at `05:47:50Z`, the #637 spec
   ran `~05:51:11-15Z`. No overlap.
3. The failing job ran 28m35s against the 30-minute cap and concluded `failure`, not `cancelled` — a
   real test failure, not a timeout.

## Mechanism (hypothesis, not yet proven)

The locator resolves, the element is visible/enabled/**stable**, and then it is **detached** during the
click. That is a re-render of the scrollback between resolve and click, not a slow UI.

The spec reads a **shared, concurrently-written scrollback** — its own header comment says so:

> mentions from other specs (#280 leaves `vjt-grappa: 280 ping` behind)

So any line arriving from another spec on the shared subject can re-render the pane and swap the button
node underneath the pending click. Tap 3 is the most exposed of the three: it is the one where the
button is *about to legitimately unmount* (snap-to-bottom hides it), so a re-render there is
indistinguishable from the expected teardown.

Also relevant: `5ac794b6` already had to burst-pace this spec's sends to fit the 25-minute ceiling, so it
is a spec that is known to sit under timing pressure.

## What the cure must NOT be

[...]

Issue #499, opened 2026-07-27
test-infra: NetworkCircuit's global ETS singleton bleeds across tests serially — max_cases:1 + async:false only close the concurrent case


vjt:
Surfaced while triaging a red `check.sh` during the #449 merge. The failure is **not** in #449 (its diff touches no admission/login/circuit file) and reproduced with nothing else running; both isolation reruns were clean (`login_test.exs:606` alone → 0 failures; the whole file → 25/25).

## What happens

`Grappa.Admission.NetworkCircuit` is an application-wide singleton whose state lives in a named ETS table with **no per-test sandbox** — the moduledoc says so itself (`lib/grappa/admission/network_circuit.ex:61-71`). The documented mitigation is `max_cases: 1` in `config/test.exs` plus an `async: false` convention for tests touching the module.

That mitigation closes the **concurrent** collision — two tests crashing on the same `network_id` at the same time. It does **not** close the **serial** one: a row written by an *earlier* test in the same run can still be observed by a later test, because sqlite recycles rowids, so a fresh network can be handed an id that a previous test already tripped the circuit for. The stale row falsifies the window count and the circuit reads as closed when the test expects it open.

Net effect: an intermittent red in the visitor-login area that has nothing to do with the code under test, and that vanishes on an isolated rerun — the shape most likely to be waved through as "just a flake" or, worse, to have a real regression hidden behind it.

## Why it is worth fixing rather than tolerating

A non-deterministic red in the ship gate costs a full serialized gate cycle every time it fires, and it trains everyone to discount reds in that file. The gate only has value if a red means something.

## Direction (to be confirmed by whoever picks it up)

The fix belongs in test infrastructure, not in production code — `NetworkCircuit`'s runtime shape is deliberate. Candidates: reset the ETS table between tests that touch it (a `setup` calling the existing operator-driven clear, which already exists as an unconditional ETS delete), or give the affected tests network ids that cannot collide with recycled rowids. Prefer whichever keeps the production module untouched.

Deliberately filed rather than fixed mid-merge — it is a pre-existing gap, not something #449 introduced.

vjt:
**Stale-premise audit** (backlog sweep against `origin/main` `8ed4f2e0`, 2026-08-06). Evidence only — nothing closed, relabelled or edited here; triage is vjt's. Full sweep: #258.

**Verdict: PARZIALE — and the stated mechanism does not explain the stated repro.**

[...]

Issue #223, opened 2026-07-12
Auth sessions never pruned/revoked — unbounded sessions growth (stale rows stay valid)


vjt:
## Summary
On prod, a single long-lived user has **39 `sessions` rows with `revoked_at IS NULL`**, the oldest ~2 months old. Every web login / socket reconnect inserts a new `sessions` row, but nothing ever revokes or prunes the old ones.

## Impact
- Unbounded growth of the `sessions` table per active user.
- Stale auth material stays valid: a two-month-old, long-unused session can still authenticate a WebSocket. Security/hygiene concern.
- "Active sessions" enumeration (admin/me views, `SessionRegistry` reasoning) is polluted by dead rows.

## Current state
- `Accounts.revoke_session/1` and `Accounts.revoke_sessions_for_user/1` exist but are only invoked on explicit rotation paths.
- There is no background reaper for auth sessions. Contrast `Grappa.Visitors.Reaper`, which sweeps expired visitors — auth sessions have no equivalent lifecycle.

## Proposed directions (design open)
- **Idle-expiry**: auto-revoke sessions whose `last_seen_at` is older than a TTL.
- **Supersede-on-reconnect**: when a new session is created for the same `client_id`, revoke the prior one.
- **Periodic reaper** mirroring the visitor Reaper, revoking/deleting rows past `last_seen_at + N`.

## Repro
Query `sessions WHERE user_id = <uid> AND revoked_at IS NULL` for any user active over several weeks — the row count grows without bound.

vjt:
Shipped in the same 4am CEST COLD deploy (server 9cd5e6a3). Accounts.Reaper supervised child now live — prunes expired/stale auth sessions on a sweep, closing the unbounded-growth + stale-valid-rows gap.

PR #664, opened 2026-08-02
test(#653): slice A — deterministic setup for 3 e2e render-race flakes (#639/#519/#531)

vjt:
## Slice A of epic #653 — the three e2e render-race flakes

Diagnosis-first result (Phase 0): #499 (NetworkCircuit ETS bleed) is **ruled
out** as the epic's cause — it is real but already pervasively mitigated
(`AdmissionStateHelpers.reset_all/0` in ~20 touching test setups). The 8
symptoms are **3 distinct mechanisms**, not one. This PR is bucket **A** —
the e2e DOM render-race flakes, each cured by making the SETUP deterministic
(re-resolve a handle legitimately about to unmount / atomic observation),
never by weakening an assert or blindly bumping a timeout.

| # | spec | root cause (SETUP) | cure |
|---|------|--------------------|------|
| #639 | `issue360-scroll-to-bottom-mention-badge.spec.ts` tap 3 | button `<Show when={!atBottomNow()}>` unmounts the instant the pane hits bottom; a settling scroll flips `atBottomNow` between the visibility check and the click → detach mid-click → Playwright waits out 10s for a node that never remounts | `tapScrollToBottom` re-resolves per attempt, treats already-unmounted as snap-landed (button gone == at bottom == the tap's goal); applied to tap 3 only |
| #519 | `ux-2-mobile-archive.spec.ts` (@webkit) open-members step | shared `openMembersDrawer` fixture: a post-PART selection redirect swaps the focused window (and re-renders the topic bar) between the `count()` probe and the `.click()` → hamburger detaches → 30s timeout | fixture loops, re-picking the correct opener each attempt until `.shell-members.open` is visible or a bounded deadline throws loudly |
| #531 | `ux-6-a-mobile-members-scroll.spec.ts:108` (@webkit) | `overlay-open` is app-managed by a refcount (`overlayScrollLock.applyClass`); a push/pop settling during boot fires `applyClass()` and strips the class the test added, BETWEEN the two `page.evaluate`s (add / read) | add the class AND read the whole chain in ONE synchronous evaluate — no microtask/effect can interleave |

**Post-conditions unchanged in all three.** No assert weakened; the two
deadlines are retry ceilings, not slow-click timeout bumps.

## Verification

- Scoped (STACK lane): `--grep` the three titles, `--repeat-each=3` →
  **9 passed (2.1m)** (#639 chromium ×3, #519 webkit ×3, #531:108 webkit ×3).
- Full `scripts/integration.sh` (ship gate): **in flight** — result appended
  here when it lands. These are intermittent load flakes: a single green run
  is no proof the race is gone, so the evidence is mechanism-soundness (cures
  target the proven root causes above) + no full-suite regression from the
  shared-fixture change.

Merge is the orchestrator's call. Slice C (ExUnit DB-pool `assert_receive` +
the residual #499 `reset_all` gap in `grappa_channel_test`) is the next slice.

Refs #653

### DIFF
--- a/cicchetto/e2e/fixtures/cicchettoPage.ts
+++ b/cicchetto/e2e/fixtures/cicchettoPage.ts
@@ -731,13 +731,36 @@ export async function composeSend(
 // synthetic) over `.tap()` for the same WebKit synthesis-race reason as
 // closeMembersDrawer below.
 export async function openMembersDrawer(page: Page): Promise<void> {
-  const topicHamburger = page.getByLabel(/open members sidebar/i);
-  if ((await topicHamburger.count()) > 0) {
-    await topicHamburger.first().click();
-  } else {
-    await page.getByTestId("shell-chrome-rail-opener").click();
+  const drawer = page.locator(".shell-members.open");
+  // #653/#519 — re-resolve per attempt instead of a single probe-then-click.
+  // The opener that renders depends on the focused window kind (channel →
+  // TopicBar hamburger `<Show when={windowIsJoined}>`; non-channel → rail
+  // opener), and under full-gate load a settling selection redirect (e.g. the
+  // post-PART close-watcher moving focus) can swap the focused window — and
+  // re-render the topic bar — BETWEEN the count probe and the click. The
+  // hamburger then detaches mid-click; Playwright's built-in detach-retry
+  // waits out its whole timeout for a node that has unmounted (the window is
+  // now a non-channel one with no hamburger). Looping re-picks the correct
+  // opener each attempt and absorbs the transient churn; the post-condition
+  // (`.shell-members.open` visible) stays exactly as strict, and a genuine
+  // failure still surfaces loudly when the deadline elapses. Green happy path
+  // is unchanged (one probe + click + visibility assert on the first pass).
+  const deadline = Date.now() + 15_000;
+  for (;;) {
+    if (await drawer.isVisible().catch(() => false)) return;
+    try {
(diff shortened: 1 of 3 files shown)


vjt:
### Full `scripts/integration.sh` ship gate — result

**581 passed / 1 failed** (all 583 tests executed; the final summary line was cut by a teardown-time kill of the run, not a test error).

The single red is **out of scope** and not in this diff:

```
✘ 400 [chromium] tests/recover-identity-real.spec.ts:193:3
   #581 recover-identity (real services) › ...recovers a held registered nick via RECOVER → +r (31.4s)
```

That is **#581/#623** (IDENTIFY-after-RECOVER protocol-sequencing — a separate unit, explicitly out of scope for epic #653 per the triage). This branch's diff touches only the three e2e files, none near recover/services.

All three slice-A specs passed in the full run — including the shared-fixture spec and every sibling that uses `openMembersDrawer`, so **zero regression** from the `openMembersDrawer` change:

- `issue360-scroll-to-bottom-mention-badge.spec.ts` (chromium) ✓ #223
- `ux-2-mobile-archive.spec.ts` (@webkit) ✓ #532
- `ux-6-a-mobile-members-scroll.spec.ts` — all 6 tests ✓ #562–#567 (incl. :108, the registered flake)

Plus scoped `--repeat-each=3` earlier: 9/9 green.

Merge is the orchestrator's call; #623 (pre-existing red) is tracked separately.

Issue #667, opened 2026-08-02
cicchetto: channel-mode description table does not match what the ircd advertises


vjt:
## What is wrong

`cicchetto/src/lib/channelModes.ts` carries a static `MODE_DESCRIPTIONS` table whose
header comment claims it is "The bahamut/Azzurra channel-mode set". It is not. One entry
is actively wrong, one describes a mode the ircd does not have, and seven modes the ircd
actually advertises have no copy at all.

Reported on IRC by Essency/Sonic ("+d delayed"), confirmed by vjt.

## Evidence

Authoritative surface is what the ircd advertises in ISUPPORT — `src/s_misc.c:1275`
of the running ircd (`azzurra/bahamut`):

```
CHANMODES=bz,k,l,BcdijmMnOprRsStuU
```

So: type A `b z`, type B `k`, type C `l`, type D `B c d i j m M n O p r R s S t u U`.

### 1. `+d` is described as the opposite of what it does

`channelModes.ts:41`

```ts
d: { label: "delayed", desc: "delayed-join related" },
```

On this ircd `+d` is `MODE_NONICKCHG` — it forbids nick changes inside the channel.
Nothing to do with delayed join.

- `src/channel.c:3856` — `case 'd': mode.mode |= MODE_NONICKCHG;`
- `src/channel.c:1220` — display table entry `MODE_NONICKCHG, 'd'`
- `include/struct.h:1212` — `#define MODE_NONICKCHG 0x40000`

Correct copy: **no nick changes · members cannot change nick while in the channel**.

### 2. `+D` does not exist on this ircd

`channelModes.ts:40`

```ts
D: { label: "delay join", desc: "hide joins until the user speaks" },
```

There is no `case 'D'` in the mode parser and no `D` in `CHANMODES`. Delayed join is a
mode other ircds have (`+D` on InspIRCd/charybdis); this one does not. The entry should
be deleted, not corrected — a description for a nonexistent letter is worse than none,
because it is what made a reader assume `+d` was its lowercase sibling.

### 3. Seven advertised modes have no copy

Advertised in type D but absent from `MODE_DESCRIPTIONS`: `B j M O S u U`.

They do not crash — `modeDescription/1` falls back to `mode +<letter> / no description
available` — but the modal renders a toggle the operator cannot read. From the ircd:

| letter | constant | meaning |
|---|---|---|
[...]

PR #669, opened 2026-08-02
fix(#667): reconcile channel-mode copy with the HelpServ CMODE helpfile

vjt:
## What

`cicchetto/src/lib/channelModes.ts`'s `MODE_DESCRIPTIONS` claimed to be
"the bahamut/Azzurra channel-mode set" but was written from general IRC
knowledge, not from what the ircd actually advertises
(`CHANMODES=bz,k,l,BcdijmMnOprRsStuU`). One entry was actively wrong, one
described a mode that does not exist, and seven advertised modes had no copy.

## The fix: verbatim helpfile, not the C source

The copy now comes **verbatim** from the network's HelpServ `CMODE` helpfile
(`azzurra/services` → `run/data/helpfiles/us/helpserv/cmode`) — the text
users are already told — rather than paraphrasing the ircd C source.
Paraphrasing the C source (a *constant name* is not a description) is exactly
how `+d` (`MODE_NONICKCHG`) was labelled "delayed" and `+u` (`MODE_NOSPAM`,
which blocks PART/QUIT) read as a "spam filter".

### Type B/C/D reconciliation
- **`+d`** → nick-change restriction (op + voice exempt), no longer "delayed".
- **`+D` deleted** → no `case 'D'` in the ircd, not in CHANMODES; the phantom
  entry is what made a reader assume `+d` was its lowercase sibling. Deleted,
  not rewritten.
- **added `j M O S u U`** → the six advertised-but-missing letters, all with
  helpfile copy.
- **`+u`** → blocks PART/QUIT, not spam.
- **`+r` / `+R` split** → `+r` = services "registered with ChanServ" marker;
  `+R` = only registered nicks may join.
- **`c i k l m n p s t` aligned** to the helpfile too.

### Deliberately not changed
- **`+B`** (`MODE_HIDEBANS`) is advertised but **absent from the helpfile** —
  left on the generic fallback pending a copy decision. No invented wording.
- **`+C`** kept (real ircd mode w/ helpfile copy, but unreachable via
  `availableModes/1` since it is not advertised — harmless).
- Type A (`b z`) + PREFIX (`o h v`) exclusions untouched.

The copy stays cic-side UI text — **no wire field, no new wire token**.

## Tests (TDD, test-first)
New vitest acceptance in `channelModes.test.ts`:
- every advertised type B/C/D letter (except `+B`) resolves to non-fallback
  copy,
- every reconciled letter's `desc` is the verbatim `us` helpfile text,
- `+d`/`+u`/`r`/`R` read correctly,
- `+D` falls through to the fallback (entry deleted),
- `+B` stays on the fallback (the open decision).

## Gates
biome (`--write` clean), `bun run check`, `tsc --noEmit` standalone (exit 0),
full vitest (**3825 passed / 213 files, exit 0**).

Refs #667

### DIFF
--- a/cicchetto/e2e/tests/issue216-channel-modes-on-join.spec.ts
+++ b/cicchetto/e2e/tests/issue216-channel-modes-on-join.spec.ts
@@ -70,8 +70,13 @@ test("#216 — channel modes set before join are visible on join, and tapping op
     await expect(modal.locator(".mode-modal-toggle").first()).toBeVisible();
 
     // The "secret" (+s) toggle is a known bahamut flag mode → present in
-    // the available list derived from ISUPPORT.
-    await expect(modal.getByText("secret")).toBeVisible();
+    // the available list derived from ISUPPORT. Target the toggle by its
+    // accessible name ("secret (+s)", the button's aria-label) — a bare
+    // getByText("secret") now matches BOTH the label span and the
+    // HelpServ-verbatim description ("Channel is secret …", #667).
+    await expect(
+      modal.getByRole("button", { name: "secret (+s)" }),
+    ).toBeVisible();
 
     // The "topic lock" (+t) toggle is ACTIVE (pressed) — the peer set it.
     const topicLock = modal.getByLabel(/topic lock/i);
--- a/cicchetto/src/__tests__/channelModes.test.ts
+++ b/cicchetto/src/__tests__/channelModes.test.ts
@@ -165,9 +165,10 @@ describe("channelModes reconciled with the HelpServ CMODE helpfile (#667)", () =
   const ADVERTISED_TYPE_D = "BcdijmMnOprRsStuU".split("");
 
   // Verbatim copy from `run/data/helpfiles/us/helpserv/cmode`, minus the
-  // leading "  x  - " marker. THIS is the acceptance source of truth.
-  // (No `B`: advertised but absent from the helpfile — the one open
-  // decision, asserted on the fallback below.)
+  // leading "  x  - " marker. THIS is the acceptance source of truth for the
+  // verbatim entries. (No `B` here: it is absent from the helpfile, so it
+  // carries authored copy instead — pinned separately in HIDEBANS_COPY and
+  // asserted below, NOT verbatim.)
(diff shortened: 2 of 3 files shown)


vjt:
## Gate verdict: **RED, and one of the two is ours.** Do not merge.

`582 passed (22.4m)`, **2 failed**.

### 1. `issue216-channel-modes-on-join.spec.ts:74` — **CAUSED BY THIS PR**

```
strict mode violation: getByTestId('mode-modal').getByText('secret') resolved to 2 elements:
  1) <span class="mode-modal-toggle-label">secret</span>
  2) <span class="mode-modal-toggle-desc">Channel is secret (does not show in /WHOIS or /LI…</span>
```

The new HelpServ-verbatim copy for `+s` contains the word *secret* in the **description**, and the
label already was *secret*. The spec's `getByText("secret")` matched exactly one node before this
PR and matches two now. Clean causal chain, no flake in it.

**The product is right and the copy is right — the LOCATOR was loose.** So fix the locator, and fix
it by making it *more* precise, not less:

- ✅ target the control by its accessible name — `modal.getByRole("button", { name: "secret (+s)" })`
  — which is what the assertion actually means ("the +s toggle is present in the modal"), or scope to
  `.mode-modal-toggle-label`.
- ⛔ **NOT `.first()`** — that is picking an arbitrary winner out of an ambiguous match, i.e. exactly
  the weakening this repo forbids. ⛔ **NOT reverting or paraphrasing the copy**: verbatim-from-HelpServ
  is the whole point of #667.

This is also a small warning about the rest of the table: any other spec asserting on a bare mode word
inside `mode-modal` can now hit the same ambiguity. **Grep the e2e suite for the other mode words you
introduced copy for and check them systematically** — a gate is a sample, not a list.

### 2. `recover-identity-real.spec.ts:193` → assert line 268, 31.8s

The tracked #623 flake, unrelated to this PR: fails at the assert, burns the full 30s wait. Third+
sighting today (also hit `main` at `5a3f968a`).

### Posture

The standing #623 rule is *"only red is line 268 ⇒ merge and cite #623; **any other red ⇒ STOP**"*.
There is another red, and it is ours. Fix #1, re-gate, then merge.

Issue #534, opened 2026-07-29
e2e flake: channel-directory.spec.ts:61 — stray forward GET /messages on #bofh trips the #81 guard


vjt:
## Symptom

`cicchetto/e2e/tests/channel-directory.spec.ts:61` ("browse, no /messages fetch (#81 guard), search filter, one-click join") fails intermittently under **full-suite load**:

```
Error: GET /messages must NOT fire when selecting kind=list — grappa-irc#81 guard
Expected length: 0
Received length: 1
Received array:  ["https://nginx-test/networks/bahamut-test/channels/%23bofh/messages?after=7777&limit=200"]
    at tests/channel-directory.spec.ts:116:9
```

## Evidence gathered (2026-07-29, during the #525 gate)

- **Failed once** in a full `integration.sh` on the `525-ascii-fold` branch (551 passed / 1 failed).
- **Did not reproduce**: 5× `--repeat-each` in isolation on the same branch → 5/5 green; a second full `integration.sh` on the same branch → 552/0 green, this spec ✓.
- **Baseline**: green in 7/7 prior full-suite logs on the same host that evening (`485-integration{,2,3,4}`, `w1-509-integration`, `gate-proof-498-on-524`, `integration513-gate2`).
- **Not caused by the fold change** it was first seen under: the stray request is for `#bofh`, and the spec's other fixture is `#e2edir-<hex>`. Neither contains `[ ] \ ~`, so `canonical_channel/1` returns byte-identical values under both the rfc1459 and ASCII tables. `ChannelDirectory.Wire.mark_featured/2` — the one fold-touching site in this path — is inert for these names.

## What the failure actually is

The stray fetch is a **forward gap-fill** (`after=7777`), on the **autojoin channel**, not on the `$list` window under test. The spec arms its request collector *after* waiting for `#bofh`'s initial scrollback to land, then asserts that **zero** `/messages` requests fire for the rest of the test. Any live activity in `#bofh` (a peer is connected and the seed traffic is real) can legitimately trigger a forward fetch inside that window — so the assertion is load-sensitive by construction rather than wrong.

## Ask

Make the setup deterministic rather than weakening the assert — the `#81` guard has real value and must keep failing when the `$list` selection genuinely fetches scrollback. Sketch:

- Scope the collector to the **`$list` window's** channel key instead of asserting a global zero, so an unrelated `#bofh` gap-fill cannot trip it; or
[...]

Issue #530, opened 2026-07-28
e2e flake: nick-follow-query:48 — peer NICK relabel fails under full-gate load (green 5/5 iso)


vjt:
`cicchetto/e2e/tests/nick-follow-query.spec.ts:48` — *query window follows a peer NICK change — relabels, keeps history, routes with no 401* (chromium) failed at ~6.7s in the full `integration.sh` gate for #529, a change that touches only `HomePane.tsx` + `default.css` + its own test. No plausible path from a HomePane restyle to #373 rename logic.

**Isolation result: 5/5 green** on a fresh stack. Full-gate load is the differentiator, so this is registered as a flake rather than a regression — same class as #519 / #520 / #522.

Registering it because it was NOT in the known-flake set, which cost a gate verdict: an unregistered red has to be treated as real until proven otherwise, and that is the correct default. The point of this issue is that the next person who sees it can compare against a baseline instead of re-deriving one.

**Do not close this by relaxing the assertion** — the spec covers the #373 rename-order barrier (`query_windows_list` broadcast AFTER the DM history migration), which was a real bug. The fix is to make the setup deterministic under load, exactly as ruled for #522.

vjt:
Consolidated into #653 — "the test suite is not load-stable under the full gate". Triaged 2026-08-02: this report and seven others all describe the same signature (green in isolation, red under full-gate load), so they are being worked as one phenomenon rather than eight spec-level patches. The repro detail here stays readable and #653 links back to it; reopen if the load explanation turns out not to cover this one.

vjt:
## "Full-gate load is the differentiator" is refuted by measurement — it reproduces in isolation

This issue registered `nick-follow-query` as a load flake on the strength of **5/5 green in
isolation**, concluding that full-gate load is what distinguishes red from green. That
inference does not survive a direct re-measurement on today's main (`2617474b`), taken on a
**quiet host with the spec isolated** — no full gate, no competing lane.

### Isolated runs, chromium, `--grep` on this spec alone

| commit | iterations | target reds (`:109`) |
|---|---|---|
| `2617474b` (main) | 3 | 1 |
| `2617474b` (main) | 6 | 2 |
| `e1892e80` (previous main) | 6 | 2 |

**3 target reds in 8 valid iterations in isolation (~37%)**, and 2/6 on the preceding main.
The differentiator is therefore not load: the spec fails at roughly one in three *without*
any.

(One further red in the first run is excluded from the count: it died at 3.0 s on
[...]

Issue #520, opened 2026-07-28
e2e flake: freshness-on-activation:71 (#159 guard) — suppressed-channel row leaks past the Phase-3 premise gate under full-gate load (green 3/3 iso)


vjt:
## Summary

`cicchetto/e2e/tests/freshness-on-activation.spec.ts:71` — `#159 — tab RE-SELECT after a socket-stays-open gap re-fetches the missed row` (project `chromium`) fails in the full `scripts/integration.sh` gate but passes **3/3 in isolation**. Same iso-green / gate-red-under-load profile as #519 and #507. This is the regression guard for the (closed) feature bug #159, **not** the feature itself.

## Observed (full gate, 2026-07-28)

Phase 3 **premise gate** (spec line 99) fails — the during-gap row is present when it must be absent:

```
expect(locator).toHaveCount(expected) failed
Locator:  locator('[data-testid="scrollback-line"][data-kind="privmsg"]')
            .filter({ hasText: 'msg-159-sel-during-gap' })
Expected: 0
Received: 1
Timeout:  5000ms
```

The spec opens a per-channel delivery gap via `__cic_suppressChannelDeliveryForTests` (subscribe.ts) for `#bofh` only, then a peer posts `msg-159-sel-during-gap` and asserts the row does **not** render (proving the gap is real) *before* re-activation drives the catch-up re-fetch under test. In the full gate the suppressed row rendered anyway (count 1), so the premise gate failed before the actual fix was ever exercised.

## Isolation

3/3 green (both tests in the file):

```
scripts/integration.sh --project chromium --grep "#159" --repeat-each 3
```

Deterministic pass in isolation → the suppression hook and assertion logic are sound; the failure is a full-gate ordering / pollution hazard where the row leaks past suppression.

## Related

- Premise-gate-races-delivery precedent (closed): #186.
- Same iso-green / gate-red-under-load profile: #519 (sibling ux-2 flake), #507.
- Feature this spec guards (closed): #159.

## Likely direction

Make the Phase-3 gate robust to cross-test residue: assert the suppression hook actually took effect before posting, and/or key the gap on a value that cannot pre-exist, rather than a bare settle-window + `toHaveCount(0)`. Investigation open.

vjt:
[...]

Issue #195, opened 2026-07-06
cicchetto: replace long-press-to-close with an explicit 'Leave #channel?' confirm modal (revert long-press)


vjt:
### Problem
Closing / leaving a channel is currently bound to a **long-press** gesture. It's poor UX: easy to mis-trigger, no confirmation, so a channel can get left by accident.

### Requested change
- **Revert the long-press-to-close gesture.**
- On close/leave, show an explicit confirmation modal: **"Do you want to leave #channel?"** with **Yes** / **Cancel** (no destructive default).

### Rationale
Prevent accidental `PART`s; make leaving a channel a deliberate, confirmed action.

_Filed per channel request (vjt)._

vjt:
Field report (Android, per channel request): the current hold-to-close gate (#172, `holdToClose.ts`, `HOLD_TO_CLOSE_MS = 500`) reads as **broken** to touch users — a plain tap on the close × does nothing, so users perceive "the X stopped working a few days ago" rather than "close now needs a 500ms hold". This is the exact UX regression this issue proposes to fix.

Reproduced on Android Chrome: quick tap → `swallowClick` preventDefaults the trailing click and the 500ms timer never elapses → no close, no feedback. Only a sustained hold closes.

Reinforces the plan here: revert the long-press gesture and replace it with an explicit "Leave #channel?" confirm modal (Yes / Cancel, non-destructive default). A visible modal also gives the tap the immediate feedback the silent hold-gate lacks.

vjt:
Field report escalation (Android, via #sniffo user): holding the X for even ~10s does **not** close the tab — the hold-to-close handler from #172 apparently never fires on Android touch. So on Android the X is fully non-functional: neither tap nor long-press closes a channel tab. That is a blocking regression, not just a changed-UX annoyance — Android users currently cannot close/leave a channel from the tab at all.

Bumping priority accordingly. The revert-to-modal fix here also resolves this, since it removes the long-press path entirely.

vjt:
Correction / scope narrowing: a second Android user reports the hold-to-close **works** for them (channel closes on ~1s press). So this is **not** a blanket Android regression — two Android users, one works, one does not. The failing case is narrower: likely a specific Android version / browser (WebView vs Chrome vs Firefox) or a competing gesture intercepting the long-press.

Needs a proper repro (Android version + browser + build) before treating it as blocking. Dropping back from P1 until repro isolates it. The revert-to-modal fix still resolves it either way by removing the long-press path.

[...]

PR #679, opened 2026-08-02
fix(#653): bucket B — deterministic e2e setup for #277/#520/#530/#534

vjt:
Epic #653 **bucket B** — the e2e setup/protocol non-determinism cluster
(#277 #520 #530 #534). All four are the same profile: **green in
isolation, red under full-gate load**. Cure pattern (locked by the epic):
make the SETUP deterministic — unique-per-run id, observable-settle wait,
bounded retry on an OBSERVABLE signal. **No assert weakened, no blind
sleep/retry/timeout bump.**

### #277 — testnet nick-release race (`resetSubject` 500s on 433)
The `_vjtReset` teardown restarts vjt's Session.Server → reconnect →
re-register `vjt-grappa`. Under load an earlier spec's upstream session
hasn't released the nick, so the re-register hits 433 and the reset 500s
(`{:client_exit, {:nick_rejected, 433, ...}}`), reddening the NEXT spec's
beforeEach (issue195:71 was the victim). Fix: retry the whole reset while
the server reports the nick still taken (500 + `nick_rejected`) with
bounded/capped backoff until it frees (204). NOT a blind pre-sleep — every
other non-204 (404/504/non-nick 500) throws immediately, so a real reset
failure is never masked.

### #520 — freshness-on-activation Phase-3 premise leaks under load
The premise gate asserts the during-gap row is ABSENT before the fix runs.
Under load a residual `msg-159-*-during-gap` row from a prior execution
(retry, or a slow `_vjtReset` truncate losing the DB-lock race) satisfies
the `hasText` substring. Fix: key both message bodies on a per-run UUID so
they cannot collide with any neighbour's leftover state. Suppression hook
is synchronous and already effective; iso unaffected.

### #530 — nick-follow-query relabel fails under load
`event_router.ex` emits `{:peer_nick_renamed}` (the relabel under test)
ONLY when the renamer is a tracked member of a shared channel
(`channels != []`). `peer.join` awaits only the peer's own echo, not
grappa's observation of it; under load grappa processes the #bofh JOIN
late → no rename effect → no relabel. Fix: after `peer.join`, wait for the
peer's JOIN line to render in our #bofh scrollback — grappa broadcasts it
from the SAME apply that adds the peer to `state.members`, so it is the
observable proof the shared-membership precondition holds. #373
rename-order assertion untouched.

### #534 — stray `#bofh` gap-fill trips the #81 `/messages` guard
The guard asserts ZERO `/messages` while selecting the `$list` window. A
legitimate forward gap-fill on the live autojoin `#bofh`
(`/channels/%23bofh/messages?after=...`) fires inside the collection
window and trips a global-zero collector. Fix: scope the collector to the
`$list` window's channel key — a `kindHasScrollback("list")` regression
would fetch `/channels/%24list/messages`, so recording only that path
ignores the `#bofh` gap-fill while still failing on a genuine `$list`
fetch. NOT a `toHaveLength(<=1)` relaxation nor a blanket substring filter.

### Gates
- biome (e2e) clean.
- Per-spec iso `--project chromium --grep "#159|NICK change|#81 guard|#195" --repeat-each 3` — exit 0 (all repeats green).
- **Full `scripts/integration.sh` × 2 consecutive: 584 passed (21.7m) then 584 passed (22.2m), 0 failed / 0 flaky both runs.** All four target specs green in both. Chromium + webkit-iphone-15. recover-identity lines 163/193 green (no out-of-scope red materialised).

Refs #653

### DIFF
--- a/cicchetto/e2e/fixtures/grappaApi.ts
+++ b/cicchetto/e2e/fixtures/grappaApi.ts
@@ -665,17 +665,39 @@ export async function resetSubject(
     );
   }
 
-  const res = await fetch(`${GRAPPA_BASE_URL}/admin/test/reset-subject`, {
-    method: "POST",
-    headers: {
-      authorization: `Bearer ${adminToken}`,
-      "content-type": "application/json",
-    },
-    body: JSON.stringify(body),
-  });
-  if (res.status !== 204) {
+  // #277/#653 — the reset restarts vjt's Session.Server, which reconnects to
+  // bahamut-test and re-registers `vjt-grappa`. Under full-gate load an
+  // earlier spec's upstream session may not have released the nick yet, so
+  // the re-register hits 433 ERR_NICKNAMEINUSE and the reset 500s with
+  // `{:client_exit, {:nick_rejected, 433, "vjt-grappa"}}`. That 433 is the
+  // OBSERVABLE "nick not free yet" signal from the testnet — retry the whole
+  // reset on it with bounded backoff until bahamut releases the ghost (204).
+  // This is NOT a blind pre-sleep: we retry ONLY while the server reports the
+  // nick still taken, and ONLY that one signal — every other non-204
+  // (404 user_not_found, 504 timeout, any other 500) throws IMMEDIATELY, so a
+  // real reset failure is never masked. Known bahamut ghost-race tail (the
+  // #268 class; see feedback_integration_bahamut_ip_autokill_flake).
+  const maxAttempts = 8;
+  for (let attempt = 1; attempt <= maxAttempts; attempt++) {
+    const res = await fetch(`${GRAPPA_BASE_URL}/admin/test/reset-subject`, {
+      method: "POST",
+      headers: {
+        authorization: `Bearer ${adminToken}`,
+        "content-type": "application/json",
+      },
+      body: JSON.stringify(body),
+    });
+    if (res.status === 204) return;
+
     const text = await res.text().catch(() => "<no body>");
-    throw new Error(`resetSubject(${userName}) failed: ${res.status} ${text}`);
+    const nickStillTaken = res.status === 500 && text.includes("nick_rejected");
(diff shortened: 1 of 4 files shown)


vjt:
## Gate: `584 passed`, **1 failed** — re-running the job before judging it

```
✘ 300 [chromium] › tests/issue580-send-snap-independent-of-post.spec.ts:104:3
      › issue #580 — own send snaps to the bottom independent of the POST
      › send whose POST fails still snaps to the bottom (case 1) (6.4s)

  Error: expect(locator).toBeInViewport() failed
  > 176 |     await expect(sentLine).toBeInViewport();
```

**I am not archiving this as "spec not touched".** This PR's four files are
`grappaApi.ts`, `channel-directory.spec.ts`, `freshness-on-activation.spec.ts`,
`nick-follow-query.spec.ts` — and `grappaApi.ts` is a **shared auto-fixture**. A fixture-level
change runs for *every* test, so its blast radius is the whole suite and it surfaces in whatever
spec happens to be running. Matching a flake by spec name is exactly the mistake #277 itself
taught us not to make; the match has to be by mechanism.

So, by mechanism: the #277 change (`74d2ed8a`) retries `resetSubject` **only** on the 433
nick-still-taken signal, with every other non-204 throwing immediately. It adds latency only when
bahamut genuinely has not released the ghost. A slower reset delaying a still-settling session into
a scroll assertion is a *plausible* path to a `toBeInViewport()` miss — but plausible is not
proven, and I am not going to argue a code path against a single sample.

**A red that reproduces beats any code-path argument; a red that does not reproduce beats any
statistic.** So I re-ran the failed job (`gh run rerun 30764667860 --failed`):

- **green** ⇒ transient, and this PR merges.
- **red on the same spec** ⇒ causal, and it goes back to w2 to fix at the setup level —
  not by weakening `toBeInViewport()` and not by bumping a timeout.

#653 stays `cooking` until that settles.

Issue #442, opened 2026-07-26
auth: two-factor for account login — TOTP, then passkeys/WebAuthn


vjt:
Account login is single-factor today: `POST /auth/login` takes `{identifier, password?}` and, when `IdentifierClassifier` routes the identifier to the account path, `mode1_login/3` checks the password via `Accounts.get_user_by_credentials/2` and mints a session bearer. There is no second factor anywhere in that path, and a stolen password is a full account takeover — including every network credential the user has attached, i.e. their IRC identity on every server grappa connects to on their behalf.

Requested on IRC (#it-opers) and accepted for the roadmap by @vjt: **TOTP and passkeys**, in that order of implementation risk.

## Scope

**In scope — account login only.** Visitors have no durable secret (`password_encrypted == nil` for anon visitors) and their attach story is the share-token flow, not a password. A second factor there has nothing to protect. Registered-visitor login (`visitor_login/6`, password optional) is deliberately left out of this issue until the visitor/account boundary is revisited.

### 1. TOTP (RFC 6238)

- Per-user secret, enrolled from the settings drawer: QR + manual key, confirm with one code before the factor is armed.
- Login becomes two-step for enrolled users: password verified → short-lived pending-2FA state → code verified → session bearer minted. **The bearer must not exist before the second factor passes**; a partially-authenticated token is the classic way this gets bypassed.
- Recovery codes, one-shot, shown once at enrollment. Without them a lost phone means an operator-side reset, which on a self-hosted instance means "the admin edits the DB".
- Replay: a consumed code must not be accepted again inside its window.

### 2. Passkeys / WebAuthn

Larger, and it is not "TOTP with a different widget":
- Needs a WebAuthn ceremony (registration + assertion) with credential storage per user, and an Elixir library decision — there is no WebAuthn support in the tree today.
- **The RP ID is bound to the origin.** grappa is self-hosted; each instance is its own relying party, and a passkey enrolled on one instance is meaningless on another. This must be stated in the self-hosting docs or it will be reported as a bug.
- shottino (the terminal client) cannot run a WebAuthn ceremony. If passkeys become the only factor a user has, that user can no longer log in from the CLI. Either a TOTP factor stays mandatory as the CLI path, or shottino gets a device-authorization flow (browser approves, terminal polls). This is a product decision, not an implementation detail — it needs an answer before passkeys ship.

### Out of scope here

[...]

Issue #340, opened 2026-07-19
perf(grappa): serialize+batch scrollback writes through a dedicated writer (SQLite single-writer saturation)


vjt:
Durable fix for the write-contention behind the #336 incident (recurred today 14:11 under a single client flood).

## Problem
Under a burst of `PRIVMSG` from one client, `Grappa.Scrollback.persist_event/1` fires one INSERT per event. All `pool_size: 10` connections contend on SQLite's single file-writer; Ecto pool checkouts queue-timeout (~160ms observed), the session GenServer raises `DBConnection.ConnectionError reason=:queue_timeout` and crashes → mass disconnect + reconnect wave. #336 makes persist *degrade instead of crash* (robustness). **This issue removes the contention at the source** so bursts stop saturating the pool.

## Already tuned (not the lever)
`config/runtime.exs:63-96`: `journal_mode: :wal`, `synchronous: :normal`, `busy_timeout: 30_000`, `pool_size: 10` (pinned by the 2026-05-22 REV-B/C3 review). PRAGMA-level tuning is done — the fix is architectural.

## Proposed
Route **all** scrollback writes through a single **dedicated writer process** (GenServer) with a bounded mailbox:
- Session processes `cast` events fire-and-forget — never block on DB latency, never crash on it (mirrors the already-hardened fire-and-forget `:reply` effect in `server.ex` ~3669).
- The writer **batches** pending inserts into one transaction (coalesce by count/time window, e.g. N events or M ms). Under burst you do 1 batched txn instead of 10-way pool contention; only one connection ever writes, so writes never exhaust the pool.
- Retry/backoff on transient errors inside the writer; backpressure or drop-oldest if the mailbox exceeds a cap (bounded memory).

## Explicitly NOT migrating to Postgres/MySQL now
WAL + a single serialized batched writer is the canonical high-throughput SQLite pattern and keeps the lean single-file / low-RSS ops story. Revisit an RDBMS only if a single batched writer still can't keep up (buys concurrent writers at the cost of a migration + an extra daemon to run/backup; for Ecto/Elixir, Postgres > MySQL).

## Refs
- `config/runtime.exs:63-96` (pool + PRAGMAs)
- `lib/**/scrollback.ex` `persist_event/1` (the `:persist` effect, ~:125-128)
- `server.ex` ~3669 (fire-and-forget `:reply` precedent)

Priority note: the acute crash is P0 and already covered by #336; this is the P1 durable throughput fix.

[...]

PR #688, opened 2026-08-02
auth: add TOTP second factor for accounts

abonforti:
## Summary

Implements the TOTP portion of #442 for durable account logins. Passkeys/WebAuthn remain a separate follow-up.

- add RFC 6238 enrollment with QR/manual secret and confirmation before arming
- return a short-lived post-password challenge; mint no bearer before TOTP/recovery verification
- encrypt TOTP secrets with the existing Cloak vault
- generate 10 one-shot 128-bit recovery codes and persist SHA-256 hashes only
- prevent TOTP replay with an atomic last-used-step update
- rate-limit second-factor failures per IP/account
- revoke other sessions atomically when enabling or disabling TOTP
- require the account password to disable TOTP
- add Settings > Security UI and generated wire error tokens

Refs #442

## Automated tests

- `scripts/test.sh test/grappa/accounts/totp_test.exs test/grappa_web/controllers/totp_controller_test.exs test/grappa_web/controllers/auth_controller_test.exs:176 test/grappa_web/controllers/auth_controller_test.exs:208 test/grappa_web/controllers/auth_controller_test.exs:227` — 14 tests, 0 failures
- rollback test forces session-revocation failure and verifies secret, enabled state, recovery codes, and revocation all roll back
- `scripts/test.sh test/grappa/migrations/add_user_totp_test.exs` — upgrades a legacy DB stopped at `20260729130000`, populated with 5,000 users and 5,000 sessions; verifies row counts, password hashes, sessions, null TOTP defaults, recovery table, and account lookup remain intact
- `scripts/bun.sh x vitest run src/__tests__/Login.test.tsx src/__tests__/auth.test.ts` — 48 tests, 0 failures
- `scripts/bun.sh x tsc --noEmit`
- `mix compile --warnings-as-errors` via repo wrapper
- targeted `mix credo --strict` via repo wrapper — no issues
- `git diff --check`

Full `scripts/check.sh` is currently blocked by pre-existing Cowboy/Cowlib dependency advisories. A broad AuthController run also exposed the existing IRC-connect expectation flake (`502 upstream_unreachable` expected, `503 connect_timeout` observed); TOTP-targeted cases are green.

## Human smoke tests

Validated on a fresh Docker instance on Linux ARM64 (`jumphost`), using a real authenticator app:

- enrolled from Settings > Security by scanning the QR code
- confirmed enrollment with a live TOTP
- logged out and logged back in with password + TOTP
- logged in with a recovery code, then verified reuse was rejected
- disabled TOTP using the account password
- verified subsequent password login no longer requested a second factor
- verified the same-origin production-style static bundle kept the WebSocket connected


vjt (changes requested):
Reviewed at `613447f2`, line by line rather than on the CI signal. Read: `totp.ex`, `totp_recovery_code.ex`, the migration, `accounts.ex`, `auth_controller.ex`, `totp_controller.ex`, the router, and the cicchetto side. I did **not** run the suite — no Elixir toolchain on this host, so everything below is read from the code and from the CI logs, nothing is a test result I produced.

## What is right

- `matching_step/3` compares with `Plug.Crypto.secure_compare/2`, ±1 step window, and the HOTP truncation is the RFC 6238 one.
- Replay is closed by the conditional `update_all` on `totp_last_used_step < ^step` — a monotonic guard, not a read-then-write race.
- Recovery codes: 128 random bits, stored only as SHA-256, consumed by `delete_all` with a `{1, _}` match, so one use each.
- Enrollment secret is round-tripped inside a signed `Phoenix.Token` (600 s) and the embedded `user_id` is re-checked against the caller. The unarmed secret never has to be parked server-side.
- `:two_factor_not_enabled` and `:two_factor_replayed` are collapsed into `:invalid_two_factor` in `verify_second_factor/3` before they can reach `FallbackController`, so the two error atoms with no fallback clause cannot escape as a 500. Deliberate or lucky, it holds.
- Enabling and disabling both revoke every other bearer.
- Front end: the secret and the recovery codes stay in signals, never in `localStorage`, never logged. The new `qrSvgWithLabel` escapes the attribute it interpolates.

## Blocking

**1. `Repo.BusyRetry.run/1` is called from inside an open transaction.**

`Accounts.confirm_totp_enrollment/5` and `Accounts.disable_totp/3` both open a
`Repo.transaction`, and inside it call `revoke_other_sessions_for_user/2`, which wraps its
`update_all` in `Repo.BusyRetry.run/1`. Every call site of that function is inside a transaction.

That module's own moduledoc states the invariant it relies on:

> The backoff sleep runs after the failed checkout was already released, so it holds no connection — bounded backpressure on the flooding caller, not a held-conn leak (#340).

Inside a transaction the connection is checked out for the whole transaction, so that sentence
stops being true: the retry loop's `Process.sleep` — up to the full 1500 ms budget — runs while
this caller still holds the connection and, once any write has happened, the SQLite writer.
A transient contention becomes a self-inflicted stall of the single writer, which is the exact
pathology the module was extracted to avoid.

On top of that the retry cannot succeed: an `Exqlite.Error` raised inside a transaction leaves
that transaction aborted, so re-running the same statement burns the budget and returns
`{:error, :db_unavailable}` regardless.

Suggested shape: drop the `BusyRetry` wrapper from `revoke_other_sessions_for_user/2` and let the
outermost non-transactional caller own the retry, or drop the outer `Repo.transaction` and keep
the retry where it can actually release the connection.

**2. CI is red on a failure this PR introduces.**

`Doctor` fails with `Failed Modules: 1` on `Grappa.Accounts.TOTPRecoveryCode`
(0% doc coverage — `@moduledoc false` plus an undocumented `changeset/2`).

The other red step, the dependency audit (`cowlib 2.18.0`, `cowboy 2.17.0` advisories), is **not**
from this PR: `main` carries byte-identical versions and this branch does not touch `mix.exs` or
`mix.lock`. It needs its own decision — bump or ignore — on `main`, not here.

The branch is also 5 commits behind `main`; worth merging `main` in so the run reflects current `main`.

## Should fix

**3. The test named "confirm arms encrypted secret" does not test encryption at rest.**

It asserts `armed.totp_secret_encrypted == @rfc_secret`, i.e. the value after `Grappa.EncryptedBinary`
has already decrypted it on load. That assertion passes whether or not the bytes on disk are
ciphertext. It matters here more than usual: this is the **first** `update_all` write to a Cloak
field in the codebase (no other call site exists), so `update_all`'s dump behaviour on a custom
Ecto type is exactly what a reader wants proven. Add an assertion over raw SQL that the stored
blob is not the base32 secret.

**4. The post-password `challenge_token` is neither one-shot nor bound to the client.**

`Phoenix.Token.sign(endpoint, @totp_challenge_salt, user.id)`, 300 s, verified and then left
valid. Within its window it can be replayed or handed to another client, and it carries nothing
tying it to the IP, user agent or client id that passed the password check. Binding it to
`current_client_id` / the source IP, or marking it consumed on first successful verify, closes that.

## Nits

5. The `else` branches in `confirm_totp_enrollment/5` and `disable_totp/3` are dead: the nested
   `Repo.rollback/1` unwinds straight past them.
6. `disable_totp/3` puts the Argon2 verification inside the outer transaction. It lands before the
   first write, so no lock is held yet, but it is fragile to reordering.
7. A successful TOTP does not clear the `FailureWindow` counter, so a user who fumbles nine codes
   and then gets one right stays one failure away from a lockout for the rest of the 15 min window.

vjt:
Follow-up on blocking point 1, since it was raised that this path is already covered by tests.

The functional tests do pass, and that is not what the finding disputes — it only bites under
`SQLITE_BUSY`, and nothing on this path injects one:

- `test/grappa/accounts/totp_test.exs` — 0 occurrences of `arm_faults` / `inject_transient_faults`
- `test/grappa_web/controllers/totp_controller_test.exs` — 0
- `test/grappa_web/controllers/auth_controller_test.exs` — the only fault-injection test is the
  pre-existing visitor-login one (`"visitor login → 503 db_unavailable (not 500) under sustained
  transient DB busy"`), which does not reach `confirm_totp_enrollment/5` or `disable_totp/3`.

So the enrollment and disable paths are exercised only in the no-contention case, where the retry
loop never runs and the difference between "inside a transaction" and "outside" is invisible.

The precedent in the codebase also points the other way. `Accounts.revoke_sessions_for_visitor/1`
on `main` is the direct sibling of the new `revoke_other_sessions_for_user/2` — same wrapper, same
bulk revoke — and it sits at the login door, outside any transaction. Its comment leans on exactly
that:

> A single idempotent UPDATE (a re-run re-matches `is_nil(revoked_at)`, so the now-already-revoked rows set to zero), so a retried statement is safe.

"Safe to retry" is a statement about a statement re-run on its own, not about one re-run inside a
transaction that the first raise already aborted.

A test that arms a transient fault so it fires on the revoke inside `confirm_totp_enrollment/5`
would settle it either way. If it comes back green with the connection released as the moduledoc
describes, I withdraw the point.

abonforti:
Addressed the review in `a08b3298` (after merging current `main`):

- moved `BusyRetry.run/1` around the complete enrollment/disable transaction; session revocation no longer retries while holding a transaction checkout
- moved password verification before the disable transaction
- added transient-contention regression tests for both enrollment and disable
- added a raw SQL assertion proving the stored TOTP secret is ciphertext
- bound the post-password challenge to source IP and `X-Grappa-Client-Id`
- clear the TOTP `FailureWindow` after successful second-factor verification
- documented `TOTPRecoveryCode.changeset/2`, fixing Doctor

Local Docker validation:
- targeted TOTP/account/controller/rate-limit tests: 21 passed
- challenge-binding regression: passed
- `mix doctor`: 300 modules passed, 0 failed
- `mix format --check-formatted`: passed

A full `mix test --warnings-as-errors` run was started and showed no failures before being intentionally stopped because the long integration-heavy suite is delegated to CI.

vjt:
Re-reviewed at `a08b3298`.

Every point from the first pass is addressed, and addressed in the right way:

- `BusyRetry.run/1` now wraps the whole `Repo.transaction`, so the raise unwinds the transaction
  and releases the connection *before* the backoff sleep. That is the shape the module's contract
  assumes. The two fault-injection tests (`inject_transient_faults(1)` on enrollment and on
  disable) are exactly the missing coverage, and they assert the session-revocation side effects
  survive the retry — so they pin the whole transaction being replayed, not just a return value.
- The at-rest assertion is now real: `SELECT totp_secret_encrypted FROM users` plus
  `refute stored_secret == @rfc_secret`. That closes the open question about `update_all` dumping
  through a Cloak type.
- Challenge tokens are bound to `{user_id, ip, client_id}` and the binding is verified, with a
  cross-client test that asserts a 401 *and* `session_count() == 0`.
- Argon2 moved out of the transaction, and a successful TOTP clears its failure window.

Two things left, neither of them the concurrency issue:

**1. CI is red on Credo (`--strict`, exit 8), and it is this change:**

```
[F] → Function body is nested too deep (max depth is 2, was 3).
      lib/grappa/accounts.ex:603:9  #(Grappa.Accounts.confirm_totp_enrollment)
      lib/grappa/accounts.ex:621:26 #(Grappa.Accounts.disable_totp)
```

The `BusyRetry.run(fn -> Repo.transaction(fn -> case … end) end)` stack is three deep. Extracting
the transaction body into a private function per call site fixes both. Doctor and dialyzer now
pass, and the dependency-audit noise from the previous run is not failing this one.

**2. `TOTP.disable/1` is public and skips the password check.**

The split left a `@doc false`, `@spec disable(User.t()) :: User.t()` clause that performs the
whole teardown — secret cleared, recovery codes deleted — with no credential check, on a module
the `Accounts` boundary exports. The only correct caller is `Accounts.disable_totp/3`, which
verifies Argon2 first, but nothing in the code says so; the next caller to reach for
`TOTP.disable(user)` gets a passwordless 2FA teardown and the compiler will not object.

Either make it private and give `Accounts` a transaction-body helper, or name it for what it is
(`disable_without_credential_check!/1`) so a call site has to be deliberate.

Minor, while you are in there: `TOTP.disable/2` now has no production caller — `disable_totp/3`
does its own Argon2 and calls the arity-1 clause — so it is reachable only from
`totp_test.exs:62-63`. And `FailureWindow.clear/2` writes ETS directly while `record_failure/4`
is serialized through the GenServer; harmless here (worst case a stale count) but it is the one
write that skips the owner process.

Happy to re-check once Credo is green.

abonforti:
Addressed the follow-up in `57433e9e`:

- extracted enrollment/disable transaction bodies; `mix credo --strict` now passes with 0 issues
- removed both public `TOTP.disable/1` and the now-unused `TOTP.disable/2`; teardown is private inside the password-verified `Accounts.disable_totp/3` flow
- routed `FailureWindow.clear/2` through its GenServer owner
- updated disable coverage to exercise the public authenticated Accounts API

Local Docker gates:
- compile with warnings as errors: passed
- Credo strict: 0 issues across 568 files
- Doctor: 300 passed, 0 failed
- targeted TOTP/controller/rate-limit tests: 22 passed, 0 failed

Integration suite was not run locally.

Issue #284, opened 2026-07-17
cicchetto: password field always-visible + optional (drop two-step / Advanced gating)


vjt:
### Goal (re-scoped 2026-07-17 by vjt — supersedes the two-step login design)

**Cancel the two-step login.** No conditional second screen, no credential-existence check. Instead: put the
**password field back OUTSIDE the Advanced section, ALWAYS visible, ALWAYS optional.** Simplify.

### Why the two-step gating is unnecessary (verified in code)

The conditional ("only prompt for a password when the identifier has a password credential") existed to avoid asking for
a password that would be wasted. But it is **not** wasted: grappa already fires an entered password to NickServ as an
IDENTIFY on first connect **even when there is NO pre-existing credential**. Verified:

- `lib/grappa/visitors/login.ex:509` — the entered password threads through the fresh-visitor login path
  (`spawn_and_await(visitor, network, input.password, timeouts)`).
- `lib/grappa/visitors/session_plan.ex:96-102` (`with_login_identify/2`) — a fresh anon plan carrying a password is
  rewritten to `auth_method: :nickserv_identify`.
- `lib/grappa/irc/auth_fsm.ex:376-379, 559-562` — at numeric 001, `maybe_nickserv_identify` sends
  `PRIVMSG NickServ :IDENTIFY <pw>`.
- `lib/grappa/session/server.ex:3626-3640` — on `+r` MODE observed, `Visitors.commit_password/3` persists it.

So any user entering nick+password authenticates first-try; **no branch drops the password**. The two-step conditional
adds friction for zero benefit.

### Spec (simplified)

- Password field: **always rendered on the main login form** (not behind Advanced), labelled **optional**.
- Advanced section keeps only ident + realname (as before).
- Empty password → anonymous / no-identify login (unchanged).
- Non-empty password → fired to NickServ IDENTIFY at 001 (existing path, above).

### Deploy scope

cic-only (login form layout). No server change — the identify path already exists.

vjt:
Shipped to prod grappa.chat in the 4am batch COLD deploy (origin/main 2fa3d749): migrate (themes table + system_user) + cic bundle rebuild + BEAM restart + theme gallery seeded (13 built-ins). Public 200, dual-net announce posted.

PR #696, opened 2026-08-02
auth: add passkey login modes

abonforti:
Refs #442

## Summary

- add WebAuthn passkey registration and per-account credential storage
- let users choose password + passkey or passwordless passkey login
- require recovery codes before passwordless activation; activation remains unchanged until the user confirms the saved codes with a passkey assertion
- consume recovery codes once and provide `mix grappa.reset_passkeys` for operator recovery
- revoke other sessions when changing authentication mode
- bind ceremonies to RP origin, account, client/IP, purpose, and one-shot challenges
- keep credential discovery client-side so browser password managers such as Proton Pass can handle assertions
- document that each self-hosted instance is a separate RP and passwordless disables shottino login

## Beta status

Treat passkeys as beta for the first release. Desktop Firefox + Proton Pass received a complete human smoke test. Responsive iPhone viewport passed, but real-device Safari/iOS and Android provider flows were not available locally. User feedback from both iOS and Android passkey providers is requested before removing the beta designation.

## Automated validation

- targeted ExUnit passkey, challenge-store, controller, origin-regression, migration, and env-registry tests: 13 passed
- broader passkey/migration set before the final origin regression: 7 passed
- frontend `biome check` + `tsc --noEmit`: passed; only existing warnings
- Vitest auth + Login: 48 passed
- Doctor: 307 modules passed, 0 failed, 100% spec coverage
- Credo strict full-tree run: one alias-order finding; corrected and rechecked on the affected file with 0 issues
- populated legacy DB migration test preserves users and existing TOTP recovery rows

The local integration suite was intentionally skipped because of its runtime cost; CI remains the integration authority.

## Human smoke tests

Using Firefox, Proton Pass, Docker-backed Grappa, and a localhost SSH tunnel:

- passkey enrollment succeeded
- password + passkey mode activation succeeded
- password + passkey login succeeded after logout
- passwordless recovery codes appeared before activation and mode remained unchanged until confirmation
- passwordless activation succeeded
- password login was rejected in passwordless mode
- passkey-only login succeeded
- recovery-code fallback succeeded
- reuse of the consumed recovery code was rejected
- iPhone responsive viewport flow/layout succeeded

## Product behavior

Passwordless means passkey is primary. Password and TOTP login are disabled; recovery codes are the only user fallback. Losing both passkey and recovery codes requires operator reset. Shottino cannot authenticate in passwordless mode.


vjt:
Merge authorized by the maintainer with "if there are changes to make, we make them later":
this goes in as soon as CI is green, and any corrections land in follow-up PRs.

Beta exit criteria, asked for by staff in channel and written down here so it does not stay
verbal: two reports from **real devices**, one iOS and one Android, on staging. Minimum path
for each: passkey enrollment, activation of passwordless mode with the recovery codes, logout,
login with the passkey alone, fallback with one recovery code, and confirmation that the same
code cannot be reused. Until both reports are in, the beta designation stays.

Note for whoever tests: passkeys are bound to origin and device, so the ones registered on
staging do not work in production and every tester has to enroll from their own phone — a
shared test session does not cover this.

vjt:
CI is red on two jobs. Triage with the measured causes (run `30772452383`).

### 1. `test + lint + audit` — Sobelow, High Confidence

```
Misc.BinToTerm: Unsafe `binary_to_term` - High Confidence
File: lib/grappa/accounts/webauthn.ex   Line: 145   Function: credential_tuple/1
{passkey.credential_id, :erlang.binary_to_term(passkey.public_key, [:safe])}
```

The gate is `fail on Medium+`, so this one finding is enough to keep the pipeline red
(the two `Traversal.FileModule` hits in `uploads/reaper.ex` are Low and do not count).
The input comes from a column written only by our own registration path and `[:safe]`
is already there, so the options are either a targeted skip
(`# sobelow_skip ["Misc.BinToTerm"]`) with the reasoning written next to it, or
serializing the COSE key as CBOR — `:cbor` is already a dependency — and dropping
`term_to_binary` from storage altogether.

### 2. `dialyzer` — 8 errors, all in the new code

```
lib/grappa/accounts.ex:620: extra_range + missing_range on consume_recovery_code/2
lib/grappa_web/controllers/passkey_controller.ex:169,178: invalid_contract
lib/grappa_web/controllers/passkey_controller.ex:170,179: no_return
lib/grappa_web/controllers/passkey_controller.ex:171,180: call authenticate will not succeed
```

The six controller errors have **a single cause**, and it is a stale typespec:

```elixir
# lib/grappa/accounts/webauthn_challenge_store.ex:7
@type purpose :: :registration | :authentication | :mode_change
```

The controller calls `WebAuthn.authenticate(params, :passwordless, ...)` and
`(..., :second_factor, ...)`, and those atoms are passed verbatim to
`WebAuthnChallengeStore.take/2`. Neither is in the union, so Dialyzer concludes that
`take/2` can never succeed; from there `authenticate/3` has no local return, and the
two controller functions become `no_return` with invalid `@spec`s.

**At runtime it does work**, and your smoke test is not lying: `begin_authentication/5`
does the `put` with the **same** atom that `authenticate/3` later hands to `take`, so
the match in `handle_call({:take, ...})` lands. It is the union that fell behind the
new purposes — it needs `:passwordless` and `:second_factor` added (and it is worth
checking whether `:authentication` still has any caller).

The seventh and eighth are the `@spec` of `consume_recovery_code/2`: it declares
`:ok | {:error, :invalid_recovery_code}`, but the function re-propagates whatever error
it gets from `RecoveryCodes.consume/2` and `Repo.BusyRetry.run/1` — which includes the
database-unavailable case — while `:invalid_recovery_code` is produced by nobody. It
should be realigned with what actually happens, not widened to `term()`.

None of this touches the ceremony logic: it is one security gate to decide on and two
specs to realign. The merge is still authorized by the maintainer, but only on green CI.

vjt:
CI triage for `2a6045c0` ("fix(auth): pass passkey CI gates"): Dialyzer and Sobelow are now **green** — the `@type purpose` and `@spec` fixes worked. `shottino` green too. What is left is `test + lint + audit`: **4 failures out of 4961 tests**, all in the repo's own drift gates, none in the WebAuthn ceremonies themselves. `cicchetto + grappa + azzurra-testnet` is still running.

**1. `error_tokens_drift_test.exs:33` — REST wire token not declared**

```
Emitted by lib/grappa_web/** but MISSING from the type (add them):
  ["passkey_required"]
```

Fix: add `"passkey_required"` to `GrappaWeb.ErrorTokens.rest_error_token/0`.

**2. `fallback_controller_test.exs:273` — `@spec` ↔ clause lockstep**

```
call/2 clauses missing from @spec union (add them to the spec): [:passkey_required]
```

Same root cause as (1): the new error atom reached the clause heads but not the spec union. Fix: add `:passkey_required` to the `call/2` error union.

**3. `application_supervision_tree_test.exs:33` — undocumented supervised child**

```
CLAUDE.md "Top-level supervision tree" is missing supervised child(ren):
  ["Grappa.Accounts.WebAuthnChallengeStore"]
```

Fix: add `Grappa.Accounts.WebAuthnChallengeStore` to the fenced tree block under `## Architecture` in `CLAUDE.md`, with its one-line why-note. Separately (not enforced by the gate) add its ordering why-comment in `application.ex`.

**4. `add_user_totp_test.exs:44` — pre-existing test broken by the new column**

```
** (Exqlite.Error) no such column: u0.passkey_mode
SELECT u0."id", ..., u0."passkey_mode", ... FROM "users" AS u0 WHERE (u0."name" = ?)
code: SmokeRepo.get_by(Grappa.Accounts.User, name: "legacy-04200")
```

This one is not a missing declaration, it is a structural conflict. That test migrates a seeded legacy DB **only up to `@totp_version = 20_260_802_190_000`** and then reads a row through the `Grappa.Accounts.User` schema. The schema is always at HEAD, so as soon as this PR adds `passkey_mode` in a later migration, the HEAD schema selects a column the pinned-version database does not have yet.

Recommended fix — drop the schema from the assertion so the fixture stops depending on future migrations:

```elixir
assert ["legacy-secret"] ==
         query!("SELECT password_hash FROM users WHERE name = ?", ["legacy-04200"]).rows
         |> List.flatten()
```

Migrating this test up to the passkey version instead would make it green today but would move the breakage to whoever adds the next column; the raw-SQL read is version-agnostic by construction. Either way the migration itself is fine — the failure is in the test's read path, not in the upgrade.

Merge is already authorised by vjt, conditional on green CI. I am watching the checks: say the word and I can push the four fixes above myself, otherwise I will re-triage on your next push.

vjt:
## e2e triage — head `6a7b947b`: 2 real regressions + a CI timeout that is not this PR's fault

`dialyzer`, `shottino` and `test + lint + audit` are **green** on this head — the anti-drift gates from the previous round are all closed. The `cicchetto + grappa + azzurra-testnet` job is red for two separate reasons; only one of them is a code problem.

### 1. Two real regressions, both caused by the new login buttons

Same two specs failed on **two consecutive runs** of this branch (`30772452358` @ `23:43`, `30774024337` @ `00:27`), on different heads. Not a flake. The control: run `30774056306` (`676-nick-fallback`) executed on the same runner image in the same time window and both specs were **green** there, so it is not environmental either.

I could not apply the `docs/TESTING.md` iron rule (3× isolated re-run) — the testnet images `ghcr.io/vjt/{bahamut,services}:main` are published amd64-only and the box I am on is aarch64. The determinism evidence above is the substitute, and the failure report from run `30772452358` is complete, so the root causes are not guesswork.

#### 1a. `issue281-account-switch-no-replay.spec.ts:111` — strict mode violation

```
Error: locator.click: Error: strict mode violation: locator('.login-advanced-toggle') resolved to 3 elements:
    1) <button ... aria-expanded="false" class="login-advanced-toggle" aria-controls="login-advanced">▸ Advanced</button>
    2) <button type="button" class="login-advanced-toggle">Sign in with passkey</button>
    3) <button type="button" class="login-advanced-toggle">Use recovery code</button>
```

The two new buttons in `Login.tsx` borrow `login-advanced-toggle` for its styling, but that class is a **selector contract**, not a skin. `main` already uses it twice — the disclosure toggle and the "Back" button in the TOTP challenge branch — and that is fine only because those two branches are mutually exclusive, so at most one is ever in the DOM. This PR is the first to put three of them in the *same* branch, which is what breaks the strict locator.

**Fix (production, not the test):** give the alternative-auth buttons their own class, e.g. `login-alt-auth`, and copy the four style declarations. No test change needed. `issue204-foolproof-login.spec.ts:61` also reaches for this class, via `document.querySelector` — it survives by taking the first match, but it is one more reason not to overload the name.

#### 1b. `login-advanced-scroll-reachability.spec.ts:103` — `realname` no longer reachable

```
Expected: in viewport
Received: viewport ratio 0
Locator:  getByLabel(/real name/i)
```

This is the `#284` guard: at maximum scroll on a 390×480 viewport, both `Connect` **and** the last disclosure field above it must be on screen, so that a regression clipping only the top of the disclosure cannot hide behind a visible `Connect`. The two new buttons sit *after* `Connect`, so the bottom-scroll position now ends on "Use recovery code" and pushes `realname` off the top edge.

**Fix:** move the alternative-auth block **above** the Advanced toggle — `Connect` should stay the last element of the form. That is also the better UI: "Sign in with passkey" is a peer of the primary action, not a footnote after it. Please do not relax the spec — it defends a real clipping bug.

### 2. The 30-minute timeout is pre-existing debt, not a passkey problem

```
The job has exceeded the maximum execution time of 30m0s
```

The job was killed at test **589 of 590**, with only the two failures above. The workflow comment says:

> the suite settled at ~23min of real work, which left a ~2min margin

That is no longer true. The green control run `30774056306` did **590 passed (25.5m)** of tests and the job took **29m55s** wall clock — five seconds under the cap. Any branch that loses a handful of seconds to runner variance now goes red for reasons unrelated to its diff, and the report never uploads because the cancellation kills the artifact step (which is why this triage had to be reconstructed from the previous run's log).

By the workflow's own stated rationale — *"a timeout is a runaway-job backstop, not a performance budget"* — `timeout-minutes` should go to 45. That is a separate change and I have not touched it; worth its own issue if you would rather shard the suite than raise the cap.

### Summary

| Check | State | Action |
|---|---|---|
| `dialyzer` | pass | — |
| `shottino (build + tests)` | pass | — |
| `test + lint + audit` | pass | — |
| `e2e` — 2 specs | **real regression** | 1a + 1b, both in `Login.tsx` |
| `e2e` — job timeout | pre-existing | raise `timeout-minutes`, separate change |

Both fixes are small and confined to `Login.tsx` plus four lines of CSS. Say the word and I will push them to this branch.

vjt:
Pushed `532c03bb` — the remaining e2e red was one spec out of 590, and it is a height-budget problem, not a selector one.

**What failed:** `login-advanced-scroll-reachability.spec.ts:138` — the *tall viewport* case (1024x900, "open Advanced → every field + Connect is visible with no scroll"). Not a timeout: the job died at 28m17s, under the (now 45m) cap.

**Measured, not guessed** (Chrome via CDP, dev build, viewport pinned to 1024x900, disclosure open):

| variant | card height | Connect rect |
|---|---|---|
| `main` (no alt-auth) | 875px | fully in view |
| two stacked alt-auth rows | 983px | top=931 — entirely below the fold |
| both on one shared row | 932px | 880–924 — half clipped |
| **alt-auth riding the Advanced row** | **874px** | **822–866, fully in view** |

The card ships with **25px of slack** against a 900px viewport — less than one 44px tap target. So any new full-width row breaks that spec, and stacking tighter does not buy enough: at 932px Connect is only ~45% visible, which passes `toBeInViewport()` (ratio > 0) but plainly violates what the spec is guarding. The fix is to add no row at all.

**What `532c03bb` does:** the passkey / recovery buttons now share the existing 44px Advanced-toggle row — toggle on the leading edge, the pair on the trailing one. Zero added height, so both this spec and the 390x480 one keep main's exact geometry. At 390px width the row measures 264px against a 264px content box, i.e. no wrap. Labels shortened to "Passkey" / "Recovery code" to fit the 22rem card at that width.

`.login-advanced-toggle` still renders exactly one element per branch, so issue281's strict-mode click stays satisfied. `tsc --noEmit` and `biome check src` both exit 0 locally (37 pre-existing warnings). e2e is the CI's call — the testnet images are amd64-only and this box is aarch64.

**Unrelated but worth flagging:** nothing under `cicchetto/e2e/tests/` references the passkey or recovery-code UI at all, so the two new entry points ship without e2e coverage. Probably a follow-up issue rather than something to bolt onto this PR.

Issue #512, opened 2026-07-28
e2e: migrate ~19 admin specs to the openAdminConsole primitive (latent drawer-close click race)


vjt:
## The latent race

Opening the admin console goes through the settings drawer: the
`admin-console-entry` button's onClick fires `onClose()` (close drawer) +
`onOpenAdmin()` (mount AdminPane). The drawer then **slides out over a 200ms
`transform: translateX(0→100%)` transition** (`cicchetto/src/themes/default.css`),
z-index 100, anchored right.

Every admin spec today hand-rolls `page.getByTestId("admin-console-entry").click()`
and then interacts with the admin pane **immediately** — it only awaits
`admin-pane` visible, never the drawer's slide-out completing
(`.settings-drawer.open` → 0). So a caller that clicks a **right-side** admin tab
(e.g. the 9th tab, *Settings*) can land the click on the still-sliding drawer
instead of the tab. The tab's onClick is a **pure synchronous SolidJS signal**
(`AdminPane.tsx`), so a swallowed click simply never switches the tab — this is a
click-**delivery** race, NOT a product defect. Real users are unaffected (they
see the drawer close, then click; the switch is instant).

The race is **pre-existing and latent**. It surfaced as a ~9% flake in
`ux-6-b-admin-settings.spec.ts` (Settings = 9th/10 tabs = right side) only once
#508's iOS-scoped input font floor perturbed the layout timing enough to widen
the window. Root-cause proof: A/B on #508's single font hunk — with 14px inputs
4/45 iso fails, with 16px restored 0/70 (main + A/B), Fisher p≈0.022.

## Fix already shipped (#508)

#508 added the **single-door primitive** `openAdminConsole(page)` in
`cicchetto/e2e/fixtures/cicchettoPage.ts` (sibling of `openSettingsDrawer` /
`closeSettings`): it opens the drawer, clicks the entry, **awaits
`.settings-drawer.open` → 0**, then awaits `admin-pane` visible. `ux-6-b` now
uses it. The wait lives in ONE place.

## This issue: migrate the rest

Replace the hand-rolled `admin-console-entry.click()` (or `.tap()`) sequence with
`openAdminConsole(page)` in the remaining specs. Mechanical but per-spec (the
post-open interaction + click-vs-tap differ), not a blind find-replace:

- [ ] u5-admin-networks-live-counts
- [ ] admin-credentials
- [ ] m-z-admin-cluster-journey (2 sites)
- [ ] admin-users
[...]

PR #699, opened 2026-08-03
test(#653): settle the mobile drawer + drive the rail launcher off app state

vjt:
Refs #653, #519, #531, #512

## One failure, one seam

Three separately-filed e2e flakes are the same defect:

| was | spec | reported shape |
|-----|------|----------------|
| #519 | `ux-2-mobile-archive` @webkit | members hamburger detaches mid-click |
| #531 | `ux-6-a-mobile-members-scroll:108` @webkit | overlay `touch-action` suspend |
| — | `ux-4-z-cluster-journey` @webkit | `openRailMenu`'s launcher click never opens the menu (failed on `main` @ `88bd4b12`) |

`openMembersDrawer`'s post-condition was `.shell-members.open` **visible** — true the instant the class lands, i.e. the START of the 200ms `translateX` slide, not its end. Every caller's next act is a click INSIDE that drawer.

## Why Playwright's actionability does not already cover this

`stable` (two consecutive animation frames with an identical box) **is** checked, and the hit-target interceptor verifies the target of the pointer **down** — but `mousedown` and `mouseup` are two separate protocol round-trips and nothing re-verifies the second. Move the element between them and the two land on different nodes, so WebKit synthesizes **no `click` at all** (click fires on the nearest common ancestor).

The handlers here are synchronous signals, so a swallowed click is silent: the affordance simply never opens and the caller's assert burns its full timeout. Under full-gate CPU load that inter-message gap stretches from ~1ms to hundreds — exactly the green-in-isolation / red-under-load signature all three reports carry.

## The cure — two halves, both setup-determinism

No assert weakened, no timeout bumped to hide timing.

1. **`openMembersDrawer` settles the slide-in** — `toBeInViewport({ ratio: 1 })`, the open-side mirror of `openAdminConsole`'s `not.toBeInViewport()` close-side wait. Ratio 1 is reachable only once the transform has settled at `translateX(0)`. Kept deliberately OUTSIDE the #519 retry loop: the openers are **toggles**, so re-clicking on a settle failure would close the drawer we just opened.

2. **`openRailMenu` drives the launcher off the app's own truth** — the rail keeps re-laying-out while stores hydrate, so a click can still be swallowed. `aria-expanded` mirrors the `open()` signal the menu renders from; re-issuing ONLY while it reads `"false"` is what makes that safe on a toggle (a blind retry would close a menu that opened late). The per-attempt assert is **stricter** than the 5s it replaces, and an app that reports expanded without mounting the menu still fails loudly at the deadline.

## Evidence

Blast radius is every spec reaching any rail affordance — **86 spec files** go through these doors — so the evidence is the full gate, not a scoped grep. The epic's bar is *"repeatably green across CONSECUTIVE runs"*:

| run | result |
|-----|--------|
| gate 1 | `Running 588 tests` → **588 passed (23.4m)**, rc=0 |
| gate 2 | `Running 588 tests` → **588 passed (27.9m)**, rc=0 |

Tick counts reconciled against the declared total in both; zero failed, zero flaky.

## Note for the epic

`#519`'s cure (`a1378e7e`) and `#531`'s (`f9baba72`) were already ancestors of `88bd4b12` — the tree that produced the third failure. This PR is the seam-level cure the two per-symptom patches could not reach, and the first two-consecutive-green evidence the epic asks for.

## Test plan

- [x] `scripts/integration.sh` full gate, run 1 — 588 passed
- [x] `scripts/integration.sh` full gate, run 2 — 588 passed
- [x] scoped `--project webkit-iphone-15 --grep "UX-2 —|UX-4-Z"` — 2/2, confirming `ratio: 1` is reachable and not an unsatisfiable assert

### DIFF
--- a/cicchetto/e2e/fixtures/cicchettoPage.ts
+++ b/cicchetto/e2e/fixtures/cicchettoPage.ts
@@ -747,7 +747,7 @@ export async function openMembersDrawer(page: Page): Promise<void> {
   // is unchanged (one probe + click + visibility assert on the first pass).
   const deadline = Date.now() + 15_000;
   for (;;) {
-    if (await drawer.isVisible().catch(() => false)) return;
+    if (await drawer.isVisible().catch(() => false)) break;
     try {
       const topicHamburger = page.getByLabel(/open members sidebar/i);
       if ((await topicHamburger.count()) > 0) {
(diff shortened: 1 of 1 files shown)


vjt:
Merged citing #623: the sole red was `recover-identity-real.spec.ts` **assert line 268** (31.6s vs a ~9s pass — the tracked setup race), with 589 passed alongside it. Two consecutive full local gates were green before this PR opened (588 / 23.4m and 588 / 27.9m).

Issue #693, opened 2026-08-02
cicchetto: returning after a long absence lands 200 rows into the past, never at the tail — both recovery paths are cursor-anchored


vjt:
## Symptom

Reported from live use: after a long absence (order of a day or more), reopening the client
does **not** show the recent scrollback. The pane sits on old rows, and scrolling to the
bottom leaves you parked at a stale position instead of at the live tail.

## Root cause (derived from the code path, not from a live repro)

Both recovery paths are anchored at the **read cursor** and capped at 200 rows, so the client
loads the region `[cursor .. cursor+200]` — the OLDEST end of the gap — and never the tail.

- Reconnect recovery: `refreshScrollback` resolves the resume cursor
  (`cicchetto/src/lib/reconnectBackfill.ts:getResumeCursor` — live high-water mark, else the
  server read cursor) and fetches `?after=<cursor>` with `REFRESH_LIMIT = 200`
  (`cicchetto/src/lib/scrollback.ts:619,661`).
- Cold load: with a read cursor present, `loadInitialScrollback`
  (`cicchetto/src/lib/scrollback.ts:298`) takes the anchored branch, same shape, same cap —
  the `#156` comment at `:440-445` names the leftover region explicitly: `[cursor+200 .. tail]`.
- The only thing that pages toward the tail afterwards is `loadNewer`
  (`cicchetto/src/lib/scrollback.ts:470,482`), and it is driven by the user
  **scrolling to the bottom**, 200 rows per gesture.

So the absence length itself is not the trigger — the **gap size** is. Any absence that
accumulates more than `REFRESH_LIMIT` messages in a window produces the symptom, and a long
absence is simply the reliable way to accumulate one. On a busy channel the gap can be
thousands of rows, i.e. tens of scroll-to-bottom gestures to walk forward to the present.

## Why a full page reload does not fix it

A reload re-enters `loadInitialScrollback`, which — because the read cursor is still there —
takes the *same* cursor-anchored branch. The client comes back up in the same place it was
stuck in. Any "reload after a long absence" behaviour has to be paired with the fix below,
or it just replays the stale window.

## The primitive for the fix already exists

`Grappa.Scrollback.count_after/5` deliberately does **not** apply the `@max_limit` cap
(`lib/grappa/scrollback.ex:528-530`), so the true gap size is knowable server-side rather
than inferred from a full-cap page.

## Proposed shape (open for discussion)

When the gap is larger than can be drained in one page, stop treating "resume contiguously
[...]

Issue #161, opened 2026-07-01
cic: newest messages unreachable after anchored fetch — no scroll-to-bottom forward paging (#156 regression)


vjt:
## Summary

After #156 (anchored scrollback fetch, commit `2ff2a24`), the **newest** messages in a channel can become unreachable when a read cursor is present. There is no scroll-to-bottom event to page *forward*, so any rows past the anchored window never load.

## Root cause

`loadInitialScrollback` (cursor-present arm) fetches the region **around** the cursor:

- `listMessagesAfter(cursor, 200)` — unread region, ASC, capped at `@max_http_limit` (200)
- `listMessages(cursor + 1)` — before-context

It no longer fetches the server tail (newest rows). The commit's own ">200 cap" note assumes the very newest rows "stream in via the WS join-ok `refreshScrollback`" — but:

1. `refreshScrollback` (`scrollback.ts:386`) also calls `listMessagesAfter(cursor, REFRESH_LIMIT=200)` from the same resume cursor, so it hits the **same 200 cap**. When true unread > 200, neither path reaches the tail.
2. There is **no symmetric forward-paging handler**. `loadMore` (`scrollback.ts:246`) pages OLDER rows on scroll-to-top (oldest-id paging). Nothing pages NEWER rows on scroll-to-bottom.

Result: the gap between `cursor + 200` and the true newest row is unreachable — you can't scroll down into it, and the WS refresh doesn't backfill it. Latest messages are inaccessible.

## Repro

1. Channel with > 200 unread relative to the frozen read cursor.
2. Open the channel. Anchored fetch loads `[cursor .. cursor+200]` + before-context.
3. Scroll to the bottom of the pane. Newest rows (id > cursor+200) never appear; no fetch fires.

## Fix direction (not implemented — triage only)

Add a forward-paging handler symmetric to `loadMore`: on scroll-to-bottom, if the highest loaded id < server tail, `listMessagesAfter(highestLoadedId, N)` and merge via `mergeIntoScrollback`. Needs an in-flight guard + end-latch mirroring `loadMoreInFlight` / `loadMoreExhausted`.

## Refs

- Regressed by #156 / `2ff2a24`, decision log `fa66d67` (`docs/DESIGN_NOTES.md`)
- `cicchetto/src/lib/scrollback.ts`: `loadInitialScrollback` (~L227-246), `loadMore` (L246), `refreshScrollback` (L386)

Reported per channel request.

vjt:
[...]

PR #700, opened 2026-08-03
fix(#693): a resume too big for one page lands at the tail, not the cursor

vjt:
Refs #693

## The symptom

After a day away, reopening cic shows stale scrollback and scrolling to the bottom never reaches
the live tail.

## Root cause

Both recovery paths anchored at the **read cursor** and capped at 200, so the client loaded
`[cursor .. cursor+200]` — the *oldest* end of the gap — and only `loadNewer` walked forward, 200
rows per scroll gesture. The trigger is **gap size**, not absence of a cursor: a small gap has
always worked, a day-sized one never could.

## What changed

**Server — a new wire token.** `Scrollback.count_after/5` (deliberately uncapped, and it had *no*
production caller; its doc still credited the join reply, which had moved to `count_after_split/5`
via `WindowCounts`) grew a required `hide_presence` positional → `/6`, and got a door:

    GET /networks/:slug/channels/:chan/messages/count?after=<id>  →  {"count": N}

⚠️ **This is a NEW WIRE TOKEN.** Per the cross-stack rule the ship runs the cic check + build gate;
the `DESIGN_NOTES.md` entry is written (search `#693`).

**cic lib (`scrollback.ts`).** Both resume paths probe the count at the anchor they are about to
use. Above one page (`PAGE_LIMIT = 200` — which also collapsed three copies of that literal) they
REPLACE the pane with the newest page (`anchorAtTail`), record `{missed, resumeFrom}` in
`farBehindByChannel`, and roll the high-water mark to the tail. A probe that throws returns `null`
and the caller keeps the pre-#693 behaviour — older servers degrade, they never guess.

**cic UI (`ScrollbackPane.tsx`).** A pinned bar at the top of the pane: *\"N unread — jump back\"*
plus a dismiss ×. The in-pane divider is suppressed while far behind: its count would describe the
~50 loaded rows, not the thousands missing.

**The cursor freezes while far behind** (`selection.ts setCursorIfAdvances`) and the badge keeps
the server seed instead of counting the pane. Dismiss is the one deliberate exit.

## The one review finding deliberately declined

A code-review agent raised ten issues. Nine are fixed in `55ea1937`. The tenth is declined on
evidence:

> Uncapped `COUNT(*)` behind an unmetered GET (`RequestBudget` passes reads through). **Declined:**
> capping it would defeat the issue's own requirement for the TRUE count, and it is not a new cost
> class — the `/me` cold-load already fans out ~2N uncapped `count_after_split/5` queries per login
> through the same unmetered door, and `index/2` on this very route does an index scan plus 200 row
> loads plus preload plus JSON render per call. Flagged for vjt rather than silently capped.

## e2e — +0, and why

The e2e fixture surface (`cicchetto/e2e/**`) belongs to another worker this cycle and the lane
instruction forbids touching it. The behaviour is pinned at the unit level (loader decisions,
cursor freeze, pane rendering). But jsdom cannot see the one thing that matters visually — that the
pinned bar is actually **visible** on a pane parked at the tail. An e2e is **owed** and should be
filed as a follow-up for whoever holds the fixture lane next.

## Gates

Both suites re-run detached on the final tree, from the worktree root:

* `./scripts/check.sh` — **exit 0**: `8 doctests, 50 properties, 4965 tests, 0 failures`
* `./scripts/bun.sh run test` — **exit 0**: `Test Files 215 passed (215)`, `Tests 3907 passed (3907)`
* biome + a **standalone** `tsc --noEmit` (the `bun run check` short-circuit means biome-red skips
  tsc entirely) — both exit 0.

Rebased onto `origin/main` @ 365bc2ec, no conflicts.

### DIFF
--- a/cicchetto/e2e/tests/issue161-forward-paging.spec.ts
+++ b/cicchetto/e2e/tests/issue161-forward-paging.spec.ts
@@ -1,42 +1,51 @@
 // #161 — the NEWEST messages become unreachable after the #156 anchored
 // fetch when unread exceeds the 200-row server cap.
 //
-// Root cause this pins: `loadInitialScrollback` (scrollback.ts) cursor-
-// present arm fetches the region AROUND the read cursor —
-// `listMessagesAfter(cursor, 200)` (the unread region, capped at the
-// server `@max_http_limit` = 200) + `listMessages(cursor + 1)` (before-
-// context). When true unread > 200 the after-page stops at `cursor + 200`,
-// so the very newest rows are NOT loaded. #156 assumed they "stream in via
-// the WS join-ok refreshScrollback" — but that path ALSO caps at 200 from
-// the same resume cursor, so it never reaches the tail either. And there
-// was NO forward-paging handler: `loadMore` pages OLDER rows on scroll-to-
-// TOP; nothing paged NEWER rows on scroll-to-BOTTOM. The gap
-// [cursor+200 .. true newest] was unreachable — the latest messages
-// inaccessible.
+// Original root cause: `loadInitialScrollback` (scrollback.ts) cursor-present
+// arm fetched the region AROUND the read cursor — `listMessagesAfter(cursor,
+// 200)` (capped at the server `@max_http_limit`) + `listMessages(cursor + 1)`
+// for before-context. With true unread > 200 the after-page stopped at
+// `cursor + 200`, so the newest rows were never loaded, and the WS join-ok
+// `refreshScrollback` capped from the same cursor so it never reached the tail
+// either. #161's answer was `loadNewer`: forward paging on scroll-to-bottom,
+// walking [cursor+200 .. tail] one 200-row page per gesture.
 //
-// The fix: a forward-paging verb (`loadNewer`) symmetric to `loadMore`,
-// fired from `ScrollbackPane.onScroll` when the pane nears the bottom of
-// the loaded content. It pulls `listMessagesAfter(highestLoadedId, 200)`
-// and merges via `mergeIntoScrollback`, page-by-page, until the true
-// server tail is reachable. (The no-storm / growing-tail-latch guarantee
-// is covered deterministically in scrollback.test.ts — asserting it here
-// would depend on distinguishing loadNewer's `?after=` from
-// refreshScrollback's identical request shape, which is flaky.)
+// #693 SUPERSEDED THAT MECHANISM FOR THIS SCENARIO, and this spec is rewritten
+// to the contract that replaced it. The observation behind #693 is that
+// forward-paging out of a day-sized gap is not a fix an operator can feel: it
+// still opens the pane at the OLDEST end of the gap and asks for one scroll
+// gesture per 200 rows. So when the gap exceeds one page, the resume no longer
+// anchors at the cursor at all — it probes the true (uncapped) unread count via
+// `GET /networks/:slug/channels/:chan/messages/count?after=<id>`, REPLACES the
+// pane with the newest page, and surfaces the abandoned region as a pinned
+// "N unread — jump back" bar. The in-pane divider is deliberately SUPPRESSED
+// while far behind (`ScrollbackPane` gates `injectMarker` on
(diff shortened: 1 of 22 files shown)


vjt:
Rebased onto `origin/main` (now at the passkey merge), and the red integration run is addressed.

## What the red actually was

Not a timeout — the 29m58s was job wall-clock. `issue161-forward-paging.spec.ts:147` failed at
11.7s: `expect(locator('[data-testid="unread-marker"]')).toBeAttached()` → element not found.

It was a real regression, and the spec was right to catch it. That spec plants a cursor with **240
unread rows** and waits on the unread-marker as, in its own words, *"the anchored fetch's own
signal"*, commenting that the pane scrolls to the marker *"NOT the tail"*. Above one page this PR
suppresses that divider **by design** — `ScrollbackPane` gates `injectMarker` on
`farBehindByChannel()[key()] === undefined` — because its count would describe the ~50 loaded rows
instead of the hundreds missing. The wait could never be satisfied.

## Why it was rewritten rather than repatched

Patching the selector would have papered over the real change. #161's *mechanism* — forward-page
out of the gap, one 200-row gesture at a time — is exactly what this PR concluded is not a fix an
operator can feel, so in this scenario there is no longer an oldest-end anchor to page forward
from. What survives is #161's *guarantee* — "the newest message is reachable" — now satisfied by
construction. The spec asserts that instead, and every assertion is picked to fail if this PR
regresses:

* the newest row is rendered **on arrival**, with no scroll gesture;
* the jump-back label carries a count **greater than the 200-row page limit** and no greater than
  the real number of rows after the cursor. A capped count would read exactly 200, so this is the
  assertion that proves the uncapped `.../messages/count?after=` door is actually consulted.
  Deliberately not an exact equality — the count honours the subject's hide-presence preference
  while `fetchAllMessagesAsc` returns every row;
* the in-pane divider is **absent** — the inverse of what the spec used to wait for, and the
  clearest single signal that the far-behind arm was taken.

The far-behind bar takes over as the readiness signal, by the same reasoning the unread-marker
held it: it renders only once the resume has probed the count, judged the gap undrainable, and
swapped in the newest page. `loadNewer` is untouched and still pages forward for gaps the tail
anchor does not claim; its no-storm / growing-tail-latch mechanics stay covered deterministically
in `src/__tests__/scrollback.test.ts`.

## One pre-existing red fixed on the way

`grappa.gen_wire_types --check` — the last gate in `scripts/check.sh` — fails on `origin/main`:
`c6045a9c auth: add passkey login modes (#696)` added `:passkey_required` to
`GrappaWeb.ErrorTokens` without regenerating the committed `cicchetto/src/lib/wireTypes.ts`. Not
from this PR, but this branch cannot be green without it, so it rides as its own commit
(`467f3c07`) rather than buried in the spec rewrite. Verified byte-identical to generator output,
`--check` re-run clean. (Regenerating in place is not possible from the dev container — the source
mount is read-only — so the content came from
`scripts/mix.sh run -e 'IO.write(Mix.Tasks.Grappa.GenWireTypes.generate())'` and was applied by
hand.)

## Gates on the rebased tree

* `./scripts/check.sh` — **exit 0**: `8 doctests, 50 properties, 4998 tests, 0 failures`, dialyzer
  `Total errors: 0`, doctor passed, wire types in sync, bats ran.
* `./scripts/bun.sh run test` — **exit 0**: `Test Files 215 passed (215)`, `Tests 3907 passed (3907)`.

The e2e itself is **not verified locally** — the STACK lane is held elsewhere, and this worktree
has no `e2e/node_modules`, so the rewritten spec is not even typechecked here. It is gated on this
PR's CI run, deliberately.

vjt:
Merging with one tolerated red: `recover-identity-real.spec.ts` failing at the **assert on line 268** (`element(s) not found`, 31.6s — the ~31s timeout vs the 9s pass is the setup-race signature). That is #623, not this PR: every other check is green, and #623's root cause has since been measured — a late 433 on the re-NICK that main's one-shot rule treats as terminal — with a fix in gate.

PR #715, opened 2026-08-03
ci(#714): run the e2e suite on dependency and image changes too

vjt:
Closes #714.

The integration workflow's path filter listed `mix.exs`/`mix.lock` but not `cicchetto/package.json` or `cicchetto/bun.lock`. An Elixir dependency bump therefore ran the full e2e matrix; a JavaScript dependency bump ran no JavaScript at all.

That is not an overstatement — `ci.yml`'s three jobs (`test + lint + audit`, `dialyzer`, `shottino`) are Elixir and bash only:

```
$ git grep -lE 'vitest|npm ci|npm run|biome|tsc --noEmit' origin/main -- .github/workflows/
(no matches)
```

and the SPA is built nowhere except inside this suite's compose stack (`scripts/integration.sh` → `docker compose build` of `cicchetto-build-test` + `playwright-runner`).

Concretely: #708 bumped **vite**, the bundler, and merged green on `mix credo` and dialyzer. #707 bumps `@solidjs/router` across a **major** (0.16.2 → 1.0.0) and is being held open until this lands.

Also adds the root `Dockerfile` and `compose.yaml` — they define the images the suite boots. The e2e-local Dockerfiles and compose files already fall under `cicchetto/e2e/**`.

**Cost:** dependency-bump PRs now wait on the ~30 min suite. Intended trade.

**Not included:** the fast JS job proposed in #714 (`bun install && tsc --noEmit && biome check && vitest run` on `cicchetto/**`) — that is a new job rather than a filter fix, and it is a complement to this, not a substitute. Say the word and I'll open it separately.

**Note on this PR's own CI:** it touches `.github/workflows/integration.yml`, which is in the filter, so the e2e suite runs on it — the change is self-exercising.

Issue #714, opened 2026-08-03
ci: cicchetto dependency bumps merge with zero JavaScript executed


vjt:
## What

A pull request that only touches `cicchetto/package.json` / `cicchetto/package-lock.json` merges with **zero JavaScript executed anywhere in CI** — not a test, not a type-check, not a lint, not even a build.

## Evidence

`.github/workflows/ci.yml` has three jobs and all of them are Elixir + bash:

- `test + lint + audit` — `mix compile --warnings-as-errors`, `mix format`, `mix credo`, `mix sobelow`, `mix hex.audit`, `mix deps.audit`, `mix doctor`, `mix coveralls.json`, bats, shellcheck, `infra/cloud/check-drift.sh`.
- `dialyzer` — Elixir.
- `shottino (build + tests)` — Elixir.

Grepping every workflow for any JS toolchain invocation returns nothing:

```
$ git grep -lE 'vitest|npm ci|npm run|biome|tsc --noEmit' origin/main -- .github/workflows/
(no matches)
```

The SPA is only ever built or executed inside the e2e stack: `integration.yml` → `scripts/integration.sh` → `docker compose build` of the `cicchetto-build-test` and `playwright-runner` services.

And `integration.yml` is path-filtered (identical list on `push` and `pull_request`):

```yaml
paths:
  - "lib/**"
  - "cicchetto/src/**"
  - "cicchetto/e2e/**"
  - "config/**"
  - "priv/**"
  - "scripts/integration.sh"
  - "scripts/testnet.sh"
  - "scripts/_lib.sh"
  - "mix.exs"
  - "mix.lock"
  - ".github/workflows/integration.yml"
```

`cicchetto/package.json` and `cicchetto/package-lock.json` are absent. So the filter is **asymmetric**: an Elixir dependency bump touches `mix.lock` and therefore runs the full e2e matrix, while a JavaScript dependency bump touches only the npm lockfile and runs nothing that can observe it.

## Why it matters now

Merged today under this hole, all three checks green, no JS executed:

- #704 / #705 / #706 — CI action bumps (harmless, but equally unverified)
- #708 — `vite` 8.1.5 → 8.2.0 — **the bundler itself**
- #709 — `jsdom` 30.0.0 → 30.0.1
- #710 — `phoenix` 1.8.5 → 1.8.9 — the JS client, runtime code

[...]

Issue #103, opened 2026-06-26
Supply-chain: pin oven/bun + nginx:alpine base images to digests


vjt:
`oven/bun:1` (used by `scripts/bun.sh`) and `nginx:alpine` (compose `--profile prod` nginx service) are moving major tags. Pin to digests (`oven/bun:1@sha256:...`) for reproducible builds.

vjt:
Still valid, queued on vjt's call (2026-08-02 triage: "fixa il fixabile"). Verified against `origin/main`: `Dockerfile.release:23` is still `FROM oven/bun:1-alpine` with no digest.

vjt:
Shipped in **v0.10.0** (release commit `b4047f04`, tagged 2026-08-02). Pull that tag to get this — https://github.com/vjt/grappa-irc/releases/tag/v0.10.0

Issue #515, opened 2026-07-28
cicchetto: pin biome — floating ^2.4.13 turns clean CSS red on a minor bump


vjt:
## Problem

`cicchetto/package.json` pins `@biomejs/biome` to a **floating** range
(`^2.4.13`). A newer 2.x minor resolved in the container ships a stricter
formatter that now flags the #508 `input:where(:not(...))` denylist
selector in `default.css` as a **formatter error** — so `bun run check`
exits 1 on **clean main**, with no code change of our own.

A formatter on a floating version range is a time bomb: `bun run check`
can go red between sessions purely because a dependency floated, and any
reformat we apply chases a moving target (the next minor may re-indent it
differently again).

## Evidence

- Clean main (d3fc7e1f): `bun run check` -> `Found 1 error` (formatter
  drift, default.css:378) + 28 pre-existing warnings -> exit 1.
- The immediate drift was formatter-fixed as the pre-existing-error commit
  on the #510 branch (CLAUDE.md: fix pre-existing first). This issue
  tracks the ROOT cause, not that one file.

## Fix

Pin `@biomejs/biome` to an EXACT version (drop the caret), and bump it
deliberately in its own commit so a formatter change is a reviewed diff,
never a silent between-session red.

vjt:
Still valid, queued on vjt's call (2026-08-02 triage: "fixa il fixabile"). Verified against `origin/main`: biome moved to `^2.5.6`, but it is **still floating** — the caret is the defect, not the version behind it. A newer minor can still turn clean CSS red without a commit touching it.

vjt:
Shipped in **v0.10.0** (release commit `b4047f04`, tagged 2026-08-02). Pull that tag to get this — https://github.com/vjt/grappa-irc/releases/tag/v0.10.0

PR #716, opened 2026-08-03
ci: run cicchetto types, lint and unit tests on every push

vjt:
Complement to #715, opened separately as agreed: #715 fixes the path filter so `integration.yml` fires on JS lockfile changes; this adds the fast unconditional gate.

## Why

`ci.yml`'s three jobs (`test + lint + audit`, `dialyzer`, `shottino`) are Elixir and C only. The SPA is built **only** inside `integration.yml` → `scripts/integration.sh` → `docker compose build`, which is path-filtered. Measured on `origin/main`:

```
git grep -lE 'vitest|bun install|bun run|biome|tsc --noEmit' origin/main -- .github/workflows/
→ (no matches)
```

So a dependency bump that broke types, lint or the unit suite could reach main with three green Elixir checks. #708 (the bundler) merged exactly that way. That is #714.

## What

A `cicchetto` job in `ci.yml` (no path filter → runs on every push and PR):

- `bun install --frozen-lockfile` — a lockfile disagreeing with `package.json` fails here instead of quietly resolving to something the build and e2e runner never install
- `bun run check` — `biome check src && tsc --noEmit`, the same gate the local path runs
- `bun run test` — vitest

It runs in the digest-pinned `oven/bun:1` image `compose.yaml` already uses for `cicchetto-build-test` (#103 supply-chain), so CI type-checks with the exact runtime that builds the bundle and one bump moves both.

## Not a substitute for the e2e

#715 makes the e2e run on JS dependency changes (~30 min, conditional). This is the ~2 min gate that always runs. Both are wanted: this one catches type/lint/unit breakage on any change; the e2e catches behaviour.

`ci.yml` has no path filter, so this PR exercises the new job on itself.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -229,3 +229,55 @@ jobs:
       # so adding a docs step here costs nothing extra.
       - name: Build docs (no warnings)
         run: mix docs
+
+  # cicchetto is TypeScript, so none of the gates above say anything about
+  # it. Before this job, CI ran zero lines of JavaScript on a JS change:
+  # the SPA is built only inside integration.yml (docker compose ->
+  # cicchetto-build-test + playwright-runner), which is path-filtered.
+  # A dependency bump that broke types, lint or the unit suite could reach
+  # main with three green Elixir checks (#714 — the path filter itself is
+  # fixed in #715; this job is the complement, not the substitute: it runs
+  # unconditionally and in ~2 min, where the e2e runs conditionally in ~30).
+  #
+  # Deliberately its own job, same reasoning as shottino: no BEAM
+  # toolchain, no deps cache, and a failure should read as "cicchetto
+  # broke" rather than as a red mark on the Elixir suite.
+  cicchetto:
+    name: cicchetto (types + lint + unit)
+    runs-on: ubuntu-latest
+    # The same digest-pinned oven/bun the cicchetto-build-test service uses
+    # in compose.yaml (#103 supply-chain). Checking types with the exact
+    # runtime that builds the bundle means one bump moves both, instead of
+    # CI and the build image drifting apart on separate bun versions.
+    # Refresh on an intentional bump:
+    #   docker buildx imagetools inspect oven/bun:1 --format '{{.Manifest.Digest}}'
+    container:
+      image: oven/bun:1@sha256:e10577f0db68676a7024391c6e5cb4b879ebd17188ab750cf10024a6d700e5c4
+    defaults:
+      run:
+        working-directory: cicchetto
+    steps:
+      # No submodules: bats-core and friends are Elixir/shell-side only.
+      - uses: actions/checkout@v7
+
+      - name: Cache bun install
+        uses: actions/cache@v6
+        with:
+          path: ~/.bun/install/cache
+          key: ${{ runner.os }}-bun-${{ hashFiles('cicchetto/bun.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-bun-
+
+      # --frozen-lockfile so a lockfile that disagrees with package.json
+      # fails here instead of silently resolving to something else than
+      # what the build and the e2e runner install.
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
+
+      # `check` is biome check src && tsc --noEmit — the same gate the
+      # local pre-commit path runs, so CI and the developer see one rule.
+      - name: Biome + typecheck
+        run: bun run check
+
+      - name: Unit tests (vitest)
+        run: bun run test


vjt:
## Review — verified against the files, not against the PR text

Reviewed on `origin/main` + this PR's head. Every claim below was checked by reading the referenced
file or by executing the pinned image locally. Ordered by severity.

**Bottom line: this job would land green.** My checkout is byte-identical to `origin/main`
(0 commits ahead, empty diff), and there both gates pass: `bun run check` → RC 0, `bun run test` →
213 files / 3822 tests passed, `bun install --frozen-lockfile` → RC 0 (lockfile in sync).

---

### 🟠 MEDIUM — `bun run check` short-circuits, so one CI round surfaces one signal

`ci.yml:280` runs `bun run check`, which `cicchetto/package.json` defines as:

```json
"check": "biome check src && tsc --noEmit"
```

`&&` means a biome failure stops the line and **tsc never runs** — the job reports the lint error and
says nothing about types. Fix the lint, push, wait, discover the type error on the next round.

This repo has already been bitten by exactly this: #515 (`feedback_cic_check_biome_short_circuits_tsc`
— biome red left `main` type-blind). And the best argument for this whole job is a type error:
#696 added `:passkey_required` to `GrappaWeb.ErrorTokens`, the regenerated union made it a known
token, and `friendlyApiError.ts` had no `case` arm — `TS2345: Argument of type '"passkey_required"'
is not assignable to parameter of type 'never'`. A job that can silently skip tsc is a weaker version
of the gate you are building.

**Suggested change** — two steps, so both signals land in one round:

```diff
-      - name: Biome + typecheck
-        run: bun run check
+      - name: Biome (lint + format)
+        run: bunx biome check src
+
+      # Not `bun run check`: that is `biome check src && tsc --noEmit`, so a
+      # lint failure would skip the typecheck entirely and CI would report
+      # half the truth (#515). `!cancelled()` keeps both signals in one round.
+      - name: Typecheck
+        if: ${{ !cancelled() }}
+        run: bunx tsc --noEmit
```

Cleaner variant if you want local and CI to keep sharing one definition: split
`"lint": "biome check src"` + `"typecheck": "tsc --noEmit"` in `package.json`, make
`"check": "bun run lint && bun run typecheck"`, and have CI call the two halves.

---

### 🟠 MEDIUM — the job never builds the bundle

`tsc --noEmit` is not `vite build`. The incident that motivated #714 was a **bundler** bump (#708,
vite 8.1.5 → 8.2.0), and this job would not have caught it.

With #715 merged, a *dependency* bump does get built (the e2e's `cicchetto-build-test` runs
`bun install --frozen-lockfile && bun run build`). But a change to `cicchetto/vite.config.ts` is in
neither gate: not in #715's path filter, and not built here. Options: add a `bun run build` step
here (~30s), or widen #715's filter to `cicchetto/**` — I have suggested the latter on that PR.

---

### 🟡 LOW — the cache path is HOME-dependent, and a miss is silent

`ci.yml:266` uses `path: ~/.bun/install/cache`. I inspected the pinned image config: it sets only
`BUN_INSTALL_BIN=/usr/local/bin` — no `BUN_INSTALL`, no `BUN_INSTALL_CACHE_DIR` — so bun's default
`$HOME/.bun/install/cache` is correct, and container jobs give the cache action and the run steps the
same `HOME`. So this is sound as written. But a never-hitting cache is a slow no-op, not a red, so
nothing will ever tell you if that stops being true.

Cheap hardening, and it matches what the rest of the repo already does
(`compose.yaml:196`, `cicchetto/e2e/compose.yaml`, `scripts/bun.sh` all set it explicitly):

```diff
     container:
       image: oven/bun:1@sha256:e10577f0db68676a7024391c6e5cb4b879ebd17188ab750cf10024a6d700e5c4
+    env:
+      # Pin the cache dir instead of leaning on $HOME, so the actions/cache
+      # path and bun's default cannot drift apart silently.
+      BUN_INSTALL_CACHE_DIR: /github/home/.bun/install/cache
```

---

### 🟡 LOW — `git` is not in the pinned image

Ran the pinned digest locally:

```
git    MISSING
tar    /usr/bin/tar
gzip   /usr/bin/gzip
zstd   MISSING
bun=1.3.14
```

Consequences, both benign today:

* `actions/checkout@v7` (`ci.yml:261`) falls back to downloading the repo via the REST API and prints
  a warning. It works, but the workspace is **not a git repository** — any future git-dependent step
  added to this job (a drift guard, a `git diff` check) will fail confusingly. Worth a one-line
  comment on the checkout step, or `apt-get install -y git` if you would rather have a real clone.
* `zstd` missing → `actions/cache@v6` falls back to gzip. Slightly slower, not a failure.

I specifically checked the one thing that *could* have made this fatal: `cicchetto/biome.json` sets
`vcs: { enabled: true, clientKind: "git", useIgnoreFile: true }`. It does not need the git binary or
a `.git` directory — my local `bun run check` runs in this same image with only `cicchetto/` mounted
(no git, no `.git`) and exits 0.

---

### 🟡 LOW — nothing enforces this job

`GET /repos/vjt/grappa-irc/branches/main/protection` → `404 Branch not protected`. There are no
required status checks, so a red `cicchetto (types + lint + unit)` blocks nothing mechanically. Not a
defect in this PR — just worth knowing that the gate's value rests entirely on someone honoring it.

---

### 🟡 LOW — the e2e will not run on this PR

`ci.yml` is not in `integration.yml`'s path filter, so the e2e suite does not fire here. Correct
outcome — this PR changes no runtime code — but stating it so it is not mistaken for a missing check.

---

### ✅ Clean — verified, not assumed

* **The digest-pin claim is literally true.** `oven/bun:1@sha256:e10577f0db68676a7024391c6e5cb4b879ebd17188ab750cf10024a6d700e5c4`
  is byte-identical in `cicchetto/e2e/compose.yaml:495` (`cicchetto-build-test`), `compose.yaml:180`
  (`cicchetto-build`), `scripts/bun.sh`, and this job. One bump really does move all of them.
* **Action majors match the siblings.** `checkout@v7` and `cache@v6` are what `test`, `shottino` and
  `dialyzer` already pin in this same file — no half-migrated state.
* **Concurrency is fine.** `ci.yml` declares it at workflow level, so the new job inherits it; a
  per-job group would be the inconsistency, not its absence.
* **No submodule needed.** `.gitmodules` has `cicchetto/e2e/infra`, `vendor/bats-core`,
  `frontends/shottino/vendor/libdatachannel`. `tsconfig.json` is `include: ["src"]` and
  `vitest.config.ts` is `include: ["src/**/*.{test,spec}.{ts,tsx}"]`, so nothing this job runs
  reaches into a submodule. Omitting `submodules:` is right.
* **`--frozen-lockfile` behaves.** bun 1.3.14 in the pinned image; it errors (not warns) on
  lockfile/manifest disagreement, and the same flag is what `cicchetto-build-test` uses — CI and the
  build image install the same way.
* YAML at the PR head parses; the job registers as `jobs.cicchetto` alongside `test`, `shottino`,
  `dialyzer`.

vjt:
main is green again (#720 merged as `572f8742`). Rerunning the shottino job, which was red here only by inheritance.

vjt:
## Correction to my review above — one piece of evidence was mis-sourced

My comment claimed *"my checkout is byte-identical to `origin/main` (0 commits ahead, empty diff)"*
and hung the "this job lands green" evidence on a local run there. That reading was wrong:
`git rev-list --count origin/main..HEAD` returning 0 only proves HEAD is an **ancestor** of
`origin/main`, not that it equals it. My checkout was **75 commits behind**, so the local
`bun run check` / `bun run test` / `bun install --frozen-lockfile` results I quoted (3822 tests) were
measured on a stale tree — not on this PR's merge base. My mistake, and the kind that is invisible
unless you check the direction of the range.

**The conclusion was right; here is the authoritative evidence for it.** Run
[30801877011](https://github.com/vjt/grappa-irc/actions/runs/30801877011) on `388f883b` (this PR on
`main`), per-job conclusions from the API:

```
== test + lint + audit: success
== shottino (build + tests): success
== cicchetto (types + lint + unit): success
== dialyzer: success
```

The new job is green on real `main`, and it installed with `--frozen-lockfile` there, which also
re-establishes the lockfile-in-sync claim on the correct tree.

Nothing else in the review depends on that checkout: every file citation was read via
`git show origin/main:<path>`, the digest comparison came from the same source, and the image probes
(`git` missing, `zstd` missing, no `BUN_INSTALL*` env, bun 1.3.14) were executed against the pinned
digest directly. The two MEDIUM findings — `bun run check` short-circuiting `biome && tsc`, and the
job never running `vite build` — stand as written and are now on `main`.

Issue #713, opened 2026-08-03
ChannelDirectoryTest asserts a sub-second freshness window against a second-precision column


vjt:
Intermittent CI red, most recently on run 30778184957 (`test + lint + audit` →
`Tests with coverage`, the only failing step in that job).

## Signature

```
1) test replace_start nukes, ingest inserts, finalize stamps captured_at (Grappa.ChannelDirectoryTest)
   test/grappa/channel_directory_test.exs:36
   match (=) failed
   code:  assert %{status: :fresh, total: 3, entries: entries, captured_at: ca} = Dir.list(s, nid, ttl_ms: 1000)
   right: %{status: :stale, total: 3, ..., captured_at: ~U[2026-08-03 02:01:18Z]}
   stacktrace:
     test/grappa/channel_directory_test.exs:42: (test)
```

## Root cause — the production code is correct; the test is not

`ChannelDirectory.Entry` declares `field :captured_at, :utc_datetime`
(`lib/grappa/channel_directory/entry.ex:56`), i.e. **second precision**, backed by
a `TEXT` column (migration `20260626120000_create_channel_directory.exs:61`).
`finalize/2` therefore *must* truncate — `DateTime.truncate(DateTime.utc_now(), :second)`
at `lib/grappa/channel_directory.ex:119` — or Ecto refuses the cast. Same at
`replace_start/2` (`:88`). Nothing to fix there.

The consequence is that a snapshot stamped at `…:18.940` is stored as `…:18.000`
and is **born up to 999 ms old**. The test at `:42` then asserts `status: :fresh`
with `ttl_ms: 1_000`, so the entire freshness budget can already be spent before
`list/3` is even called. In the failing run the stamp is `02:01:18Z` and the
assertion fires at `02:01:19.03` → derived age ≥ 1000 ms → `:stale`.

**A sub-second TTL is not representable against a second-precision column.** The
test is asking a question the storage cannot answer, and it passes only when
`finalize` happens to land early in a wall-clock second. That is why it is green
on most runs and on `main`.

Production is unaffected: `ttl_ms/0` is 48h (`@ttl_ms`, `channel_directory.ex:39`),
so a 999 ms truncation is noise against a 172,800,000 ms window.

## Blast radius

`ttl_ms: 1_000` appears ~9 times in `test/grappa/channel_directory_test.exs`, but
only **line 42** asserts `status: :fresh`. The rest assert `:refreshing`, `:empty`,
[...]

PR #718, opened 2026-08-03
test(#713): stop asking a second-precision column a sub-second question

vjt:
Refs #713

Test-only. No production code changes — per the call on the issue, this is a
defect in the test, not in `ChannelDirectory`.

## The defect

`ChannelDirectory.Entry` declares `field :captured_at, :utc_datetime` — second
precision — so `finalize/2` **must** truncate or Ecto refuses the cast. A
snapshot stamped at `…:18.940` is therefore stored as `…:18.000` and is born up
to 999 ms old. The tests then asserted `status: :fresh` against `ttl_ms: 1_000`,
so the whole freshness budget could be spent before `list/3` was called. Whether
CI went green depended on where in the wall-clock second `finalize` landed —
not on the code under test.

## Reproduced, not assumed

Aligning `finalize` to ~50 ms before a second boundary and reading just past it
fails **100% of the time**, with the exact CI signature:

```
match (=) failed
code:  assert %{status: :fresh} = Dir.list(s, nid, ttl_ms: 1000)
right: %{status: :stale, ...}
```

The same alignment passes under the new window. The probe was temporary and is
not in the diff — it existed to prove the diagnosis and to prove the fix
actually addresses it, rather than just moving a number until CI stopped
complaining.

## The fix

Both TTLs are named now, because the next person writing a case here needs to
know the floor exists:

* **`@ample_ttl_ms 60_000`** — applied to **all nine** call sites, not only the
  one that was failing. Leaving `1_000` in the others would keep the trap armed
  and invite the value to be copied into the next freshness assertion. Every
  case in this file is about *which branch* `list/3` takes; none is about the
  boundary itself.
* **`@expired_ttl_ms -1`** — for the one case that wants `:stale` on purpose.
  `0` is **not** equivalent: the comparison is `age_ms <= ttl_ms`, so a stamp
  and a read both landing in the same millisecond as the second boundary would
  score `:fresh`. Far rarer than the bug above, same class, fixed while here.

`test/grappa/session/directory_test.exs` carried the same assertion and is worse
exposed — an `assert_receive` with a 1s timeout sits between `finalize` and the
read, so most of the budget can be gone before it even asks. Widened, with the
reason inline.

## Deliberately not changed

The truncation, and `list/3` itself. The issue also floated rejecting a
`ttl_ms` below the column's granularity at the boundary, so the next caller
cannot ask an unanswerable question. That is a production change and this is a
test defect — it stays on #713 for a separate call.

## Gates

* `scripts/test.sh` on both touched files — **exit 0**: `1 property, 13 tests, 0 failures`
* `scripts/mix.sh format --check-formatted` on both — **exit 0**

Full `check.sh` is not quoted here because `main` currently fails its last gate
(`gen_wire_types --check`, #696 fallout); that fix is `467f3c07` on #700 and
reaches everyone from there.

### DIFF
--- a/test/grappa/channel_directory_test.exs
+++ b/test/grappa/channel_directory_test.exs
@@ -12,6 +12,10 @@ defmodule Grappa.ChannelDirectoryTest do
   `captured_at` still NULL), `:fresh` (finalised within TTL), `:stale`
   (finalised but older than TTL).
 
+  An injected TTL must stay clear of the storage granularity: `captured_at`
+  is second-precision, so a sub-second window makes the branch depend on the
+  clock rather than on the code. See `@ample_ttl_ms` / `@expired_ttl_ms`.
+
   The property test pins the keyset-paging invariant that's easy to
   break: walking the cursor visits every row exactly once with no
   overlap, across the tie-heavy `:users` sort.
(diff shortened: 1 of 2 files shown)


vjt:
main is green again (#720 merged as `572f8742`). Rerunning the shottino job, which was red here only by inheritance.

Issue #769, opened 2026-08-03
e2e: issue281 account-switch replay leaks a /messages/count fetch after the identity purge


vjt:
## What happened

PR #741 (`w2/623-recover-reclaim-sequence`) ran the `integration` job three times on the **same head**
`f54f45af`, and produced **two reds on two DIFFERENT specs, then a green**:

| attempt | started | result |
|---|---|---|
| 1 | 2026-08-03T11:13Z | ✘ `issue281-account-switch-no-replay.spec.ts:53:3` — one leaked fetch, `/networks/bahamut-test/channels/%24server/messages` (account A's history replayed under B's bearer) |
| 2 | 2026-08-03T11:46Z | ✘ `issue580-send-snap-independent-of-post.spec.ts:104:3` case 1 — `toBeInViewport()` failed (a send whose POST fails must still snap to the bottom). **`issue281` passed this run.** |
| 3 | 2026-08-03T12:18Z | ✅ fully green |

Run: https://github.com/vjt/grappa-irc/actions/runs/30808553415

## Why this is filed rather than dismissed as a flake

- **#741's diff cannot touch either spec.** It is recover-identity (`lib/grappa/session/*`) plus a
  *type-only* addition in `cicchetto/src/lib/api.ts` (one new `RecoverResultReason` union member).
- **Neither spec is in the tracked-flake set** (#653 and the separately tracked #279 #254 #291/#339 #506
  #522 #552 #585 #623).
- **Both failures live in the same family: scroll / scrollback.** `issue281`'s offending request is a
  scrollback `/messages` fetch; `issue580` asserts the snap-to-bottom position.

## Hypothesis (not proven)

#700 (#693) rewrote exactly that surface shortly before: `cicchetto/src/lib/scrollback.ts` (+200),
`ScrollbackPane.tsx`, "stop the tail anchor from destroying the position it protects", plus the
jump-back boundary row. A destabilised family surfaces on **whichever spec loses the race**, which would
explain why the failing spec CHANGES between runs while the head does not.

Counter-evidence: main was green once at `7ead6ade` (post-#700), and attempt 3 was green.

## What would settle it

Run `issue281` + `issue580` repeatedly at plain `origin/main` (STACK lane) and compare against the same
loop at the commit before #700. **Establish the baseline before blaming a branch.**

## Triage note for whoever picks this up

Do **not** match this by spec name — it will not match. Key it off the **family + the mechanism**
[...]

Issue #717, opened 2026-08-03
cicchetto: installed PWA on Android+Firefox hangs on the splash at launch (browser tab never does)


vjt:
## Problem

On Android + Firefox, the **installed PWA** hangs on the CRT splash at launch. A plain browser tab on the same phone, same instance, same account never does (~10 consecutive tries at first report). The splash stays up **indefinitely** — it does not recover on its own; the only way out is to force-kill the app and relaunch.

This is the trigger #687 refers to as "not yet root-caused and tracked separately". #687 is the *feedback* defect (a mute splash cannot distinguish loading from dead, and the boot fetches carry no timeout/abort/retry, so any hang is permanent). This issue is the hang itself.

## Repro

Reported by a self-hoster on IRC, and **re-confirmed today** on the same setup:

- Google Pixel 8 Pro, Android 17
- Firefox 153.0.2
- **Installed PWA only** — the same URL in a browser tab is clean
- Splash persists until the app is killed and restarted; no self-recovery
- Instance is self-hosted and reachable only over a private network, so there is no third-party URL to poke at

## What is already known from the code (`origin/main` @ `42ee2013`)

- The service worker does **not** intercept API calls: `cicchetto/src/service-worker.ts` registers `precacheAndRoute` plus a `NavigationRoute` that only matches `request.mode === "navigate"`. REST (`cors`/same-origin) and the WS upgrade pass through untouched.
- `registerSW()` in `cicchetto/src/main.tsx:317` uses the default deferred-until-`window.load` timing, with `onRegisterError`/`onRegisteredSW` wired to the #119 stacked-error region — so a registration *failure* would surface as a typed `sw-registration` warn entry rather than silence. Worth checking whether the hung launch shows one.
- The two boot fetches (`me()` in `cicchetto/src/lib/api.ts`, plus `channelsBySlug()`) have no `AbortSignal`, timeout or retry (#687) — which is why a hang is terminal rather than slow.

## First measurement to take

The cheap discriminator, before any hypothesis: attach desktop Firefox to the phone (`about:debugging` → USB → the PWA's own process) and answer **one** question —

1. does `GET /me` (and the channels fetch) **ever leave the client**? If no request appears, the app is stuck *before* the fetch (SW startup / precached shell / registration promise), not on the network.
2. if it does leave, does it get a response, a redirect chain, or nothing at all?

[...]

Issue #687, opened 2026-08-02
cic: the CRT splash must show boot stages, not stay mute while a boot fetch hangs


vjt:
## Problem

`CrtSplash` is a **mute** fallback. In `cicchetto/src/Shell.tsx` it renders as
`<Switch fallback={<CrtSplash />}>`, i.e. purely on `selectedChannel() === null`,
and the auto-select effect that clears that condition waits on two resources:

- `user()` — seeded by `me()` (`cicchetto/src/lib/api.ts`, `fetch("/me")`)
- `channelsBySlug()`

Neither boot fetch carries an `AbortSignal`, a timeout, or a retry. If one hangs,
the splash stays up forever and shows exactly the same thing it shows during a
healthy 200 ms boot: nothing. The user cannot tell "still loading" from "dead",
and the only recovery is to kill and relaunch the app.

Reported on IRC by a self-hoster: on Android + Firefox, the **installed PWA**
hangs on the splash on the first launch every time (a plain browser tab on the
same phone never does — ~10 consecutive tries). The PWA-relaunch trigger is not
yet root-caused and is tracked separately; this issue is the missing feedback,
which is a defect on its own regardless of what makes the fetch hang.

Precedent: #193 (client stuck on splash after a BEAM restart) — same dead-end,
different door.

## Wanted

Not a progress bar. A short boot log in the existing CRT register, one line per
stage, marked done as it resolves:

```
fetching my info... done
fetching networks... done
connecting...
```

Plus a state change when a stage does not resolve: after N seconds the stuck
line should say so and offer a retry, instead of the screen staying identical
to a healthy boot.

## Notes

- The service worker is **not** involved: `cicchetto/src/service-worker.ts` has
  no `fetch` handler, so it never intercepts the API calls.
- Timeout + backoff retry on the two boot fetches is the natural companion
  change — it is what gives the splash a "this stage failed" state to render.

Verified against `origin/main` @ `043a71e2`.

vjt:
Gated: vjt asked for this to land after the in-flight code review, so it is de-queued for now. Re-queue once the review work clears.

vjt:
Finding from #717 investigation — it changes what a fix for this issue has to do.

**A boot fetch that fails FAST lands in the same dead end as one that hangs**, so a
timeout alone does not close this issue: a timeout that rejects reproduces the bug
exactly.

The chain, verified against `origin/main` (throwaway vitest with a real
`createResource` and this issue's own predicate, since deleted):

[...]

Issue #674, opened 2026-08-02
cicchetto: auto-refresh the running client on new bundle instead of requiring a banner click


vjt:
## Request

Update the running client automatically instead of requiring a manual click on the refresh banner. Requested via chat: "why do I have to click reload by hand?"

## Current behaviour

The pieces are all there; the last step is deliberately manual.

- The service worker is `registerType: "autoUpdate"` (`cicchetto/vite.config.ts:92-95`), so a deploy that bumps asset hashes bumps the SW bytes, the new SW installs, and `activate` evicts the old precache — with no prompt. That part is already automatic.
- What `autoUpdate` does **not** do is touch the page that is already open: the tab keeps executing the JS bundle it loaded at boot. A long-lived PWA session therefore runs old code indefinitely after a deploy.
- Detection of that mismatch is live, not poll-based: `bootBundleHash` is read once from the `<script src="/assets/index-<hash>.js">` tag, `serverBundleHash` arrives over the user topic on join and on `POST /admin/cic-bundle-changed`, and `shouldShowRefreshBanner()` is the strict-inequality compare (`cicchetto/src/lib/bundleHash.ts:87-91`).
- The banner is projected by `errorBanners.ts:145-153` as an `info` entry with a `Refresh` action wired to `performRefresh()` → `window.location.reload()`.

So today: detection automatic, application manual.

## Why the manual step exists (and why "just auto-reload" is not the whole answer)

A reload throws away in-memory UI state on a client the user may be actively typing into. At minimum that is the compose draft and scroll position. This is not hypothetical — #666 covers a case where losing/duplicating draft content is the actual bug — so a blind `location.reload()` on deploy would trade a minor annoyance for data loss, and would do it at exactly the moment an operator is shipping.

## Proposed shape

Auto-refresh **when it is provably safe**, keep the banner otherwise:

- Primary gate, as proposed in the original request: **the compose textbox does not have focus.** It is the cheapest honest signal that the user is not mid-sentence, and it needs no timer.
- Focus alone is not sufficient on its own, though: text typed and then clicked away from is unfocused but not disposable. So pair it with **no non-empty draft in any window** and **no modal/dialog open**. Tab hidden should also qualify as safe.
- Otherwise leave today's banner exactly as it is — the user is mid-task and owns the timing.

Worth considering as a complement (or as the more thorough fix): persist compose drafts and scroll position across reload. If a refresh becomes non-destructive, the safety gate above mostly stops mattering and auto-update can be the default.

[...]

PR #776, opened 2026-08-03
cic(#717): make a failed boot recoverable — bounded fetch, one root door, error boundary

vjt:
Refs #717.

## Scope, stated up front

**This does NOT fix the #717 trigger, and #717 stays open.** The report is "the installed PWA on Android + Firefox hangs on the CRT splash forever; a browser tab on the same phone never does". The trigger is not root-caused: the issue's own first measurement is device-side (desktop Firefox attached to the phone via `about:debugging`) and nobody has taken it.

What lands here is the **recoverability** half — the part decidable from the code, and the part that turns the next user report into evidence instead of another dead end.

## The measured mechanism

`CrtSplash.loading()` is `!user() || channelsBySlug() === undefined`. Those are `createResource`s fed by `me()` / `listNetworks()` / `listChannels()`. When one of those fetches **rejects**, the resource enters state `errored`, and reading an errored Solid resource **re-throws**.

cicchetto mounted **no `ErrorBoundary` anywhere** — a grep for the literal returned zero mounts. So the DOM kept the last painted frame, which on a cold boot is the splash, forever. The only way out was force-killing the app.

Three things were needed, and each is load-bearing:

1. **A bound on the transport** (`lib/bootFetch.ts`) — the three boot GETs had no timeout, no abort, no retry.
2. **An error context at root creation** (`lib/moduleRoot.ts`) — a throw from a computation in a bare module-level `createRoot` propagates out of `runUpdates`, which nulls `Effects`/`Updates` before rehandling. The queued render effects are **discarded**, so no render-tree boundary ever sees it.
3. **A user-facing recovery affordance** (`BootErrorBoundary.tsx`) — a console line is not a way out.

**A timeout alone cannot fix this, and that is the point.** Bounding a hang converts it into a rejection, and a rejection is exactly the state that freezes the UI. Either half shipped alone would look like a fix and change nothing the operator sees.

## One door for module roots

The first cut put the error context in `identityScopedStore` alone. Review caught that with a measured matrix: `activeWindows.ts` — a bare root whose memo reads `channelsBySlug()` and `networks()`, subscribed before the first render — runs **earlier** in the Updates queue than the store's memo, so it aborted the cycle first and the `listNetworks`/`listChannels` failures still froze the splash. Only `/me` was actually fixed.

Two patterns, and the boot fell through the gap between them. CLAUDE.md is explicit ("half-migrated creates two patterns; no exclusion lists"), so the context moved to `moduleRoot` — **one door** — all 30 module-lifetime roots across 26 files were migrated, and `__tests__/moduleRootGuard.test.ts` fails the build if a bare `createRoot` reappears.

Test files are out of that rule on a real distinction, not a convenience: their roots take the `dispose` callback and are scoped to the case, which is precisely what `createRoot` is for.

## The reset loop is deliberately OUTSIDE the context

`identityScopedStore` wraps only `build` in `withErrorContext`; its `on(token)` effect is registered after. Inside, a reset that throws would be caught and logged, and because the loop aborts at the throw the store's **remaining** resets would be silently skipped on logout or rotation — #281's failure mode, whose missed identity purge produces the 404 burst that trips the host's fail2ban jail and firewall-bans the client. Never widen the catch to swallow more.

## Review found a blocking defect in the retry, and it is fixed

Three review rounds ran; each of the first two found a real defect. The third found the retry was a **no-op on the first press** for any failure downstream of `/me`.

`reset()` is synchronous; the `user → networks → channelsBySlug` cascade is not. Refetching only the root left `networks` errored with nothing in flight, so the re-render rethrew and the boundary bounced straight back into its own fallback. Requests went out, but nothing changed under the user's finger and recovery needed a **second** press — the force-kill-the-app behaviour this issue exists to remove.

It was reproduced by execution before being believed, and falsified afterwards: reverting to the single verb turns the new case red and leaves the other three green. That is exactly how it hid — the existing retry spec covered only the `/me` failure, the one link where refetching the root alone happens to work.

## Reload must not purge the cache offline

The failure screen carries a `Reload` beside `Retry`, using `performRefresh` (the existing SW-aware verb from #674/#695) **only when online**. That verb deletes every cache before reloading, which #674 can do safely because a bundle-hash advertisement means the client is online by construction. This screen inverts the precondition: the likeliest reason it is up is that the network is gone, and purging the precache then lands the installed PWA on the browser's offline error page with the app shell destroyed — strictly worse than the frozen splash. Offline it does a plain reload. Reuse the verb, not the noun.

## Boundary with #687

#687 owns the splash's UI: the staged boot log and a per-stage stuck state during a **healthy** boot. This change adds no competing layer — the failure screen is a terminal state that replaces the Shell subtree, deliberately bare, and must not become the place staged-progress copy accretes.

## Known residuals, recorded in DESIGN_NOTES

- `bootFetch` is attached to the FUNCTION, not the boot call site, so `me()` from `mountBadgeReconcile` and the WS-driven refetches inherit the retry budget. Unintended, harmless, accepted.
- 8s per attempt x 3 attempts + backoff = worst case ~26.5s before the failure screen appears; `channelsBySlug` fans out with `Promise.all` so each network gets its own budget.
- Recovery depends on some component **inside** the boundary reading the specific errored resource. True today (`CrtSplash`, `Sidebar`, `BottomBar`), but it is a new load-bearing invariant with no test.

## Test plan

- [x] `scripts/bun.sh run check` — rc=0
- [x] `scripts/bun.sh x tsc --noEmit` standalone — rc=0 (biome red short-circuits tsc, so run separately)
- [x] `scripts/bun.sh run test` — 221 files / 3950 tests green
- [x] `scripts/bun.sh run build` — rc=0
- [x] `scripts/integration.sh` FULL suite — 589 passed / 1 failed
- [ ] Device verification on the reporter's Android PWA — **not done, and cannot be from here**

**No Elixir is touched by this diff, so `scripts/check.sh` was deliberately not run.** Not "it passed" — it was not run.

### About the one integration red

The full-suite failure is `issue580-send-snap-independent-of-post.spec.ts:104:3` case 1 (`toBeInViewport()`), which is **#769, a defect on main, not this branch**. Proven by baseline rather than asserted: the same spec fails at plain `7ead6ade` in a detached worktree containing none of this diff (`--grep 'issue #580' --repeat-each 10` → 1 failed / 9 passed in 34.8s). Full evidence posted on #769, including that the failure is a scroll-position failure — the line is in the DOM with a real `data-msg-id`, it is simply never scrolled into view.

`issue281` was also looped on both arms (5x each, main naked and this branch): 10/10 green, so it did not reproduce and is not covered by that finding.

### About cic suite stability

`focus-rule.test.ts` "incoming PRIVMSG on a channel does not change focus" timed out (flat 5000ms) in **one of three** post-rebase full runs; the other two are 3950/3950 green, it is 3/3 green isolated, and plain `c1a0e535` is 3932/3932 green. The failure mode is the clock, not the focus assertion — if this diff had broken the invariant the assertion would fail, not time out. Flagging it rather than burying it: 2-of-3 is not proof of innocence, and the test looks load-fragile.

### DIFF
--- a/cicchetto/src/BootErrorBoundary.tsx
+++ b/cicchetto/src/BootErrorBoundary.tsx
@@ -75,28 +75,46 @@ const BootFailure: Component<{ error: unknown; onRetry: () => void }> = (props) 
             resources) loops straight back here. Force-killing the app is what
             #717 exists to eliminate; this is what replaces it.
 
-            OFFLINE IT MUST NOT PURGE. `performRefresh` is the right SW-aware
-            verb when a stale bundle is the problem (#674's banner, #695's
-            stale-resume) — but it deletes every cache before reloading, and
-            #674 can only do that safely because a bundle-hash advertisement
-            means the client is online by construction. This screen inverts that
-            precondition: the likeliest reason it is up at all is that the
-            network is gone. Purging the precache and reloading offline lands
-            the installed PWA on the browser's offline error page with the app
-            shell destroyed — strictly worse than the frozen splash. Reusing the
-            verb, not the noun: the shared part is "reload", the cache purge is
-            the 20% that does not fit. */}
+            IT MUST NOT PURGE WITHOUT PROOF THE SERVER ANSWERS. `performRefresh`
+            is the right SW-aware verb when a stale bundle is the problem
+            (#674's banner, #695's stale-resume) — but it deletes every cache
+            before reloading, and #674 can only do that safely because a
+            bundle-hash advertisement means the client is online by
+            construction. This screen inverts that precondition: the likeliest
+            reason it is up at all is that the network is gone. Purging the
+            precache and reloading then lands the installed PWA on the browser's
+            offline error page with the app shell destroyed — strictly worse
+            than the frozen splash. Reusing the verb, not the noun: the shared
+            part is "reload", the cache purge is the 20% that does not fit.
+
+            `navigator.onLine` ALONE CANNOT GATE THAT, which is why the error is
+            read too. It is trustworthy in one direction only: false means
+            definitely no link, true means only "an interface exists". Behind a
+            captive portal, on dead mobile data with the radio still attached,
+            or on a Wi-Fi association with no route, it reads true — and those
+            are exactly the conditions this screen comes up in. The guard was
+            weakest in its own motivating case.
+
+            So the purge needs POSITIVE evidence the server is reachable, and
+            an `ApiError` is precisely that: it exists only because a response
+            came back with a status. A transport failure is not one — `bootFetch`
+            has already spent three attempts and the server never answered.
+            Anything else (a mid-session render throw) is connectivity-unknown,
+            and unknown takes the safe branch by the asymmetry of harm: purging
+            offline destroys the app shell, while not purging costs one more
+            reload and leaves #674's banner to offer the new bundle once the
+            link is back. */}
       <button
         type="button"
         class="boot-failure-reload"
         data-testid="boot-failure-reload"
         onClick={() => {
           setReloading(true);
-          if (isOffline()) {
-            window.location.reload();
+          if (props.error instanceof ApiError && !isOffline()) {
+            void performRefresh();
             return;
           }
-          void performRefresh();
+          window.location.reload();
         }}
       >
         {/* `performRefresh` can take up to ~2s waiting on controllerchange.
--- a/cicchetto/src/BootErrorBoundary.tsx
+++ b/cicchetto/src/BootErrorBoundary.tsx
@@ -1,0 +1,136 @@
+import { type Component, createSignal, ErrorBoundary, type JSX } from "solid-js";
+import { ApiError } from "./lib/api";
+import { performRefresh } from "./lib/bundleHash";
+import { isOffline } from "./lib/connectivity";
+import { friendlyApiError } from "./lib/friendlyApiError";
+import { refetchUser } from "./lib/networks";
+
+// #717 — keep a failed boot recoverable.
+//
+// The boot chain is three resources (`user` → `networks` → `channelsBySlug`,
+// see lib/networks.ts) and every consumer reads them through a predicate:
+// CrtSplash's is `!user() || channelsBySlug() === undefined`. When one of the
+// underlying fetches REJECTS, the resource enters state `errored` and reading
+// it RE-THROWS. That throw lands mid-render, and before this component
+// cicchetto mounted no ErrorBoundary at all — so the throw was swallowed, the
+// DOM kept the last painted frame (the splash), and the only way out was to
+// force-kill the app. That is the reported #717 symptom on the installed
+// Android/Firefox PWA.
+//
+// Why this is not solved by the timeout in `lib/api.ts`: a timeout REJECTS.
+// Bounding the slow path converts a hang into a rejection, which is the exact
+// state that freezes the UI. The bound and this boundary are two halves of one
+// fix — retry absorbs the transient failure, the boundary catches the terminal
+// one.
+//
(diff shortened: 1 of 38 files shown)


vjt:
## Update: rebased onto `4441d67e`, plus one more review fix

Force-pushed `2488b4a5`. PR is **CLEAN / MERGEABLE** again — #773's merge had left it conflicting, so no CI was running at all.

The rebase conflict was `main.tsx`: #674 added a comment block above the root this branch converts from `createRoot` to `moduleRoot`. Both kept.

### New commit: the Reload button's purge guard was wrong in its own motivating case

Review round three flagged the `isOffline()` gate, and it was right — this is the feature not working, so it landed here rather than in the follow-up issue.

`performRefresh` deletes **every cache** before reloading. It was gated on `isOffline()`, i.e. `!navigator.onLine`, which is trustworthy in one direction only: `false` means definitely no link, `true` means merely "an interface exists". Behind a captive portal, on dead mobile data with the radio still attached, or on a Wi-Fi association with no route, it reads `true`. Those are exactly the conditions this screen comes up in — so on the reporter's Android PWA with the data link dead, the tap would have purged the precache and reloaded into the browser's offline error page **with the app shell destroyed**, which is strictly worse than the frozen splash this whole PR exists to replace.

The purge now needs positive evidence the server answers: `error instanceof ApiError && !isOffline()`. An `ApiError` exists only because a response came back with a status; a transport failure is not one, and by then `bootFetch` has spent three attempts without the server ever answering. Both conditions, because a status that came back earlier does not mean the link survived.

Accepted trade-off, documented in DESIGN_NOTES: a stale bundle causing a mid-session render throw is no longer cured by this button. That is #674's door, and it is the cheap failure — unknown connectivity takes the plain reload by the asymmetry of harm.

Falsified: restoring the `isOffline()`-only gate turns the transport-failure case red and leaves the other two reload cases green, which is precisely why the old gate looked correct.

### The other four review findings are NOT here

Split to #779, deliberately, to keep this PR scoped to recoverability. None is a live defect: the `moduleRoot` guard enforces a spelling rather than the invariant (a rootless module-scope `createEffect` reproduces #717 and the regex cannot see it — verified by execution, zero instances today), its anti-vacuity case can be satisfied by a comment, it has no escape hatch for a legitimately disposable root, and `bootFetch`'s per-attempt timeout also bounds the body read whose abort is never retried.

### Gates on `2488b4a5`

- `scripts/bun.sh run check` rc=0
- `scripts/bun.sh x tsc --noEmit` standalone rc=0
- `scripts/bun.sh run test` — **222 files / 3986 tests green**
- `scripts/bun.sh run build` rc=0
- No Elixir touched, so `scripts/check.sh` was **not run**. Not "it passed".

### Read this before trusting a red cic suite here

The full vitest suite is **order/load sensitive on an unchanged tree** — filed as #781. Two of four full runs on this branch went red in `userTopic-rotation.test.ts` / `archive.test.ts` / `windowClose.test.ts` / `focus-rule.test.ts`, all of which are green in isolation (`userTopic-rotation` 5/5, `focus-rule` 3/3, the trio together 31/31).

It is not this branch's production code, and that is proven rather than asserted: removing this branch's **four new test files** while leaving the production diff fully in place gives 218 files / 3965 green — identical to main naked at `4441d67e`. Restoring them gives 222/3986 green. The trigger correlates with test-file scheduling, not with code under test.

One of those failures deserves its own look and is called out in #781: `AssertionError: expected "vi.fn()" to be called 1 times, but got 2 times` — a **double join** of the user topic. That is not a slow clock, and a duplicate user-topic join is the shape #281's replay bug takes.

### Unrelated e2e reds, for completeness

The integration suite's `issue580` red is #778 (split out of #769, which is now `issue281` only — they are two different bugs, proven by rate and mechanism, not one "family"). Both reproduce locally at plain main; neither involves this branch.

Issue #736, opened 2026-08-03
cic: the passkey surface has no tests, and the Login mock omits two functions it calls


vjt:
Found by the codebase-review campaign on `main` (`v0.10.0..388f883b`), `cicchetto/src` slice. Verified against the files at `388f883b`.

**File:** `cicchetto/src/__tests__/Login.test.tsx:11`-`:18`
**Severity:** MEDIUM

## Problem

The passkey surface shipped with no unit coverage at all. The new tests in this range cover two TOTP happy paths (`auth.test.ts`) and one TOTP form test. Nothing covers `lib/passkeys.ts` (including the base64url round-trip, a silent-corruption candidate), `PasskeySettings`, `TotpSettings`, the passkey→TOTP fallback branch in `auth.login` (the most conditional logic in the slice), or any failure path — cancel, expiry, wrong code.

The mock actively blocks anyone from adding that coverage. `vi.mock("../lib/auth", …)` declares `login`, `verifyTotp`, `logout`, `token`, `isAuthenticated`, `setToken` — but `Login.tsx` also calls `auth.loginWithPasskey` (line 331) and `auth.loginWithRecoveryCode` (line 343).

## Failure scenario

The first person to write a test that clicks "Passkey" or "Recover account" gets `TypeError: auth.loginWithPasskey is not a function` instead of a meaningful assertion. Meanwhile every other finding filed from this review against the auth surface — the form-nesting bug, the missing feature detection, the missing in-flight guard, the silent TOTP degrade — is invisible to the suite.

## Suggested fix

Add the two missing entries to the mock factory, then cover at minimum: the `passkeys.ts` encode/decode round-trip, the `auth.login` passkey→TOTP fallback with `challenge_token` both null and non-null, and one cancel path per component.

vjt:
Shipped in **v0.11.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.11.0

The tag carries `.deb`, `.rpm` and `pkg.tar.zst` assets, so self-hosters have something to pull.

Note this is a **release, not a deploy**: the hosted service is unchanged and still runs the previous build until the code-review queue closes.

Issue #712, opened 2026-08-03
test(cicchetto): passkey and recovery-code login entry points have no e2e coverage


vjt:
The passkey / recovery-code entry points merged in #696 (squashed as `c6045a9c`) ship with **zero end-to-end coverage**.

### Evidence

On `origin/main`, across the 330 spec files under `cicchetto/e2e/tests/`:

```
git grep -il "passkey\|recovery.code\|login-alt-auth" origin/main -- cicchetto/e2e/
# (no output)
```

The UI itself is on main — `login-alt-auth` is present in `cicchetto/src/Login.tsx` and `cicchetto/src/themes/default.css` — so this is a genuine gap, not a naming mismatch in the grep.

### Why it matters

The two regressions that held #696 back for three CI rounds were both caught by *unrelated* specs that happen to touch the login card:

- `issue281-account-switch-no-replay:111` — strict-mode violation, because the new buttons reused the `.login-advanced-toggle` class and the visitor branch then rendered three of them.
- `login-advanced-scroll-reachability` — the buttons added a row to the card, pushing `Connect` out of the viewport at 390x480 and again at 1024x900.

Both failures were about *neighbouring* elements. Nothing asserts that the new controls themselves render, are reachable, or do anything when clicked. The next change to the login card can silently break them.

### Suggested coverage

1. Visitor login form renders both entry points (passkey + recovery code) and they are in the viewport at 390x480 and 1024x900.
2. Clicking each one navigates to / opens the expected flow.
3. Both are keyboard-reachable in tab order, `Connect` still last.
4. A guard that the login card height stays within the shortest supported viewport with Advanced expanded — the 1024x900 case only had 25px of slack when measured, which is less than one tap target, so any future row added here breaks it again.

Follow-up to #696 / refs #442.

vjt:
Manual verification (reported on IRC by Hypnotize, 2026-08-03 10:19 CEST): the passkey login flow was exercised end-to-end on **iOS with ProtonPass** against aLe's self-hosted instance — registration and login both work.

This is currently the *only* evidence that the WebAuthn ceremony works on real hardware: CI does not execute it (no e2e spec references the passkey/recovery entry points, see the issue body), and the ceremony itself cannot be driven by the headless Chromium/WebKit projects without a virtual authenticator.

That does not close this issue — the automated coverage described above is still missing — but it does mean the shipped feature is known-good on at least one real platform.

vjt:
[...]

PR #795, opened 2026-08-03
test(cic): cover the passkey surface and unblock the Login mock (#736)

vjt:
Addresses #736 (code-review campaign, `cicchetto/src` slice).

## The mock

`Login.test.tsx`'s `vi.mock("../lib/auth", …)` declared six functions;
`Login.tsx` calls eight. The two missing ones — `loginWithPasskey` and
`loginWithRecoveryCode` — are the whole #442 alternate-auth row, so the
first test to click either button got `TypeError: … is not a function`
instead of an assertion.

**Fixing the mock turned nothing red: 28/28 before, 36/36 after.** That is
the finding, not a relief — the omission was not hiding a failing
behaviour, it was *preventing the surface from being tested at all*. No
existing test leaned on the false factory.

## The coverage that was missing

| file | what it now pins |
|---|---|
| `lib/__tests__/passkeys.test.ts` (new, 15) | base64url ⇄ ArrayBuffer round-trip, URL-alphabet direction, stripped `=` pad, both option rebuilds (`display_name` → `displayName`), null `user_handle`, absent `getTransports`, both cancel guards |
| `__tests__/auth.test.ts` (+7) | all four outcomes of the `auth.login` passkey → TOTP degrade, plus `loginWithPasskey` / `loginWithRecoveryCode` |
| `__tests__/Login.test.tsx` (+8) | both alt-auth buttons: sanitization, empty-field refusal, ApiError vs plain cancel Error, recovery round trip |
| `__tests__/PasskeySettings.test.tsx` (+2) | registration happy path + a cancelled ceremony that leaves the pane usable |
| `__tests__/TotpSettings.test.tsx` (new, 4) | enrolment, wrong code, rejected disable password |

The round-trip test echoes the decoded challenge back out as `raw_id`, so
it is an identity check through the real encode/decode pair — neither half
is re-implemented in the test to drift against.

Both component tests assert the pane stays **usable** after the failure: a
`busy` flag stranded by a rejected ceremony disables every button until a
reload, and dropping the pending TOTP enrolment on one typo would
invalidate the QR the user already scanned into their authenticator.

## Proof the tests can fail

Nine defects injected into production code, one at a time, each reverted
after its run. Every one was killed by exactly the intended test(s):

| mutation | red |
|---|---|
| decode: swap `-`→`/`, `_`→`+` | 3 (both `-_` samples + the alphabet pin) |
| encode: keep the `=` padding | 6 |
| `getPasskey`: drop the null-credential guard | 1 (cancel) |
| `createPasskey`: `displayName: user.name` | 1 (creation rebuild) |
| `auth.login`: swallow a failed ceremony as authenticated | 1 (no-fallback rethrow) |
| `auth.login`: never degrade to TOTP | 2 (both degrade cases) |
| `Login`: skip identifier classification | 3 (sanitize + both empty refusals) |
| `PasskeySettings`: leave `busy` latched | 1 (retry) |
| `TotpSettings`: drop the enrolment on error | 1 (enrolment survives) |

No mutation survived; the tree was clean after every restore.

## Gates

* `scripts/bun.sh run test` — **225 files / 4046 tests passed, rc=0**
* `scripts/bun.sh run check` (biome + tsc) — rc=0

cic-only, no lane taken. No production code changed.

## Conflict boundaries respected

Does not touch `PasskeySettings.tsx`, `TotpSettings.tsx` or
`themes/default.css` (PR #794's files) — the pane tests drive them from
the outside and avoid #794's `.share-qr` → `.qr-frame` rename. Nothing
under `cicchetto/e2e/` (worker-2's #712).

No defects found worth filing: the surface behaved correctly under every
failure path exercised.

### DIFF
--- a/cicchetto/src/__tests__/Login.test.tsx
+++ b/cicchetto/src/__tests__/Login.test.tsx
@@ -8,9 +8,16 @@ import { ApiError } from "../lib/api";
 // raw HTTP layer so the form test stays focused on form behavior, not
 // transport. The auth-store ⇄ api wiring is covered separately in
 // auth.test.ts.
+// #736 — the factory MUST declare every auth function `Login.tsx` reaches.
+// It used to omit `loginWithPasskey`/`loginWithRecoveryCode` (the #442
+// alternate-auth row), so the first click on either button raised
+// `TypeError: … is not a function` instead of a meaningful assertion — the
+// whole passkey surface was untestable from here.
 vi.mock("../lib/auth", () => ({
   login: vi.fn(),
   verifyTotp: vi.fn(),
+  loginWithPasskey: vi.fn(),
+  loginWithRecoveryCode: vi.fn(),
   logout: vi.fn(),
   token: vi.fn(() => null),
   isAuthenticated: vi.fn(() => false),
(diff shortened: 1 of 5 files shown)


vjt:

PR #796, opened 2026-08-03
e2e: cover the login card's passkey / recovery-code entry points

vjt:
Follow-up to #696 / refs #442, addressing the gap reported in #712. Test-only: one new spec file, zero changes under `cicchetto/src/`.

## Summary

- The alt-auth controls shipped with **zero** direct coverage. Both regressions that held #696 back for three CI rounds were caught by specs that only *neighbour* them — `issue281-account-switch-no-replay` (strict-mode violation once the buttons reused `.login-advanced-toggle`) and `login-advanced-scroll-reachability` (the extra row pushed `Connect` out of the viewport). This guards those two at their **cause**: the toggle selector stays a single match, and the pair keeps riding the Advanced row (tops within 2px) instead of stacking a row that eats the 25px of slack the 1024x900 card has left.
- Adds the behaviour nobody was watching: an empty identifier is refused client-side with **no** request; a filled one fires `POST /auth/passkeys/options` with exactly `{identifier}`; recovery posts `{identifier, recovery_code}` — the post-#770 snake_case wire — and both refusals surface as the `friendlyApiError` copy the user actually reads. A real Tab walk pins the order with `Connect` still last, collapsed and expanded.
- Written against **current `origin/main` (`123720d5`, post-#770)**, so the wire assertions match the regenerated shape rather than a remembered one.

## Deliberately out of scope

The WebAuthn ceremony itself. The options step 401s for an identifier with no passwordless account, so click → request → visible-error runs end to end without a virtual authenticator and stays browser-agnostic. A real assertion needs CDP (chromium-only) and belongs to a passkey *registration* spec.

## Not assumed green — mutation-proven

A build carrying four targeted breaks (Passkey re-using `.login-advanced-toggle`; the pair moved out of the row; the empty-identifier guard dropped; `Connect` moved above the row) put **6 of 7 tests red** while the untouched passkey-request test stayed green — so the mutations were targeted, not a blanket smash.

That run also falsified one of my own asserts: a `toBeInViewport` on the revealed recovery field went red on card **geometry**, not on recovery behaviour. It is gone — that guard belongs to the two geometry tests here and to the neighbour spec, which drives a real wheel gesture instead of a static check.

## Test plan

- [x] Scoped run green: 7/7, rc=0 read from a file
- [x] Mutation run: 6 red / 1 green, matching a pass/fail matrix registered before the run
- [x] Full `scripts/integration.sh`: **597 passed, 0 failed** (27.0m), rc=0 — not a scoped `--grep`
- [x] Count diff across the gate: 590 → 597, with all 7 new titles present as `✓` in the full log
- [x] Re-runnable: per-run-unique identifier, no seeded account state, no sleeps

### DIFF
--- a/cicchetto/e2e/tests/login-alt-auth-entry-points.spec.ts
+++ b/cicchetto/e2e/tests/login-alt-auth-entry-points.spec.ts
@@ -1,0 +1,266 @@
+// #712 — the login card's alternate-auth entry points (Passkey / Recovery
+// code, shipped in #696 / refs #442) had ZERO direct coverage.
+//
+// ## Why this spec exists
+//
+// Both regressions that held #696 back for three CI rounds were caught by
+// specs that only *neighbour* these buttons:
+//
+//   * `issue281-account-switch-no-replay` — strict-mode violation, because
+//     the new buttons reused the `.login-advanced-toggle` class and the
+//     visitor branch then rendered three of them.
+//   * `login-advanced-scroll-reachability` — the buttons added a ROW to the
+//     card, pushing Connect out of the viewport at 390x480 and at 1024x900.
+//
+// Nothing asserted that the controls themselves render, are reachable, or do
+// anything when clicked. This spec closes that: it guards the two failure
+// modes that already happened (the selector contract, the card geometry) AND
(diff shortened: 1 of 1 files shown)


vjt:

Issue #738, opened 2026-08-03
cic: the directory search box never shows a filter set by /list <pattern>


vjt:
Found by the codebase-review campaign on `main` (`v0.10.0..388f883b`), `cicchetto/src` slice. Verified against the files at `388f883b`.

**File:** `cicchetto/src/DirectoryPane.tsx:171` (`searchText` init) and the comment at `:172`-`:177`; `cicchetto/src/lib/compose.ts:1089`
**Severity:** MEDIUM
**Related:** #677

## Problem

`searchText` is hard-initialised to `""`, justified by this comment:

> the search key is cleared on window close … so searchText always mounts "" and the store agrees

That premise is no longer true. `/list <pattern>` calls `setQuery(networkSlug, cmd.pattern)` from `compose.ts:1089` — the store's query is settable from OUTSIDE the pane, with no window-close in between.

The sibling control got this right: #677 rehydrates the sort toggle from `directorySort(slug)` (line 179) with the explicit rationale that a control must not lie about the store's view. There is no `directoryQuery` export to do the same for the filter — the module exports `directorySort` but no query counterpart.

## Failure scenario

The operator types `/list rust`. `$list` opens, the store's view is `q: "rust"`, and the pane renders only matching channels — under an EMPTY search box. The operator sees a short, arbitrary-looking channel list with no visible reason for it, and cannot clear the filter except by typing a character into the empty box and deleting it.

## Suggested fix

Export a `directoryQuery(slug)` sibling of `directorySort` and seed `searchText` from it in the same effect that seeds `activeSort`. Fix the comment's premise while there.

vjt:
Shipped in **v0.11.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.11.0

The tag carries `.deb`, `.rpm` and `pkg.tar.zst` assets, so self-hosters have something to pull.

Note this is a **release, not a deploy**: the hosted service is unchanged and still runs the previous build until the code-review queue closes.

Issue #732, opened 2026-08-03
cic: the channel directory has no error path and no request-ordering guard


vjt:
Found by the codebase-review campaign on `main` (`v0.10.0..388f883b`), `cicchetto/src` slice. Verified against the files at `388f883b`.

**File:** `cicchetto/src/lib/channelDirectory.ts:53` (`fetchInto`), `:58` (the unguarded write); callers at `cicchetto/src/DirectoryPane.tsx:211`, `:258`
**Severity:** MEDIUM
**Related:** #677

## Problem

Two defects in the same function, with the same root: the write path has no guard and no failure path.

**No error handling.** Neither `fetchInto` nor `loadMore` catches — there is no `catch` anywhere in the module — and both callers invoke them as `void`. A rejection is an unhandled promise rejection, `pages[slug]` stays undefined, and `<Show when={page()}>` renders nothing: no message, no retry, and the mount effect only re-fires on a slug change. #677 made this sharper by dropping the cached page on close, so a transient failure that previously showed a stale snapshot now shows an empty pane. CLAUDE.md: *"No silent-swallow at boundaries."*

**No request ordering.** `onSearchInput` fires `setQuery` on every keystroke with no debounce, and line 58 writes `setPages(prev => ({...prev, [slug]: page}))` unconditionally when its GET resolves. `loadMore` got an explicit anti-clobber guard (line 95, commented "so a fetchInto that landed while this GET was in flight isn't clobbered"); the reverse direction — fetchInto vs fetchInto, and a late fetchInto vs an append — is unguarded.

## Failure scenario

**Blank pane:** `$list` is reopened while the bouncer is restarting. The pane shows the header and the search box forever. The operator cannot tell whether the network has zero channels or the request failed.

**Stale results:** type `ru` then `rust` quickly; the `ru` GET resolves last, so the pane shows `ru` results under a box reading `rust`, and `next_cursor` now belongs to the wrong view, so load-more pages the wrong query. Same shape when a progress-ping refresh lands after the user has appended three pages: all of them silently vanish.

## Suggested fix

Add a per-slug error field to the store (mirroring the per-row `error` signal in `DirectoryRow`) rendered with a retry affordance, routed through `friendlyApiError`; and stamp each `fetchInto` with a monotonic request id (or the `(sort, q)` tuple), dropping the write when it is no longer the newest. Debounce `onSearchInput` while there.

vjt:
[...]

Issue #677, opened 2026-08-02
Channel directory: never paginates past page 1, and the search key outlives its input


vjt:
## Symptom

The channel directory shows only the first ~100 channels and never loads more. Scrolling to the bottom of the list does nothing.

Reported on #grappa against a Libera session: the header reports 6585 channels, the list holds about 100. vjt reproduced it — "arrivando in fondo non pagina".

## Root cause — the client never asks for page 2

Traced on `origin/main`; every layer below the store is already correct, the wiring simply stops one step short.

- **Server paginates correctly.** `Grappa.ChannelDirectory` has `@default_limit 100` (`lib/grappa/channel_directory.ex:60`), queries `limit + 1` rows (`:167`) and hands back a `next_cursor` from `paginate/3` (`:211`).
- **The HTTP client already supports the cursor.** `listDirectory` (`cicchetto/src/lib/api.ts:1904`) takes `opts.cursor` and forwards it as a query param.
- **The store drops it.** `fetchInto` in `cicchetto/src/lib/channelDirectory.ts:35` calls `api.listDirectory(t, slug, { sort: view.sort, q: view.q })` — no `cursor`, ever. Its own comment is explicit: *"fetches page 1 of the current view … Pagination / load-more is the pane's concern (task E3); the store always re-fetches from the top on any event."*
- **The pane never implements E3.** `DirectoryPane.tsx` has no load-more path: no `IntersectionObserver`, no sentinel row, no `next_cursor` consumption. Its single `onScroll` handler (`:281`) exists only to save `scrollTop` for scroll preservation across re-renders (`:38`, `:190`).

So `next_cursor` is produced by the server, typed on the wire (`wireTypes.ts:437`), plumbed through `api.ts` — and then read by nobody. Task E3 was never landed, and the first page is all a user can ever see. The "about 100" the reporter counted is exactly `@default_limit`.

## Fix

Consume `next_cursor`: append-on-scroll in `DirectoryPane` (sentinel + `IntersectionObserver`) driving a store action that fetches with `cursor` and **appends** rather than replaces.

Points that need care:

1. **Append vs replace.** Every existing store entry point (`loadDirectory`, `setSort`, `setQuery`, and the three ping hooks `onDirectoryProgress` / `onDirectoryComplete` / `onDirectoryFailed`) calls `fetchInto`, which *replaces* the page. A load-more must not go through that path or it will wipe the accumulated rows.
[...]

PR #797, opened 2026-08-04
cic(#738): make the directory search box follow the store's filter

vjt:
Issue #738. `/list <pattern>` calls `setQuery(networkSlug, cmd.pattern)` from
`compose.ts:1239`, so the directory's filter is settable from OUTSIDE the pane
with no window close in between. `searchText` was hard-initialised to `""` on
the premise that the filter is only ever cleared on close and therefore
"always mounts empty and the store agrees" — a premise that died when the
pattern arg was wired. The operator typed `/list rust`, got a short filtered
list under an empty box, and could only clear it by typing a character and
deleting it again.

## The fix

Follow the sibling control. #677 rehydrates the sort toggle from
`directorySort(slug)` precisely so a control cannot lie about the store's
view; the store had no query counterpart, so this adds `directoryQuery(slug)`.

**One place it departs from the sort pattern, deliberately.** Sort is only ever
set from inside the pane, so sampling it once at mount is sound. The filter is
not: `compose.ts` selects the `$list` window FIRST and calls `setQuery` after,
and a second `/list <pattern>` reaches a pane that is already mounted. A
one-shot seed would leave both cases broken. So the box FOLLOWS the store via
an effect. It keeps its local signal rather than binding to the store directly
because typing must LEAD the debounced GET (#732); the store echoes the same
text back ~250ms later, so the mirror write is a no-op.

The stale comment is gone rather than left to justify the next regression — an
assertion whose premise had quietly expired is what let this survive review.

## Mutation evidence

Every leg was broken and watched go red (`bun.sh run test`, iso):

| Mutation | Result |
|---|---|
| remove the mirror effect | **2 red** — mount + after-mount |
| `directoryQuery` returns `""` | **1 red** — the store test |
| revert to `createSignal(directoryQuery(slug))` seed **and** keep the effect | **0 red** → the seed was dead code, dropped (2nd commit) |
| bind the input straight to the store | **0 red** → that guard test could not fail, dropped (2nd commit) |

Two of those four killed the *test* rather than the code. jsdom keeps whatever
value `fireEvent.input` wrote to the DOM, so a test asserting "the typed text
survives" stays green even with the input bound to the store — it was a mirror,
and its honest content is already covered by the #732 debounce suite. Solid
flushes the mirror effect inside `render()`, before the first paint, which is
why the separate mount seed provably bought nothing.

## The sort-control check you asked for

`setSort` has exactly one caller — `DirectoryPane.tsx` itself. No compose verb,
no userTopic dispatch, nothing outside the pane. **No finding**: the sort
toggle's one-shot rehydration has no outside-set hole today. It would have the
identical bug the moment one is added, but there is nothing to fix now and I
have not touched it.

## Gates

* `scripts/bun.sh run check` → rc=0 (biome + tsc)
* `scripts/bun.sh run test` → rc=0, **229 files / 4125 tests**
* `scripts/bun.sh run build` → rc=0
* Working tree clean before and after.

cic-only; no server code, no lane taken, docker/e2e untouched.

## Inherited red

This branch's `integration` job will show the same
`tests/nick-follow-query.spec.ts` failure `origin/main` (`4f92701d`) currently
has. Not from this change — nothing here touches the server or e2e.

### DIFF
--- a/cicchetto/src/DirectoryPane.tsx
+++ b/cicchetto/src/DirectoryPane.tsx
@@ -4,6 +4,7 @@ import { token } from "./lib/auth";
 import {
   directoryError,
   directoryPage,
+  directoryQuery,
   directorySort,
   isLoadingMore,
   loadDirectory,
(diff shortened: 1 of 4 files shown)


vjt:

Issue #731, opened 2026-08-03
cic: DirectoryPane focuses the raw /LIST casing while every sibling folds


vjt:
Found by the codebase-review campaign on `main` (`v0.10.0..388f883b`), `cicchetto/src` slice. Verified against the files at `388f883b`.

**File:** `cicchetto/src/DirectoryPane.tsx:98` (`onJoin`) and `:115` (already-joined branch)
**Severity:** MEDIUM

## Problem

`channel_directory.name` is the documented verbatim-casing exception — the directory stores the `/LIST` spelling. Every sibling focus site folds before handing the name to `setSelectedChannel`:

- `lib/channelJoin.ts:37` — `channelName: canonicalChannel(rawChannel)`, with a comment stating that focus and the already-joined lookup use the FOLDED key.
- `lib/compose.ts:550` — folds the first `/join` element, added for #510 precisely because "a mixed-case first element targets a key no `window_states` entry matches, opening the empty phantom window".

`DirectoryPane` passes `props.entry.name` RAW at both call sites — even though its own `isJoined()` at line 77 folds via `channelKey`. This is the CLAUDE.md fold invariant: keys fold, display stays raw.

## Failure scenario

`/LIST` returns `#Sniffo` (bahamut preserves creation casing). Tapping the row joins successfully, but the selection becomes `{channelName: "#Sniffo"}` while `channelsBySlug` and `window_states` hold `#sniffo`. The sidebar shows no active row, and tapping the `#sniffo` sidebar row is then treated as a switch to a DIFFERENT window rather than a re-tap of the current one.

## Suggested fix

Wrap both `setSelectedChannel` calls in `canonicalChannel(props.entry.name)`, exactly as `channelJoin.switchTo` does. Keep `entry.name` raw for the `postJoin` wire argument and for the visible label.

vjt:
Shipped in **v0.11.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.11.0

The tag carries `.deb`, `.rpm` and `pkg.tar.zst` assets, so self-hosters have something to pull.

Note this is a **release, not a deploy**: the hosted service is unchanged and still runs the previous build until the code-review queue closes.

Issue #510, opened 2026-07-28
cicchetto: /join #a,#b focuses a phantom window named "#a,#b"


vjt:
## Symptom

`/join #bugs,#devel` in cicchetto opens (and focuses) a single phantom window literally named `#bugs,#devel` — header shows `#bugs,#devel  (no topic set)`, body `no messages yet`, composer placeholder `message #bugs,#devel`. The two real channels *are* joined upstream and do appear in the sidebar; only the focused window is bogus.

Reported on IRC with a screenshot after a `/join #bugs,#devel` on the production instance.

## Root cause — client side only, the server is correct

Server handles the RFC1459 list properly (#382):

- `GrappaWeb.ChannelsController.create/2` validates via `validate_channel_list/1` (every element must be a valid channel, no partial JOIN).
- `Grappa.Session.send_join/4` does `String.split(channel, ",")` → `fold_join_channels/1` → forwards **one** multi-target JOIN line and opens a `:pending` window **per channel**.

cicchetto does not split. In `cicchetto/src/lib/compose.ts`, `case "join"`:

```ts
await postJoin(t, networkSlug, cmd.channel, cmd.key);
...
setSelectedChannel({ networkSlug, channelName: cmd.channel, kind: "channel" });
```

`cmd.channel` is the raw, unsplit string (`slashCommands.ts` `join:` returns `{ kind: "join", channel, key }` with `channel` = first token verbatim). The POST is fine — the server splits it — but `setSelectedChannel` is handed `"#bugs,#devel"`, so focus lands on a channel key that no `window_states` entry will ever match. Hence the empty, topic-less window.

## Aggravating factor: the parser's own error message points at the bug

`slashCommands.ts` rejects the bare-name comma form and suggests the sigil form:

```
/join: bare names with commas are ambiguous — spell each channel out (e.g. /join #foo,#bar)
```

So a user who types `/join bugs,devel` gets an error that instructs them to type exactly the input that produces the phantom window. Previously the comma form simply failed; now it fails *with a suggestion into a worse failure*.

## Suggested fix

In `compose.ts` `case "join"`, split `cmd.channel` on `,` and focus the **first** element (the POST body stays the unsplit string — the server contract already accepts the list):

```ts
const targets = cmd.channel.split(",");
await postJoin(t, networkSlug, cmd.channel, cmd.key);
setSelectedChannel({ networkSlug, channelName: targets[0], kind: "channel" });
```

[...]

PR #798, opened 2026-08-04
cic(#731): fold the /LIST casing before focusing a directory row

vjt:
Issue #731.

## The defect

`channel_directory.name` is the documented verbatim-casing exception — it stores the `/LIST` spelling, and bahamut preserves the casing a channel was created with. Both focus sites in `DirectoryRow` handed that raw name straight to `setSelectedChannel`, while the row's OWN `isJoined()` folded through `channelKey`.

So tapping `#Sniffo` joined fine, but the selection became `#Sniffo` while `channelsBySlug` and `window_states` held `#sniffo`. The sidebar showed no active row, and tapping the `#sniffo` sidebar row afterwards read as a switch to a DIFFERENT window instead of a re-tap of the current one.

## The fix

Route both `setSelectedChannel` calls through `canonicalChannel` — the same helper `channelJoin.switchTo` uses, and the one `compose.ts` folds `/join`'s first element with (added for #510, filed for this exact phantom-window shape). Two call sites, one import.

This is the CLAUDE.md fold invariant, not a new rule: **only the KEY folds**. The `postJoin` wire argument and the rendered label keep the raw spelling — the server does its own casemapping, and a channel's casing is presentation.

## Tests

Three new cases under `focus folds the /LIST casing (#731)`, driven by a `#Sniffo` entry:

- join-then-foreground focuses `#sniffo` while `postJoin` still receives `#Sniffo`
- tapping an already-joined mixed-case row focuses `#sniffo`
- the row label still renders `#Sniffo`

The folded expectation is a **literal**, not `canonicalChannel(...)`: routing both sides through the same helper would keep the test green if the helper itself regressed.

**Mutation-proven, one leg at a time** — a shared "both legs broken" red would not have shown that each assert is individually load-bearing:

| mutation | result |
|---|---|
| both legs raw (pre-fix) | 2 failed / 43 passed |
| `onJoin` raw only | 1 failed — the join test, and only it |
| `onActivate` raw only | 1 failed — the already-joined test, and only it |
| fix in place | 45 passed |

## Gates

Run sequentially from the worktree root, output to file, rc read from file:

- `bun run check` (biome + tsc) — **rc=0**
- `bun run test` — **rc=0**, `Test Files 229 passed (229) / Tests 4125 passed (4125)`
- `bun run build` — **rc=0**, `precache 21 entries`

cic-only change; no server, e2e, or docker surface touched.

## Notes

- Branched off `origin/main` at `4f92701d`. `main`'s `integration` job is currently red on `tests/nick-follow-query.spec.ts` out of the five-PR merge — inherited, unrelated to this diff, which touches no server or e2e code.
- Expect a rebase: PR #797 (#738) also touches `DirectoryPane.tsx`, in different hunks (the pane-level search box; this one is inside the `DirectoryRow` sub-component).

### DIFF
--- a/cicchetto/src/DirectoryPane.tsx
+++ b/cicchetto/src/DirectoryPane.tsx
@@ -13,7 +13,7 @@ import {
   setSort,
   triggerRefresh,
 } from "./lib/channelDirectory";
-import { channelKey } from "./lib/channelKey";
+import { canonicalChannel, channelKey } from "./lib/channelKey";
 import { friendlyApiError } from "./lib/friendlyApiError";
 import { closeToPreviousWindow, setSelectedChannel } from "./lib/selection";
 import { windowStateByChannel } from "./lib/windowState";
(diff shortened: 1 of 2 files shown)


vjt:

PR #801, opened 2026-08-04
docs: retract the harmful rsync cure for a fresh worktree's empty e2e submodule

vjt:
## Summary

Docs-only. `docs/TESTING.md` trap 5 carried advice that is both **stale** and **actively harmful**, and two entries under "When the test stack itself is broken" pointed at it.

**The anti-cure.** Trap 5 said: rsync `cicchetto/e2e/infra` from the main checkout. That copies main's `.git` *pointer file*, whose relative gitdir `../../../.git/modules/cicchetto/e2e/infra` resolves to `<worktree>/.git/modules/…`. A linked worktree's `.git` is a **file**, so that path is literally `Not a directory` — measured:

```
$ ls -d $W/.git/modules/cicchetto/e2e/infra
ls: .../.git/modules/cicchetto/e2e/infra: Not a directory
```

Every superproject git command then dies `fatal: not a git repository` (exit 128, plain `git status` included). A worktree's real pointer targets its OWN module dir — different path, so main's copy can never be right:

```
main:     gitdir: ../../../.git/modules/cicchetto/e2e/infra
worktree: gitdir: ../../../../../.git/worktrees/<name>/modules/cicchetto/e2e/infra
```

And `submodule update --init` only works on an EMPTY infra dir, so the rsync forecloses the real cure. Init FIRST, never rsync-then-init.

**The stale premise.** Trap 5 claimed the auto-init re-clones over SSH and hangs. #592 (PR #616) shipped `-c protocol.file.allow=always` on 2026-08-01; the init clones from the superproject's local module store over `file://` — no network, no SSH. Re-measured on a genuinely fresh worktree at `4f92701d`: offline, rc 0, seconds.

Trap 5 now documents what the scripts actually do (`testnet.sh:41` guard, `:52` init; `bats.sh:37` for `vendor/bats-core`), states the symptom is an **infra bootstrap failure, not a test red**, gives the one-line manual equivalent plus the `git submodule status` leading-space/`-` tell, and marks the rsync as the anti-cure with its recovery path. Also records that `cicchetto/` is a plain subdirectory, not a submodule.

## Verification

Every claim read out of the code on this branch's merge-base (`4f92701d`), not from memory:

- `scripts/testnet.sh:41` — guard is `[ ! -d "$E2E_DIR/infra/bahamut" ]`, a content-presence check; init + flag at `:52`–`:53`.
- `scripts/bats.sh:37`–`38` — same pattern for `vendor/bats-core`.
- PR #616 `MERGED` 2026-08-01T18:35:09Z; the flag is present on `main`.
- Ran the shipped init in a fresh worktree with an empty `cicchetto/e2e/infra`: cloned offline from the local module store, rc 0, `git submodule status` flipped `-` → leading space.
- `.gitmodules` read directly (three submodules; `cicchetto/` absent).

No gates apply — docs-only, no product code, nothing under `scripts/` or `cicchetto/` touched.

## Test plan

- [x] `git status --porcelain` clean before and after; diff is `docs/TESTING.md` only (+50/-14)
- [x] Offline submodule init reproduced end-to-end on a fresh worktree
- [x] rsync breakage mechanism proven by path resolution, not asserted

### DIFF
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -148,7 +148,7 @@ The authoritative source is the comment block at the top of each
 `scripts/integration.sh` orchestrates an all-in-one docker compose
 stack (`cicchetto/e2e/compose.yaml`):
 
-* **azzurra-testnet** (git submodule at `cicchetto/e2e/infra/`): hub + leaf-v4 + leaf-v6 + services. Bahamut IRCd + Anope-shape services so CAP/SASL/NickServ behave like real Azzurra.
+* **azzurra-testnet** (git submodule at `cicchetto/e2e/infra/`): hub + leaf-v4 + leaf-v6 + services. Bahamut IRCd + Anope-shape services so CAP/SASL/NickServ behave like real Azzurra. Note `cicchetto/` itself is **not** a submodule — it is a plain subdirectory of this repo; `.gitmodules` lists only `cicchetto/e2e/infra`, `vendor/bats-core` and `frontends/shottino/vendor/libdatachannel`.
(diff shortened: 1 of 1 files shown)


vjt:

Issue #799, opened 2026-08-04
cic: the INVITE-row [Join] CTA focuses an unfolded channel param


vjt:
Found while auditing the other `setSelectedChannel` call sites for #731 (PR #798). Verified against `origin/main` at `4f92701d`.

**File:** `cicchetto/src/ScrollbackPane.tsx:1264` (`handleJoinChannel`)
**Severity:** LOW-MEDIUM — narrow exposure today, but a latent regression trap

## Problem

`handleJoinChannel` posts the join and then focuses the channel **raw**:

```ts
void postJoin(t, props.networkSlug, channel, null).then(() => {
  setSelectedChannel({ networkSlug: props.networkSlug, channelName: channel, kind: "channel" });
});
```

Its argument comes from the INVITE row's `[Join]` CTA at `ScrollbackPane.tsx:564`, which reads `params[1]` off the stored INVITE event row.

This is the only `kind: "channel"` selection site that forwards a server message param into selection without `canonicalChannel`. All three siblings fold: `channelJoin.switchTo` (`lib/channelJoin.ts:35`), `compose.ts` `/join` (`:680`), and `DirectoryPane` (`:117`/`:137`, just fixed in #731).

## Why it is narrow today

Ingress already folds it: `EventRouter`'s `:invite` clause (`lib/grappa/session/event_router.ex:296`) normalises param 1. So a freshly received INVITE is folded before it is persisted.

## Why it is still a defect

1. **That fold is the #537 fix.** The clause's own comment says `:invite` was previously lumped into the param-0 group, which "folded the wrong arg and left the channel raw-cased". Per the #525 posture, `refold_identifiers_ascii` does not rewrite stored values — so pre-#537 INVITE rows in prod scrollback still carry a mixed-case `params[1]`. Clicking `[Join]` on such a row joins fine (the wire is casing-tolerant) but focuses key `slug #Sbiffo` while `window_states` / `channelsBySlug` hold `#sbiffo`: no active sidebar row, empty phantom pane. Exactly the #731 failure shape.
2. **The fold lives only at ingress.** Any future path that reaches this handler with an unfolded name breaks silently, with no test to catch it. The siblings all fold at the selection boundary regardless of what ingress does — this one does not, so the codebase has two patterns for the same problem (CLAUDE.md "total consistency or nothing").

## Suggested fix

[...]

Issue #525, opened 2026-07-28
Fold is rfc1459 but Azzurra is CASEMAPPING=ascii — nicks and channels the ircd keeps distinct merge into one window


vjt:
## Summary

The whole stack folds nicks **and** channels with the **rfc1459** table (`A-Z` plus the four "national" chars `[ ] \ ~` → `{ } | ^`). Azzurra does **not** use that mapping: it advertises `CASEMAPPING=ascii` in 005 **and** implements plain ASCII folding in the ircd. So grappa **over-folds**: two identifiers the ircd keeps distinct collapse onto one grappa key — the exact window-merge failure #364 set out to prevent, in the opposite direction.

Raised by Sonic on IRC, then measured rather than assumed.

## Evidence

**1. What the ircd advertises.** Every 005 from Azzurra carries `CASEMAPPING=ascii` (68/68 occurrences in the bot's raw log; zero `rfc1459`). Source side, `src/s_misc.c` builds the token literally as `CASEMAPPING=ascii`.

**2. What the ircd actually does.** bahamut's fold table (`src/match.c`, `tolowertab[]`, used through `#define tolower(c)` in `include/common.h`) maps `A-Z` → `a-z` and leaves `0x5b 0x5c 0x5d 0x5e` (`[ \ ] ^`) and `0x7e` (`~`) **untouched**. It is ASCII folding in the table as well as in the advertisement — not the historical DALnet rfc1459 table.

**3. Live check against the production network.** Joining both spellings from one connection yields **two separate channels**, each with its own JOIN echo and its own 353/366:

```
> JOIN #vjtclaude[t]
< :vjt-claude!… JOIN :#vjtclaude[t]
< :ruby.azzurra.chat 353 vjt-claude = #vjtclaude[t] :@vjt-claude
< :ruby.azzurra.chat 366 vjt-claude #vjtclaude[t] :End of /NAMES list.
> JOIN #vjtclaude{t}
< :vjt-claude!… JOIN :#vjtclaude{t}
< :ruby.azzurra.chat 353 vjt-claude = #vjtclaude{t} :@vjt-claude
< :ruby.azzurra.chat 366 vjt-claude #vjtclaude{t} :End of /NAMES list.
```

Under `canonical_channel/1` both fold to `#vjtclaude{t}` → **one** window key for two distinct upstream channels.

## Impact

Server-side (`lib/grappa/irc/identifier.ex`, shared `fold_rfc1459/1`):

[...]

PR #803, opened 2026-08-04
cic(#799): fold the INVITE-row channel before focusing it

vjt:
Issue #799. Branched from `origin/main` at `4f92701d`.

## The defect

`ScrollbackPane.handleJoinChannel` forwarded the INVITE row's `params[1]`
straight into `setSelectedChannel`. Selection and `window_states` are keyed
FOLDED, so a mixed-case param foregrounds a key the sidebar cannot match — no
active row, empty phantom pane. Same failure shape as #731 in `DirectoryPane`.

Narrow today because ingress folds it (`EventRouter`'s `:invite` clause). But
that fold IS the #537 fix, and per the #525 posture `refold_identifiers_ascii`
does not rewrite stored values, so a pre-#537 INVITE row still sits in prod
scrollback with its raw casing. More structurally: the three siblings
(`channelJoin.switchTo`, `compose.ts` `/join`, `DirectoryPane`) fold at the
SELECTION boundary regardless of ingress; this one relied on ingress — two
patterns for one problem, and a silent trap for any future path in.

## The fix

`canonicalChannel(channel)` for the `setSelectedChannel` argument **only**. The
`postJoin` argument and the rendered row label keep the invite's spelling —
the key/display/wire split.

## The lying mock

`vi.mock("../lib/channelKey")` in `ScrollbackPane.test.tsx` stubbed
`canonicalChannel` as the **identity**, so no test in this file could ever
observe a missing fold. It now folds `A-Z` like prod, spelled out inline rather
than re-exporting the real `asciiFold` — a mock that delegates to the module it
stands in for cannot catch that module regressing.

`channelKey` stays deliberately non-folding, and the comment now says why: the
fixtures here key raw (`freenode NickServ`), and making it fold orphans them
(tried it — it reddened the unrelated `DOES count peer notice rows` unread-marker
test). Out of scope; recorded rather than papered over.

## Mutation proof

Both legs, plus the mock itself. Each mutation applied alone against the
committed tree; the named test is the only failure in all three.

| Mutation | Result |
|---|---|
| Revert the fold (`channelName: channel`) | RED — 1 failed / 166 passed |
| Fold the wire too (`postJoin(..., canonicalChannel(channel), ...)`) | RED — 1 failed / 166 passed |
| Restore the identity `canonicalChannel` mock | RED — 1 failed / 166 passed |
| Unmutated | GREEN — 167 passed |

The third row is the one that matters for the mock change: per the #736 lesson,
a mock fix that turns nothing red is not a fix.

## Gates

cic-only, run from the worktree root. No lane taken, nothing touched on
docker / e2e / integration.

- `bun run check:fix` then `bun run check` — **exit 0** (biome clean, so `tsc
  --noEmit` actually ran rather than being short-circuited)
- `bun run test` — **exit 0**, 229 files / **4124 passed**
- `git status --porcelain` empty before and after

jsdom is blind to layout, but this defect is store-level (selection key vs
sidebar key), so unit tests are the right instrument. No e2e.

## Not done here

`ArchiveModal.tsx` passing `target` unfolded (the issue's "also seen, NOT
filed" section) is untouched, as instructed.

### DIFF
--- a/cicchetto/src/ScrollbackPane.tsx
+++ b/cicchetto/src/ScrollbackPane.tsx
@@ -14,7 +14,7 @@ import LusersCard from "./LusersCard";
 import { isContentKind, ownNickForNetwork, postJoin, type ScrollbackMessage } from "./lib/api";
 import { token } from "./lib/auth";
 import { confirmJoinChannel } from "./lib/channelJoin";
-import { channelKey, decodeChannelKey } from "./lib/channelKey";
+import { canonicalChannel, channelKey, decodeChannelKey } from "./lib/channelKey";
 import { type TopicJoinLine, topicByChannel, topicJoinLine } from "./lib/channelTopic";
 import { isDocumentVisible } from "./lib/documentVisibility";
 import { highlightPatterns } from "./lib/highlightList";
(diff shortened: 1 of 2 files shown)


vjt:

Issue #802, opened 2026-08-04
Server-side WHOIS cache: the rail should ask grappa, not the ircd


vjt:
Follow-on from #800 (#606's rail auto-WHOIS reddened main by charging the
operator's next PRIVMSG upstream latency). #800 asks *when* to send the WHOIS.
This issue argues that is the wrong question.

## Problem statement

**There is no server-side store for a WHOIS answer.** Every rail-card
population therefore requires a fresh upstream WHOIS, because there is no other
source. Two days of #606/#613 fixes were all about *when* to spend that command
— TTL, no-TTL, retry window, rename migration, ordering. The missing piece is
**where to put the answer**, not the right moment to fetch it.

Generalised root cause: **a client-side pane decided to spend a server-side
connection budget it cannot see.** cic has no visibility into bahamut's
fake-lag `since`, into what the session sent a second ago, or into how close
the connection is to the recvQ parse gate. #606's own commit noted the rule
"binds future prefetch surfaces too" — this is that. Any future pane that wants
peer metadata will hit the identical wall unless the answer lives on the server.

## Evidence — verified on `origin/main` @ `4f92701d`

`whois_pending` is an **in-flight accumulator only**:

* `lib/grappa/session/server.ex:783` — typedef, "drained by 318 RPL_ENDOFWHOIS
  into a `{:whois_bundle, target, accum}` effect"
* `lib/grappa/session/server.ex:1215` — initialised `%{}`, "NOT persisted across crashes"
* `lib/grappa/session/server.ex:2060` — primed on `{:send_whois, …}`, i.e. only when
  somebody *asks*
* `lib/grappa/session/event_router.ex` — 311/312/313/317/319 fold into the entry
* `lib/grappa/session/server.ex:5112` — `apply_effects([{:whois_bundle, target, accum} | rest], state)`
  broadcasts to the user topic and **drops the entry.** Nothing stores it.

So the premise holds for WHOIS. **Two corrections to how it should be framed:**

### 1. cic *does* cache — at the wrong layer

`cicchetto/src/lib/railWhois.ts` is a per-nick cache with no TTL, migrated
old→new on rename, and it works. But it is a Solid signal inside
`identityScopedStore` — **no `localStorage`, no `sessionStorage`, no IndexedDB.**
It is in-memory, per-tab, per-page-load, and cleared on identity change.

That is the actual reason the fetch keeps recurring: every reload, every new
tab, every device starts cold and re-pays upstream for an answer grappa already
[...]

Issue #800, opened 2026-08-04
#606 rail auto-WHOIS charges the user's next message upstream latency — red main on nick-follow-query


vjt:
`origin/main` @ `4f92701d` is RED on `tests/nick-follow-query.spec.ts` (#373).
Bisected to **#613 / #606** (query-window rail context). Filing the evidence so
it survives the session; **no fix is proposed here — the choice is vjt's.**

```
✘ 379 [chromium] › tests/nick-follow-query.spec.ts:49:1 ›
    query window follows a peer NICK change — relabels, keeps history, routes with no 401 (6.7s)
Error: IrcPeer: timeout waiting for privmsg from vjt-grappa containing "#373 followup <sha>" (5000ms)
1 failed | 590 passed (25.5m)
```

Reproduced in three separate CI runs post-merge (main `4f92701d`, PR #795 head,
PR #796 head) and green in the last run that gated pre-merge main (PR #794).
**Not a flake.**

## Why the merge hid it

Five PRs landed in one move onto `123720d5` — #613 → #789 → #790 → #791 → #792.
Each was individually green, but the other four were each gated against
`123720d5` **without** #613, so the combination was never gated. That is not
where the bug is, though: **#613 alone is the regression.** The other four
merely added ~1s and pushed a spec that was already passing by 0.7s over the
5s cliff.

## Evidence 1 — CI durations, one variable each

| commit | content | spec duration | result |
|---|---|---|---|
| `123720d5` | pre-merge baseline | 2.7s | ✓ |
| `93796aa8` | #794 (baseline + #734/#735, no #613) | 3.7s | ✓ |
| `2ac821cc` | **#613 alone on baseline** | **5.7s** | ✓ (by 0.7s) |
| `4f92701d` | main, all five | 6.7s | ✘ |

## Evidence 2 — proof by displacement (local, same host, chromium, `--repeat-each 3`)

| build | durations | total |
|---|---|---|
| baseline `123720d5` | 2.0 / 3.4 / 4.4 | 23.8s |
| main `4f92701d` | 7.4 / 10.0 / 9.3 | 48.3s |
| main **minus one line** — the `requestRailWhois(...)` call at `cicchetto/src/RailContext.tsx:71` | 2.0 / 5.5 / 4.5 | 22.9s |

Deleting that single call restores the baseline. Nothing else was touched, and
the spec was never modified.

## Evidence 3 — mechanism, off grappa's own outbound wire

[...]

Issue #782, opened 2026-08-03
cic: explicit refresh control for the query rail WHOIS card


vjt:
Split out of #613 / #606 so that PR can land the bug fix without carrying a new UI affordance.

## Context

#613 removes `RAIL_WHOIS_TTL_MS` and the staleness-refetch path entirely. After it lands, the rail fetches WHOIS **once, on first select of a query window**, and caches it for the life of the window. It never refetches on its own.

That is deliberate. The refetch-on-staleness path was one of the ways the rail put closely-spaced commands on the connection, which blows bahamut's fake-lag ceiling (`src/s_bsd.c:1657` reads the socket only while `cptr->since - now < 10`) and leaves the user's next PRIVMSG sitting in the kernel buffer — measured at ~8s in the #613 e2e. Deleting the TTL removes that burst by construction rather than rate-limiting it.

## The gap this issue closes

With the TTL gone there is **no way for a user to refresh the card at all**. A deliberately-stale rail is acceptable — away/channels/idle do not justify an unrequested re-fetch — but stale *with no recourse* is not the end state we want.

Credit: proposed by Sonic in #grappa (2026-08-03), who argued a single fetch plus an explicit button is more honest than silent re-fetching. Agreed.

## Scope

- An explicit user-triggered refresh control on the query rail WHOIS card.
- Needs: placement, loading/disabled state while in flight, and behaviour when the peer is offline or the WHOIS errors.
- Must reuse the existing per-nick in-flight dedup (`PENDING_TTL`) so a mashed button cannot itself produce the burst this whole change exists to remove.

## Explicitly NOT in scope

- Any time-based refetch. No TTL comes back.
- Any automatic refresh on window focus, reconnect, or nick change. A rename migrates the cached entry old→new per the #373 identity-migration rule (done in #613); it does not refetch.

## Ordering

Should land in the same release as #613, so users never ship-and-sit with an unrefreshable card.

vjt:
## The button is not only a UX affordance — it is user-visible to the peer

Verified empirically (vjt, #grappa 2026-08-03), correcting an earlier claim in that channel that WHOIS produces no notification.

A user carrying umode `+y` who is whoised receives, directly:

```
-raccooncity.azzurra.chat- vjt-claude!~claude@... is doing a WHOIS on you
```

Source, `src/s_user.c:2200` inside `m_whois`:

```c
if (IsUmodey(acptr) && acptr != sptr)
[...]

PR #805, opened 2026-08-04
cic(#800): drop the rail's speculative WHOIS — the pane stops spending a budget it cannot see

### LINKED ISSUE #606: cicchetto: query windows get a rail context — heading + whois card, fetched on select (the deferred half of #474)

vjt:
Fixes the red `main`. Approved by vjt: *"ok leva la chiamata speculativa"*.

> **Correction (2nd commit).** An earlier version of this PR — and of the
> DESIGN_NOTES entry it adds — stated the bahamut fake-lag mechanism as
> established fact. It is not: it comes from reading bahamut source, not from
> measuring a running ircd, and the `5000ms` in the failing assertion is the
> **test helper's own wait budget**, not a bahamut latency. Caught by Sonic and
> vjt. The measured evidence below is unchanged and still carries the argument.

## What this does

Deletes the `createEffect` in `RailContext.tsx` that called
`requestRailWhois` when a query window was selected. That is the whole
behavioural change.

**This is NOT a revert of #613.** The rail card, the `renameRailWhois` #373
migration, the `whoisCard` / `railWhois` store split and the
`source: "user" | "rail"` wire field all stay. The cache still fills from the
user's own `/whois` — #606's origin routing already delivers a `source: "user"`
bundle for the nick the rail is showing into that store.

## Why — what is MEASURED

One extra WHOIS on the session's upstream connection delays the operator's
**next message** by seconds.

| build | `nick-follow-query` |
|---|---|
| `123720d5` pre-merge baseline | 2.7s ✓ |
| `2ac821cc` #613 alone | 5.7s ✓ (by 0.7s) |
| `4f92701d` merged main | 6.7s ✘ |

Locally, `--repeat-each 3`, chromium: main **7.4 / 10.0 / 9.3**; main minus this
one call **2.0 / 5.5 / 4.5**; baseline **2.0 / 3.4 / 4.4**. The red is
intermittent — ~7 of 9 post-merge runs, 0 of 4 pre-merge — which is exactly why
the duration delta and the displacement result are the evidence, not any single
run.

## Why — what is INFERRED, and not confirmed

*Where* those seconds are spent inside the ircd is **open**. The leading
hypothesis, from reading bahamut source, is fake-lag: each command costs
`since += 2 + len/120` and the recvQ drain is gated on `since - now < 10`
(`parse.c:236`, `s_bsd.c:1657`). It fits — grappa emitted the followup 357ms
after the previous send, so the delay is not on our side — but nobody has
instrumented the ircd to show it. Closing it means measuring the ircd, not
re-reading the source.

**The rule does not depend on it.** It follows from the displacement result
alone: **cic cannot see the connection's upstream cost — not what the session
sent a moment ago, not what it cost, not how close it is to whatever ceiling
exists — so it does not get to decide to spend a command.** Every timing fix
#606 reached for (TTL, retry window, migrate-before-select ordering) was
answering the wrong question. If the fake-lag reading turns out wrong, the rule
does not move.

## Tests

Two tests in `RailContext.test.tsx` pin the rule at the surface that broke it:

* `issues NO WHOIS when a query window is selected`
* `issues NO WHOIS when a rename swaps the focused query's nick`

They replace the two that asserted the opposite. **Both were watched RED against
the shipped code before the call came out** — 2 failed / 8 passed. A test that
passes with the call restored would be worthless, so this is the evidence that
it does not.

**On the rename leg, precisely.** That test's `called 2 times` counts **effect
firings against a mocked store**, not commands. On the wire it stayed at one:
the instrumented run showed `WHOIS Guest…` and no `WHOIS NickTmp…`, because
`subscribe.ts` migrates the rail cache old→new *before* `followQueryNick` swaps
the selection, so the second call short-circuits on a hit. The second entry
point is still worth pinning — a cache miss there (empty bundle, peer offline,
retry window elapsed) makes that call a real command — but it was a near miss,
not a second WHOIS. And the rename is not why this spec is the one that fell:
it fell because it is the only spec that opens a query window and then sends a
timed message.

## Deliberately left alone

`requestRailWhois` keeps its tests and its de-dupe rules and now has **no
production caller.** That is a choice, not an oversight: it is the seam #782's
explicit fetch control attaches to. Stated in the module moduledoc rather than
left to be rediscovered.

## Also recorded (no code)

What shipped was never what was asked for. The request was to populate when the
rail is **opened** on mobile / the window is **viewed** on desktop.
Fetch-on-select is neither — on mobile the rail is off screen, so #606 spent
upstream budget filling a card the user could not see.

The button itself is **#782** and is not built here. The server-side cache that
would make the question mostly go away is **#802** and is not built here either.

## Gates

* `scripts/bun.sh run check` (biome + tsc) — **rc=0**
* `scripts/bun.sh run test` — **rc=0**, 229 files / 4122 tests passed
* Integration deliberately left to this PR's CI rather than the local stack.

⚠️ Given the intermittency, a green integration here is **consistent with** the
fix but does not on its own prove it. The duration deltas and the displacement
run are the evidence.

Related: #800 (evidence), #613 / #606 (the fetch), #782 (the button), #802 (the
cache).

### DIFF
--- a/cicchetto/src/RailContext.tsx
+++ b/cicchetto/src/RailContext.tsx
@@ -17,9 +17,9 @@ import WhoisCard from "./WhoisCard";
 //              single-slot `whoisCard` store the user-issued /whois owns) and
 //              carries no × affordance (persistent, like the server card).
 //              It RENDERS what is known and asks for nothing — #800 removed
-//              the fetch-on-select, because cic cannot see the upstream
-//              fake-lag budget a WHOIS spends and was charging it to the
-//              operator's next PRIVMSG. The cache fills from the user's own
+//              the fetch-on-select, because that WHOIS measurably delayed the
+//              operator's next message and cic cannot see the upstream cost it
+//              was spending. The cache fills from the user's own
 //              /whois (#606 routes a `source: user` bundle for the shown nick
 //              here); #782 adds the explicit fetch control.
 // It renders NOTHING for kinds with no context content (channel already has
--- a/cicchetto/src/RailContext.tsx
+++ b/cicchetto/src/RailContext.tsx
@@ -1,15 +1,6 @@
-import {
-  type Component,
-  createEffect,
-  createSignal,
-  Match,
-  on,
-  onCleanup,
-  Show,
-  Switch,
-} from "solid-js";
+import { type Component, createSignal, Match, onCleanup, Show, Switch } from "solid-js";
 import { networkBySlug } from "./lib/networks";
-import { railWhoisFor, requestRailWhois } from "./lib/railWhois";
+import { railWhoisFor } from "./lib/railWhois";
 import { selectedChannel } from "./lib/selection";
 import ServerInfoCard from "./ServerInfoCard";
 import WhoisCard from "./WhoisCard";
@@ -20,12 +11,17 @@ import WhoisCard from "./WhoisCard";
 // grafts the matching context content:
 //   * server → ServerInfoCard (connection facts already in the store)
 //   * query  → the query context (#606, the deferred half of #474): a
-//              heading + a WHOIS card for the conversation partner,
-//              auto-fetched on select. The card REUSES the same `WhoisCard`
-//              presentation as the scrollback overlay but is fed by the
-//              per-nick `railWhois` cache (NOT the single-slot `whoisCard`
-//              store the user-issued /whois owns) and carries no × affordance
-//              (persistent, like the server card).
+//              heading + a WHOIS card for the conversation partner. The card
+//              REUSES the same `WhoisCard` presentation as the scrollback
+//              overlay but is fed by the per-nick `railWhois` cache (NOT the
+//              single-slot `whoisCard` store the user-issued /whois owns) and
+//              carries no × affordance (persistent, like the server card).
+//              It RENDERS what is known and asks for nothing — #800 removed
+//              the fetch-on-select, because cic cannot see the upstream
+//              fake-lag budget a WHOIS spends and was charging it to the
+//              operator's next PRIVMSG. The cache fills from the user's own
+//              /whois (#606 routes a `source: user` bundle for the shown nick
+//              here); #782 adds the explicit fetch control.
 // It renders NOTHING for kinds with no context content (channel already has
 // the MembersPane above; home/admin/list/mentions have none). Built as a
 // container, not a hardcoded server card, so the rail is the per-kind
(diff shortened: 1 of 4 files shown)


vjt:
Landed inside #809, which is now merged to `main`.

Both commits of this PR ride in there:

- `db9520da` — `cic(#800): the rail renders what is known, it does not go asking`
- `f05a1318` — `docs(#800): demote the fake-lag mechanism from fact to hypothesis`

The SHAs above are the post-rebase ones, so the head of this PR (`c90b9fa2`) is **not** an ancestor of `main` — that is expected, not a miss. Verified by content:

```
$ git log origin/main --oneline --grep='#800'
28702ccb cic(#782): ask for the rail WHOIS when the card is on screen, not on select
f05a1318 docs(#800): demote the fake-lag mechanism from fact to hypothesis
db9520da cic(#800): the rail renders what is known, it does not go asking
```

Why they were folded in: on base `800-drop-rail-autofetch` the workflows never fired at all (`ci.yml` / `integration.yml` declare `pull_request: branches: [main]`), so the stack was running zero CI. Re-basing #809 onto `main` made its diff the union of #805 + #782 — which is the gate that was actually wanted, since #805 alone is red on a spec it never touched and #782 is what makes it green again. All five check-runs were success at `035ce5ac`.

Closing as superseded.

PR #807, opened 2026-08-04
irc(#800): S7 — instrument the fake-lag bank inside grappa, per send

vjt:
S7 of the #800 plan. **Instrumentation, not a fix.** Nothing about what
grappa sends, or when, changes.

## The point

#805's rule stands on the displacement result and needs no ircd-internal
mechanism. What is still open is WHERE the seconds go, and the leading
answer — bahamut's fake lag, `since += 2 + len/120`, recvQ drain gated on
`since - now < 10` (`parse.c:236`, `s_bsd.c:1657`) — has only ever been
**read in C**. Twice this cycle an inference from that source was stated
as fact and had to be retracted. A third reading cannot close it.

This closes the half that lives on our side of the socket: the bank
becomes a number grappa observes about itself, per frame, on a live
connection.

## What it emits

One `[debug]` line per outbound frame:

```
command=NICK sent_bytes=18 commands_10s=1 penalty_10s_s=2.0 bank_model_s=2.0 headroom_s=8.0 [debug] upstream send
command=USER sent_bytes=35 commands_10s=2 penalty_10s_s=4.0 bank_model_s=4.0 headroom_s=6.0 [debug] upstream send
command=PING sent_bytes=13 commands_10s=3 penalty_10s_s=6.0 bank_model_s=6.0 headroom_s=4.0 [debug] upstream send
```

**Per send, not aggregated** — the bank at the moment of *that* frame, so
a slow iteration and a fast one can be compared line by line on the same
run without any post-processing.

**Two numbers, not one.** #800 claims the cost bites only on a connection
already near the ceiling; that is a claim about `headroom_s`, and a bank
printed alone cannot settle it either way.

## Measured vs modelled — the boundary is in the field names

| field | kind |
|---|---|
| `sent_bytes`, `commands_10s` | **MEASURED** — what this process put on the wire, and how much of it is still inside the ten-second horizon |
| `penalty_10s_s`, `bank_model_s`, `headroom_s` | **MODELLED** — bahamut's published arithmetic evaluated locally |

The modelled fields are **not** a reading off an ircd and must never be
quoted as one. Because every line also carries the raw verb and byte
count, a reader who thinks the model charges the wrong thing (bahamut may
well exempt `PONG`; opers are exempt outright) can re-derive the bank
from the raw lines instead of trusting the arithmetic.

## Two limits, written down rather than discovered later

* **Above the gate the model diverges.** It advances the bank when *we
  write*; bahamut advances it when *it parses*, and past the gate it
  stops parsing — so the real bank saturates near the ceiling while ours
  keeps climbing. A modelled bank far above 10 reads as "we wrote much
  more than the server could drain", not as a server-side figure. Below
  the gate the two track. Pinned by a test.
* **The parameters of a frame are never logged.** A PRIVMSG's params are
  the message body and an AUTHENTICATE's are a base64 credential. The
  verb and the byte count are what the model needs; the content is not.
  Pinned by two tests.

## Where the accounting hooks in

At the transport choke point, so the registration burst, the liveness
self-PING and the FSM's handshake replies count too — not only
session-issued commands. A bank fed just the commands convenient to count
would under-report, and **an under-report is exactly what would retire
the hypothesis for the wrong reason**. The brief scoped this to
`handle_call({:send, _})`; widening it is deliberate and is the one place
this PR exceeds the brief.

The fake IRC server's own tally is the oracle: a test asserts the model
charged for exactly the frames the server received.

## Cost, and where you read it

`Logger.debug/2` never evaluates its arguments when the level is off, and
production runs at `:info` — so prod is silent and pays only the fold.
The dev/compose stack the integration testnet runs has no level override
and defaults to `:debug`, which is where the measurement gets taken.
`LOG_LEVEL=debug` turns it on anywhere else. A live node can also read
the current bank straight off `:sys.get_state/1` with no log at all.

## Pre-registered verdicts (agreed before any number existed)

* **Bank near the ceiling when the spec fails** ⇒ fake lag confirmed as
  the mechanism; the fix is scheduling in `IRC.Client` (user-originated
  ahead of speculative — the `source: "rail" | "user"` tag from #606
  already exists, so no new wire field is needed).
* **Bank far from the ceiling** ⇒ the accumulation story falls, and #800
  needs correcting: it contains an assertion nobody has measured. This is
  the more valuable outcome of the two.
* **Same bank across a fast and a slow iteration of the same run** ⇒ fake
  lag is exonerated outright and the cause is elsewhere. The shape
  already argues for this — 28ms / 3024ms / 4989ms with the WHOIS present
  in all three, and arrivals clustering at 3031 / 6066 (≈2×3033) / 5002 /
  5010 ms, look like thresholds, not a bucket draining.

### The two outcomes do NOT weigh the same

Written down here **before** any number exists, because it is exactly the
kind of thing that gets forgotten once a table is on the screen. It
follows from the measured/modelled split above: the bank is bahamut's
arithmetic **re-evaluated by us, locally**. It is not the ircd's bank.

* **Bank far from the ceiling ⇒ STRONG falsification.** Even bahamut's
  own formula, borrowed and evaluated against the real frames, fails to
  predict a throttle. The hypothesis dies on its own terms. This verdict
  is solid and needs nothing further.
* **Bank near the ceiling ⇒ WEAK confirmation.** It shows that *our
  model* predicts a throttle, not that the ircd applied one. That is
  still a reading of the source — merely executed instead of read.

So the near-the-ceiling case closes nothing by itself. Settling it means
asking the ircd rather than our copy of its maths: **S8** — send a PING
immediately after the PRIVMSG and time the PONG on grappa's own socket.
Not built here; recorded so whoever reads these numbers in a month finds
how much they weigh sitting next to them.

## Tests

`test/grappa/irc/fake_lag_test.exs` (13, pure) — the cost floor, integer
division at the 120-byte step, stacking, real-time drain, the 10s window
boundary at 9999/10000/10001 ms, bounded growth, headroom, and the
above-the-gate divergence. Includes the arithmetic half of w1's
objection: 500 commands paced at 3s apart leave the bank at the floor, so
`since` cannot accumulate across a 25-minute run.

`test/grappa/irc/client_outbound_cost_test.exs` (3, `async: false` —
lowers the global log level, per the `SessionLogHygieneTest` precedent).

`client_test.exs` — the oracle test plus the verb-extraction seam.

**Mutation-proven, 2/2 killed:**
* dropping `:headroom_s` from the `config/config.exs` allowlist → the log
  test fails. That allowlist is a real silent-failure mode: an unlisted
  key is discarded at FORMAT time, so the code looks right and the
  operator gets a blank field.
* reverting the choke point to account only session sends → the oracle
  test fails 1 vs 3, and the handshake test fails.

## Gates

`scripts/check.sh` — **rc=0**. 8 doctests, 50 properties, **5081 tests, 0
failures**; 262 bats ok, none failing. Format, Credo strict, Dialyzer
(`:underspecs`), Sobelow, deps.audit all clean.

Integration/e2e **not run** — the STACK lane is not mine. The measurement
itself is the next step and needs that lane.

Related: #800 (evidence), #805 (the rule), #782 (the button), #802 (the
cache).

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -526,7 +526,20 @@ config :logger, :console,
     # allowlist beyond `Meta.known_keys/0` (the sync test is a subset
     # check: known_keys ⊆ metadata). `:new_nick` pre-exists above.
     :old_nick,
-    :rows_migrated
+    :rows_migrated,
+    # #800 S7 — `Grappa.IRC.Client`'s per-frame outbound-cost line, at
+    # DEBUG (silent in production, which runs at :info). `:sent_bytes` and
+    # `:commands_10s` are MEASURED — what this process put on the wire and
+    # how much of it is still inside the ten-second horizon. `:penalty_10s_s`,
+    # `:bank_model_s` and `:headroom_s` are MODELLED by `Grappa.IRC.FakeLag`
+    # from bahamut's published `2 + len/120` arithmetic and must never be
+    # quoted as a reading off an ircd. `:command` (the verb, pre-existing)
+    # rides the same line; the parameters deliberately do not.
+    :sent_bytes,
+    :commands_10s,
+    :penalty_10s_s,
+    :bank_model_s,
+    :headroom_s
   ]
 
 import_config "#{config_env()}.exs"
--- a/lib/grappa/irc/client.ex
+++ b/lib/grappa/irc/client.ex
@@ -100,7 +100,7 @@ defmodule Grappa.IRC.Client do
   """
   use GenServer
 
-  alias Grappa.IRC.{AuthFSM, Identifier, Message, Parser}
+  alias Grappa.IRC.{AuthFSM, FakeLag, Identifier, Message, Parser}
 
   require Logger
 
@@ -164,7 +164,11 @@ defmodule Grappa.IRC.Client do
           liveness_idle_ms: pos_integer(),
           liveness_timeout_ms: pos_integer(),
           idle_timer: reference() | nil,
-          ping_timer: reference() | nil
+          ping_timer: reference() | nil,
+          # #800 S7 — what this connection's own outbound traffic would
+          # cost it under bahamut's flood throttle. Diagnostic only:
+          # nothing reads it to make a decision.
+          fake_lag: FakeLag.t()
         }
 
   # `:socket` is intentionally NOT enforced — pre-connect (between `init/1`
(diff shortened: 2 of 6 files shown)


vjt:

Issue #750, opened 2026-08-03
passkey_origin is read with a runtime Application.get_env inside a controller


vjt:
Found by the periodic codebase review, slice `test/**` + `cicchetto/e2e` since `v0.10.0`. This is a production finding surfaced by the test that leans on it.

    defp passkey_origin,
      do: Application.get_env(:grappa, :passkey_origin, GrappaWeb.Endpoint.url())

`lib/grappa_web/controllers/auth_controller.ex:459-460`, called per request.

CLAUDE.md: "`Application.{put,get}_env/2`: boot-time only, runtime banned … neither read nor written from any GenServer callback, controller, context function, plug body, or release task." The prescribed shape for a non-process DI seam is the `boot/0` + `:persistent_term` resolver with a `Mix.env() == :test`-gated `put_test_*/1` — precedent: `Grappa.Admission.Config`, `Grappa.HttpHosts`, `Grappa.Uploads`.

The test cements the banned pattern from the other side: `test/grappa_web/controllers/auth_controller_test.exs:74` injects with `Application.put_env(:grappa, :passkey_origin, "https://login.example:8443")`.

**Why it matters here specifically.** The RP origin IS the WebAuthn anti-phishing binding. Resolving it from mutable global state on every login means nothing structurally prevents a second writer from repointing it.

**Fix.** Move `passkey_origin` to a `boot/0` reading env once into `:persistent_term`, resolve from there at request time (keeping the `Endpoint.url()` default), and switch the test to the gated `put_test_*/1` helper.

vjt:
Closing as **not a defect** — deliberately NOT as "shipped in v0.11.0".

This one arrived dead: the described defect did not exist on main when it was picked up. Naming a release it never needed would misrepresent both the issue and the release, so it is being closed on its own terms.

Issue #740, opened 2026-08-03
cic: passkeys shipped with no DESIGN_NOTES entry, and a cloned login-button rule


vjt:
Found by the codebase-review campaign on `main` (`v0.10.0..388f883b`), `cicchetto/src` slice. Verified against the files at `388f883b`.

Two documentation/duplication defects with the same shape: a second copy that will drift from the first.

## 1. The passkey surface shipped with no DESIGN_NOTES entry, and the log says it does not exist

`docs/DESIGN_NOTES.md` — the only relevant entry, *"2026-08-02 — #442 — account TOTP second factor"*, ends with "Passkeys, recovery-code regeneration, and email recovery remain out of scope". The commits under review ship passkey registration, second-factor mode, passwordless mode, recovery codes and a mode-change ceremony. No later entry mentions passkeys.

**Failure scenario:** a future session reads the decision log, concludes passwordless login does not exist, and either re-litigates or re-implements the design — the drift CLAUDE.md's "Directions over code" rule exists to prevent.

**Fix:** add an entry covering the mode ladder, the recovery-token handshake before passwordless activation, the `rp_id`/origin configuration and the "a passkey belongs to this instance's domain" constraint; and amend the #442 out-of-scope sentence.

## 2. `.login-alt-auth` is a verbatim clone of `.login-advanced-toggle`

`cicchetto/src/themes/default.css:875`-`885` — all nine declarations are byte-identical to `.login-advanced-toggle` (836-846). The same pattern repeats at line 2449, where `.channel-clickable` is a third copy of the `.nick-clickable` (2409) / `.scrollback-invite-join` (2469) inline-transparent-button reset.

**Failure scenario:** a theme tweak to the quiet text-button (padding, `font-size`, `min-height: var(--tap-min)`) lands on one class and is silently skipped on the other. The Advanced toggle and the Passkey/Recovery buttons sit side by side in the same flex row, so `align-items: center` makes the height mismatch immediately visible.

**Fix:** extract one `.login-quiet-button` (and one inline-button reset for the scrollback trio) applied alongside the existing per-instance classes, leaving only the genuine deltas local.

vjt:
Shipped in **v0.11.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.11.0

The tag carries `.deb`, `.rpm` and `pkg.tar.zst` assets, so self-hosters have something to pull.

[...]

Issue #735, opened 2026-08-03
cic: the new security-pane controls have no tap-target sizing or accessible names


vjt:
Found by the codebase-review campaign on `main` (`v0.10.0..388f883b`), `cicchetto/src` slice. Verified against the files at `388f883b`.

**Files:** `cicchetto/src/PasskeySettings.tsx:130,159,164,167,195,199`; `cicchetto/src/TotpSettings.tsx:114,139,158,176`
**Severity:** MEDIUM

## Problem

None of the buttons added to the two new security panes carry a class, and `default.css` has no generic `button` rule — only per-class rules plus `.compose-box button` and `html.is-ios .admin-tab-panel button`. So every control on this pane renders at the user agent default size, in the UA font and colour, on a themed dark drawer.

Text inputs are fine: the global `input:where(...)` rule from #497/#508 gives them `--tap-min: 44px`. The buttons inherit nothing.

Note the project sets the html root font-size to 14px, so `rem`-based sizing under-shoots the 44px iOS minimum — absolute px is the correct unit here.

Secondary: each per-passkey "remove" button has the same accessible name with no reference to WHICH passkey it deletes.

## Failure scenario

On an iPhone, "remove" next to each passkey is a roughly 20px-tall target adjacent to other tappables, with no confirmation step behind it. A mis-tap deletes the wrong credential. A screen-reader user hears a list of identical "remove" buttons.

## Suggested fix

Give the new controls the drawer's existing button classes (or add a `.settings-section button` rule) with `min-height: var(--tap-min)`, and give each remove button an `aria-label` including the passkey name.

vjt:
Shipped in **v0.11.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.11.0

The tag carries `.deb`, `.rpm` and `pkg.tar.zst` assets, so self-hosters have something to pull.

Note this is a **release, not a deploy**: the hosted service is unchanged and still runs the previous build until the code-review queue closes.

Issue #734, opened 2026-08-03
cic: the TOTP enrolment QR is rendered into a class that has no CSS rule


vjt:
Found by the codebase-review campaign on `main` (`v0.10.0..388f883b`), `cicchetto/src` slice. Verified against the files at `388f883b`.

**File:** `cicchetto/src/TotpSettings.tsx:124`
**Severity:** MEDIUM

## Problem

The enrolment QR is rendered into `class="share-qr"`. No `.share-qr` rule exists in `cicchetto/src/themes/default.css` — the only matches are `.share-modal-qr` (line 9601) and `.share-modal-qr svg` (line 9612).

The name was evidently taken from `ShareSessionModal.tsx:208`, where `share-qr` is the `data-testid` and `share-modal-qr` is the actual class. That real class supplies the `12rem` box, the white background and the `svg { width:100%; height:100% }` sizing — and `lib/qr.ts`'s own doc comment states the caller must provide "a *sized* container — no fixed px width/height".

## Failure scenario

The TOTP enrolment QR renders at the user agent's default replaced-element size, on a dark themed card, with no white frame and no quiet zone. Depending on the browser it comes out under- or over-sized, and on dark themes it is at risk of being unscannable — the exact hazard `qr.ts` documents. A user who cannot scan the code cannot enrol the second factor.

## Suggested fix

Use `class="share-modal-qr"`, or add a dedicated sized rule if the security pane wants its own box.

vjt:
Shipped in **v0.11.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.11.0

The tag carries `.deb`, `.rpm` and `pkg.tar.zst` assets, so self-hosters have something to pull.

Note this is a **release, not a deploy**: the hosted service is unchanged and still runs the previous build until the code-review queue closes.

PR #813, opened 2026-08-04
#740: record the passkey design, and give the cloned button shapes one declaration each

vjt:
Code-review finding #740. Two defects of the same shape — a second copy that will drift from the first — in two commits.

Every cited line was re-verified against `main` @ `942f642b` before being touched. Both defects are alive; neither had drifted.

## 1. The decision log denied that passkeys exist

`docs/DESIGN_NOTES.md`'s only passkey mention was #442's closing *"Passkeys, recovery-code regeneration, and email recovery remain out of scope"* — while #442 and #736 shipped registration, both modes, recovery codes and the mode-change ceremony.

That sentence is **amended in place**, not left standing with a correction further down. A decision log is a coherent account, not a wall of caveats: the next session reads the sentence, not the thread. It now points forward to the new entry.

The new entry states **behaviour, not the config-reading mechanism** — #750 is moving `passkey_origin` onto a `:persistent_term` seam in parallel, and an entry describing today's plumbing would be wrong by the time anyone reads it. Where it does touch mechanism it was verified on `main` at the time of writing.

Covered: the three-mode ladder and where `:second_factor` sits relative to TOTP; the uniform `invalid_credentials` oracle a passwordless account answers a password with, and why that uniformity is the point; why arming passwordless costs two steps and cannot be folded into one — including why the recovery token is *encrypted* rather than signed, and why `set_mode/4` re-checks at the context boundary an invariant the route shape already implies; the origin binding, with the two consequences worth knowing before rediscovering them the hard way — **moving a live deployment's public origin invalidates every passkey already registered**, and a `:passwordless` account whose origin moved has no password to fall back on.

🔴 **One thing the finding did not ask for, added because it is the sharpest edge on this surface:** the recovery codes are **one account-level set shared with TOTP** (`Accounts.RecoveryCodes` — *"shared by TOTP and passkeys"*, and `replace/2` deletes all rows before inserting). So arming passwordless **replaces the codes a TOTP user was already holding**, and their old printout silently stops working. It is the one step here that invalidates something the user was told to keep safe, and it was written nowhere.

## 2. Two cloned button shapes

`.login-alt-auth` was nine declarations byte-identical to `.login-advanced-toggle`; the two sit in the same flex row under `align-items: center`, so a padding or `min-height` tweak landing on one and not the other is a visible height mismatch. `.nick-clickable` / `.channel-clickable` / `.scrollback-invite-join` were three copies of the same "render a `<button>` as inline text" reset.

Each becomes one rule — `.login-quiet-button` and `.scrollback-inline-button` — worn **alongside** the per-instance class. The class names are load-bearing beyond paint: `.login-advanced-toggle` and `.login-alt-auth` are selector contracts the e2e clicks as strict single matches (issue281), and `keepKeyboard.ts` keys its `SELECTABLE_TEXT_EXCLUDE` off the scrollback trio. Composing a second class disturbs neither; the counts are pinned by a test.

Only genuine deltas stay local: colour, padding, `font-weight`, and the #250 `user-select: text` the two text tokens carry and the [Join] CTA deliberately does not. Where a class had no delta left, its rule is **deleted** rather than left empty, with a comment saying where a future delta goes.

**Not done as a `.scrollback :where(button)` floor in #735's shape**, which was the tempting precedent from this morning: the scrollback also holds the scroll-to-bottom badge and the far-behind jump/dismiss pair, and `display: inline` would break them. The shape belongs to these three controls, not to their container.

### The guard is the clone's absence, not the shared rule's presence

Asserting "a shared rule exists" passes with the clones still sitting beside it. So the tests assert the per-instance rules **no longer declare** the shared properties, plus a signature scan for a fourth copy — and the markup assertions check the shared class is actually **worn**, which is the #734 failure mode inverted (a rule with no element behind it).

- **13 CSS assertions, 12 failed before the fix** (the 13th pins the deltas that must stay local, and correctly passed throughout).
- **The scan found a real false positive of its own**: an earlier draft keyed on `padding` + `min-height` and flagged `.archive-modal-group-summary`, which is a bold accent flex row merely honouring the same 44px HIG floor. It now keys on the shape (transparent background + mono face + that padding).
- **Markup assertions verified by mutation** — stripping the shared classes from all four call sites turns 4 tests red across `Login`, `ScrollbackPane` and `MircText`; restored after.

jsdom applies no stylesheet, so all of this proves what the cascade is *asked* to do, never what a browser paints.

## Gates

From the worktree root: `check:fix` (fixed 1 file, committed) · `check` **rc=0** · `tsc --noEmit` standalone **rc=0** · `test` **rc=0, 233 files / 4203 tests**.

Server-side code untouched; the only non-cic file is `docs/DESIGN_NOTES.md`. No stack or compile lane used.

### DIFF
--- a/cicchetto/src/Login.tsx
+++ b/cicchetto/src/Login.tsx
@@ -449,7 +449,7 @@ const Login: Component = () => {
                   <div class="login-alt-auth-row">
                     <button
                       type="button"
-                      class="login-advanced-toggle"
+                      class="login-quiet-button login-advanced-toggle"
                       aria-expanded={advanced() ? "true" : "false"}
                       aria-controls="login-advanced"
                       onClick={() => setAdvanced((v) => !v)}
@@ -458,14 +458,14 @@ const Login: Component = () => {
                     </button>
                     <button
                       type="button"
-                      class="login-alt-auth"
+                      class="login-quiet-button login-alt-auth"
                       onClick={() => void onPasskeyLogin()}
                     >
                       Passkey
                     </button>
                     <button
                       type="button"
-                      class="login-alt-auth"
+                      class="login-quiet-button login-alt-auth"
                       onClick={() => setRecoveryMode((value) => !value)}
                     >
                       Recovery code
@@ -591,7 +591,7 @@ const Login: Component = () => {
                   </button>
                   <button
                     type="button"
-                    class="login-advanced-toggle"
+                    class="login-quiet-button login-advanced-toggle"
                     onClick={() => {
                       setTotpChallenge(null);
                       setTotpCode("");
--- a/cicchetto/src/MircText.tsx
+++ b/cicchetto/src/MircText.tsx
@@ -99,7 +99,11 @@ const renderChannel = (
 ): JSX.Element => {
   if (!onChannelClick) return channel;
   return (
-    <button type="button" class="channel-clickable" onClick={() => onChannelClick(channel)}>
+    <button
+      type="button"
+      class="scrollback-inline-button channel-clickable"
+      onClick={() => onChannelClick(channel)}
+    >
       {channel}
     </button>
   );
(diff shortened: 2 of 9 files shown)


vjt:

Issue #788, opened 2026-08-03
cic: a scrollback verb can put a request on the wire under a REVOKED bearer


vjt:
Split out of #769. **#769 itself turned out to be a spec race** (account A's own
join-ok backfill landing inside the spec's judging window) and is fixed in
PR #786. This is the separate, real defect that investigation surfaced.

## The defect

Every async verb in `cicchetto/src/lib/scrollback.ts` captures the bearer once
at entry and then awaits, with no re-check:

```ts
const t = token();
if (!t) return;
...
const page = await listMessagesAfter(t, slug, name, cursor, PAGE_LIMIT);
...
const gap = await probeGap(t, "reconnect-refresh", slug, name, anchor);   // t may be revoked by now
```

The nine `onIdentityChange` resets clear STATE; they cancel nothing in flight.
So a rotation or a detach during any of those awaits leaves the continuation
holding a bearer the server has already revoked — and it sends it.

`refreshInFlight` and `jumpInFlight` are not in the reset list either, so the
per-key locks survive the rotation too.

## Reproduced in a browser, not argued from source

With the join-ok backfill delayed 400ms (`page.route`, scratch edit) and the
issue281 spec's drain disabled, on the real cic bundle:

```
identity timeline: [A=5a2b3240 @24.7, null @925.1, null @962.7, B=e937314f @1149.3]
probe            : reconnect-refresh #bofh anchor=200 staleBearer=TRUE @935.2
on the wire      : GET /networks/bahamut-test/channels/%23bofh/messages/count?after=200
                   Authorization tail 5a2b3240  ← account A, revoked at 925.1
```

The detach landed at 925.1; the request went out at 935.2 — **10ms after its
own bearer was revoked**, through entirely production code paths.

## Which sites, and which cannot

* `reconnect-refresh` — awaits a page before probing. **Affected** (measured above).
* `resolve-jump-target` — awaits two. **Affected.**
* `initial-load` — NO await between `token()` and the request (`getReadCursor`
  is a synchronous signal read). **Structurally cannot** be stale.

Non-probe verbs (`loadMore`, `loadNewer`, `sendMessage`, `jumpToUnread`,
`refreshScrollback`'s own page fetch) share the shape and want the same audit.

## Why it matters beyond a 401

This is the #281 harm class. A request for network A's channel issued around an
identity switch 404s if the new identity is not attached to that network, and
the host `http-404` fail2ban jail bans the client IP at the firewall — a routine
[...]

Issue #281, opened 2026-07-17
cicchetto: account switch replays previous session's JOINs/history-fetches → 404 storm self-bans the client IP


vjt:
## Summary

Switching accounts on an already-open cicchetto session (detach from account A, log back in as account B on the same client) leaves the client holding account A's cached network/channel state. On the new session the client replays that stale state — JOINs over the websocket and history-fetches over REST — against networks/channels that account B's session is **not** attached to. The server correctly answers `404` to each phantom `GET /networks/{net}/channels/{chan}/messages`, but the client fires the whole cached list in a burst (~30 requests in ~2s), which trips the host's `http-404` fail2ban jail and **bans the client's public IP at the firewall** — cutting off the web session itself and every other service on the host.

Self-inflicted lockout triggered purely by a normal account switch.

## Repro

1. Log in to cicchetto as account **A** (multiple networks, e.g. `libera` + `oftc`, several channels each).
2. Detach.
3. Log back in on the **same client** as a different account **B** (whose session is attached to different / no networks).
4. Client hydrates from cached state and fires history-fetch for **A's** networks/channels.

## Observed

Burst of ~30 requests in ~2 seconds, all `404`, from the client IP (redacted):

```
GET /networks/libera/channels/%23ai/messages?after=0&limit=200      -> 404
GET /networks/libera/channels/%23elixir/messages?...                 -> 404
GET /networks/libera/channels/%23grappa/messages?...                 -> 404
GET /networks/oftc/channels/%23openwrt/messages?...                  -> 404
GET /networks/libera/featured                                        -> 404
GET /networks/oftc/featured                                          -> 404
...  (libera + oftc: ~11 channels + $server + query windows)
```

Threshold on the `http-404` jail is exceeded -> the client's IPv4 is banned at the firewall -> total loss of access (including the cicchetto session that caused it).

## Root cause

Account switch does not reset client-side session state. The cached network/channel list from the previous account survives the re-login and is replayed against a session that does not own it.

## Proposed fixes

- **Client (primary):** on logout/detach or on login as a different account, purge cached networks/channels/session state. Do not replay the previous account's JOINs or history-fetches. Hydrate strictly from the new session's `networks` list.
- **Server (secondary/defensive):** consider returning `200` with an empty page instead of `404` for a well-formed history request against a network/channel the session isn't attached to — a 404 storm from a legitimate-but-confused client should not be indistinguishable from a scanner. Alternatively rate-limit/collapse client-side.
[...]

Issue #159, opened 2026-06-29
[P0] cic silently stops rendering new messages on one channel; no activation/visibility catch-up re-fetch exists, so the gap only heals on a full reload


vjt:
## Severity: P0

Silent message loss in a live client. The messages ARE on the server (other
clients + REST scrollback have them), but this cic session never renders
them and gives the operator **no signal** that anything is wrong — the pane
just looks quiet. The only recovery the user found was force-closing the PWA
and reopening (a full reload). Users miss messages and do not know it. A
chat client that silently drops a channel's live feed is worse than one that
visibly errors, because the operator trusts an empty pane to mean "nothing
new."

## Symptom / repro

Verbatim report:

> On one channel, my cic client STOPPED receiving/rendering new messages.
> The messages WERE present on the server (other clients and scrollback had
> them). I had to FORCE-CLOSE the app and reopen to see the new messages. We
> need some sort of freshness check on window activation, and I THINK we had
> that already — but it failed now.

Inferred deterministic repro (the gap class, see findings):

1. Open cic, focus channel `#chan`; it renders fine and receives live.
2. A per-channel live-delivery gap opens **without a socket-level
   disconnect** — e.g. the per-channel Phoenix subscription stops receiving
   while `socketHealth().state` stays `"open"` (server-side fan-out severed,
   a single wedged/`errored` channel that never completes a rejoin `"ok"`, a
   join `timeout` that only logged, or a `GrappaChannel` process restart that
   re-subscribed PubSub without a socket close).
3. Peers keep posting to `#chan`; the server persists + broadcasts every
   row. This cic session is no longer a live subscriber for that topic, so
   the rows never arrive.
4. Operator switches away from `#chan` and back, and/or backgrounds and
   re-foregrounds the tab. **Nothing re-fetches.** The pane stays frozen at
   the last row it received live.
5. Full reload (force-close + reopen) recovers everything.

### Does a freshness-on-activation path exist today? Partly — and not for this case.

There are **two** scrollback-fetch verbs, with very different triggers:

**(a) `loadInitialScrollback` — LOAD-ONCE, on pane selection.**
`cicchetto/src/lib/scrollback.ts:146-244`. Gated by the `loadedChannels` Set
(`scrollback.ts:150-151`): once a channel has been loaded this session, the
verb returns immediately. It is called from the selection-change effect
`cicchetto/src/lib/selection.ts:471-522` (the actual call at
[...]

PR #817, opened 2026-08-04
cic(#788): a continuation that outlived its identity does nothing further

vjt:
Refs #788.

## The defect

`identityScopedStore`'s resets clear scrollback STATE on an identity
transition; they cancel nothing in flight. Every async verb in
`cicchetto/src/lib/scrollback.ts` captured `token()` at entry and then awaited
without re-checking, so a rotation or detach landing inside one of those awaits
left the continuation running under an identity that no longer existed — and it
sent. Measured in a browser on the real bundle: with the join-ok backfill held
400ms, a `/messages/count` went out 10ms after its own bearer was revoked.

The harm is not the 401. It is the #281 class through another door: a request
for the old identity's channel 404s when the new identity is not attached to
that network, and the host's `http-404` fail2ban jail bans the client IP at the
firewall. A routine account switch self-bans the operator.

## The census

15 awaits, 9 async verbs. 11 now carry the check; 4 are exempt and say so in
place — three whose await is their last act (`probeGap`, the two tail calls
into `anchorAtTail`), and `resolveJumpTarget`, which returns a pure value its
caller checks before spending. Every `setReadCursor` is either pre-await or
post-guard.

The issue named `probeGap`-after-`listMessagesAfter`. The class is wider in two
directions the issue did not name:

* **the store, not just the wire.** The same continuation poisons without
  touching the network: the page it already fetched merges into a map the
  rotation just purged, so the new identity renders the old one's scrollback.
  That happens *between* the resolved fetch and the next request — which is why
  the guard sits after each await and not in front of each REST call. The
  wrapper shape the issue floated as the "reuse the verb" answer cannot see it.
* **three exits, not one.** `api.ts` throws on any non-ok response, so a bearer
  revoked mid-flight arrives as a 401 throw rather than a clean resolve — the
  likelier path. `loadInitialScrollback`'s catch releases the load-once gate:
  past a rotation it releases the NEW identity's gate, whose next re-select
  reads `wasLoaded` false, takes the fresh-open arm that deliberately does not
  fire `refreshScrollback` (#159), and skips one live-delivery catch-up. The
  four `finally` blocks release per-key locks that, after the reset cleared the
  Set, belong to whoever replaced us.

## The mechanism

One shared predicate, `identityMoved(t) => token() !== t`, at every exit from
an await that touches state. Not hoisted into `identityScopedStore`: that
factory owns the identity TRANSITION, this owns a verb's own captured identity,
which the factory never sees. `displayPrefs` and `customTheme` already inline
the same comparison to drop a stale RESPONSE — this is the same predicate one
step earlier, to refuse a stale REQUEST.

## The two locks

`refreshInFlight` and `jumpInFlight` join the reset list. Declared beside their
verbs rather than beside the other five Sets, they were missed — and left held
they are a real defect: an in-flight refresh for A keeps A's key until its
`finally`, and B's refresh for that key short-circuits meanwhile, so B's window
silently never backfills.

An earlier draft justified the unconditional `finally` delete with
"`appendToScrollback` dedupes by id, worst case one redundant GET". True of the
merge paths, false of `jumpToUnread`, which REPLACES the key's rows. The
conditional release retired the trade instead of documenting it wrongly.

## Tests

Seven cases in `scrollbackIdentityRotation.test.ts`, each verified RED by
removing exactly the guard it covers and watching only it fall: the stale
probe, the stale page landing in the purged store, the surviving lock, the
cold-open cursor baseline, the released-and-stamped in-flight refresh, the
released load-once gate on a 401, plus a control case that fails loudly if the
"full page" the stale-probe case releases ever stops being full — which would
otherwise turn that assertion into one about a branch nobody entered.

They mock `../auth` with a REAL Solid signal. The flat `token: () => value`
stub the sibling suites use is not reactive, so `createEffect(on(token))` never
fires and the rotation these cases turn on would not exist.

## Honest notes

* `sendMessage`'s post-await cursor POST was already unreachable past a
  rotation — but only via the store purge emptying the pane so the #50
  anti-poison gate declines. A guarantee owed to an unrelated invariant is one
  refactor from evaporating, so it is now stated rather than inherited.
* Out of scope, noticed while reading: `readCursor.ts` is NOT identity-scoped,
  so its cursor map survives a rotation. May well be deliberate (join-reply
  hydration re-seeds it), but nothing says so.

## Gates

cic-only. `bun run check` 0, `tsc --noEmit` 0, `bun run test` 0 — 233 files,
4195 tests.

### DIFF
--- a/cicchetto/src/lib/__tests__/scrollbackIdentityRotation.test.ts
+++ b/cicchetto/src/lib/__tests__/scrollbackIdentityRotation.test.ts
@@ -2,7 +2,7 @@
 // further: no request on the wire, no write into the store.
 //
 // Every async verb in `scrollback.ts` captures `token()` at entry and then
-// awaits. The nine `onIdentityChange` resets clear STATE; they cancel nothing
+// awaits. The `onIdentityChange` resets clear STATE; they cancel nothing
 // in flight. So a rotation or a detach landing inside one of those awaits
 // leaves the continuation holding a bearer the server has already revoked —
 // and it sends it. Reproduced in a browser on the real bundle (see the issue):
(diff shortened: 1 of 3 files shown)


vjt:

Issue #804, opened 2026-08-04
cic: ArchiveModal is the one DM-open site that does not resolve the peer casing


vjt:
Split out of #799's "also seen, NOT filed" section, after auditing the call
sites. **Severity: LOW** — inert today, and inert for a reason that lives in
another module. Filed for consistency, not for a repro.

## The count

`openQueryWindowState` has 9 non-test call sites. Eight resolve the peer
casing first:

| Site | Argument |
|---|---|
| `MembersPane.tsx:152` | `canonical` |
| `WhoModal.tsx:213` | `canonical` |
| `NamesModal.tsx:78` | `canonical` |
| `ScrollbackPane.tsx:1242` | `canonical` |
| `UserContextMenu.tsx:109` | `canonical` |
| `lib/pushTarget.ts:60` | `canonical` |
| `lib/compose.ts:774` (`/msg`) | `canonical` |
| `lib/compose.ts:867` (`/query`) | `canonical` |
| **`ArchiveModal.tsx:121`** (`handleSelectEntry`) | **`target`, raw** |

The same handler then passes that raw `target` to `setSelectedChannel`.

That is the whole argument: **total consistency or nothing**. One site out of
nine reaching a shared verb by a different route is a pattern fork, and per
CLAUDE.md the next person to touch this copies whichever is closer.

## Correction to what #799 implied

I originally framed "the raw value reaches `openQueryWindowState` too" as the
aggravating detail. Checking it, that leg is **effectively inert**, and the
issue is weaker than I first stated:

- `openQueryWindowState` (`lib/queryWindows.ts:80`) already folds internally —
  its `alreadyOpen` guard is a `nickEquals` compare, so a raw target cannot
  fork a duplicate row.
- `canonicalQueryNick` returns the nick **unchanged** when no window is open.
  And `list_archive` (`lib/grappa/scrollback.ex:907`) rejects any target
  present in the active keyset via a **folded** compare — so an archive entry
  by construction has no open window, which is exactly the case where the
  eight siblings' resolver is a no-op too.

In the normal path, therefore, ArchiveModal and its siblings behave
**identically**. Recording this so nobody re-derives it.

## What actually diverges

`setSelectedChannel` receives whatever casing SQLite happened to pick. The
archive query groups on the fold but selects the raw column:

```elixir
|> group_by([m], Identifier.nick_fold(fragment("COALESCE(?, ?)", m.dm_with, m.channel)))
[...]

PR #820, opened 2026-08-04
cic(#804): resolve the archive peer's casing like the other eight sites

vjt:
Fixes the one `openQueryWindowState` call site out of nine that did not
resolve the peer casing first. Issue #804.

## What

`ArchiveModal.handleSelectEntry` passed the archive row's raw `target`
to both `openQueryWindowState` and `setSelectedChannel`. It now runs it
through `canonicalQueryNick` — like `ScrollbackPane.handleNickClick`,
`MembersPane`, `WhoModal`, `NamesModal`, `UserContextMenu`,
`lib/pushTarget.ts`, and both `lib/compose.ts` sites — and uses the
resolved value for BOTH legs.

## Why, given it is inert today

The inertness is real but borrowed: `list_archive/3` excludes any target
present in the active keyset via a **folded** compare, so an archive
entry by construction has no open window — the one state where the eight
siblings' resolver is a no-op too. A *server* module's invariant is
holding a *client* selection path upright, nothing states the coupling
and no test pins it.

`dm_with` is stored RAW (#121/#372) and the archive query groups on the
fold while selecting one arbitrary row's spelling, so the moment the
exclusion changes — or a window opens between the archive fetch and the
row tap — selection holds `VJT-Peer` while the sidebar row is
`vjt-peer`: the #731 / #799 phantom-pane shape.

The governing rule is CLAUDE.md's *total consistency or nothing*: one
site out of nine reaching a shared verb by another route is a pattern
fork, and the next person to touch this file copies whichever is
closest.

## Notes

- **Not a fold.** `canonicalQueryNick` resolves to the existing window's
  RAW `targetNick`; folding there would break `Sidebar.tsx`'s raw `===`
  match and the ChannelKey derivation.
- The `kind: "channel"` leg is untouched — `messages.channel` stores the
  folded channel, so there is no raw casing to carry.
- The test file's `queryWindows` mock previously exported only
  `openQueryWindowState`. It now also carries `canonicalQueryNick` as a
  faithful stand-in over the production `nickEquals`, **not** an
  identity stub: an identity stub returns whatever the component passes
  and would have been blind to exactly this defect (the #799 lesson).

## Test

New case in `ArchiveModal.test.tsx`: an archive entry spelled
`vjt-peer` while a window is open as `VJT-Peer` must land selection —
and the window-open push — on `VJT-Peer`. Fails on `main` (selection
receives `vjt-peer`), passes here.

## Gates

- `bun run test` — 232 files / 4189 tests, rc=0
- `bun run check` (biome + tsc) — rc=0; `tsc --noEmit` standalone rc=0

cic-only change; no server code touched.

### DIFF
--- a/cicchetto/src/ArchiveModal.tsx
+++ b/cicchetto/src/ArchiveModal.tsx
@@ -14,7 +14,7 @@ import { mentionCounts } from "./lib/mentions";
 import { networks } from "./lib/networks";
 import { normalizeNick } from "./lib/nickEquals";
 import { createOverlayLock } from "./lib/overlayScrollLock";
-import { openQueryWindowState } from "./lib/queryWindows";
+import { canonicalQueryNick, openQueryWindowState } from "./lib/queryWindows";
 import { eventsUnread, messagesUnread, setSelectedChannel } from "./lib/selection";
 import NickText from "./NickText";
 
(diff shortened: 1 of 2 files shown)


vjt:

Issue #779, opened 2026-08-03
cic: four non-blocking findings from #717 review round 3 (moduleRoot guard + bootFetch)


vjt:
Deliberately kept OUT of #717 / PR #776 so that PR stays scoped to recoverability. The one finding from the same round that WAS blocking (the `isOffline()` discriminator) landed there, because it was the feature not working.

None of these is a live defect today. All four are verified by reading, two by execution where noted.

---

## 1. The `moduleRoot` guard enforces a spelling, not the invariant

`cicchetto/src/__tests__/moduleRootGuard.test.ts:38` — `const BARE_ROOT = /\bcreateRoot\s*\(/;`

The invariant #717 needs is "every module-lifetime computation has an error context". The guard enforces "no bare `createRoot`". A module-scope `createEffect` / `createMemo` / `createResource` with **no root at all** has `Owner === null`, so `handleError` finds no handler and rethrows out of `runUpdates` — the identical cycle abort, the identical frozen last frame, and the regex reports the file clean. Solid only emits a dev-mode `console.warn`.

**Verified by execution**: a bare module-scope `createMemo` that throws in the same cycle as a render-tree memo → the `ErrorBoundary` fallback never renders and the stale node stays in the DOM. That is the #717 signature.

Zero instances today — `grep -rnE '^(const |let |var )?create(Effect|Memo|Resource|Computed)\(' src` returns nothing, and the only `createRoot(` sites in non-test `src` are the two factories. So this is hardening. It is worth doing because it is the *easier* thing for the next author to write than the evasions the test's header already anticipates.

Evasions weighed and judged not worth covering: aliased import (`createRoot as makeRoot`) — nobody aliases this; multi-line `createRoot\n(` — biome would not emit it. Namespace import (`solid.createRoot(`) and re-export are already caught by the `\b`.

## 2. The anti-vacuity case can be satisfied by a comment

`cicchetto/src/__tests__/moduleRootGuard.test.ts:79-82`

The second case runs `BARE_ROOT.test(readFileSync(f))` over the **whole file**, without the comment filtering the first case applies. `identityScopedStore.ts:10` contains the literal `//   const exports_ = createRoot(() => {` in its pattern-history comment — so if that module ever stopped calling `createRoot` (say it is refactored to delegate to `moduleRoot`), the vacuity check would still pass on prose alone.

[...]

PR #821, opened 2026-08-04
cic(#779): moduleRoot guard checks the invariant; bootFetch's budget covers the body

vjt:
The four findings from #717 review round 3. Issue #779. Every line was
re-verified against the files as they stand on `main` (the findings were
written at `388f883b`) — all four premises still hold, with **one
correction** noted below.

Two commits: the transport fix, then the guard.

## 4 — `bootFetch`'s per-attempt timeout also bounded the body read, and that abort was never retried

The one with a real consequence. `AbortSignal.timeout(8_000)` starts at
dispatch and stays live on the `Response`, but `fetch` resolves at the
HEADERS and every caller (`me`, `listNetworks`, `listChannels`) does
`await res.json()` after `bootFetch` returned. A body still streaming at
the deadline aborted **outside** the retry loop — an `AbortError` the loop
never saw, so an attempt the policy explicitly calls retryable was
silently not retried.

Fixed by draining a `res.clone()` inside the `try`. That pulls the body
read into the attempt (the abort becomes an ordinary retryable transport
failure) and leaves the returned `Response` fully buffered, so the
caller's own read can no longer be cut by our signal.

**Cloning, not reconstructing** — `new Response(await res.arrayBuffer(),
…)` would have worked too but loses identity, `res.url`, and trips the
null-body statuses. The original `Response` comes back, so the existing
`resolves.toBe(ok)` assertions still hold and `readError` is untouched.

Test: a first attempt whose body stream errors mid-way, a good second one.
Before: one `fetch` call and a `Response` whose `json()` rejects. After:
two calls and a readable body.

## 1 — the guard enforced a spelling, not the invariant

A module-scope `createEffect` / `createMemo` / `createResource` with no
root at all has `Owner === null` — the throw leaves `runUpdates`, the
queued render effects are discarded, the last frame freezes. The #717
signature exactly, and `/\bcreateRoot\s*\(/` reports the file clean.

The predicate is now two rules: an **owning root** (a `createRoot` whose
callback ignores `dispose`) and an **unowned computation** (a reactive
primitive at column 0 with no root on the line). Module scope is detected
as "starts at column 0" — a heuristic, pinned by biome owning the
formatting, and its failure direction is a false positive the author fixes
by moving the call into the root it belonged in.

Zero offenders in the tree today, so the classifier is also **unit-tested
against fixtures**: a guard that has never seen an offender proves nothing
about what it would catch. Those fixtures are the RED — the disposable-root
case and the module-scope-memo case both failed against the old predicate
before the rewrite.

## 2 — the anti-vacuity case was satisfiable by a comment

It ran the raw regex over the whole file, so `identityScopedStore.ts:10`'s
`//   const exports_ = createRoot(() => {` would have kept it green on
prose alone. It now runs through the same classifier as the walk, which
applies the prose filter. Pinned by a fixture case.

## 3 — no escape hatch for a legitimately disposable root

`createRoot(dispose => …)` for a detached or portal render is normal Solid
and correct in production code. The old guard flagged it, leaving the next
author choosing between `moduleRoot` (wrong — it never disposes, so they
leak a root) and an exclusion list, which CLAUDE.md forbids. Binding
`dispose` IS the invariant now: a root with a handle ends, a root without
one is forever and must come through the door.

**Correction to the issue.** #779.3 says the test-file exemption's own
justification is the right predicate — "every one of them takes the
`dispose` callback". That claim is **false**: `home.test.ts:42`,
`railWhois.test.ts:186`, `whoisCard.test.ts:61` and `subscribe.test.ts:1186`
each open a root that does not. So the dispose predicate could not replace
the path exemption without either editing four unrelated test files or
opening the exclusion list the finding was written to avoid.

The exemption therefore stays, with its justification corrected to the true
one: it is the ERROR CONTEXT that puts test roots out of scope — a test
module's root dies with the worker and a throw surfaces as a failing test,
so the runner IS the context this guard exists to supply. The four leaking
roots are a test-hygiene matter, not a #717 one; flagged here rather than
filed, since the freeze is on review findings.

## Gates

- `bun run test` — 232 files / 4194 tests, rc=0
- `bun run check` (biome + tsc) — rc=0; `tsc --noEmit` standalone rc=0

cic-only; no server code touched. `docs/DESIGN_NOTES.md` carries the
durable rationale. Refs #717, PR #776.

### DIFF
--- a/cicchetto/src/__tests__/moduleRootGuard.test.ts
+++ b/cicchetto/src/__tests__/moduleRootGuard.test.ts
@@ -2,7 +2,8 @@ import { readdirSync, readFileSync, statSync } from "node:fs";
 import { join, relative, resolve } from "node:path";
 import { describe, expect, it } from "vitest";
 
-// #717 — a bare `createRoot` must not come back.
+// #717 — a module-lifetime reactive computation must not run without an error
+// context.
 //
 // The first cut of #717 gave the error context to `identityScopedStore` alone
 // and left every other module root bare. `activeWindows.ts` — a module-level
@@ -15,7 +16,7 @@ import { describe, expect, it } from "vitest";
 //
 // CLAUDE.md: "Half-migrated creates two patterns — Claude copies whichever is
 // closer." So the migration is total and this test is what keeps it total. A
-// new root goes through `moduleRoot`; nothing else may call `createRoot`.
+// new root goes through `moduleRoot`; nothing else may own one.
 //
 // Sibling of biomePin.test.ts / versionSource.test.ts — vitest runs from the
 // cicchetto dir, so `src` is at cwd.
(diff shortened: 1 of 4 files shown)


vjt:

Issue #749, opened 2026-08-03
TOTP: three untested holes (throttle blocking, challenge-token expiry, enrollment label)


vjt:
Found by the periodic codebase review, slice `test/**` + `cicchetto/e2e` since `v0.10.0`.

**1. The TOTP door's throttle-blocking branch has no test.** The only `:totp_login` test seeds nine failures straight into ETS (`test/grappa_web/controllers/auth_controller_test.exs:291`) and then asserts the CLEAR path (`:307` `assert :ok = FailureWindow.check(:totp_login, key, 1)`). No test drives N wrong codes through `POST /auth/totp/verify` and asserts the next one is refused, so `{:error, :limited} -> {:error, :too_many_attempts}` (`auth_controller.ex:465-466`) is never reached. Same for `:passkey_recovery` (`passkey_controller.ex:191`, limit 10).

The template already exists one describe over: `auth_controller_test.exs:1168` ("the 11th attempt is 429 too_many_attempts even with the correct password") does exactly the right thing for the PASSWORD door. The TOTP door is the sibling that never got it.

Failure scenario: `check_totp_throttle/2` sits mid-`with` and is trivially droppable — unlimited brute force over a 1-in-10^6 space, suite green.

**2. The 2FA challenge token never expires in a test.** `@totp_challenge_max_age_seconds 300` (`auth_controller.ex:70`) and the `{:error, :expired} -> {:error, :two_factor_challenge_expired}` clause (`auth_controller.ex:147`) have zero coverage — `grep two_factor_challenge_expired test/` is empty. Worth documenting while there: the challenge is a `Phoenix.Token`, i.e. REUSABLE inside its 300s window; only the TOTP code is one-shot.

Failure scenario: `max_age:` is dropped from the `Phoenix.Token.verify/4` call and a post-password token lifted from a log stays valid forever.

**3. The enrollment-URI test does not assert the account.** `test/grappa/accounts/totp_test.exs:100` is titled "new enrollment URI contains issuer, account, and secret" but asserts only the `otpauth://totp/` prefix, `issuer=`, and `secret=` (a same-struct echo). The label — `URI.encode("#{issuer}:#{name}")`, `totp.ex:27` — is unasserted, as are `algorithm=SHA1` / `digits=6` / `period=30`.

Failure scenario: the label is built from `user.id` instead of `user.name`; every enrolled account shows as a UUID in the authenticator app.

vjt:
[...]

Issue #676, opened 2026-08-02
Auto-append `_` on nick collision at connect instead of failing the login


vjt:
## Problem

A visitor who logs in with a nick already taken on the upstream network gets a failed login, not a session. There is no automatic fallback: the user must go and pick a different nick by hand.

This came out of a real session on #grappa: a user connected with the same nick they were already using from their own client on Libera, hit the collision, and had to find the nick field in `settings → general` to get in. The UI did tell them the nick was in use — the friction was entirely "now what do I type, and where".

## Current behaviour (verified on `origin/main`)

- Registration collides → AuthFSM has no recovery path for a passwordless session, so it stops the Client with `{:nick_rejected, 433, _}`.
- `Grappa.Visitors.Login.classify_down/1` (`lib/grappa/visitors/login.ex:748`) maps that to `:nick_in_use`.
- `auth_controller.ex:512` → `fallback_controller.ex:653` renders it as HTTP `409 {"error": "nick_in_use"}`.
- cic shows the friendly copy and the login does not complete.

So the 433 is classified correctly and reported honestly — the gap is that we stop there instead of retrying.

## Proposal

On a 433 during registration, retry with `<nick>_` (and keep appending on repeated collisions, up to a small bound), complete the login under the fallback nick, and let the user rename afterwards from settings. Classic IRC-client behaviour, and it turns a dead end into a session that is already usable.

## Points to settle before implementing

1. **Retry bound and shape — DECIDED (vjt, #grappa 2026-08-02).** First retry is `<nick>_`. If `<nick>_` is *also* taken, do **not** keep stacking underscores — fall back to a random suffix, which is far less likely to collide again on a busy network. A bound on the total number of attempts is still required: unbounded retries against a hostile or duplicated nick space is a respawn-flood risk (the same trap the e2e fixtures document around shared-leaf 433 autokill). Open sub-detail: the shape and length of the random suffix, which interacts with `NICKLEN` below.
2. **NICKLEN.** Appending can exceed the upstream `NICKLEN`; the ircd may silently truncate, which can re-collide. Truncate on our side using the network's advertised `NICKLEN` before sending.
3. **Tell the user.** After a successful fallback the client should say what happened — "`<nick>` was taken, you are `<nick>_`" — otherwise the rename never happens and the underscore becomes permanent by accident.
[...]

Issue #40, opened 2026-06-03
Surface 'Nick already in use' on connect (currently shows 'handshake did not complete')


vjt:
When connecting with a nickname that already exists on the network, the client shows `handshake did not complete` instead of the actual IRC error `Nick is already in use`.

The 433 numeric (ERR_NICKNAMEINUSE) arrives during registration but the connect code path swallows it as a generic handshake failure, leaving the user without actionable info.

Surface the real error to the UI instead.

Repro:
- Open landing page, pick a nickname currently in use on Azzurra.
- Observe: `handshake did not complete`.
- Expected: `Nick is already in use` (or equivalent), so user knows to pick another nick.

Reported by vjt via #sniffo while nextime was trying grappa and hit it.

vjt:
**Triage 2026-06-08 — still open.**

The visitor AuthFSM returns `{:nick_rejected, 433, nick}` as a stop reason, but `lib/grappa/visitors/login.ex`'s `login_error` type doesn't include it and the client `friendlyApiError.ts` has no `nick_already_in_use` code — so it surfaces as the generic "handshake did not complete". Fix: thread 433 through `login_error` → API → a dedicated client message ("Nick already in use").

— automated triage (claude, on vjt's account)

PR #826, opened 2026-08-04
test(#749): put the TOTP door's three unasserted claims under test

vjt:
Issue #749. All three cited sites verified alive against current main
(`8ae917a6`) before writing anything — line numbers had drifted from
`388f883b` but every defect is real. One half of one premise had not:
see the bottom.

## What changed

**1 — the throttle's blocking arm.** `check_totp_throttle/2` sits
mid-`with` in `verify_totp/2` and its `{:error, :limited}` arm was
unreachable from the suite: the one `:totp_login` test seeded nine
failures into ETS and asserted the CLEAR path. New test drives ten wrong
codes through `POST /auth/totp/verify` and demands 429 on the eleventh
*even with the correct code* — the shape the mode-1 password describe
already uses. A second test proves the window is keyed per (ip, user):
one account's exhausted bucket must not refuse a fresh one.

Two details worth flagging:
- The eleven requests reuse ONE challenge token. The challenge is a
  `Phoenix.Token`, so it is **reusable for its whole 300s window** —
  only the TOTP code is one-shot. That is the intended UX (a typo must
  not cost the password round trip) and precisely why the window needs a
  counter. Documented on the describe.
- The wrong code is derived from the production generator, excluding
  `matching_step/3`'s ±1 window plus one step either side so a 30s
  boundary crossed mid-loop cannot turn it valid. `"000000"` is a real
  code once every 10^6 enrollments, and the loop rolls the dice ten
  times.

**2 — the challenge window.** `grep two_factor_challenge_expired test/`
was empty. Two tests: one token minted 301s ago is refused as expired,
one minted inside the window still mints the bearer. The second is not
decoration — age is the subject, so these mint directly and duplicate
the controller's private salt. Running the SAME helper with a current
timestamp and demanding a 200 means a drifted salt, payload shape, or
binding turns the control red, instead of the expiry test passing for
the wrong reason (a wrong salt reads as `invalid_two_factor`, never as
an accepted login).

**3 — the enrollment label.** The URI test asserted a prefix, the
issuer, and an echo of its own secret. It now parses the URI instead of
substring-matching it, asserts the whole label
(`Grappa (irc.example):alice`), and pins `algorithm=SHA1` / `digits=6` /
`period=30` — `code_at/2` is not written against RFC defaults, so a URI
that omits them enrolls an app whose codes this server refuses. The
secret must additionally decode as 20 raw Base32 bytes, so it is no
longer a same-struct echo.

## Mutation matrix

Every mutation applied to production code, run, and reverted. Re-verified
against the final version of the wrong-code helper.

| mutation | result |
| --- | --- |
| drop `check_totp_throttle/2` from the `with` | **red** — only "the 11th attempt is 429…" |
| `max_age:` → `:infinity` in `Phoenix.Token.verify/4` | **red** — only "a challenge older than its window…" |
| `new_enrollment(%User{id: name}, …)` (label from the id) | **red** — only "new enrollment URI labels the account…" |

## One premise did not survive measurement

The issue extends finding 1 with *"Same for `:passkey_recovery`
(`passkey_controller.ex:191`, limit 10)"*. That half is **already
covered**, twice over, so nothing was added for it:

| mutation | result |
| --- | --- |
| drop the `:passkey_recovery` account-bucket check (limit 10) | **red** — the existing "recovery throttling survives respelling the identifier" catches it |
| drop the check at the literal cited line 191 (which is `:passkey_login_options`, limit 30 — not `:passkey_recovery`) | **red** — the existing "login_options is throttled even while it keeps succeeding" catches it |

Both existing tests drive real attempts through the endpoint and assert
the 429, which is exactly the shape the issue asks for. Reporting rather
than inventing a substitute.

## Pre-existing flake found while measuring, NOT touched here

`POST /auth/login (visitor via nick) every nick candidate in use → 409
nick_in_use (#40)` fails intermittently — about 2 reds in ~14 runs of
this file, and it reproduces on the **unmodified `origin/main` version**
of the file (1 red in 5 runs, 49 tests), so it is not this branch's. It
is in the #676 433-fallback-ladder path. I did not capture the failure
text before it went quiet, so I am not guessing at the mechanism. Flagged
for a separate issue rather than widening this PR.

## Gate

`scripts/check.sh` rc=0 from the worktree root: credo `found no issues`,
dialyzer `Total errors: 0` / `done (passed successfully)`, doctor passed,
sobelow clean, ExUnit `8 doctests, 50 properties, 5105 tests, 0 failures`,
bats 271 ok / 0 not ok.

Test-only change; no production code is touched.

### DIFF
--- a/test/grappa/accounts/totp_test.exs
+++ b/test/grappa/accounts/totp_test.exs
@@ -115,13 +115,34 @@ defmodule Grappa.Accounts.TOTPTest do
     assert %DateTime{} = Repo.get!(Session, other.id).revoked_at
   end
 
-  test "new enrollment URI contains issuer, account, and secret" do
+  # The label is what an authenticator app puts under the user's thumb, so
+  # it has to name the ACCOUNT, not merely be present: built from `user.id`
+  # instead of `user.name` the URI still parses, still carries the issuer,
+  # still round-trips the secret, and every enrolled account shows up as a
+  # UUID. The parameters are pinned for the same reason — RFC 6238 leaves
+  # SHA1/6/30 as defaults, but `code_at/2` is not written against defaults,
+  # so a URI that omits or contradicts them enrolls an app that computes
+  # codes this server will refuse.
+  test "new enrollment URI labels the account and pins the RFC 6238 parameters" do
     user = user_fixture(name: "alice")
     enrollment = TOTP.new_enrollment(user, "Grappa (irc.example)")
 
-    assert String.starts_with?(enrollment.provisioning_uri, "otpauth://totp/")
-    assert enrollment.provisioning_uri =~ "issuer=Grappa+%28irc.example%29"
-    assert enrollment.provisioning_uri =~ "secret=#{enrollment.secret}"
+    assert %URI{scheme: "otpauth", host: "totp", path: path, query: query} =
+             URI.parse(enrollment.provisioning_uri)
+
+    assert URI.decode(path) == "/Grappa (irc.example):alice"
+
+    params = URI.decode_query(query)
+    assert params["issuer"] == "Grappa (irc.example)"
+    assert params["algorithm"] == "SHA1"
+    assert params["digits"] == "6"
+    assert params["period"] == "30"
+
+    # Not just an echo of the struct: the URI has to carry a secret the
+    # authenticator can actually key on — 20 raw bytes, Base32, unpadded.
+    assert params["secret"] == enrollment.secret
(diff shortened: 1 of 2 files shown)


vjt:
Landed via the union PR #832 as `d309513c` — closing by content, not by GitHub: the cherry-pick rewrote the SHA, so GitHub cannot mark this merged itself. Verified on `main`: `test(#749): put the TOTP door's three unasserted claims under test`. Nothing here is unlanded.

Issue #751, opened 2026-08-03
passkey + TOTP migration tests assert columns, never constraints or reversibility


vjt:
Found by the periodic codebase review, slice `test/**` + `cicchetto/e2e` since `v0.10.0`.

`test/grappa/migrations/add_user_passkeys_test.exs` and `add_user_totp_test.exs` prove the tables/columns exist (`add_user_passkeys_test.exs:53` `assert %{rows: [[0]]} = query!("SELECT COUNT(*) FROM user_passkeys")`; `add_user_totp_test.exs:115` `"PRAGMA table_info(users)"`) and, for TOTP, that a 5000-row table survives the migration — that last check is real and worth keeping.

What no test checks is every security-relevant constraint the migrations create:

- `unique_index(:user_passkeys, [:credential_id])`
- the `on_delete: :delete_all` FK to users
- the `null: false` columns
- `unique_index(:user_totp_recovery_codes, [:user_id, :code_hash])`

Neither test runs `:down`, and neither re-runs `:up` to prove idempotency.

**Failure scenario.** The `credential_id` unique index is dropped: two users register the same credential id and `Repo.get_by(Passkey, credential_id: ..., user_id: ...)` (`webauthn.ex:98`) starts resolving ambiguously. Or the FK loses `on_delete: :delete_all` and account deletion orphans passkey rows — dead credentials outliving the account they authenticated.

**Fix.** Assert against `PRAGMA index_list` / `index_info` / `foreign_key_list` for each expected constraint, and add an up → down → up round trip.

vjt:
Shipped in **v0.11.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.11.0

The tag carries `.deb`, `.rpm` and `pkg.tar.zst` assets, so self-hosters have something to pull.

Note this is a **release, not a deploy**: the hosted service is unchanged and still runs the previous build until the code-review queue closes.

PR #829, opened 2026-08-04
test(#751): assert what the passkey/TOTP migrations REFUSE

vjt:
Issue #751. Both cited sites verified against current main (`fa519aa7`)
before writing: the tests really do stop at tables-and-columns, no
constraint is asserted anywhere, and neither file runs `:down` or a
second `:up`. Nothing here was already dead.

## Shape of the assertions

Behavioural, not catalogue reads. A `PRAGMA index_list` assertion only
confirms the declaration was written; what matters is that SQLite
**refuses**. So each test performs the incident the constraint exists to
prevent and demands the error — the shape `check_constraints_test.exs`
already established in this repo. Both smoke repos run
`foreign_keys: :on`, so the FK is genuinely enforced here rather than
merely recorded.

- `unique_index(:user_passkeys, [:credential_id])` — a second account
  claiming the same credential id must be refused. That is exactly the
  issue's scenario, and it also rules out a unique over
  `[credential_id, user_id]`, which would permit it.
- the `on_delete: :delete_all` FK — deleting an account must take its
  passkeys with it, with a second account's key left standing as the
  control.
- `unique_index(:user_totp_recovery_codes, [:user_id, :code_hash])` —
  **three inserts, not one.** A duplicate rejection alone is also
  satisfied by a unique on `[user_id]` (which would break enrollment,
  since an account holds ten codes) and by one on `[code_hash]` (which
  would let one account lock another out). Same account/different hash
  and different account/same hash must both be ACCEPTED for the
  composite to be pinned. Both narrowings are in the matrix below.
- `null: false` — the one PRAGMA read. Seven inserts to prove seven
  columns would be noise; asserted as a whole set so a column losing the
  constraint and a column gaining it are equally loud.

## Reversibility: verified before tested, not assumed

The brief was right to flag this. Both migrations are `change`, and the
passkey one's middle step is a table RENAME
(`user_totp_recovery_codes` → `user_recovery_codes`) — the step most
likely to strand a database halfway. I probed it first: it reverses
cleanly, `:down, step: 1` restores the old name, drops `user_passkeys`
and removes `passkey_mode`, and a second `:up` replays. So both files now
round-trip **up → down → up**, and the second up re-runs the *refusal*
rather than re-checking columns — a table restored without its unique
index would satisfy any column-shaped check.

The 5000-row survival check in `add_user_totp_test.exs` is untouched;
everything here sits beside it.

## Mutation matrix

Each applied to the migration, run, reverted. 10 tests across both files.

| mutation | red |
| --- | --- |
| drop `unique_index(:user_passkeys, [:credential_id])` | "one credential id belongs to at most one account" + the passkey round trip |
| passkey FK `on_delete: :delete_all` → `:nothing` | "deleting an account takes its passkeys with it" |
| drop `null: false` from `user_passkeys.credential_id` | "every passkey column but the optional one is NOT NULL" |
| drop `unique_index(:user_totp_recovery_codes, [:user_id, :code_hash])` | "a recovery code is unique per account…" + the TOTP round trip |
| narrow that index to `[:user_id]` | "a recovery code is unique per account…" |
| narrow that index to `[:code_hash]` | "a recovery code is unique per account…" |
| recovery FK `on_delete: :delete_all` → `:nothing` | "deleting an account takes its recovery codes with it" |

No collateral reds. The two round-trip tests going red alongside their
constraint test is intended: they re-assert the same refusal after the
replay, which is the point of the second `:up`.

## Two facts worth recording

- **The rename does not rename the index.** After
  `user_totp_recovery_codes` → `user_recovery_codes`, the unique index is
  still called `user_totp_recovery_codes_user_id_code_hash_index`. That
  is why nothing here asserts index *names*.
- **SQLite does not imply NOT NULL for a non-INTEGER PRIMARY KEY**, so
  the `binary_id` `id` column reads as nullable in `pragma_table_info`.
  An engine quirk, not a defect of these migrations — pinned by omission
  from the expected set, with a comment, so the next reader does not
  "fix" it.

## Gate

`scripts/check.sh` rc=0 from the worktree root: credo `found no issues`,
dialyzer `Total errors: 0` / `done (passed successfully)`, doctor passed,
sobelow clean, ExUnit `8 doctests, 50 properties, 5112 tests, 0 failures`,
bats 0 not-ok. Both migration files also run green 3× in isolation.

Test-only change; no production code and no migration is modified.

### DIFF
--- a/test/grappa/migrations/add_user_passkeys_test.exs
+++ b/test/grappa/migrations/add_user_passkeys_test.exs
@@ -10,6 +10,12 @@ defmodule Grappa.Migrations.AddUserPasskeysTest do
   @totp_version 20_260_802_190_000
   @passkey_version 20_260_803_090_000
 
+  @timestamp "2026-01-01T00:00:00Z"
+  @user_a "cccccccc-cccc-4ccc-8ccc-cccccccccccc"
+  @user_b "dddddddd-dddd-4ddd-8ddd-dddddddddddd"
+  @key_1 "eeeeeeee-eeee-4eee-8eee-eeeeeeeeeeee"
+  @key_2 "ffffffff-ffff-4fff-8fff-ffffffffffff"
+
   setup do
     path = Path.join(System.tmp_dir!(), "grappa-passkey-migration-#{System.unique_integer([:positive])}.db")
     Application.put_env(:grappa, SmokeRepo, database: path, pool_size: 1, foreign_keys: :on)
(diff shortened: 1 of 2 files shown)


vjt:
Landed via the union PR #832 as `c5cbad9a` — closing by content (cherry-pick rewrote the SHA, so GitHub cannot close it itself). Verified on `main`: `test(#751): assert what the migrations REFUSE, not what they declare`. Nothing here is unlanded.

Issue #818, opened 2026-08-04
cic: readCursor's map is not identity-scoped, so read state survives an account rotation


vjt:
Spotted by w1 while closing #788 (stale-bearer guards in `scrollback.ts`). Out of that PR's perimeter — noticed, deliberately not touched, filed here.

## The defect

`cicchetto/src/lib/readCursor.ts` is not identity-scoped. The nine `onIdentityChange` resets that clear the other per-account stores do not cover its cursor map, so **the map survives a rotation or a detach**: the read cursors of the account that just left stay in memory and are applied to the windows of the account that just arrived.

This is the same shape as #788 — a store outliving the identity that populated it — but a different mechanism: #788 was continuations *in flight* holding a revoked bearer, this is state *at rest* that nobody purges.

## Failure scenario

Account A is read up to message 5000 in `#chan`. The session rotates to account B, which shares that channel and has read nothing. B's sidebar shows `#chan` as read to A's cursor: the unread divider sits in the wrong place and a genuinely unread backlog renders as already seen. The inverse is worse — B's lower cursor applied to A's window re-marks read history as unread and can drive a spurious MARKREAD upstream.

## Not answered here

Whether the fix is to route the store through the existing `identityScopedStore` factory (uniform with its siblings, but that factory owns the *transition*, and the cursor map may want to persist across a reload for the SAME identity), or to add an explicit purge to the `onIdentityChange` set. Worth a look at what the server-side cursor already guarantees — read state is server-owned per `(subject, network, channel)`, so the client map is a cache and the question is really "how does the cache learn the subject changed".

Adjacent, same review: #788 (merged/in PR #817) guarded the in-flight half.

vjt:
Shipped in **v0.11.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.11.0

The tag carries `.deb`, `.rpm` and `pkg.tar.zst` assets, so self-hosters have something to pull.

Note this is a **release, not a deploy**: the hosted service is unchanged and still runs the previous build until the code-review queue closes.

PR #830, opened 2026-08-04
cic(#818): refuse a /me hydration that outlived its identity

vjt:
Refs #818.

## The reported defect is not live — verified on the file, at `fa519aa7`

The issue says readCursor's map survives a rotation because the nine
`onIdentityChange` resets do not cover it. The premise is right (the module
does not use the factory) but the conclusion is not: `readCursor.ts` has
carried **its own** `on(token)` arm since C7.3, and its predicate is the
factory's, line for line —

```
if (prev != null && t !== prev) clearReadCursors();   // readCursor.ts
if (prev != null && t !== prev) for (const r of resets) r();   // identityScopedStore.ts
```

So neither offered option changes behaviour: **(a)** routing through the
factory and **(b)** an explicit purge both address the identity TRANSITION,
and the transition already purges. The open persistence question dissolves
too — the map is memory-only and the legacy `rc:` localStorage keys are
purged at module load, so "persist across a reload for the same identity" is
already answered *no* by construction. No product decision to make here; had
there been one I'd have stopped and asked.

## What IS live: the write that lands after the purge

`networks.ts`'s `/me` fetcher is keyed on `token`. `createResource` discards
a stale VALUE, but it cannot cancel an await — when account A's `/me` settles
past a rotation, the fetcher's side-effects still run.

Measured, not argued: hold A's `/me`, rotate to B, then release A, and
`getReadCursor("freenode", "#grappa")` is **5000 — A's cursor, under B**.
And it is never corrected: a channel B has never read answers the join reply
with `read_cursor: null`, which `applyJoinReply` deliberately no-ops on. The
same await carries `setBadge`, so B inherits A's badge count as well. That is
precisely the issue's failure scenario, reached through a door the issue did
not name: state **in flight**, not state at rest.

This is #788's shape, so it takes #788's rule — *check after the await, not
in front of the request*. The check sits in the caller rather than inside
`applyMeEnvelope` because (1) it guards the whole side-effect block, badge
included, and (2) `readCursor` cannot know which identity produced an
envelope handed to it. **The condition that would flip that choice:** a
second caller of `applyMeEnvelope`. Then the guard must move inside with the
captured token in its signature — the way `setReadCursor` already owns its
virtual-window and positive-int guards so every caller inherits them. Today
there is exactly one caller.

I did **not** extract `identityMoved` into a shared helper: it would be the
fourth inline site (`scrollback.ts`, `displayPrefs`, `customTheme`), and
"total consistency or nothing" means migrating all four in one pass — a
cross-module refactor outside this perimeter and poor timing during a freeze.
Worth doing as its own change.

## A second, quieter defect: the arm was never covered

`readCursor.test.ts`'s header has listed *"7. on(token) cleanup arm wipes on
logout/rotation"* since C7.3. There is no such case. Deleting the arm
outright leaves all 25 readCursor tests green — and the **entire 4217-test
suite** green. The one thing standing between the cache and the reported bug
had zero protection. Three cases now pin it: rotation, logout, and the
cold-start login that the `prev != null` half must NOT wipe.

## Mutation proof (one at a time, against the full suite)

| mutation | result |
|---|---|
| remove `on(token)` arm in `readCursor.ts` | **2 failed** — exactly the rotation + logout cases (4219 pass) |
| remove `if (token() !== t)` in `networks.ts` | **1 failed** — exactly the hydration case (4220 pass) |
| neither (shipped state) | 4221 / 4221 |

`bun run check` (biome + tsc) and `bun run build` both clean.

### DIFF
--- a/cicchetto/src/__tests__/networks.test.ts
+++ b/cicchetto/src/__tests__/networks.test.ts
@@ -529,5 +529,60 @@ describe("networks resources", () => {
         expect(networks.channelsBySlug()).toEqual({});
       });
     });
+
+    // #818 — the identity TRANSITION was already purged (the arm above, plus
+    // readCursor's own `on(token)`); what leaked is a write arriving AFTER it.
+    // `createResource` drops the stale VALUE but cannot cancel the await, so
+    // the fetcher's side-effects still run when account A's `/me` settles
+    // post-rotation — seeding B's cursor map (and badge) with A's numbers.
+    it("a /me still in flight from the OLD identity hydrates neither cursors nor badge", async () => {
+      localStorage.setItem("grappa-token", "tok-a");
+      await seedStubs();
+      const api = await import("../lib/api");
+      const auth = await import("../lib/auth");
+      const networks = await import("../lib/networks");
+      const readCursor = await import("../lib/readCursor");
+      const badge = await import("../lib/badge");
+
+      let releaseA: ((m: MeResponse) => void) | undefined;
+      const meA = new Promise<MeResponse>((resolve) => {
+        releaseA = resolve;
+      });
+      vi.mocked(api.me).mockImplementation((t: string) =>
+        t === "tok-a"
+          ? meA
+          : Promise.resolve({
+              kind: "user",
+              id: "u2",
+              name: "bob",
+              is_admin: false,
+              inserted_at: "x",
+              read_cursors: {},
+              badge_count: 0,
+            } as MeResponse),
+      );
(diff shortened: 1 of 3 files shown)


vjt:
Landed via the union PR #832 as `3bbcdc79` — closing by content (cherry-pick rewrote the SHA, so GitHub cannot close it itself). Verified on `main`: `cic(#818): refuse a /me hydration that outlived its identity`. Nothing here is unlanded.

Issue #759, opened 2026-08-03
shottino/call: media fds are closed while libdatachannel callback threads are still live


vjt:
Found by the periodic codebase review, slice `frontends/shottino` since `v0.10.0`.

Teardown order in `call/main.c:1205-1211`:

    media_stop(&call.send_audio);
    media_stop(&call.send_video);
    media_mix_free(&call.amix);
    media_mix_free(&call.vmix);
    session_release(&call.pub);
    for (int i = 0; i < CALL_MAX_PEERS; i++) session_release(&call.sub[i]);
    rtcCleanup();

`media_stop` closes the leg socket (`media.c:671-672`) holding NO lock. The peer connections are still alive — they are only torn down two lines later — so the RTP callbacks can still run, and they guard with a TRYLOCK:

    if (pthread_mutex_trylock(&c->vlock) != 0) return;

`main.c:293` and `:312`. The trylock SUCCEEDS during teardown precisely because teardown never takes `vlock`, so the "callback is skipped while we tear down" protection the lock exists for does not apply here.

**Failure scenario.** A peer's RTP arrives during hangup. The callback passes a `leg` whose `fd` was just closed but not yet set to `-1` (`media_feed`'s `leg->fd < 0` guard is the only check, and the close and the invalidation are two separate non-atomic stores). `session_release` immediately opens new sockets, so the descriptor number is recycled and the `sendto` targets an unrelated fd. Narrow window and no memory corruption, but the ordering is unsound and the guard reads as protection it is not providing.

**Fix.** Take `vlock` around the `media_stop`/`media_mix_free` block, or move `session_release` + `rtcCleanup()` ABOVE the media teardown so no callback thread can still be running when the fds go.

Two latent siblings found nearby, not worth their own issues but worth fixing while in here:
- `main.c:368` `shown[at] = 0;` — `at` accumulates `snprintf`'s WOULD-BE length and the loop guard (`at + 40 < sizeof(shown)`) is smaller than a record's worst case (~46 bytes). Unreachable at `CALL_TILE_MAX 3` (worst case `at ≈ 159` of 320); raising the cap makes it an OOB stack write. Clamp on the return value.
[...]

PR #841, opened 2026-08-05
shottino(#759): stop the writers before releasing what they write

vjt:
Fixes the teardown race reported in #759, plus the two latent siblings the
issue asked to be fixed while in here.

## The race

`shottino-call`'s hangup released the media legs before it stopped the threads
that write them. `media_stop` closed each leg's socket while the peer
connections were still alive two lines below, so RTP arriving mid-hangup ran on
libdatachannel's thread against a leg being dismantled — `media_feed` tests
`fd < 0` and then `sendto`s it as two steps, `media_stop` closes and blanks the
fd as two more, and `session_release` opens a socket for its DELETE right
after, so the descriptor number comes back and the write lands on an unrelated
fd.

**Which shape, and why.** The issue offered `vlock` around the media block, or
moving `session_release` + `rtcCleanup()` above it. I took the ordering.

The `vlock` reads like the protection against this and is not: the callbacks
take it with TRYLOCK because it is *backpressure* — a re-tile is forking
ffmpeg, drop the datagram rather than park a media thread — and teardown never
took it, so the trylock always succeeded exactly when it mattered. A lock whose
fast path is allowed to skip it cannot be a lifetime primitive.

Locking would have shut the window. Ordering deletes it: after `rtcCleanup()`
no callback thread exists, so nothing races the close. What locking would have
cost is the part that decided it — the concurrency stays and only gets
synchronised, so every future callback has to remember the lock, and there is
still an interval in which live peer connections point at legs already freed.
The invariant is now one sentence in the code: stop everything that can touch a
resource before releasing the resource.

Accepted cost: the decoders outlive the hangup DELETEs by however long the SFU
takes to answer them — ffmpeg processes reading loopback sockets that have gone
quiet, killed a moment later.

## Sibling 1 — the grid terminator

`shown[at] = 0` where `at` accumulated `snprintf`'s would-be length, under a
guard reserving a flat 40 bytes against a 55-byte worst case. The guard is now
the return value. The builder moves to `media.c` beside the other pure builder
over the same tiles, which is what makes it testable without the libdatachannel
link, and it is now complete-or-refused: the reader adopts a grid wholesale, so
one record short says three people are in the call when four are.

## Sibling 2 — the `#` prefix

`note()` prefixed the first line and handed the rest to `vfprintf`, while the
reader skips a line only on `line[0] == '#'` and parses every other non-empty
line as an event. A note carrying an SDP answer therefore hands the event
stream one server-controlled line per line the SFU sent. Now formatted first,
written a line at a time. `emit`/`note`/`emit_event` move to `call/note.c` so
the contract has an owner that links without libdatachannel and can be tested.

## Tests

Both siblings have tests written **failing first** against the old logic
transcribed verbatim: the grid one fails three checks plus an ASan
stack-buffer-overflow on exactly that terminator write; the note one fails six,
including the injected `{"event":...}` line arriving unprefixed.

The race itself has **no test, deliberately**. It needs a live SFU, a peer
sending RTP, and a hangup inside a window of a few instructions; the only unit
test available would assert the call *order*, which mirrors the implementation
rather than checking it. Saying so rather than shipping a test a broken
implementation would still satisfy.

## Worth knowing

**No CI job compiles the call helper.** `make` builds `shottino`, `make check`
builds the suites, `make call` is opt-in behind cmake and the vendored
submodule — so `call/main.c` is compiled by nobody in CI. That is how two
`-Wformat-nonliteral` warnings sat in it unseen (the extraction gives the
varargs helpers the format attribute the rest of the tree carries; they are
gone). The changes here were syntax-checked by hand against a stub `rtc/rtc.h`
under the shipped warning set.

## The class, swept

`shottino.c` already applies this rule and documents it: `call_helper_stop`
joins the reader before closing the pipe it reads, closes the frame pipe only
after the helper is gone "so a frame in flight cannot arrive to a freed
buffer", and `main` carries a long comment about the model thread that was
neither stopped nor joined while shutdown freed the mutex and SSL context
underneath it. The helper's own teardown was the one place that had drifted.

## Verification

16/16 local suites green from a clean rebuild (`test_keys` is Linux-only and
runs in CI). Two new suites: `tests/test_callnote`, and the grid case in
`tests/test_callmedia`.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -28954,3 +28954,81 @@ not add another.
 assertions fell. The entropy-came-from-the-source half stayed green under that
 mutation, which is the honest reading: it pins the derivation, and the refusal
 assertions pin the check.
+## 2026-08-05 — #759: teardown is an order, and a trylock is not a lifetime barrier
+
+`shottino-call`'s hangup released the media legs before it stopped the threads
+that write them. `media_stop` closed each leg's socket while the peer
+connections were still alive two lines below, so RTP arriving mid-hangup ran on
+libdatachannel's thread against a leg being dismantled — `media_feed` tests
+`fd < 0` and then `sendto`s it as two steps, `media_stop` closes and blanks the
+fd as two more, and `session_release` opens a socket for its DELETE right
+after, so the descriptor number comes straight back and the write lands
+somewhere unrelated.
+
+**The `vlock` reads like the protection against this and is not.** The RTP
+callbacks take it with TRYLOCK on purpose: it is BACKPRESSURE, meaning "a
+re-tile is forking ffmpeg, drop this datagram rather than park a media thread".
+Teardown never took it, so the trylock always succeeded exactly when it
+mattered. A lock whose fast path is allowed to skip it cannot be a lifetime
+primitive, whatever the call sites look like.
+
+**Ordering over locking.** Both fixes were on the table. Wrapping the media
+block in `vlock` shuts the window; releasing the sessions and calling
+`rtcCleanup()` FIRST deletes it. The second is preferred because it removes the
+concurrency rather than synchronising it — no future callback has to remember
+the lock, and there is no interval in which live peer connections point at legs
+that have already been freed. The invariant is now one sentence: stop
+everything that can touch a resource before releasing the resource. The pump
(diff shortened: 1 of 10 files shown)


vjt:

Issue #757, opened 2026-08-03
shottino/whip: the query-only URL branch truncates silently and forwards the fragment


vjt:
Found by the periodic codebase review, slice `frontends/shottino` since `v0.10.0`.

    if (*p != '/') {
        if (*p && *p != '?' && *p != '#') return false;
        snprintf(out->path, sizeof(out->path), "/%s", *p == '?' ? p : "");
        return true;
    }
    if (strlen(p) + 1 > sizeof(out->path)) return false;
    snprintf(out->path, sizeof(out->path), "%s", p);
    /* A fragment is a client-side concept and is never sent. */
    char *hash = strchr(out->path, '#');
    if (hash) *hash = 0;

`call/whip.c:78-88`. The two branches disagree twice: the path branch REJECTS an over-long path, the query branch TRUNCATES it; the path branch strips the fragment, the query branch does not.

**Failure scenario.** `http://sfu/?<2100 bytes>` yields a path cut at 2047 bytes, so the POST addresses a different resource than the URL named — a truncated auth token in the query silently becomes a different or absent token, and `whip_resolve` later resolves a relative `Location` against the wrong base. Separately `http://h?a#frag` sends the fragment upstream, contradicting the comment three lines below it.

**Fix.** Mirror the sibling branch: length-check before writing (`strlen(p) + 2 > sizeof(out->path)` → return false) and strip at `#` in both.

vjt:
Shipped in **v0.12.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.12.0

Pull that tag to get this. Note it is a tag-only release: irc.sindro.me is deliberately held on the previous release while resource usage is measured over several days, so "shipped" here means "in a named release", not "running on our instance".

Issue #756, opened 2026-08-03
shottino/whip: a header-less HTTP response underflows hlen to SIZE_MAX (unbounded heap over-read)


vjt:
Found by the periodic codebase review, slice `frontends/shottino` since `v0.10.0`. **Highest-severity finding of the campaign — traced by hand, not inferred.**

    const char *first_eol = memchr(raw, '\n', len);
    if (!first_eol) return false;
    const char *hblock = first_eol + 1;
    size_t hlen = (size_t)(hdr_end - hblock);

`call/whip.c:221-224`. `hdr_end` points at the START of the blank-line terminator; `hblock` points PAST the status line's newline. When a response carries ZERO header lines the terminator IS the status line's newline, so `hblock > hdr_end` and the `ptrdiff_t` subtraction wraps.

**Traced with `"HTTP/1.1 200 OK\r\n\r\n"` (len 19):**

- the CRLFCRLF branch at `:209` matches at `i == 15` → `hdr_end = raw + 15`
- `memchr` finds the first `\n` at index 16 → `hblock = raw + 17`
- `hlen = (raw+15) - (raw+17) = -2` → **`SIZE_MAX - 1`**

The LF-only form `"HTTP/1.1 200 OK\n\n"` gives `hlen = SIZE_MAX` the same way.

`header_value(hblock, hlen, "Location", …)` is then bounded ONLY by that length:

    while (pos < len) {
        size_t eol = pos;
        while (eol < len && block[eol] != '\n') eol++;

`call/whip.c:163-165` — it walks off the end of the 256 KB `raw` buffer scanning heap for `\n` and `Location:`. **Failure scenario:** segfault on the over-read; or, on a hit, `memcpy(out, block + v, vlen)` copies up to 2047 bytes of adjacent heap into `out->location`, which becomes `s->resource` and is later sent in the `DELETE` — heap disclosure to the server that triggered it.

**Reachability.** Any WHIP/WHEP endpoint answering an empty `200`/`201` with no headers. The endpoint is the SFU named in a call invite, and an invite is an `http(s)://` URL arriving in an IRC message (`call_invite_parse`, `shottino.c:3165`), so the responding host is attacker-choosable — though joining is user-initiated, so this is not zero-click.

**Fix.** After `if (!hdr_end) return false;`:

    size_t hlen = hblock > hdr_end ? 0 : (size_t)(hdr_end - hblock);

[...]

Issue #755, opened 2026-08-03
shottino: the invite's room segment is the one URL component that is never encoded


vjt:
Found by the periodic codebase review, slice `frontends/shottino` since `v0.10.0`. Defense-in-depth — no remotely reachable exploit over the IRC path today; filing because the invariant is one transport change away from mattering.

Every WHIP/WHEP URL is `<base>/rtc/<room>/<nick>/{whip,whep}`. Three of those four components are percent-encoded through `call_path_nick` (`shottino.c:12867`), which is a correct RFC-3986 unreserved allowlist with sound bounds. The room is not:

    room[0] = 0;
    for (const char *p = hash + 1; *p;) {
        if (p[0] == 'r' && p[1] == '=') {
            size_t n = 0;
            for (p += 2; *p && *p != '&' && n + 1 < room_sz; p++) room[n++] = *p;

`call_invite_split`, `shottino.c:3137-3142` — every byte except `&` and NUL is copied verbatim. It goes straight into the URL:

    snprintf(rtc, sizeof(rtc), "%s/rtc/%s", page_base, room_id);

`shottino.c:13186`. And the request builder interpolates the path raw into the request line:

    "%s %s HTTP/1.1\r\n"
    ... method, url->path, hostport, ...

`call/whip.c:367-377`.

**What is actually reachable today.** An invite arrives in a PRIVMSG, so IRC framing already excludes CR, LF and space (`call_invite_parse` also stops the URL at the first whitespace and enforces an http/https scheme — that gate is genuinely strict and I am not asking for changes there). What survives is `?`, `#`, `%` and `..`: a room of `a?x` makes the POST target `/rtc/a` with a query, and `..` walks the path. Same origin, so the blast radius is a misdirected publish rather than an exfiltration.

**Why file it anyway.** The encoding invariant is "everything that goes into this path is encoded", and this is the one hole in it. The day the room reaches this code from anywhere that is not an IRC line — a config value, a CLI flag, a `--room` argument, a future WS transport — CR/LF becomes reachable and `whip.c:367` turns it into HTTP request splitting. The fix costs one call.

**Fix.** Run `room` through `call_path_nick` (or a sibling path-segment encoder) at the point of parse in `call_invite_split`, and reject any control character in `whip_request` as a belt-and-braces second gate.

vjt:
Shipped in **v0.12.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.12.0

[...]

PR #843, opened 2026-08-05
shottino(#757): refuse an over-long query instead of cutting it, and strip the fragment on both branches

vjt:
`whip_url_parse` splits into a branch for a URL that carries a path and
one for a URL that carries only a query, and the two disagreed twice:

- the path branch length-checked and returned `false`; the query branch
  handed the bytes to `snprintf` and let it cut at `WHIP_MAX_URL`
- the path branch stripped at `#`; the query branch returned *before*
  reaching the strip, so `http://h?a#frag` sent the fragment upstream —
  contradicting the comment three lines below the write

Verified live on current `main` before touching anything. #755 changed
this file but not this function: the room-segment encoder and the
`whip_request` control-character gate sit elsewhere, so both halves of
this finding were still exactly as reported.

## Why refuse beats truncate

A cut query is not a shorter version of the request — it is a
**different** request. Another session id, or a token that is no longer
the token, sent out looking like the one that was asked for, and
`whip_resolve` then resolves a relative `Location` against a base that
never existed. That is the shape of #756: a value the parser silently
invents while reporting success. Refusing hands the caller the one true
statement available — this URL does not fit — instead of a plausible
wrong one.

The branches now differ only in whether they prepend the `/`, and the
single fragment strip runs after both, so neither can drift from the
other again by being edited alone.

## Tests

Added to `a_url_is_split_or_refused`, asserting the resulting path
**byte for byte** rather than the bool — a bool-only test survives
deleting the strip, since the parse still succeeds either way:

- `http://h?a=1#frag` → `/?a=1`, `http://h#frag` → `/`, `http://h?a=1` → `/?a=1`
- a query past `WHIP_MAX_URL` is refused
- the bound is pinned from both sides: the largest query that fits still
  parses (path length `WHIP_MAX_URL - 1`), one byte more is refused, so
  an off-by-one cannot pass itself off as a refusal

Measured, not assumed — each mutation kills checks:

| mutation | checks killed |
|---|---|
| drop the query-branch length gate | 2 |
| drop the fragment strip | 3 |
| off-by-one (`+2` → `+1`) | 1 |
| restore the pre-fix early return verbatim | 3 |

## Class sweep

No other branch pair in this parser disagrees the same way.
`whip_url_format`, both `whip_resolve` branches, `header_value` and the
`head` buffer in `whip_request` all refuse on overflow; `hostport` is
sized so the parser's own bounds make truncation unreachable; the
`media.c` SDP builders all return `w > 0 && w < out_sz`. The one
remaining truncation, `split_source`, is fed by a local `/set` value,
not by a stranger, and both of its callers treat it identically — no
disagreement to reconcile.

## Gate

`make check` locally: 15 of 16 suites, 11004 checks green. `test_keys`
needs `<pty.h>` and cannot build on darwin; CI runs all 16 on Ubuntu.

### DIFF
--- a/frontends/shottino/call/whip.c
+++ b/frontends/shottino/call/whip.c
@@ -77,11 +77,16 @@ bool whip_url_parse(const char *url, struct whip_url *out) {
      * malformed, so this is normalised here rather than at every use. */
     if (*p != '/') {
         if (*p && *p != '?' && *p != '#') return false;
-        snprintf(out->path, sizeof(out->path), "/%s", *p == '?' ? p : "");
-        return true;
+        /* Refused rather than cut to length, like the branch below: a
+         * truncated query is a DIFFERENT request — another session id,
+         * a token that is no longer the token — and it would go out
+         * looking like the one that was asked for. */
+        if (strlen(p) + 2 > sizeof(out->path)) return false;
+        snprintf(out->path, sizeof(out->path), "/%s", p);
+    } else {
+        if (strlen(p) + 1 > sizeof(out->path)) return false;
+        snprintf(out->path, sizeof(out->path), "%s", p);
     }
-    if (strlen(p) + 1 > sizeof(out->path)) return false;
-    snprintf(out->path, sizeof(out->path), "%s", p);
     /* A fragment is a client-side concept and is never sent. */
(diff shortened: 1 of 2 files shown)


vjt:

Issue #754, opened 2026-08-03
shottino: call/main.c (1214 lines) is compiled by no CI or release job


vjt:
Found by the periodic codebase review, slice `frontends/shottino` since `v0.10.0` (31 commits, ~7000 new lines).

The `shottino` CI job exists precisely to stop this. Its own comment:

    # Before this job nothing in CI compiled it at all: a change that broke
    # the build (or the wire narrowers) reached main unnoticed and would
    # first surface in the release workflow, on a tag

`.github/workflows/ci.yml:138-141`. But the job runs `make -C frontends/shottino` (`all: $(BIN)`, Makefile:56) and `make check`, and the media helper is a separate opt-in target:

    .PHONY: call
    call: $(CALL_BIN)
    CALL_SRC := call/main.c call/whip.c call/media.c http.c

Makefile:181-186. `call/whip.c` and `call/media.c` DO get compiled — the test binaries link them deliberately without libdatachannel (`Makefile:96-97, 103-104`), which is a good call. **`call/main.c` is the exception**: 1214 lines, the helper's whole main loop and the entire libdatachannel integration, and no target CI runs ever touches it.

Verified: `git grep -n "make call\|shottino-call\|libdatachannel" -- .github/ infra/` returns nothing. Neither `ci.yml` nor `release.yml` builds it, and no packaging path does either.

**Failure scenario.** `call/main.c` stops compiling — or grows a `-Werror` warning the rest of the tree would fail on — and main stays green. Nobody finds out until a developer runs `make call` by hand. The `-Werror` guarantee the CI job advertises simply does not extend to the largest file in the feature.

Second-order: since no packaging path builds the helper, every shipped `.deb`/Arch package takes the `install` fallback ("no shottino-call built — calls will open a browser", Makefile:224). That part looks deliberate — the opt-in exists because the helper needs cmake and a C++ toolchain — but it does mean the native call path is untested end-to-end in anything we ship.

**Fix.** Add a CI step that at minimum syntax-checks it (`$(CC) -fsyntax-only -I vendor/libdatachannel/include call/main.c`), or a full `make -C frontends/shottino call` job with `cmake` + `g++` installed and the libdatachannel build cached. Cheapest honest option is the syntax check; the full build is what actually matches the job's stated promise.

vjt:
Corroborated from the test-suite angle, plus one detail worth folding in: even if a `make call` step is added, the helper's own flags carry no sanitizer.

[...]

PR #844, opened 2026-08-05
shottino: compile call/main.c in CI (#754)

vjt:
Issue #754: `call/main.c` — 1214 lines, the helper's main loop and the whole
libdatachannel integration — is compiled by no CI, release or packaging job.

Re-verified on current `main` before touching anything: `git grep -n "make
call\|shottino-call\|libdatachannel" -- .github/ infra/` still returns nothing,
`all: $(BIN)` still excludes `$(CALL_BIN)`, and `test_whip`/`test_callmedia`
still link `whip.c`/`media.c` deliberately without libdatachannel. So
`call/main.c` is the one helper source no gate touches. Still real.

## Which option, and why

The issue offers a `-fsyntax-only` check or a real `make call` with cmake and a
cached libdatachannel build, and is honest that they are not equivalent. This is
neither of the two exactly: it is a real **compile** (`-c -o /dev/null`) of every
helper translation unit, with the same `-Werror` flag set as the client build
step above it.

- `-fsyntax-only` was the cheap option, and it is cheap partly by dropping
  diagnostics that need codegen. A compile costs the same second and does not.
- The full `make call` is what the job's promise implies, and it costs cmake,
  g++, and a multi-minute libdatachannel + libjuice + libsrtp + usrsctp build to
  cache — on a job that currently finishes in **2 minutes** end to end. Because
  this edits `ci.yml`, merging it re-runs every open PR, so the cost is paid many
  times over immediately.
- The compile needs **nothing new**: the job already does `submodules:
  recursive`, and the run log for `main` shows `libdatachannel` and its five
  deps being checked out today. The headers are on disk before the first step.

What it catches is exactly the failure the issue names: the file stops
compiling, or grows a warning the rest of the tree fails on. What it does not
catch: the C++ link driver, `-static-libstdc++`, and the four archives. That
gap is stated in both comments rather than implied away.

## Comment and reality now agree

The job advertised a `-Werror` guarantee that stopped short of the largest file
in the feature. Two comments changed so the advertisement matches:

- the job header now says what it covers, exactly — client built and tested,
  helper **compiled but never linked and never packaged**;
- the new step and the Makefile target both name the link as the deliberate
  hole.

If a later change adds the real `make call` job, both comments are the thing to
update; they are written to be falsifiable, not reassuring.

## I saw the gate fail

`make -C frontends/shottino call-compile` with the CI flags, on the untouched
tree: `rc=0`. Then two mutations appended to `call/main.c`, one at a time:

| mutation | result |
|---|---|
| `rtcNoSuchFunction(1);` — a libdatachannel symbol that does not exist | `error: call to undeclared function` → `make: *** [call-compile] Error 1`, `rc=2` |
| `static int f(int a, int b) { (void)a; return 0; }` — a mere warning | `error: unused parameter 'b' [-Werror,-Wunused-parameter]` → `rc=2` |

Restored, `rc=0`, tree clean. The second one matters more than the first: it
proves the `-Werror` half of the promise, not just the compiles-at-all half.

Verified under **both** compilers, since the local one is not the CI one: clang
on darwin (with the usual `_DARWIN_C_SOURCE` / `SOCK_CLOEXEC` shims), and gcc
14.4 in a Linux container running the byte-exact CI command, including the
`sed`-extracted `CFLAGS_DEPS`. Green on both. `ci.yml` parses (pyyaml) and the
step lands between Build and Tests.

One thing found while measuring: the shipped Makefile's own `WARNINGS` carries
`-Wformat=2`, but the CI Build step's `CFLAGS` override drops it — and under
clang the helper's varargs forwarders do trip `-Wformat-nonliteral`. This gate
therefore uses the Build step's flag set verbatim rather than inventing a
stricter one in a CI-plumbing change. Worth a separate look at whether CI should
carry `-Wformat=2`; not this PR.

## The packaging second-order point

The issue reads the packaging gap as deliberate, and I agree, with one caveat
worth writing down. `nfpm.yaml` builds one binary whose runtime deps the bundled
ERTS already pulls in; asking the .deb and AUR builders for cmake and a C++
toolchain to produce a second binary that most users will never invoke is a real
cost for a feature that degrades to "open a browser". The opt-in is the right
default and `install` already announces the fallback in plain words.

The caveat is that the fallback is the *only* path anything we ship has ever
taken, so `call_helper_path`'s discovery, the `--protocol` handshake and the
version-refusal branch are exercised by no shipped artifact. That is a testing
gap, not a packaging one, and it is not fixed by building the helper into a
package. I do not think it is worth a finding on its own yet; if the native call
path is meant to be a supported feature rather than a build-it-yourself one,
that is the conversation to have, and it is a bigger one than #754.

Scope note: packaging untouched, no `.deb`/AUR change, no new job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,6 +143,11 @@ jobs:
   # first surface in the release workflow, on a tag, when the packages
   # tried to build a client that no longer compiled.
   #
+  # What it covers, exactly: the client is built and its suites run, and
+  # the media helper's sources are COMPILED but never linked and never
+  # packaged — the shipped .deb/AUR carry the client alone and fall back
+  # to a browser for calls, so nothing here proves a helper binary works.
+  #
   # Deliberately its own job rather than a step on `test`: it shares no
   # toolchain with the BEAM jobs (no Elixir, no deps cache), it runs in
   # well under a minute, and a failure should read as "shottino broke"
@@ -177,6 +182,20 @@ jobs:
         run: |
           make -C frontends/shottino clean
           make -C frontends/shottino \
+            CFLAGS="-std=c11 -Wall -Wextra -Wpedantic -Werror -O2 $(sed -n 's/^CFLAGS_DEPS := //p' config.mk)"
+
+      # The media helper is a separate opt-in target (`make call`) because
+      # linking it wants cmake, a C++ toolchain and a multi-minute
+      # libdatachannel build. So `make` above never touched call/main.c —
+      # 1214 lines, the helper's main loop and the whole libdatachannel
+      # integration — and no other job did either (#754). Compiling it is
+      # what that guarantee costs here: the vendored headers arrived with
+      # the recursive checkout, nothing has to be built, and the flags are
+      # the ones above so the helper cannot warn where the client may not.
+      # The link stays unproven on purpose — see the Makefile comment.
+      - name: Compile the media helper (warnings are errors)
+        run: |
+          make -C frontends/shottino call-compile \
             CFLAGS="-std=c11 -Wall -Wextra -Wpedantic -Werror -O2 $(sed -n 's/^CFLAGS_DEPS := //p' config.mk)"
 
       # Suites build with ASan+UBSan (see the Makefile): this is C parsing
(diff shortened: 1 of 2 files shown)


vjt:

Issue #758, opened 2026-08-03
shottino/call: ffmpeg exec failure is undetectable, so every "is ffmpeg installed?" error is dead code


vjt:
Found by the periodic codebase review, slice `frontends/shottino` since `v0.10.0`.

`spawn_ffmpeg` (`call/media.c:82-95`) returns a valid pid whenever `fork()` succeeds. A failing `execvp` only sets the CHILD's exit status to 127, and nothing ever inspects it: the only `waitpid` calls in the helper are the blocking reaps in `media_stop` (`media.c:668`) and `media_mix_stop` (`media.c:587`), both on the teardown path. There is no `WNOHANG` poll and no `SIGCHLD` handler while a call is running.

So the start functions report success unconditionally:

    mix->pid = spawn_ffmpeg(argv, -1);
    return mix->pid > 0;

`media.c:580-581`, and the same at `:645`. `media_start_send` (`:239`) only catches `fork` failure for the same reason.

Which makes these unreachable:

    if (!ok) emit_event("error", "message", "cannot start audio playback (is ffmpeg installed?)");

`call/main.c:399`, and `"cannot open the microphone"` / `"cannot open the camera"` at `main.c:1015-1017`.

**Failure scenario.** ffmpeg is not installed — the common case, since the helper is opt-in and its README does not make ffmpeg a hard dependency. The call connects, the tile map publishes, the UI draws a call window, and the user gets silence and a black rectangle with no diagnostic anywhere. `media.h` claims "A failure shows as the process dying, which the caller sees"; nobody looks.

**Fix.** Either have `spawn_ffmpeg` carry the child's `errno` back over an `FD_CLOEXEC` pipe (write-then-`_exit` in the child; a non-empty read means exec failed), or poll `waitpid(pid, &st, WNOHANG)` on the supervisor tick and emit the error for any leg whose child has exited.

vjt:
Shipped in **v0.12.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.12.0

Pull that tag to get this. Note it is a tag-only release: irc.sindro.me is deliberately held on the previous release while resource usage is measured over several days, so "shipped" here means "in a named release", not "running on our instance".

Issue #760, opened 2026-08-03
shottino: a malformed tiles line blanks the call grid, inverting the stated all-or-nothing contract


vjt:
Found by the periodic codebase review, slice `frontends/shottino` since `v0.10.0`.

`call_tiles_parse` is deliberately all-or-nothing, and its test pins that:

    /* … the whole line is refused and the caller keeps the grid it
       already had — better a stale picture than a mislabelled one. */

The only call site does the opposite. `shottino.c:13073-13086`:

    struct call_tile tiles[CALL_MAX_PEERS];
    int fw = 0, fh = 0;
    int n = call_tiles_parse(value, &fw, &fh, tiles, CALL_MAX_PEERS);
    pthread_mutex_lock(&app->lock);
    memcpy(app->call_live.tiles, tiles, sizeof(tiles));
    app->call_live.tile_count = n;
    app->call_live.frame_w = fw;
    app->call_live.frame_h = fh;

No `n > 0` guard. On a refused line the caller adopts `n = 0` and a `0x0` frame — the grid is blanked, which is exactly the outcome the parser's contract (and the comment three lines above the memcpy, "Adopted wholesale or not at all") exists to prevent.

Second defect on the same path: `tiles` is an UNINITIALISED local. `call_tiles_parse` returns 0 on its early-exit paths without writing `out`, so the `memcpy` propagates uninitialised stack into `app->call_live.tiles`. Harmless while every reader gates on `tile_count`, but it is uninitialised-memory propagation under a mutex and MSan/UBSan would be right to shout.

**Failure scenario.** One malformed `tiles` line from the helper mid-call — a truncated pipe write, a re-grid racing a peer leaving — blanks the video grid instead of holding the last good one, and it stays blank until another valid line arrives.

**Fix.** `if (n > 0) { …apply… }`, and initialise `tiles` at declaration. The call site has no test at all — worth extracting the per-line handling of `call_reader_main` into a pure `call_event_apply(app, line)` so the `tiles` event can be driven from a test with real helper output.

vjt:
Shipped in **v0.12.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.12.0

Pull that tag to get this. Note it is a tag-only release: irc.sindro.me is deliberately held on the previous release while resource usage is measured over several days, so "shipped" here means "in a named release", not "running on our instance".

PR #847, opened 2026-08-05
shottino(#758): let a failed ffmpeg exec reach the error message written for it

vjt:
Fixes the #758 blind spot: `spawn_ffmpeg` returned a pid whenever `fork()`
succeeded, so a machine with no ffmpeg opened a call window, published a tile
map, and produced silence and a black rectangle with no diagnostic anywhere.

## Still real, verified against the current files

`call/media.c:83-98` — `execvp` failing only set the CHILD's exit status to
127; the only `waitpid`s in the helper are the blocking reaps in `media_stop`
and `media_mix_stop`, both on teardown. So `media_start_send` (`:240`) caught
only a fork failure, and both mixes `return mix->pid > 0` (`:581`, `:674`) over
a pid that was always valid. The three messages the issue calls unreachable —
`call/main.c:399`, `:1015`, `:1017` — were exactly that.

## Shape (a), the FD_CLOEXEC pipe, over shape (b), the WNOHANG tick

- **It makes the EXISTING messages reachable.** The callers were already
  written correctly; the primitive under them lied. A tick poll has no bool to
  answer, would need error messages of its own, and would leave the ones the
  issue names as dead code — i.e. not fix the defect.
- **It cannot conflate.** The write happens between `execvp` returning and
  `_exit`, so "ffmpeg is not installed" is never confused with "ffmpeg died at
  second thirty". A `WNOHANG` poll conflates the two by construction, and is
  late by up to a tick.
- **It is testable.** PATH decides the outcome, so the suite never asks whether
  the machine running it has ffmpeg. The tick lives in `call/main.c`, which no
  test links and (until #754) no job compiles.

Cost: one pipe per spawn on a path that already forks, and a read that blocks
for as long as the exec takes. shottino already holds this line elsewhere —
`call_helper_path` says "not installed" BEFORE forking, "because a child that
exits 127 into a pipe" tells nobody anything.

## The header now agrees with the code

`media.h` promised "a failure shows as the process dying, which the caller
sees" over code where nobody looked, and "returns false … if ffmpeg cannot be
started" over a function that could not return false. The first is replaced
with what is actually detected; the second is now true. What is still NOT
covered — an ffmpeg that starts and then dies, a device that will not open — is
stated in the header and in `docs/CALLS.md` rather than implied away. A comment
promising a guarantee nobody implements is the bug (cf. #754).

## Tests, and the mutation that found a bug in the test

Failing first: with the fix reverted, 7 checks fall across all three start
functions. Four mutations measured:

| mutation | caught |
|---|---|
| revert `spawn_ffmpeg` to `origin/main` | 7 FAIL |
| drop `FD_CLOEXEC` from the write end | parent hangs (CI timeout) |
| treat EOF as a failed exec (`got < 0`) | parent hangs |
| drop the reap of the failed child | **NOT caught** → fixed, see below |

The reap matters: the supervisor rebuilds both mixes on every tick, so on the
machine this issue is about an unreaped failure is one zombie per tick for the
length of the call. The suite now asserts `waitpid(-1, WNOHANG) == ECHILD`
after a failed start.

The no-CLOEXEC mutation found a defect in the TEST, not the fix: the fake
ffmpeg (`exec sleep 30`) ran under a PATH holding only the scratch directory,
could not find `sleep`, and died instantly — closing the write end and handing
the parent a clean EOF by accident, so `pid > 0` passed on a zombie. The
positive control now sets its own PATH and asserts the child is RUNNING.

## The class

Every other fork+exec in shottino already has a channel its failure reaches:
`run_cmd` returns the exit status; the recorder waits and reports ffmpeg's last
stderr line; the `claude` CLI's exec failure arrives as an empty reply and says
"is \`claude\` on PATH and logged in?". The double-forked `xdg-open` /
`notify-send` / `mpv` are fire-and-forget by design (the grandchild is an
orphan; nothing can wait for it). `call/media.c` was the only place where the
diagnostic existed and could not be reached.

## Gates

- 15/16 suites green locally (`test_keys` is Linux-only), and **16/16 in a
  Linux container running the byte-exact CI commands** — build with `-Werror`
  and `make check`.
- `test_layout` fails 2 of 2937 checks in that container — **identically on
  `origin/main`**, so pre-existing and not from this change.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -29112,3 +29112,51 @@ arrive to a freed buffer", and `main` ends with a long comment about the model
 thread that was neither stopped nor joined while shutdown freed the mutex and
 SSL context underneath it. The helper's own teardown was the one place that had
 drifted from the rule the rest of the tree follows.
+## 2026-08-05 — #758: a primitive that cannot fail makes its callers' error handling dead code
+
+`spawn_ffmpeg` in the call helper returned a pid whenever `fork()` succeeded.
+`execvp` failing set the CHILD's exit status to 127 and nothing ever inspected
+it — the only `waitpid`s in the helper are the blocking reaps on the teardown
+path — so all three start functions reported success unconditionally. On a
+machine with no ffmpeg the call connected, the tile map published, the UI drew
+a call window, and the user got silence and a black rectangle with no
+diagnostic anywhere. ffmpeg is an OPT-IN dependency of an opt-in feature, so
+this is the ordinary case rather than an exotic one.
+
+**The callers were already right; the primitive underneath them lied.**
+`media_start_send` closes its socket and returns false on -1, and both mixes
+`return mix->pid > 0`. Three error messages were written for that false branch
+and none of them could be reached. That is the general shape worth naming: a
+function that cannot report failure does not merely lose an error, it converts
+every caller's error handling into unreachable code that reads, in review, as
+though the case were handled.
+
+**The exec gets a channel of its own: an `FD_CLOEXEC` pipe.** A successful
+exec closes the write end, so the parent reads EOF; a failed one leaves the
+child alive to write its errno first, so a non-empty read means the exec
+failed and nothing else does. Chosen over polling `waitpid(WNOHANG)` on the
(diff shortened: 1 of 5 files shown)


vjt:

PR #848, opened 2026-08-05
shottino(#760): keep the grid a refused tiles line was meant to keep

vjt:
Fixes the contract violation in #760 — and does **not** fix it the way the issue words it, for a reason worth reading before the diff.

## What was verified against the current file

`call_tiles_parse` (`shottino.c:3547`) and its only call site (now `call_event_apply`, formerly inline in `call_reader_main` around `shottino.c:14681`). Both defects were still real:

- No guard on the parse result. A refused line was adopted as count `0`, frame `0x0` — the grid blanked, which is the exact outcome the parser's own comment, the comment three lines above the `memcpy`, and `docs/CALLS.md` all exist to prevent.
- `tiles` was an uninitialised local, and the parser writes nothing to it on its early exits, so the `memcpy` under `app->lock` carried uninitialised stack into `app->call_live.tiles`.

## Why `if (n > 0)` is the wrong fix

The helper publishes `{"event":"tiles","value":""}` when it stops the decoder because nobody is sending a picture — `call/main.c` `video_retile`, the `n == 0` branch. That arrives as the same `0` a malformed line does. A count-only guard would therefore hold the last grid up over a frame stream that has *stopped*: two faces still labelled, nobody on camera.

The root cause is not the missing guard, it is that `0` already meant two opposite things, so no correct guard could be written at the call site. The parser now answers three ways:

- `< 0` — refused. Keep the grid you have.
- `0` — a grid with no cells. Everybody's camera is off; clear.
- `> 0` — the tile count, and only then are `*frame_w` / `*frame_h` written.

A bare frame size (`"640x480"`, nothing after it) stays a **refusal**: it is a line cut mid-write, and the helper never emits one — when `media_grid_layout` refuses, `media_start_video_mix` returns false on `tile_count <= 0` and the helper reports an `error` event instead of a tile-less grid. The empty grid has exactly two spellings, `""` and `"<w>x<h>;"`, both pinned.

## Should a refused line be reported?

**Yes, and it now is** — one line into the call window. Unreported, a grid that silently stops following the call reaches us as "the video broke", which nobody can act on. That is the same failure mode as #758's error strings no branch could reach and #756's over-read no one could observe.

Neither latched nor rate-limited, deliberately: a `tiles` event is emitted when the grid *changes*, not per frame, so the flood a latch would guard against does not exist — and a latch is a per-call field that needs resetting and will drift. The line does **not** echo the payload; the three sibling branches in the same loop already echo helper-controlled text verbatim, and this one had nothing to add by widening that surface.

## The RED was measured, not assumed

The per-line handling moved out of the reader thread into `call_event_apply` so the decision can be driven with the helper's real JSON — a thread blocked in `fgets` on a pipe cannot be asserted about, which is why this call site had no test at all. Five mutations, each applied to the committed tree and run:

| mutation | checks failed |
|---|---|
| drop the refusal guard (the #760 defect, restored) | 9 |
| collapse the parser back to one `0` | 17 |
| drop the `= {0}` on the tiles local | 35 |
| write the guard the way the issue words it (`n <= 0`, old parser) | 17 — three of them the camera-off regression |
| keep the guard, say nothing | 2 |

The zero-init check is the one that could have been a zombie: on a quiet stack an uninitialised array is often already zero, so the test dirties the stack first through a `noinline` frame sized past the tile array. With `= {0}` reverted the assertions come back `0xAAAAAAAA`, so the check is reading real garbage rather than an accidental clean slate.

## Class sweep

One structural twin, latent not live: `admin_verb_run` (`shottino.c:5396`) `memcpy`s an uninitialised `struct admin_field fields[ADMIN_FORM_MAX]` whole into `app->overlay.form`, but it is guarded by `if (nf)` and every reader is bounded by `form_count`. Producer side, `video_retile` (`call/main.c:342`) assigns `media_grid_layout`'s count straight into `c->vmix.tile_count` and leaves `tiles` stale on a refusal — safe for the same reason. No other all-or-nothing parser in `frontends/shottino` has a call site that adopts partial output; ~30 candidates were checked and ruled out (every `wire_*` narrower is guarded, `llm_config_parse` is incremental by design, the rest pre-initialise their out-parameters).

## Gate

`make check` over 15 of the 16 suites locally (`tests/test_keys` needs `<pty.h>`, Linux-only — CI runs all 16): exit 0, 15 suites, `1302 checks passed` in `test_windows`, no new warnings.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -29161,3 +29161,69 @@ lives in the CALLER's judgement, not in the context function: `revoke_session/1`
 always returns the typed error, and the plug decides to continue. Absorbing
 inside the context would have taken the choice away from logout, which needs
 it.
+
+## 2026-08-05 — #760: a refusal and an empty answer are not the same zero
+
+`call_tiles_parse` decides which rectangle of the composited call frame
+belongs to which person, and it is deliberately all-or-nothing: a line that
+does not parse cleanly is refused whole, because a half-applied grid draws
+faces under the wrong names. That contract is stated three times — in the
+function's own comment, in the comment three lines above the `memcpy` that
+consumed it, and in `frontends/shottino/docs/CALLS.md`. The single call site
+did the opposite of all three: it adopted the refusal, count `0` and frame
+`0x0`, so one malformed line — a truncated pipe write, a re-grid racing a peer
+leaving — blanked the video grid mid-call and left it blank until the next good
+line arrived.
+
+**The fix the issue proposed would have introduced a second bug.** `if (n > 0)`
+reads as the obvious guard, and it is wrong, because `0` already had a second
+meaning: the helper publishes `{"event":"tiles","value":""}` when it stops the
+decoder because nobody is sending a picture (`call/main.c` `video_retile`), and
+that is an INSTRUCTION, not a failure. A count-only guard would hold the last
+two faces labelled over a frame stream that has stopped. The root cause is not
+the missing guard — it is that a lossy return value made the guard unwriteable:
+the caller could not distinguish "there is nothing to draw" from "I could not
+read this". So the parser now answers three ways (`< 0` refused, `0` an empty
+grid, `> 0` the tile count) and the caller does the opposite thing in each.
+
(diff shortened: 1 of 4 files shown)


vjt:

Issue #764, opened 2026-08-03
shottino: the call's identity-critical helpers have zero test references


vjt:
Found by the periodic codebase review, slice `frontends/shottino` since `v0.10.0`.

Verified by grep across `frontends/shottino/tests/` — these have NO test reference at all:

- `call_path_nick` — the URL path fold/escape
- `call_url_for_me`, `call_invite_peers` — invite construction
- `call_helper_path`, `call_helper_accept` — helper resolution
- `call_frame_box` — the frame geometry handed to the helper
- `call_vcodec_parse` / `call_vcodec_word` — the codec word round trip

`call_path_nick` is the sharpest gap. Its own comment explains why it exists:

    ESCAPE, do not squash. A nick may legally contain [ ] \ ` ^ { | },
    none of which belong in a path unescaped. Mapping them all to `_`
    would be simpler and would make foo[1] and foo{1} the SAME room —
    two different people in one call, silently.

Both ends of a call must derive the identical path or they publish past each other. That is a pure, trivially testable function — fold + percent-encode + bounds — and nothing tests it.

`call_vcodec_parse`/`call_vcodec_word` is the same word↔enum round-trip pattern as `call_ring_parse`/`call_ring_word`, which DOES get a round-trip loop. One sibling was tested, the other was not.

**Failure scenario.** The escape set changes on one side of a release; `foo[1]` encodes differently on two clients, both connect, neither hears the other, and nothing in CI notices.

**Fix.** A `call_path_nick` table test (fold, each of `[ ] \ ` ^ { | }`, a truncating bound, non-ASCII) and a `call_vcodec` parse↔word loop next to the existing ring-policy one.

vjt:
Shipped in **v0.12.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.12.0

Pull that tag to get this. Note it is a tag-only release: irc.sindro.me is deliberately held on the previous release while resource usage is measured over several days, so "shipped" here means "in a named release", not "running on our instance".

Issue #762, opened 2026-08-03
shottino: the new call tests assert shapes that a broken implementation still satisfies


vjt:
Found by the periodic codebase review, slice `frontends/shottino` since `v0.10.0`. Four assertions that do not constrain what their test names claim.

**1. The 4:3 rows formula is unconstrained.** `tests/test_windows.c:2127-2136` pins the WIDTH exactly (`CHECK_LONG(c, 20)`, `40`, `16`) but the height only by inequality:

    CHECK(r >= 6);
    CHECK(r * 2 <= c);

Production is `int r = (w * 3) / 8;` (`shottino.c:13005`). At `c = 40` the correct `r` is 15 — but 6, 8, 10 and 20 all pass both checks. Replace the formula with `r = w / 4` and every face is stretched at every terminal width, gate still green. Fix: assert the exact rows for the three widths already named (80→7, 200→15, 40→6).

**2. The loopback bind test never checks loopback.** `tests/test_callmedia.c:345`:

    CHECK(a != b);

Two simultaneously-open ephemeral sockets are distinct by kernel guarantee, so this constrains nothing. Nothing asserts the fd is bound to `INADDR_LOOPBACK` (production uses it, `call/media.c:41`) nor that the reported port matches the socket — even though the test's own comment says "the port has to be one the caller can actually tell ffmpeg about". Change `INADDR_LOOPBACK` to `INADDR_ANY` and unauthenticated RTP ports are exposed to the LAN with the test still green. Fix: `getsockname` and assert both the address and the port.

**3. The codec under test is an accident of zero-initialisation.** `tests/test_callmedia.c:23-34` — the designated initialiser never sets `.video_codec`, so it is `0` = `MEDIA_VIDEO_VP8`, and the test then asserts `a=rtpmap:96 VP8/90000`. Reorder the enum so `MEDIA_VIDEO_H264 = 0` and the assertion silently changes meaning. Fix: set `.video_codec = MEDIA_VIDEO_VP8` explicitly.

**4. The sampling draw is validated against an all-black buffer.** `tests/test_layout.c` allocates `calloc(40 * 40 * 3, 1)` and asserts only the CELL COUNT (`count_blocks(...) == 50`). Nothing asserts the draw uses the sampler, so `draw_media_region_locked` ignoring the source rect and painting the top-left tenth — the exact bug its comment names — still counts 50 cells. Fix: fill the buffer with a positional gradient and assert that corner cells differ.

vjt:
[...]

PR #849, opened 2026-08-05
shottino(#764): test the call helpers that decide who is in the room

vjt:
Addresses the #764 review finding: the call's identity-critical helpers
had no test reference at all. New suite `frontends/shottino/tests/test_call.c`
(17th in `make check`), plus its Makefile rule.

## First: the issue is partly stale, and I am not inventing a replacement gap

The findings were written at `388f883b`. Re-grepped `tests/` at
`9be2778f`:

| helper | issue says | actually |
| --- | --- | --- |
| `call_path_nick` | no reference | no DIRECT test — but reached indirectly by the `call_invite_peers` tests |
| `call_invite_peers` | no reference | **covered since**, `test_windows.c:3249-3325` (three tests) |
| `call_url_for_me`, `call_helper_path`, `call_helper_accept`, `call_frame_box`, `call_vcodec_parse`/`call_vcodec_word` | no reference | confirmed, none |

`call_invite_peers` is no longer a gap and I did not write a second set
of tests for it.

`call_path_nick`'s indirect coverage is worth naming precisely, because
it is why the gap survived: the three peer-list tests only ever pass
`vjt`, `VJT` and `sarabean`. A squash-to-`_` encoder, a raw copy and the
real percent-encoder all produce the same bytes for those, and in the
`VJT` case `irc_name_eq` skips the nick before `call_path_nick` even
runs. So nothing existing constrained either the fold or the escape set —
which is exactly what the issue says, by a different route.

`call_path_nick` also no longer holds the encoder; it delegates to the
single `call_path_encode(in, fold, …)` that #755 introduced, with `fold`
as the domain axis. The tests are written against `call_path_nick`, the
identity-side caller.

## What each test constrains

- **fold + escape set, byte for byte**, with `foo[1] != foo{1}` asserted
  as the thing that must not happen rather than as two encodings that
  happen to differ. Same invariant this tree spells `canonical_target`:
  fold `A-Z` and nothing else, and only KEYs.
- **truncation held against the WHOLE encoding.** A short buffer must
  produce a *prefix* of it — skipping the byte that did not fit and
  carrying on would also produce a shorter string, and would map `a[b`
  and `ab` to one path. Exactly-sized heap buffers, so ASan is the
  bounds assertion rather than a hand-written canary.
- **`me=` compared to `peers=`, not to a literal**, with a bracketed
  nick (`vjt[m]`). They are two call sites that must agree — the browser
  prefills from one and subscribes to the other.
- **the codec word handed to the HELPER'S parser.** This is the one that
  needed the cross-binary link (`call/media.c`): the round-trip loop
  alone cannot say it, because shottino's parser accepts strictly more
  spellings than the helper's. See the measurement below.
- **frame box floors and caps**, tied to what they are for: the reader
  thread turns the box into `cols * rows * 2 * 3` bytes per frame.
- **helper search ORDER**, and that a configured-but-unusable
  `call.helper` FAILS rather than falling through to whatever is on PATH.

## Mutation results — every RED proven load-bearing

14 mutations applied to `shottino.c`, rebuilt, re-run. 13 killed:

| mutation | outcome |
| --- | --- |
| M1 drop the fold | KILLED (4) |
| M2 squash specials to `_` | KILLED (15) |
| M3 rfc1459 over-fold (`[]\^` → `{}|~`) | KILLED (8) |
| M4 `tolower()` instead of the `A-Z` range | **SURVIVED** — see below |
| M5 escape bound `n+4` → `n+3` | KILLED (ASan heap-buffer-overflow, WRITE of size 1) |
| M6 `continue` instead of `break` on overflow | KILLED (5, the prefix check) |
| M7 `call_vcodec_word(H264)` returns `h.264` | KILLED (3) — **round trip still passed** |
| M8 configured helper falls through to the rest of the search | KILLED (3) |
| M9 drop the fit check in `call_helper_accept` | KILLED (3) |
| M10 drop the frame-box floors | KILLED (195) |
| M11 `me=` uses the raw nick | KILLED (4) |
| M12 search PATH before HOME | KILLED (2) |
| M13 drop the frame-box caps | KILLED (1345) |
| M14 append `me=` to a non-invite URL | KILLED (2) |

M7 is the one that justifies linking the second binary. `h.264`
round-trips through shottino's own parser and is refused by
`call/media.c`, so a `word()` returning it passes the parse↔word loop
and silently runs the call on the default codec while the user asked for
the other one. A literal `CHECK_STR(word, "h264")` would have caught
this particular mutation too, but only by restating what this side
already believes; the helper's parser is the actual authority and is now
what the assertion reads.

Each mutation was confirmed to have actually applied (`git diff`) — a
"survived" that is really a no-op script is not evidence.

## What I am NOT claiming

**M4 survived, and I did not paper over it.** Replacing the `A-Z` range
check with `tolower()` is *semantically identical under the C locale*,
which is the locale the test binary runs in, so the mutation is not
observable from my assertions. It is not harmless in the product:
`shottino.c:20583` calls `setlocale(LC_ALL, "")`, so under a latin-1
locale `tolower(0xC9)` is `0xE9` and `CAFÉ`/`café` would fold into one
room. Killing it would mean setting a specific locale in the test, which
is not guaranteed to be generated on the CI image — an assertion that
silently no-ops when the locale is missing is exactly the non-constraining
kind #762 is about, so I did not add one. The byte-level output IS
pinned; the locale-dependence is not.

**A latent truncation collision, measured, not fixed.** Every
`call_path_nick` call site encodes into a `char pn[128]` and the encoder
can triple its input, so a nick over 42 bytes truncates. Measured: two
distinct 61-character CJK nicks produce the *same* 126-byte path — two
people, one room, which is the failure mode the function exists to
prevent. Not reachable on the ircds this talks to (bahamut and solanum
both cap NICKLEN well under 42 and both keep nicks ASCII), so I am
calling it latent rather than live, and there is now an assertion that
pins the 42-byte headroom and fails if either number moves. No product
change — we are under a product-surface freeze and this is not reachable.

**A directory named `shottino-call` is accepted by `call_helper_accept`**
(measured: `access(X_OK)` on a directory returns 0). The consequence is
`execv` failing into `_exit(127)` while the parent believes a call
started, instead of the browser fallback. The function never claims to
reject directories — its comment names them only as the reason the
truncation fit-check matters — so this is reported, not fixed.

**`call_frame_box` can under-size the corner PiP** on terminals of 9
rows or fewer (at `rows_total=6` the frame floors at 6 rows while
`call_video_box` asks for up to 15). I did not assert a frame ≥ video
invariant, because it is false at that boundary and gerrymandering the
range to make it pass would be an assertion that constrains nothing. A
15-row PiP cannot be drawn on a 6-row terminal anyway; not a defect
claim.

## Why a new file rather than more of `test_windows.c`

`test_windows.c` is being edited by two other queued changes. This is
also the honest home for it: the identity-and-exec family, not window
bookkeeping. The pre-existing call tests (invite build/split, peer list,
ring policy) stay where they are; the new suite's moduledoc says so, so
nobody has to guess which file to add the next one to.

## Gate

`make check` — 17/17 suites, rc=0, 0 failures, 0 warnings, working tree
clean before and after. `tests/test_keys` is excluded locally
(`<pty.h>`, Linux-only); CI runs all 18 on ubuntu.

Refs #764.

### DIFF
--- a/frontends/shottino/Makefile
+++ b/frontends/shottino/Makefile
@@ -47,7 +47,7 @@ OBJS := shottino.o json.o wire.o alias.o mirc.o termcolor.o http.o media.o ircd.
 # reaches, which is the drawing primitives and nothing else. Do not trust
 # this net for changes elsewhere in that file; add a test or reason about
 # the path by hand.
-TESTS := tests/test_llm tests/test_ws tests/test_json tests/test_wire tests/test_alias tests/test_mirc tests/test_termcolor tests/test_http tests/test_media tests/test_whip tests/test_callmedia tests/test_callnote tests/test_layout tests/test_keys tests/test_ircd tests/test_commands tests/test_windows
+TESTS := tests/test_llm tests/test_ws tests/test_json tests/test_wire tests/test_alias tests/test_mirc tests/test_termcolor tests/test_http tests/test_media tests/test_whip tests/test_callmedia tests/test_callnote tests/test_layout tests/test_keys tests/test_ircd tests/test_commands tests/test_windows tests/test_call
 SANITIZE ?= -fsanitize=address,undefined -fno-omit-frame-pointer
 TEST_CFLAGS := -std=c11 -Wall -Wextra -Wpedantic -O1 -g $(SANITIZE)
 
@@ -148,6 +148,14 @@ tests/test_commands: tests/test_commands.c tests/test.h shottino.c json.c wire.c
 # question every other bug in the sidebar has turned out to be.
 tests/test_windows: tests/test_windows.c tests/test.h shottino.c json.c wire.c alias.c mirc.c termcolor.c http.c media.c ircd.c ws.c llm.c
 	$(CC) $(CPPFLAGS) $(TEST_CFLAGS) $(CFLAGS_DEPS) -o $@ tests/test_windows.c json.c wire.c alias.c mirc.c termcolor.c http.c media.c ircd.c ws.c llm.c $(LDLIBS)
+
+# test_call is the only suite linked against BOTH binaries: it compiles
+# shottino.c for the call helpers and links call/media.c so the codec
+# word shottino puts in the helper's argv can be handed to the helper's
+# own parser. Asserting that word against a literal here would only say
+# what this side already believes.
+tests/test_call: tests/test_call.c tests/test.h shottino.c call/media.c call/media.h json.c wire.c alias.c mirc.c termcolor.c http.c media.c ircd.c ws.c llm.c
+	$(CC) $(CPPFLAGS) $(TEST_CFLAGS) $(CFLAGS_DEPS) -o $@ tests/test_call.c call/media.c json.c wire.c alias.c mirc.c termcolor.c http.c media.c ircd.c ws.c llm.c $(LDLIBS)
 
 tests/test_keys: tests/test_keys.c tests/test.h shottino.c json.c wire.c alias.c mirc.c termcolor.c http.c media.c ircd.c ws.c llm.c
 	$(CC) $(CPPFLAGS) $(TEST_CFLAGS) $(CFLAGS_DEPS) -o $@ tests/test_keys.c json.c wire.c alias.c mirc.c termcolor.c http.c media.c ircd.c ws.c llm.c $(LDLIBS) -lutil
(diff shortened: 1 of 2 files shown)


vjt:

Issue #863, opened 2026-08-05
cicchetto: a whitespace-only line aborts a multi-line paste with `body: can't be blank`


vjt:
## Symptom

Pasting a multi-line block into the composer can abort mid-paste with

```
Please fix: body: can't be blank.
```

and the unsent remainder is pushed back into the draft, so the paste looks
"scrambled". Reported independently by two users.

## Cause (measured on `origin/main`, not guessed)

Client and server do not agree on what "empty line" means.

**Client** — `cicchetto/src/lib/messageLines.ts`:

```ts
export const splitMessageLines = (body: string): string[] =>
  body.split(/\r\n|\r|\n/).filter((line) => line !== "");
```

The filter is a strict `!== ""`, and the module comment states the intent
explicitly: *"A whitespace-only line is kept: it is content on the wire, not
empty."* So a truly empty line is dropped, but a line of spaces/tabs — the
common shape of a "blank" line inside pasted code, logs or quoted text — is
kept and sent as its own PRIVMSG by `sendBodyLines` (`compose.ts`).

**Server** — the whitespace-only body clears every early gate:

* `messages_controller.ex` `create/2` guard is `body != ""` → a `" "` body passes
  (an actually empty body would land on the `:bad_request` clause and produce a
  different error).
* `BodyLimit.check/1`, `validate_post_target_name/1`, the send-token bucket: all pass.
* `Session.send_privmsg/4` only checks CRLF/NUL (`Identifier.safe_line_token?/1`): passes.

It then dies at persist time. `session/server.ex` `persist_and_send_fragments/5`
persists **before** relaying upstream:

```elixir
with {:ok, message} <- Persistor.persist_and_broadcast(attrs, state, push: false),
     :ok <- send_privmsg_or_log(state.client, target, fragment) do
```

and `Grappa.Scrollback.Message.changeset/2` runs `validate_required([:body])`
for `:privmsg` (`validate_body_for_kind/1`). Ecto's `cast/4` trims string params
before the empty-value test —

```elixir
# deps/ecto/lib/ecto/changeset.ex:760,774
trim_values = Keyword.get(opts, :trim_values, &Ecto.Type.trim/2)
filter_values = fn type, value -> trim_values.(type, value) in empty_values end
```

with `@empty_values [""]` — so `" "` trims to `""`, is treated as an absent
param, the change is never applied, `:body` stays `nil`, and
`validate_required` adds `can't be blank`. That changeset error is rendered as
[...]

PR #869, opened 2026-08-05
fix(#863): a blank-looking line no longer aborts a multi-line paste

vjt:
Refs #863. cic-only — the server is not touched.

## The bug

Two users pasted a block with an indented blank line in it and watched the
paste stop halfway, with the unsent remainder shoved back into the draft and
`Please fix: body: can't be blank.` on screen.

Client and server did not agree on what "empty" means, and nothing connected
the two predicates:

- `splitMessageLines` filtered on a strict `!== ""`, keeping a whitespace-only
  line on the reasoning — stated in the module comment — that it is "content on
  the wire, not empty".
- The server accepts a body only if it is non-empty **after trimming**: Ecto's
  `cast/4` trims before its empty-value test, so `" "` arrives as an absent
  param, `:body` stays nil and `validate_required` fires.

So cic was sending a line the door would always refuse. `sendBodyLines` treats
a non-429 as fatal, so the refusal killed the rest of the paste.

## The fix

The client half moves onto the server's notion: the splitter drops a line that
trims to nothing. Whitespace that *surrounds* content is untouched —
indentation is content, and a pasted code block keeps its shape.

The direction was vjt's call between two viable ones (client drops vs server
accepts). The cost of this one was measured before it was chosen: it forces the
existing `preserves a whitespace-only line` case to be reversed.

## Two commits, on purpose

1. **The red tests, before any cure.** They assert only the constraint that
   holds whichever direction wins — a whitespace-only line must never abort the
   lines after it — plus a cross-stack pin that the two predicates agree.
2. **The cure.**

## What changed in the existing test, and why not delete it

`preserves a whitespace-only line` asserted the old intent. It is **inverted,
not removed**: it now reads `drops a whitespace-only line — the server counts
it as blank, so it is not content`, and carries a comment saying it used to
assert the opposite and why that was reversed. Deleting it would erase the fact
that the intent was reversed; a reader six months from now needs to find the
reversal, not its absence.

Two cases join it: a tab-only line (what a code paste actually produces) and an
indented line with content, which pins that the trim decides *emptiness* rather
than stripping content — a guard against the over-fix, not against the bug.

The cross-stack pin is left as written. It does not encode which answer won; it
asserts the two predicates AGREE, so it keeps holding if the server's half ever
moves. `serverBodyPredicate.ts` stays the declared place to update when it does.

## Evidence

- Before the cure: `4 failed | 167 passed (171)` on the two suites. The test
  that counts failed with
  `expected [ 'def f():' ] to deeply equal [ 'def f():', '    return 1' ]` —
  the second line never attempted.
- After: `173 passed (173)` on those suites; full vitest run **235 files, 4266
  tests, all green**; `tsc --noEmit` clean; biome clean on the touched files.
- **Mutation measured, not assumed.** Committing first, then restoring the old
  `!== ""` filter, turns **6 tests red** — the original 4 plus the 2 new ones.
  The indented-line case stays green under that mutation, correctly: it guards
  the over-fix, not the bug. Working tree restored and re-verified green.
- Both directions were probed while the tests were still red, to confirm they
  were not written to only one outcome: modelling the server as accepting
  whitespace made them pass too.

## Not measured

**No ExUnit was run — no COMPILE lane on this worker.** The server's behaviour
is MODELLED in `cicchetto/src/__tests__/serverBodyPredicate.ts`, from the
measurement in the issue and from reading the code; the file states each step
it was derived from. If that model is wrong, this fix is aligned to the wrong
thing.

One honest limit, written in the file rather than hidden: production now trims
with JavaScript's `trim()` while the server trims with Elixir's `String.trim/1`.
They agree across ASCII spaces and tabs — the samples stay in that range — but
not on exotic whitespace (U+FEFF is stripped by one and not the other). That
residual disagreement is narrower than the one this closes, and it is now
documented instead of silent.

### DIFF
--- a/cicchetto/src/__tests__/compose.test.ts
+++ b/cicchetto/src/__tests__/compose.test.ts
@@ -1,6 +1,7 @@
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { channelKey } from "../lib/channelKey";
 import { LIST_WINDOW_NAME } from "../lib/windowKinds";
+import { BLANK_BODY_ERROR, serverAcceptsBody } from "./serverBodyPredicate";
 
 vi.mock("../lib/api", () => {
   class ApiError extends Error {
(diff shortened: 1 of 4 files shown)


vjt:

Issue #837, opened 2026-08-04
cic: identityMoved is inlined in four modules — extract one predicate


vjt:
Recommended by w1 while closing **#818**, and deliberately **not** done there: it is a cross-module refactor, and #818's perimeter was a bug fix.

## What

The `identityMoved`-style guard — "the identity that started this continuation is not the identity that will receive its result" — is now written inline in **four** modules:

- `cicchetto/src/lib/scrollback.ts` (#788)
- `cicchetto/src/lib/displayPrefs.ts`
- `cicchetto/src/lib/customTheme.ts`
- `cicchetto/src/lib/networks.ts` (#818)

Four copies of one predicate is the shape that drifts: a fifth `await` site gets added, someone copies whichever spelling is nearest, and one of them eventually diverges in a way no test names.

## Why it is filed rather than queued

This is exactly the kind of change the current freeze is meant to hold back — it touches four modules and buys no user-visible behaviour. It is also the kind CLAUDE.md wants (*"implement once, reuse everywhere"*), and vjt's own framing — *"non stratificare, estendere e migliorare"* — reads as favouring the consolidation over a fifth copy.

**So: a decision about timing, not about merit.** No `status:*` label.

## When it is picked up

- One predicate, one owner; the four call sites adopt it unchanged.
- **Mutation-prove it**: removing the shared guard must turn the existing #788 and #818 tests red — those already pin the behaviour at two of the four sites, so the refactor has a real net under it.
- Worth checking while there whether the other two sites (`displayPrefs`, `customTheme`) have any test pinning their guard at all; if not, that gap is the more valuable half of this issue.

vjt:
Shipped in **v0.12.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.12.0

Pull that tag to get this. Note it is a tag-only release: irc.sindro.me is deliberately held on the previous release while resource usage is measured over several days, so "shipped" here means "in a named release", not "running on our instance".

PR #870, opened 2026-08-05
cic(#837): one identityMoved predicate, and the two guards nobody was holding

vjt:
Picks up #837. The refactor is the smaller half; the missing tests are the deliverable the issue predicted.

## What was actually missing

#788 (`scrollback.ts`) and #818 (`networks.ts`) each shipped a test for their own mid-flight guard. The other two copies — `mountDisplayPrefsSync` and `mountCustomThemeSync` — had **none**. Deleting `if (token() !== t) return;` from either module left the entire cic suite green.

So consolidating first would have moved untested code from four places to one and protected nothing: the single owner would have been just as free to be dropped, now in one edit instead of four. Tests first, then the extraction.

## Commits

1. **`a0d93f4` — pin the two unheld guards.** New rotation cases for `displayPrefs` and `customTheme`, each with a no-rotation control that releases the SAME held response (a gate that stopped delivering it fails loudly instead of passing vacuously).
2. **`c6d50d7` — give the predicate one owner.** New `cicchetto/src/lib/identityMoved.ts`; all four sites adopt it **unchanged**.
3. **`02bc316` — DESIGN_NOTES.**

## Mutation numbers

Per-site, before the extraction:

| site | red | green | pre-existing cases moved |
|---|---|---|---|
| `displayPrefs` guard removed | 2 | 18 | **0 of 17** |
| `customTheme` guard removed | 1 | 36 | **0** |

Shared predicate neutered after the extraction — 9 red across exactly the four adopters and nothing else:

| suite | red |
|---|---|
| `scrollbackIdentityRotation` (#788) | 5 |
| `networks` (#818) | 1 |
| `displayPrefs` (new) | 2 |
| `customTheme` (new) | 1 |

Restored: **235 files / 4263 tests green**, `bun run check` exit 0.

## Two design points

**Own module, not `auth.ts`.** `auth.ts` owns `token`, so it looked like the owner — but 38 test files stub `../auth`, and a partial stub would replace the guard with `undefined`: the rule under test would stop existing wherever a suite meant to control only the token. A separate module importing `token` keeps the predicate real while still resolving through those mocks. Confirmed empirically — under the mutation the #788 suite, which mocks `../auth` with its own Solid signal, goes red like the rest.

**Not folded into `identityScopedStore`.** That factory owns the identity TRANSITION (fire the resets); this owns a verb's OWN captured identity, which the factory never sees. Complements, not duplicates.

## What the two new guards hold, beyond the 401

- `displayPrefs` — A's prefs applied over B's, and on the `persisted: false` seed-up branch a PUT under A's retired bearer (#281's fail2ban harm class).
- `customTheme` — `applyResolvedPair` paints `documentElement` **and** writes the FOUC boot cache, so A's theme lands on B's screen and survives the reload that would otherwise correct it.

Neither is reachable by identity-transition cleanup: an A→B rotation never runs the logout-clear branch the existing cross-account-bleed tests cover, and nothing cancels a request already on the wire.

## What I will not claim

- **No behaviour change is asserted beyond what the suite measures.** The four adopters are byte-equivalent in logic (`token() !== t`); no browser or e2e run was done, because nothing user-visible changes and nothing outside vitest was touched.
- **The guard is still a per-verb habit, not a whole-module property.** A new `await` that captures the token and skips the check reopens the defect for its own call path. One owner makes the rule greppable, not automatic.
- **Only the four sites named in the issue were audited.** I did not sweep the codebase for other capture-then-await patterns that ought to adopt it.

## Test plan

- [x] `bun run test` — 235 files / 4263 tests green (baseline 4258 + 5 new)
- [x] `bun run check` — exit 0 (biome + tsc)
- [x] `bun run build` — tsc --noEmit + vite build clean
- [x] Mutation, per site and shared — numbers above
- [ ] cic-only change: no lane, no server suite, no e2e

### DIFF
--- a/cicchetto/src/__tests__/customTheme.test.ts
+++ b/cicchetto/src/__tests__/customTheme.test.ts
@@ -1,13 +1,18 @@
-import { beforeEach, describe, expect, it } from "vitest";
+import { createRoot } from "solid-js";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { setToken } from "../lib/auth";
 import {
+  activePair,
   applyCustomTheme,
   COLOR_KEYS,
   getAppliedThemePayload,
+  mountCustomThemeSync,
   setThemePreviewMode,
   tokenToCssVars,
 } from "../lib/customTheme";
 import { EDITOR_BASE_KEYS, EDITOR_MODE_KEYS, EDITOR_NICK_KEYS } from "../lib/themeEditor";
 import type { TokenColors, TokenPayload } from "../lib/themesApi";
+import type { ThemesWireT } from "../lib/wireTypes";
 
 // #75 producer path — apply-engine seams the editor depends on.
 //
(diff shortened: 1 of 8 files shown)


vjt:

Issue #868, opened 2026-08-05
cicchetto: re-wire the live notification path through the shared predicate (dead shouldNotify + port divergence)


vjt:
Prerequisite refactor for #866, split out per vjt's ruling on that issue's Q6/Q7:
*"fixa ovviamente e va bene issue separata fatta PRIMA. prima refactor."*

Surfaced by the #866 analysis (comment 5191514513) and **independently verified against
`origin/main`** by the orchestrator before filing. Three defects, one root: the client's
notification predicate is not the code that decides anything.

## 1. `shouldNotify` is dead code

`cicchetto/src/lib/pushTriggers.ts` exports `shouldNotify`. Its **only** import in the whole
`cicchetto/src` tree is its own test:

```
cicchetto/src/__tests__/pushTriggers.test.ts:2:import { type ShouldNotifyMessage, shouldNotify } from "../lib/pushTriggers";
```

Already on the record in `docs/DESIGN_NOTES.md` (#412), never acted on. Consequence: any rule
added to that predicate is green in CI and **inert in the browser**.

## 2. The two ports are not mirrors, and the shared fixture cannot see the gap

`lib/grappa/push/triggers.ex` has a branch the client port does not have at all:

```
triggers.ex:213:      own_row?(message, own_nick) -> false
triggers.ex:247:  defp own_row?(%Message{sender: sender}, own_nick) when is_binary(sender) do
triggers.ex:251:  defp own_row?(_, _), do: false
```

(the #532 C step). `grep -i 'own_row|ownRow' cicchetto/src/lib/pushTriggers.ts` → **no match**.

`cicchetto/src/lib/shouldNotifyTruthTable.json` holds **18 rows and none exercises that step**,
so `pushTriggers.test.ts` + `test/grappa/push/should_notify_parity_test.exs` are **both green over
a divergent predicate**. The parity gate is currently an assertion about agreement it does not test.

## 3. The live path reads no preference at all

What actually beeps and moves the tab title is `cicchetto/src/lib/subscribe.ts`
(`:311-341`, `:703-705`, `:740`), and it runs on a bare `matchesWatchlist` — no `NotificationPrefs`
anywhere on that path. Measured consequences:

- `channel_mentions: false` still beeps.
- an `action` fires push but does not move the title.
- a NOTICE DM beeps outside the server predicate entirely.

## What this issue is

Re-wire the live path through the (fixed) shared predicate, rather than delete the dead one:

[...]

Issue #866, opened 2026-08-05
cicchetto: per-conversation notification mute (per tab/query, with snooze) — every pref today is an allow-list


vjt:
## Request

Per-conversation notification control in cicchetto, the way Telegram does it: mute a
specific tab/query, either permanently or for a while (snooze: 1h, 8h, until tomorrow),
without touching the global preferences.

Filed for analysis, not as a settled design.

## Where we are today (measured on `origin/main`)

`NotificationPrefs` (`cicchetto/src/lib/userSettings.ts:21`) is:

```ts
channel_messages_all: boolean;
channel_messages_only: string[];
channel_mentions: boolean;
private_messages_all: boolean;
private_messages_only: string[];
```

with defaults `channel_messages_all: false`, `channel_mentions: true`,
`private_messages_all: true` (`userSettings.ts:33`).

The predicate lives in two mirrored ports — `cicchetto/src/lib/pushTriggers.ts`
(`shouldNotify`) and `lib/grappa/push/triggers.ex` (`should_notify?/4`) — kept in parity by
the shared fixture `cicchetto/src/lib/shouldNotifyTruthTable.json`, consumed by both
`pushTriggers.test.ts` and `test/grappa/push/should_notify_parity_test.exs`.

Channel branch (`pushTriggers.ts:93`):

```
channel_messages_all
  || channel_messages_only.includes(canonicalChannel(channel))
  || (channel_mentions && matchesWatchlist(body, ownNick, patterns))
```

DM branch (`pushTriggers.ts:80`):

```
private_messages_all || private_messages_only.includes(asciiFold(sender))
```

## The gap

1. **Every list is an allow-list; there is no deny-list.** `channel_messages_only` and
   `private_messages_only` can only *add* notifications. Nothing can subtract one.

2. **Muting a single noisy channel is therefore impossible.** Even with
   `channel_messages_all: false` and the channel absent from `channel_messages_only`, the
   third disjunct still fires: `channel_mentions` is a **global** boolean, so a highlight in
   that channel notifies anyway. The only way to silence it is to turn mentions off
   *everywhere*.

3. **Muting a single DM is impossible for the same reason.** `private_messages_all`
   defaults to `true`; one chatty correspondent can only be silenced by switching every DM
   to whitelist mode and re-listing everyone else.

4. **Nothing is time-boxed.** The prefs have no temporal dimension at all — no snooze,
   no "until", no expiry sweep.

## Design questions for analysis

[...]

PR #871, opened 2026-08-05
cic(#868): re-wire the live notify path through the shared predicate

vjt:
Prerequisite refactor for #866, per vjt's ruling on that issue's Q6/Q7
("fixa ovviamente e va bene issue separata fatta PRIMA. prima refactor" /
"ricabla"). Refs #868.

## The root

The client's notification predicate was not the code that decided anything.

`pushTriggers.shouldNotify` — the mirror of `Grappa.Push.Triggers.should_notify?/4`,
pinned by a shared truth-table fixture and two suites — had **no production
caller**. What actually beeped and moved `document.title` was `subscribe.ts`,
calling `matchesWatchlist` directly and reading **no preference at all**.

## 1. The two ports were not mirrors, and the fixture could not see it

The server has a branch the client did not: `own_row?` (#532 C), the
sender-identity gate that runs *before* the window-shape test. The client also
compared `message.channel === ownNick` **raw**, while the server folds both
sides — and `channel` is folded at the persist boundary (#537), so every
operator with an uppercase letter in their nick had inbound DMs routed into the
channel branch.

The 18 existing rows exercised the kind gate, the DM branch and the channel
branch, and nothing else. **Nine rows added; seven failed on the client and
passed on the server** before the fix. That asymmetry is the measurement.
Two of the nine are negative controls (a near-miss sender `vjtx`, `#vjt` as a
real channel) so a too-broad fix cannot buy itself green.

## 2. The fold was unguarded — found by mutation, not by reading

Swapping `canonical_target/1` → `String.downcase/1` (server) and
`asciiFold` → `toLowerCase()` (client) reddened **zero** rows. Every identifier
in the table was plain ASCII, where the two folds agree. The 2026-07-20
architecture review named this hole on the bracket axis; brackets survive
`toLowerCase` untouched, so those rows could never have caught it either. Two
non-ASCII rows (`#CAFÉ` vs `#café`, `CAFÉ` vs `café`) close it.

## 3. The live path now decides with the predicate

Measured behaviour changes:

| case | before | after |
|---|---|---|
| `channel_mentions: false` + mention | beeped, no push | silent, no push |
| `private_messages_all: false` + DM | beeped, no push | silent, no push |
| `channel_messages_all: true` | silent, push fired | beeps, push fired |
| `channel_messages_only` hit | silent, push fired | beeps, push fired |
| `action` mentioning own nick | no title bump, push fired | title bumps, push fired |
| DM whose body mentions own nick | **beeped twice** | beeps once |
| peer NOTICE DM | beeped, no push | silent, no push |

The double beep was not inferred: the assertion read *"expected 1 call, got 2"*.
The DM listener decided at its own call site and `routeMessage` decided again
over the same row; the pre-existing DM-beep test could not see it because its
body carried no mention.

`notificationPrefs.ts` is the signal DESIGN_NOTES asked for on 2026-06-21
("cic has no global notification-prefs signal to feed the full shouldNotify at
message-arrival"). Same shape as `highlightList.ts` / `aliasList.ts`: no
broadcast exists for these, so it hydrates on every user-topic (re)join and the
settings drawer feeds it the server's normalized echo. It defaults to the
**server's own defaults**, so a failed hydrate degrades to "unconfigured",
never to silence.

## One deliberate subtraction — flag for review

**A peer NOTICE no longer beeps.** UX-6-L gave it PRIVMSG's audio policy as
inbound DM-shaped traffic, but NOTICE sits outside `notify_kinds` server-side,
so the desktop was beeping on a row the phone was silent about. The window
still takes the unread. Making it audible again is a *preference*, not a
hard-coded exception — which is #866's territory. One line to reverse.

## Mutation evidence

Every branch deleted in turn; the rows that went red recorded.

* **client port, 10/10 branches guarded** — own_row 5 red, DM fold 2, kind gate 2,
  `channel_messages_all` 1, channel whitelist 2, mentions 4, `private_messages_all` 2,
  private whitelist 3, channel fold 1, sender fold 1
* **server port, 10/10 branches guarded** — identical row counts, so the fixture
  is a two-way gate, not a one-way assertion
* **re-wire, 7/7 behaviours guarded** — prefs-read 3 red, kind gate 1,
  single-decision-site 4, store default 1, hydrate 1, mirror 2, no-session guard 1

Two harness findings worth naming: the store's "starts at the server defaults"
test reddened **nothing** at first (`beforeEach` mirrored a map into the shared
signal, so it asserted the mirror, not the default — fixed by re-importing the
module per test), and one "0 red" reading was my own parser bug, caught by
re-running that mutation by hand rather than filing it.

## Scope

Decides none of #866's Q1–Q5 — data shape, snooze, mute semantics and UI stay
there. No wire change: no new field, no `protocol_version` bump. Server
untouched; the client moved to it.

## Verification

* `scripts/check.sh` exit 0 — Credo `5007 mods/funs, found no issues`;
  ExUnit `8 doctests, 50 properties, 5145 tests, 0 failures`; Dialyzer
  `Total errors: 0`; bats 295 ok. Tree clean before and after.
* `scripts/bun.sh run test` — 236 files, 4283 tests, 0 failures (baseline 235 / 4258).
* `tsc --noEmit` exit 0, `biome check src` exit 0 (48 pre-existing warnings, unchanged).
* Re-verified after rebasing onto current `main`.

**Not verified:** no e2e ran (no STACK lane), and no real browser — the beep
and `document.title` claims are jsdom-level. `playBeep` is a mock here; that a
tone is actually audible, and that the OS/PWA badge follows, are unproven by
this PR.

### DIFF
--- a/cicchetto/src/SettingsDrawer.tsx
+++ b/cicchetto/src/SettingsDrawer.tsx
@@ -20,6 +20,7 @@ import { type FontSizeKey, getFontSize, setFontSize } from "./lib/fontSize";
 import { friendlyApiError } from "./lib/friendlyApiError";
 import { detach, quit, updateIdentity } from "./lib/lifecycle";
 import { isAdmin, networks, user } from "./lib/networks";
+import { mirrorNotificationPrefs } from "./lib/notificationPrefs";
 import { popOverlay, pushOverlay } from "./lib/overlayScrollLock";
 import {
   deletePushSubscription,
@@ -359,6 +360,10 @@ const SettingsDrawer: Component<Props> = (props) => {
     try {
       const loaded = await getNotificationPrefs(t);
       setPrefs(loaded);
+      // #868 — feed the live notify path the same authoritative map the form
+      // renders, so the beep obeys a pref the moment it is read, not on the
+      // next user-topic rejoin.
+      mirrorNotificationPrefs(loaded);
       setChannelsOnlyText(loaded.channel_messages_only.join(", "));
       setNicksOnlyText(loaded.private_messages_only.join(", "));
     } catch {
@@ -465,6 +470,9 @@ const SettingsDrawer: Component<Props> = (props) => {
     try {
       const saved = await putNotificationPrefs(t, next);
       setPrefs(saved);
+      // #868 — mirror the server's NORMALIZED echo (not `next`): the whitelists
+      // come back folded, which is the form the predicate compares against.
+      mirrorNotificationPrefs(saved);
     } catch (err) {
       const code = err instanceof Error ? err.message : "save_failed";
       setPrefsError(code);
--- a/cicchetto/src/__tests__/focus-rule.test.ts
+++ b/cicchetto/src/__tests__/focus-rule.test.ts
@@ -37,6 +37,11 @@ vi.mock("../lib/api", () => ({
   // transitively) needs the classifier in its api mock.
   isContentKind: (k: string) => k === "privmsg" || k === "notice" || k === "action",
   isPresenceKind: (k: string) => !(k === "privmsg" || k === "notice" || k === "action"),
+  // #868 — this file drives real messages through subscribe.ts, which now
+  // decides the beep via `pushTriggers.shouldNotify` and so reads the
+  // notify-worthy kind set from api.ts. Same mirror the classifiers above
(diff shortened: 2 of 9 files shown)


vjt:

Issue #441, opened 2026-07-26
infra/linux has no CI gate at all: add bats coverage + a shellcheck job


vjt:
Approved by vjt in #grappa ("va bene ad aggiungere nuovi gate"), filed after merging #436/#437 as agreed.

Merging those two PRs surfaced that **this path has no automated gate whatsoever**. Concretely, verified on `main`:

- `integration.yml` is path-filtered on `lib/**`, `cicchetto/**`, `config/**`, `priv/**`, `mix.*` plus three orchestration scripts. A patch touching only `infra/linux/*.sh` **does not run it at all**.
- `ci.yml` runs the Elixir suite plus bats over `test/bin/` and `test/scripts/`. Neither directory touches `infra/linux`. Note also that `test/infra/` — which does hold script tests — is **not** in `ci.yml`; only `scripts/bats.sh` runs it locally.
- There is **no shellcheck gate anywhere in the repo**. The scripts carry inline `# shellcheck disable=` directives, so they are written to be shellcheck-clean, but nothing verifies it.

Net effect: a green CI on an installer-only change proves that Elixir still compiles, which is the one thing such a change cannot break. For #436/#437 the real evidence was a human running the installer on Debian 13, plus hand-run `shellcheck` and `bash -n`. That does not scale and does not survive the next contributor.

Proposed:

1. A **shellcheck job** covering `infra/`, `scripts/`, and `bin/`, honouring the existing inline directives (they encode real intent — e.g. `install.sh:51` deliberately single-quotes so `$HOME`/`$PATH` expand later inside the `run_as_grappa` heredoc, not at definition).
2. **bats coverage for `infra/linux`**, and add `test/infra/` to the `ci.yml` bats discovery set so it matches `scripts/bats.sh` rather than silently diverging from it.
3. Decide whether these run unconditionally or under a path filter that actually includes `infra/**` — the current filter is why the gap exists.

Filed by the orchestrator.

vjt:
Shipped in **v0.12.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.12.0

Pull that tag to get this. Note it is a tag-only release: irc.sindro.me is deliberately held on the previous release while resource usage is measured over several days, so "shipped" here means "in a named release", not "running on our instance".

PR #874, opened 2026-08-05
infra(#441): derive the shellcheck set, and give infra/linux its first tests

vjt:
Closes the gap in #441: `infra/linux` had no automated gate at all, and what gate existed elsewhere was a hand-written list.

## Coverage, in counted files

| | before | after |
|---|---|---|
| shellcheck | **11** paths named by hand in `ci.yml` | **71** derived |
| `bin/` | not linted at all | 2 files |
| `infra/linux/install.sh` | never on the list | linted **and** 8 bats cases |
| bats | 301 | 316 |

The 60 newly-linted files include every `infra/linux/install_*.sh`, all 15 `infra/freebsd/jail_*.sh`, the whole of `infra/packaging/`, and the four extensionless scripts a `*.sh` glob cannot see (`bin/grappa`, `grappa-source-alias`, the two `rc.d` services).

## Derived, not listed

`scripts/shellcheck.sh` enumerates: a regular file under `bin/`, `infra/`, `scripts/` that ends in `.sh` **or** opens with a shell shebang. `--list` prints the set, so the coverage claim above is inspectable rather than asserted.

No dialect table — every script already declares its shell (shebang, or `# shellcheck shell=` on the two sourced libs). A table here would be a second place for that to drift, i.e. the removed defect one level down.

**The `dash -n` list stays hand-written**, in its own step, labelled as such. "Which files must parse under dash" is a claim about where a file *runs*, not about what it is; deriving it honestly is separate work and I did not smuggle it in.

## Old defects the gate found — both real, both in their own commit

**1. `bin/grappa` (`8d213db`)** — two dead `local`s in `help_top`. The only genuine lint defect of the 22 initial failures.

**2. `infra/linux/install.sh` (`8d26b08`) — the serious one, and it came from the bats, not the linter.** The file's comments say the 2026-07-22 incident was closed with *"fail loud … rather than let a blank secret slip into the env file."* It fails loud **and writes the blank**:

- every generator failing → `SECRET_KEY_BASE=`, `SECRET_SIGNING_SALT=`, `GRAPPA_ENCRYPTION_KEY=` all blank;
- **only `gen_encryption_key` failing → exit 0.** Unit started, healthcheck passed, `grappa is up and healthy` printed, Cloak vault keyed on the empty string.

Mechanism: `set_env_if_blank KEY "$(gen ...)"` puts the failure in *argument* position — `gen_raw`'s `exit 1` exits the subshell, the enclosing command's status is `set_env_if_blank`'s, so neither `set -e` nor `pipefail` sees it; the redundant second `| tail -n1` made even the substitution's status 0. The sole reason the first case aborted is VAPID's empty-check, the one empty-test outside a substitution. Fixed with capture-CHECK-write at all five generator sites — the shape the VAPID branch already used.

## What the other 21 failures were (none silently excluded)

- **15** missing the `# shellcheck source=scripts/_lib.sh` directive `scripts/deploy.sh` already carried. Unfollowed, shellcheck also cannot see the lib's `set -euo pipefail`, so it flagged every `cd "$REPO_ROOT"` (SC2164) — one directive cleared both.
- **6** deliberate idioms, annotated **at the site with the reason**, never dropped from the set: the `su -c '...'` bodies whose single quotes are the point; literal backticks in a printf'd help string; `CDPATH= cd`; and — file-wide on the two `rc.d` services — the four codes that *are* the rc.subr contract (`/etc/rc.subr` absent on a linter host, `*_cmd`/`rcvar` read by rc.subr, `: ${var:="x"}`, `local` in ash). Scoped to those codes; everything else in those files stays linted.

Happy to convert any of the rc.d mutes into per-line annotations or a fix if you'd rather — that block is the only judgement call here.

## Shellcheck version

Measured under **shellcheck 0.10.0**, digest-pinned (`koalaman/shellcheck:v0.10.0@sha256:2097951f…`) and used by CI too, so the runner's preinstalled binary no longer decides what "clean" means. Same posture as the #103 base-image pins; refresh recipe in the script header.

## Tests

`test/infra/install_linux_test.bats` (8) — every assertion is about the env file's **contents or mode**, never call order. Six characterize behaviour already correct (PHX_HOST required; the operator's host overwriting the template's `grappa.example.org`, the 2026-07-22 defect; the four-line VAPID generator landing whole, the 2026-07-24 one; the final **0640** after the `tmp+mv` rewrite that regressed twice; re-run keeps secrets, re-applies config). Two pin the fix — **red against the pre-fix script, green after**.

`test/infra/shellcheck_gate_test.bats` (7) — the derivation rule in a sandbox tree (a new extensionless shebang script joins; a `.sh` with no shebang joins; `nginx.conf`/`.pl`/`PKGBUILD` stay out; an empty set fails loud; the gate lints itself), plus the anti-regression: `ci.yml` must invoke the script and name no shellcheck paths. Proved non-vacuous by re-adding such a line — the case goes red.

## What I will not claim

- **`install.sh` was never executed end-to-end against a real host.** Every delegate, `sudo`, `mix`, `openssl`, `systemctl` and `curl` is stubbed. What only a real Debian install exercises — the toolchain build, systemd `Type=exec`, nginx — is still covered by nothing but a human.
- **`chown` is stubbed** (the suite is not root), so the *ownership* half of `root:grappa` is unverified. `chmod` is deliberately real, so the **mode** assertion is a genuine `stat`.
- **CI's shellcheck version was not measured, it was replaced.** I never established which version `ubuntu-latest` shipped; pinning makes the question moot rather than answering it.
- **The 60 newly-covered files are lint-clean, not correct.** shellcheck found one real defect in them; it is not a proof of anything else.
- **The rc.d mutes are a judgement call**, not a measurement. I read the four codes as rc.subr contract and wrote the reason down; I did not run those services.

## Test plan

- [x] `scripts/shellcheck.sh` — exit 0 over 71 derived files (shellcheck 0.10.0, pinned)
- [x] `scripts/bats.sh` (full set) — exit 0, **316 ok / 0 not ok** (was 301)
- [x] Mutation: 2 install.sh cases red against the pre-fix script, green after
- [x] Mutation: the `ci.yml` anti-regression case red when a bare `shellcheck <path>` line is re-added
- [ ] No Elixir gate run — no `lib/` change, and `scripts/check.sh` would take the COMPILE lane

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,28 +101,39 @@ jobs:
         # arg wiring the bare runner doesn't need.
         run: vendor/bats-core/bin/bats test/bin/ test/infra/ test/scripts/
 
-      - name: Shellcheck + POSIX-lint deploy scripts (#503 anti-drift)
-        # The shared deploy lib (infra/lib/deploy_common.sh) MUST stay
-        # strict POSIX sh so dash can run it on the FreeBSD jail (/bin/sh).
-        # Lint the lib as sh, the three consumers with -x (they source the
-        # lib), and dash -n parse the POSIX pieces so a reintroduced
-        # bashism in the shared algorithm fails CI instead of silently
-        # breaking the jail deploy — the exact copy-paste-drift class #503
-        # exists to kill.
+      - name: Shellcheck every shell script under bin/ infra/ scripts/ (#441)
+        # DERIVED, not listed. The list this replaces named eleven paths by
+        # hand and had been correct exactly once: `infra/linux/install.sh` —
+        # the file whose defects filed #441 — was never on it, `bin/` was not
+        # linted at all, and the next script added would have been uncovered
+        # for the same reason. scripts/shellcheck.sh enumerates instead
+        # (`--list` prints the set), so coverage is a property of being a
+        # shell script rather than of someone remembering.
+        #
+        # The linter is a digest-pinned container, not the runner's
+        # preinstalled shellcheck: findings change between minors, so a
+        # laptop and whatever ubuntu-latest ships that month must not
+        # disagree about what "clean" means.
+        run: scripts/shellcheck.sh
+
+      - name: POSIX-parse the shared deploy lib + its sh consumers (#503)
+        # A DIFFERENT property from the lint above, with a different tool:
+        # the shared deploy lib (infra/lib/deploy_common.sh) must stay strict
+        # POSIX sh so dash can run it as the FreeBSD jail's /bin/sh. dash -n
+        # parses; shellcheck lints. A reintroduced bashism in the shared
+        # algorithm fails here instead of silently breaking the jail deploy —
+        # the copy-paste-drift class #503 exists to kill.
+        #
+        # Still a hand-written list, deliberately: #441 derived the SHELLCHECK
+        # set, and this one is not the same question — "which files must parse
+        # under dash" is a claim about where they run, and the honest
+        # derivation for it is a separate piece of work.
         run: |
-          shellcheck -s sh infra/lib/deploy_common.sh infra/docker/assert-abi-lockstep.sh infra/docker/release-entrypoint.sh infra/docker/get.sh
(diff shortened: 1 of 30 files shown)


vjt:

Issue #766, opened 2026-08-03
A second_factor account with no TOTP can never spend its recovery codes


vjt:
## What

Recovery codes are one flat, account-level set shared by TOTP and passkey login. But only two doors redeem them, and neither serves an account in passkey `second_factor` mode with TOTP disarmed:

- `GrappaWeb.PasskeyController.recover/2` → `recover_resolved/4` matches only `%User{passkey_mode: "passwordless"}`. Any other mode falls through to the catch-all clause, which records a throttle failure and returns `:invalid_two_factor`.
- The TOTP door requires `TOTP.enabled?(user)`.

So such an account holds ten valid, armed recovery codes it has no way to use. Lose the passkey and the account is unreachable, even though the documented fallback exists and looks armed from the settings UI.

## Why now

Found while scoping the recovery-code teardown rule. That work settled on the conservative line — `second_factor` **counts** as a factor that depends on the set, so a teardown must not delete the codes while it is armed. That is the right call for lockout safety, but it sharpens the contradiction: we now deliberately preserve a credential for a mode that cannot redeem it.

## Not answered here

Whether the fix is to widen the `recover/2` door to `second_factor`, or to stop offering codes for a mode that cannot use them. The first restores the fallback the UI implies; the second is a smaller change but removes a way back in. Worth a deliberate decision rather than picking the convenient one.

vjt:
## Ruling (vjt, 2026-08-05): **clear the recovery codes when they are unusable.**

That is branch **(b)** of the "Not answered here" section: we stop keeping an armed set for a mode that
cannot redeem it, rather than widening `PasskeyController.recover/2` to `second_factor`.

**Consequence to implement deliberately, not discover later.** (b) removes a way back in, and it directly
reverses the conservative teardown line this issue was born from — that rule preserves the codes *because*
`second_factor` counts as a factor depending on them. Concretely, once codes are cleared for that state:

- an account in passkey `second_factor` mode with TOTP disarmed has **no fallback at all**; losing the
  passkey means the account is unreachable by any self-service door;
- the settings UI must **stop implying a fallback exists** for that state — today it shows ten armed codes.
  Clearing them silently while the UI still advertises them would be strictly worse than the current bug;
- the teardown rule and its rationale (recorded in `docs/DESIGN_NOTES.md`) must be updated in the same
  change, or the codebase will carry two contradictory statements about when codes may be destroyed.

[...]

PR #875, opened 2026-08-05
auth(#766): the recovery set decides the door, not the factor that is armed

vjt:
Refs #766.

## What was broken

An account in passkey `second_factor` with TOTP disarmed held ten valid,
armed recovery codes that **no door would read**:

- `PasskeyController.recover/2` matches only `%User{passkey_mode: :passwordless}`;
  everything else falls to the catch-all that records a throttle failure and
  answers `:invalid_two_factor`.
- The post-password door was offered only when `TOTP.enabled?/1`, and
  `TOTP.verify/3` rolls the exchange back with `:two_factor_not_enabled` when
  `totp_enabled_at` is nil.

The state is reachable through the supported UI path, which is what makes it a
defect: a passwordless account is the only mode that mints codes without TOTP,
and switching it to second factor **keeps** them, because
`RecoveryCodes.drop_if_orphaned/1` deliberately counts `second_factor` as
armed. `Accounts.reset_totp/1` reaches the same state from the other side.

## The fix asks about the SET, not about the factor

The recovery set is ONE flat account-level credential shared by every factor, so
a door keyed on `passkey_mode` or on `TOTP.enabled?/1` is asking a question that
was never about it.

- `Accounts.recovery_codes_armed?/1` gates the post-password challenge alongside
  `TOTP.enabled?/1` — the door exists whenever there is something to spend at it.
- `Accounts.verify_second_factor/3` spends a recovery code when TOTP is disarmed.
  It branches on `TOTP.verify/3`'s **own** not-enabled answer rather than a
  `TOTP.enabled?/1` pre-check, so a disarm racing the redemption cannot land
  between the question and the answer.

`challenge_token` is not repurposed: `/auth/totp/verify` has always spent a
recovery code as readily as a TOTP one, and cic's form is already labelled
"Authenticator or recovery code". Only the condition under which the token is
minted widens — no client change is required, and one cic test pins the degrade
on the token rather than on `totp_available` (the two have stopped being the
same fact).

## What was deliberately NOT widened

The issue proposed widening `PasskeyController.recover/2` to `second_factor`.
That door mints a session from a recovery code **alone** — the right trade for a
passwordless account (one factor in, one factor out, since password login is
refused for the mode) and the wrong one for a `second_factor` account, whose
password IS the first factor. Widening it would have closed a lockout by
demoting a two-credential account to a one-credential one, on an unauthenticated
endpoint. A test pins the refusal so a later "obvious" fix fails the suite
instead of quietly landing.

This is also the honest answer to the preferred "single `passkey_mode`-blind
redemption path": the doors differ because the FACTOR COUNT differs.

Nothing is destroyed and no teardown rule moves.

## Evidence

- **RED first, on the right rows.** The matrix failed on exactly the two
  `second_factor_without_totp` rows; the other three states were green before
  the fix.
- **Both production halves mutation-checked individually.** Reverting the
  challenge predicate alone → 2 red. Reverting the redemption branch alone → 2
  red.
- **Throttle parity measured, not asserted.** The widened branch shares the
  `:totp_login` window: the eleventh attempt is 429 *even with a genuine code*,
  and the same code then opens the door once the counter is cleared — the second
  half is what stops the 429 from being satisfied by a code that was simply
  unredeemable.
- **The guard is the class.** `recovery_code_doors_test.exs` enumerates every
  state whose login demands a second factor (passwordless, second factor with
  and without TOTP, TOTP alone) and asserts each can spend an armed code and
  that the code is consumed by doing so.
- `scripts/check.sh` exit 0 — `8 doctests, 50 properties, 5140 tests, 0
  failures`, dialyzer `Total errors: 0`, bats `done (passed successfully)`.
  cic: 4259 tests green, `tsc --noEmit` clean, biome clean.

## Reported, not fixed

`Accounts.disable_totp/3` deletes the whole recovery set unconditionally instead
of going through `drop_if_orphaned/1`, so a user who disables TOTP while passkey
`second_factor` is armed loses the fallback that rule says to keep — the mirror
image of this issue. It is a teardown change, which this issue explicitly froze,
so it is written down in DESIGN_NOTES rather than touched here.

### DIFF
--- a/cicchetto/src/__tests__/auth.test.ts
+++ b/cicchetto/src/__tests__/auth.test.ts
@@ -167,6 +167,32 @@ describe("auth signal store", () => {
       expect(auth.token()).toBeNull();
     });
 
+    // #766 — the token and `totp_available` are no longer the same fact. The
+    // server hands out a challenge whenever the account can spend SOMETHING
+    // at that door, and for an account whose only remaining fallback is its
+    // recovery set that means a token with `totp_available: false`. The
+    // degrade must key on the token, because keying on `totp_available`
+    // rethrows and leaves an account with a valid code no way to type it.
+    it("degrades on a recovery-only challenge, where TOTP is unavailable", async () => {
+      const api = await import("../lib/api");
+      const { getPasskey } = await import("../lib/passkeys");
+      vi.mocked(api.login).mockResolvedValue({
+        two_factor_required: true as const,
+        passkey_options: { challenge_id: "pk-1", public_key: { challenge: "AQID" } },
+        totp_available: false,
+        challenge_token: "challenge-123",
+      });
+      vi.mocked(getPasskey).mockRejectedValue(new Error("Passkey authentication cancelled"));
+      const auth = await import("../lib/auth");
+
+      await expect(auth.login("alice", "secret")).resolves.toEqual({
+        kind: "totp",
+        challengeToken: "challenge-123",
+        passkeyError: new Error("Passkey authentication cancelled"),
+      });
+      expect(auth.token()).toBeNull();
+    });
+
     it("degrades to the TOTP challenge when the server rejects the assertion", async () => {
       // A rejected assertion (replayed challenge, expired ceremony) is a
       // ceremony failure like any other: the account still has a second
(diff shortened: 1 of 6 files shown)


vjt:

Issue #867, opened 2026-08-05
docker run <image> start on a fresh volume dies on an unmigrated DB (no such table: admin_events) — the door after #862's


vjt:
## What

With #862's secret bootstrap in place, `docker run ghcr.io/vjt/grappa:<tag> start` on a fresh `/data` volume gets past the `SECRET_KEY_BASE` wall and then dies on an **unmigrated database**:

```
no such table: admin_events
```

Nothing runs `ecto.migrate` on the bare `docker run` path. The release image ships the migration capability (`bin/grappa eval` works — running `migrate()` and then `start` was measured to boot cleanly and serve `/healthz` → 200), but no code path invokes it when the operator simply runs the image.

This is the **second** door on the same journey, and #862 did not see it: that issue's scope ends at secrets, so fixing it alone moves the failure one step later rather than making the documented one-liner work.

## Evidence

Measured by worker-2 on 2026-08-05 while building and running a full `Dockerfile.release` image locally (not a derived probe), as part of #862:

| scenario | result |
|---|---|
| bare `docker run`, fresh volume, secrets bootstrapped | dies: `no such table: admin_events` |
| `eval migrate()` then `start` | `/healthz` → 200 |

## Why this is a design decision, not a one-liner

Auto-migrating on boot is exactly the kind of thing that must be decided rather than added:

- **It is right for the single-container self-hoster**, who has no other way to run migrations and for whom "pull the image, run it" is the whole value proposition.
- **It is wrong-by-default for a multi-instance or orchestrated deployment**, where two containers starting at once would race the migration, and where operators expect to gate schema changes on their own timeline.
- The existing deploy paths already handle this deliberately (`infra/docker/deploy.sh`, the packaging postinstall, the FreeBSD/Linux flows), so an unconditional auto-migrate in the entrypoint would duplicate a decision those paths already make — and could make it differently.

Plausible shapes, none chosen here: an opt-in env toggle (`GRAPPA_AUTO_MIGRATE=1`) defaulted on only for the bare-run path; a documented `docker run … eval 'Grappa.Release.migrate()'` step before first `start`; or an entrypoint that migrates only when it detects it created the database itself.

Whichever way it goes, the acceptance test is the same as #862's: **`docker run` on a fresh volume, following only what the docs say, must end with a serving instance** — or the docs must state the extra step plainly.

Filed from a measured run, not from reading. No label: this needs triage before it becomes work.

vjt:
[...]

Issue #862, opened 2026-08-05
docker run ghcr.io/vjt/grappa:<tag> start dies on missing SECRET_KEY_BASE and points at a script the image does not ship — bootstrap secrets in the release image with the existing gen-secrets.sh


vjt:
## What happens

```
$ docker run ghcr.io/vjt/grappa:v0.11.0 start
ERROR! Config provider Config.Reader failed with:
** (RuntimeError) environment variable SECRET_KEY_BASE is missing.
Generate one with: scripts/mix.sh phx.gen.secret

    /app/releases/0.11.0/runtime.exs:195: (file)
```

Two distinct problems in that output:

1. **The advice is wrong for this image.** `config/runtime.exs:196` tells the operator to run `scripts/mix.sh phx.gen.secret`. The release image has no `scripts/`, no `mix`, and no source tree — `Dockerfile.release` ships only `_build/prod/rel/grappa`, the SPA dist and the entrypoint. The one message a docker user sees points at something that cannot exist in their container.
2. **There is no first-boot secret bootstrap on the bare `docker run` path at all.** The image is deliberately secret-less (`Dockerfile.release`, the `ENV` block comment: *"SECRETS ARE NOT BAKED … so a bare `docker run` fails loudly until they are supplied (unit D wires the compose/one-liner env)"*), and `infra/docker/release-entrypoint.sh` only sizes the BEAM caps and `mkdir -p`s the data dirs. So the failure is by design — but the design only ever gave the operator the *deploy.sh* path, and `docker run <image>` is the first thing everybody types when they see a ghcr package.

## The wheel we already have (three of them, actually)

Secret bootstrap is implemented **three separate times** in this repo today:

| path | how | file |
|---|---|---|
| from-source Linux install | `mix phx.gen.secret` etc. | `infra/linux/install.sh` |
| `.deb` / `.rpm` | **openssl only, no BEAM** | `infra/packaging/gen-secrets.sh` (installed to `/usr/share/grappa/gen-secrets.sh`, `nfpm.yaml:152`; invoked by `infra/packaging/scripts/postinstall.sh:38`; re-runnable as `grappa gen-secrets`) |
| docker release box | host `openssl` for the 4 random secrets + `docker run … eval` in the image for VAPID | `infra/docker/deploy.sh` `write_env_file()` |

[...]

Issue #41, opened 2026-06-03
Hot-deployable migrations: in-reload ecto.migrate + AST expand/contract classifier


vjt:
## Goal

Make **additive (expand) migrations hot-deployable** — apply the schema change and code-reload the live BEAM without a cold restart, so IRC sessions are preserved. Today `Grappa.Deploy.Preflight` blanket-cold-trips **any** `priv/repo/migrations/*` change (`preflight.ex:265`), forcing a full `service grappa restart` (session reset, visible quit/rejoin on every network) for even a trivial `add column null: true`.

### 1. Run `ecto.migrate` inside the reload handler, in-process

`GrappaWeb.AdminController.reload/2` already runs in the **live BEAM with `Grappa.Repo` started and its pool open**. Run the migration there, on the existing pool, **before** the module-reload loop:

```elixir
# NOT Ecto.Migrator.with_repo/2 (it starts+stops the repo, fighting the
# supervised one). run/3 uses the already-running pool → ONE sqlite
# writer = the live BEAM. No separate eval node.
migrated = Ecto.Migrator.run(Grappa.Repo, :up, all: true)
```

- **Order: migrate → then `:code.load_file`.** For an expand migration the currently-loaded (old) code keeps running safely (it never references the new column); only after the column exists do we load new code.
- **Migrate fails → return non-2xx, do NOT reload.** Transaction rolls back (Ecto default; sqlite DDL is transactional), old code keeps running, `deploy.sh`'s `curl -fsS` sees the failure → deploy fails loudly.
- Running `migrate(:up, all: true)` unconditionally on every hot reload is idempotent (no pending = no-op) and also hardens `--force-hot`.
- sqlite note: `ALTER TABLE ADD COLUMN` is metadata-only/fast and transactional; needs WAL + a `busy_timeout` so the in-handler migrate doesn't `SQLITE_BUSY` against live readers (verify runtime config).

### 2. AST parser + expand/contract classifier in `Grappa.Deploy.Preflight`

Replace the blanket `migration?/1` cold-trip with a real classifier. **Parse the migration's `change/0`/`up/0` AST** and walk the `Ecto.Migration` DSL calls. **No `@annotation` / opt-in flag — pure AST inference.**

Build it as a **conservative ALLOWLIST** (HOT only when *every* op is provably expand; COLD on anything unrecognized — false-COLD costs a restart, false-HOT crashes live sessions, so doubt → COLD):

**HOT (expand) allowlist:**
[...]

PR #876, opened 2026-08-05
infra(#867): the bare docker run migrates itself, and refuses to boot when it can't

vjt:
`docker run ghcr.io/vjt/grappa:<tag> start` on a fresh volume died on `no such table: admin_events`. The image ships the migrator and nothing on the bare path invoked it; inside the image there is no mix, no checkout and no deploy script to invoke it with, so the operator had no door at all. #862 fixed the door before this one and, by design, stopped there — fixing it alone moved the failure one step later rather than making the documented one-liner work.

## Where the switch lives, and why not in the BEAM

`infra/docker/release-entrypoint.sh`, which exists **only for this image**. The jail, the `.deb`/`.rpm` hosts and the linux systemd box each migrate deliberately from their own deploy script BEFORE swapping code, and that ordering is load-bearing for a hot deploy (#41 — only "expand" migrations are hot-safe). An auto-migrate inside `Grappa.Application` would have been one line and would have changed all of them at once. It is the same reasoning that block already gives for sizing the BEAM caps through `ERL_ZFLAGS` instead of baking a `rel/vm.args` that would follow the release into every other substrate.

Only the verbs that boot a release trigger it (`start`, `start_iex`, `daemon`, `daemon_iex`) — `deploy.sh`'s own `eval 'Grappa.Release.migrate()'` runs through this very entrypoint and must not nest.

## Why the default is ON, and what I refuse to claim

The only shipped consumers of this image are single-container: the bare run of the docs, and `infra/docker/deploy.sh` in release mode (one container name, one named volume). Two BEAMs sharing one sqlite file is not a supported topology with or without migrations, so defaulting OFF would break the real path to protect one that is already broken.

**This is explicitly not a claim that concurrent starts are safe.** Nothing serialises them and no lock was added: a `flock` would have guarded one step of a configuration that is unsupported at every other step, and could not cover `deploy.sh`'s host-side `eval` anyway. An orchestrated deployment must set `GRAPPA_AUTO_MIGRATE=0` and run the migration as its own job — that is documented, not implied.

`deploy.sh` passes `GRAPPA_AUTO_MIGRATE=0` on its `docker run` rather than relying on "there will be nothing pending". It closes a real window — `--restart unless-stopped` on a crash-looping old container could restart it INTO a migration while `release_migrate` is running — and keeps the ordering decision in ONE place per path instead of being made twice, differently.

**Rejected: "migrate only when I created the DB myself."** The predicate lies: a crashed first boot leaves a zero-migration DB file behind and every later attempt then skips forever. It also puts the bug straight back on the update path, where the acceptance criterion is identical.

Unknown values are rejected rather than guessed — reading `true` as "not 1, so off" would reinstate the bug on a box whose operator believes they turned it on. Empty means unset, matching the secrets block above it and `config/runtime.exs`.

## Measured, not reasoned

| scenario | before | after |
|---|---|---|
| bare run, **empty** volume | container exits, `no such table: admin_events` | 79 migrations, `/healthz` **200** |
| bare run, **populated** volume | — | **0** migrations re-applied, `/healthz` 200, secrets file sha256 **identical** before and after |
| migration made to fail for real | — | exit **1**, `/healthz` never answers, volume keeps all **112** schema objects + its rows, `schema_migrations` truthful; restoring the cause brought it back to 200 with the secrets still byte-identical |

The failure case was injected by deleting two `schema_migrations` rows on a populated volume so the migrator genuinely re-ran and blew up (`duplicate column name: totp_secret_encrypted`) — Ecto's per-migration transaction rolled the attempt back. The failure mode being refused here is the one #441 had just found in `install.sh`: fail, print loudly, and exit 0 announcing a healthy host.

## The wrapper

`scripts/release-image.sh` is new because `docs/TESTING.md` forbids raw `docker` and no wrapper reached this artifact — which is precisely why #862 and #867 were both found by hand and neither was reproducible afterwards. `build`, `fresh-boot`, `warm-boot`, `oneshot`, `logs`, `down`. Unit D's "verification pending a real tag" caveat is narrowed, not deleted: the one-liner is measured against a locally built image; the published artifact and the job that pushes it are still unverified, and the command to check them once a tag cuts is written down.

## Gates

- new `test/infra/release_entrypoint_migrate_test.bats` (11 cases) — read RED against the old entrypoint first: 7 failed, 4 passed vacuously
- full host bats set: 313 ok, rc=0
- `shellcheck -s sh` + `dash -n` on the entrypoint, `shellcheck -x` on `deploy.sh` and the new wrapper


vjt:

Issue #279, opened 2026-07-16
ci flake / robustness: EventRouter route/2 emits malformed {:umode_changed, [..]} on garbage numeric-221 (RPL_UMODEIS)


vjt:
## What
The Elixir `ci` workflow (`test + lint + audit` job) went **RED on main `2d025c64`** (green on the prior 4 main SHAs) on a **seed-dependent StreamData property failure** — NOT caused by the SHA's change (#275 is cic-only, never touches EventRouter). It was green before purely because those runs drew a different seed.

- **Test:** `Grappa.Session.EventRouterPropertyTest` — `property route/2 always returns {:cont, state, [effect]} — no panics, no malformed effects` (`test/grappa/session/event_router_property_test.exs:83`).
- **Seed:** 671214 (failed after 80 successful runs).
- **Run:** [29542831205](https://github.com/vjt/grappa-irc/actions/runs/29542831205).

## Counterexample
```
message: %Grappa.IRC.Message{prefix: {:nick, "x2Q70mGcYmG_NsP_", "u", "h"},
  command: {:numeric, 221},   # RPL_UMODEIS
  params: ["]{@O& L#j|vH9_E...", "x]{K(<ShL!UV|P)...")]}     # garbage fuzz string
→ effect: {:umode_changed, [" ", "!", "\"", "$", "'", "(", ")", ",", "0", ... "~"]}   # flagged "malformed effect"
```

## Root cause (hypothesis — verify)
The **numeric-221 (RPL_UMODEIS) handler splits the mode param into a char list without validating** it's actually mode letters. On a fuzzed param full of spaces/punctuation it emits `{:umode_changed, [<every char incl " ", "!", "\"">]}`, which the property's malformed-effect check rejects. A real ircd only ever sends a clean umode string (`+iwx`), so this is **not hit in production** — but it's (a) a **latent robustness gap** and (b) an **intermittent `ci`-red source on main** (random seed → occasionally hits the garbage-221 case).

## Fix direction
Either **validate/sanitize the umode chars in the 221 handler** (drop non-mode bytes, or reject a malformed 221 param at the boundary), OR **constrain the property's `message_gen()` to realistic umode strings** for numeric 221. Prefer hardening the handler (boundary validation) over weakening the property. **Ties to queued #249** (UmodeModal from server umode set, gated on #221) — coordinate so the 221 effect shape is settled once.

[...]

Issue #249, opened 2026-07-15
cic: drive UmodeModal from the server-advertised umode set (gated on #221)


vjt:
Follow-up split out of #240 (part 3).

#240 shipped the channel-mode param editor (parts 1+2). Part 3 — driving the **user-mode** modal (\`UmodeModal\`) from the *server-advertised* umode set instead of a static letter table — is **not buildable pure-cic today**:

- The wire has no umode-availability event. \`umode_changed\` carries only the operator's *active* set, not the supported set.
- \`004 RPL_MYINFO\` (which lists supported umodes) is deliberately unparsed on the server.
- So "render every umode the server exposes" needs a **new server \`.ex\` wire event** exposing the supported umode set — which is **#221's territory** (generic upstream umode parsing, currently in a separate unmerged worktree).

**Blocked on #221** landing a supported-umode wire event. Once that exists, cic folds \`UmodeModal\`'s available set from it (same way \`availableModes\` already folds server-advertised CHANMODES for channels).

Today \`availableUmodes\` unions a static bahamut table with any active-but-unknown letter — the pure-cic best effort. Low urgency; cosmetic completeness, not a functional gap.

vjt:
Merged to `main` in **9b2754f0** (PR #323).

**What shipped:** the server now parses the supported-umode set from 004 RPL_MYINFO (param index 3) into a per-session `supported_umodes` field and broadcasts a typed `supported_umodes_changed` event on `Topic.user/1` (mirror of #216's ISUPPORT/CHANMODES plumbing). cicchetto folds the `/umode` modal's available toggles from it, keeping the static bahamut table as a graceful fallback when the server hasn't advertised (a network that omits 004 umodes, or a pre-snapshot session). The 004 `$server` display row is un-regressed (routing unchanged; the fold runs after the persist). The #301 description table is unchanged — #249 only changes WHICH letters are available.

**Deploy: HELD — NOT deployed this session.** COLD: `supported_umodes` is a `Grappa.Session.Server` state-shape change (tracked in `HotReload.LongLivedModules` → `Deploy.Preflight` auto-classifies COLD), and the cic bundle carries the fold. Rides the already-HELD #299 COLD batch (#282 / #290 / #294 / #301 / #302 / #304 / #305 / #306 / #307 / #318) — ships together in the next COLD 4AM window via `deploy-m42.sh --force-cold` (+ `--cic`).

[...]

PR #877, opened 2026-08-05
irc(#279): validate the mode-letter class at the umode ingress

vjt:
Fixes the #279 malformed-effect class at the boundary, and un-does the
half-cerotto that answered it the first time.

## The defect

`walk_umodes/3` did `MapSet.put` of ANY byte, so a fuzzed `221 RPL_UMODEIS`
param emitted `{:umode_changed, [" ", "!", "\"", …]}`. A real ircd only sends
`+iwx`, so it never fired in production — it fired in CI, on whichever seeds
drew numeric 221.

## What the boundary validates now, and on what source of truth

The mode letter **SET** is per-ircd (bahamut `+iwxs`, solanum `+DQZagiow`,
extension-registered letters) and grappa stays agnostic to it. I did **not**
validate against `supported_umodes` (004): an under-advertising server would
then silently drop a real umode, and a 221 can arrive before any 004.

What is closed is the **grammar**: an RFC-2812 mode block is signs + ALPHA.
That is the widest rule that still rejects the fuzz, so it is the contract —
`Identifier.valid_mode_letter?/1`, one predicate shared by the handlers *and*
by the property that pins them, so a future ircd letter cannot be green in one
and red in the other.

Posture: **reject the whole token**, not the bad bytes. A mode string is one
atomic statement about the session; a half-parsed one publishes a set the
server never claimed. Empty is a rejection too (it asserts nothing and would
let a truncated line WIPE the set); a bare `+` stays a valid empty snapshot.

Two extras that fell out:

* **One verdict, two readers.** The self-MODE branch reads the same bytes
  twice — the umode fold and `set_r_mode?/1`, the `+r` detector that commits a
  staged registration secret as confirmed. Gating only the fold would leave a
  malformed line able to confirm a secret. The `$server` row still persists:
  rejecting the FOLD is not rejecting the LINE.
* **004 RPL_MYINFO** is the same boundary one numeric over — un-validated it
  emitted the identical malformed effect as `:supported_umodes_changed`, which
  drives the #249 `/umode` modal's toggles.

**#249 coordination:** no effect SHAPE changed. `{:umode_changed, [letters]}`
is what it always was, so there is nothing to settle once — only the domain of
the letters got narrower.

## The property was the other half of the job

`c45bf1b3` (a #373 review commit) reddened on seed 671214 and answered it in
the test: it added a `{:umode_changed, modes}` arm asserting `is_list` +
`all is_binary`. **Half of that was legitimate** — the arm did not exist, so
before it *every* `:umode_changed` flunked, including `["i","w"]`. The other
half was the cerotto: `is_binary/1` is exactly what `" "` and `"!"` satisfy.

So "restore the pre-widening allowlist" was **not** the fix: that allowlist
rejected the legitimate case. The narrowing here is stricter than either — the
arm asserts the letter class, via the production predicate.

## Measured by displacement

With production reverted and the restricted tests kept:

* `--seed 671214` reproduces the original counterexample exactly — numeric
  221, *after 80 successful runs*, `[" ", "\"", "$", "'", ")", …]` — and the
  restricted arm catches it. 3 properties red.
* Same seed on the fixed handler: green. Also green on seeds 0/1/42/999999/123456.
* Two new targeted properties hit the 221 and 004 params on **every** generated
  example, so the class stops depending on a seed drawing a 1-in-999 numeric.
* Unit tests went red-first with the defect printed in the failure
  (`[" ", "#", "&", "9", …]`), then green.

`scripts/check.sh` exit 0: 8 doctests, 52 properties, 5164 tests, 0 failures;
dialyzer 0; sobelow/credo --strict/doctor clean; bats 0 `not ok`.

## What I refuse to claim

This does **not** close the malformed-effect class — only its umode half.
Measured, not assumed: `MODE #chan "+n!$ "` still yields
`{:channel_modes_changed, "#chan", %{modes: [" ", "$", "!", "n"]}}`, and the
property's arm for it (`is_list(entry.modes)`) is the same
certificate-of-nothing one effect over. Not fixed here because channel mode
letters consume ARGS through the ISUPPORT CHANMODES/PREFIX tables: a
whole-token rejection has to be agreed by `walk_modes/5`, `walk_channel_modes/5`
and the 324 snapshot together, with ban/key/limit arg-alignment as the risk
surface. Worth its own issue.

Nor does it claim any production symptom was fixed: no ircd sends these bytes.
The value delivered is an unvalidated ingress closed and an intermittent main
red made deterministic.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -29545,3 +29545,70 @@ than a parameter worth adding. And the guard remains a per-verb habit, not a
 whole-module property: a new `await` that captures the token and skips the
 check reopens the defect for its own call path — one owner makes the rule
 greppable, it does not make it automatic.
+
+## 2026-08-05 — #279: the mode-letter class, and the assertion that certified nothing
+
+A fuzzed `221 RPL_UMODEIS` param folded into the per-session umode set
+verbatim — `walk_umodes/3` did `MapSet.put` of ANY byte — so a garbage param
+produced `{:umode_changed, [" ", "!", "\"", …]}`. No ircd sends that, so it
+never fired in production; it fired in CI, on whichever seeds happened to draw
+numeric 221.
+
+**The production half was never fixed the first time.** `c45bf1b3` (a #373
+review commit) reddened on seed 671214 and answered it in the property test:
+it added a `{:umode_changed, modes}` arm asserting `is_list` +
+`Enum.all?(&is_binary/1)`. Half of that was legitimate — the arm did not exist
+at all, so before it EVERY `:umode_changed`, including `["i","w"]`, flunked as
+a malformed effect; the type union had genuinely outgrown the allowlist. The
+other half was the cerotto: `is_binary/1` is precisely what `" "` and `"!"`
+satisfy. So "restore the pre-widening allowlist" was not the fix either — that
+allowlist rejected the legitimate case. The narrowing that closes it is a
+STRICTER arm than either: assert the mode-letter class.
+
+**Where the class comes from.** The mode letter SET is per-ircd (bahamut
+`+iwxs`, solanum `+DQZagiow`, extension-registered letters) and grappa stays
+agnostic to it — `supported_umodes` was rejected as the validator for exactly
+that reason: an under-advertising 004 would then silently drop a real umode,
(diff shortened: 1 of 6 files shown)


vjt:

Issue #204, opened 2026-07-10
cicchetto: foolproof new login page (minimal nick-first, collapsible advanced, connect spinner, A2HS hint)


vjt:
## Summary

Redesign the cicchetto login page to be foolproof ("a prova di utonto") for first-time and non-technical users. The current login collects the connection details but does little to guide a novice; this issue collects the requested UX rework into one place.

## Requested behaviour

- **Minimal by default.** The page shows *only* the nick input up front — nothing else competing for attention.
- **Nick validation.** Validate the nick as it is typed; illegal characters are handled gracefully rather than rejected silently — e.g. spaces are substituted with `_` instead of erroring.
- **New fields: ident + realname.** Add the client-side ident and realname inputs here (see #152 — this is where those fields land on the client). They live in the advanced/collapsible area, not the minimal default view.
- **Collapsible "advanced" section.** Everything beyond the nick (ident, realname, and the rest of the connection knobs) goes into a hide-away section that only users who care open. Default state: collapsed.
- **Connecting feedback.** While the session is establishing the upstream connection, show a spinner / progress indicator with some progress updates, instead of a dead page.
- **Add-to-home-screen hint.** Matrix-style background, plus a little arrow pointing at Safari's `...` / share menu to guide users through "Add to Home Screen" (PWA install).

## Relationship

- Depends on / lands the client-side ident + realname fields from #152 (server-side identity split + live-apply reconnect primitive).

## Origin

Requested by vjt on #grappa (next-batch scope).

vjt:
## Design proposal — foolproof login (#204) · REVISED per feedback

**Revised** after vjt's change requests (this comment edited in place). Still zero code/branch — for final aesthetic sign-off.

**Changelog vs first draft:**
- **Q1 nick validation — reversed to ON-SUBMIT.** No as-typed stripping; don't fight the user while typing. On submit: field contains `@` → validate naively as an email; no `@` → sanitize the nickname + strip whitespace.
- **Q2 A2HS arrow — moved to the install-splash screen ONLY.** Not on the login screen. If the user picks "Continue from browser" → no arrow anywhere.
- **Q3** arrow is dim/subtle. **Q4** connecting copy stays cosmetic (approved).
- **Branding** — "IRC" now visible in big letters, not only "cicchetto".
- **Connecting screen** shows the network name: copy = `connecting <network> IRC network…`.

### Scope note (vjt override applied)
[...]

Issue #96, opened 2026-06-26
Phase 5: accessibility pass (sidebar tree semantics, tap targets, focus contrast)


vjt:
The channel sidebar uses raw `<ul><li><button>` with no tree semantics -- iOS VoiceOver doesn't read the network -> channel hierarchy as a tree. Phase 5 a11y audit covers this + tap-target sizing + focus-state contrast (web.dev a11y guidelines).

Phase 5 hardening cluster.

vjt:
**vjt accepted the withheld `role="tree"`** (2026-08-05): shipping the named-groups + `aria-current` + spoken-state form, without the APG composite-widget contract. The trade that decided it: a tree is one tab stop with roving tabindex, which takes each row's `×` out of the Tab order — a keyboard affordance traded for a screen-reader one. Precedent in-tree: `AdminPane` ships `role="tablist"` + 10 `role="tab"` with neither.

vjt:
Owner's call (2026-08-05): leave it as it is.

Recording why, so this does not get "fixed" by a later pass: `role="tree"` was withheld on purpose.
A tree is a **single tab stop** — adopting it takes every channel out of the Tab order and replaces
sequential navigation with arrow-key navigation inside one widget. For a sidebar people tab through,
that is a regression dressed as an a11y improvement. See bccf209d for the long form.

The rest of this issue (tap targets, focus contrast) is unaffected and stays open.

vjt:
The sidebar a11y leg shipped in **v0.12.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.12.0 — named per-network groups, `aria-current`, state in the accessible name, and focus rings, with `role="tree"` deliberately withheld (vjt accepted that).

**Leaving this issue OPEN:** the tap-target and focus-contrast legs of the original scope are not done, and vjt said explicitly they stay open. `status:soon` stripped so the board stops claiming the whole issue is awaiting a release.

vjt:
**Stale-premise audit** (backlog sweep against `origin/main` `8ed4f2e0`, 2026-08-06). Evidence only — nothing closed, relabelled or edited here; triage is vjt's. Full sweep: #258.

**Verdict: PARZIALE — two thirds landed today; the last third is a product call, not a build.**

[...]

PR #879, opened 2026-08-05
cic(#96): sidebar a11y — named groups, spoken state, focus rings (tree role withheld)

vjt:
Issue #96 asks for three things on the channel sidebar: tree semantics, tap
targets, focus contrast. The tap targets already landed on their own via
#204/#306/#459 (`--chrome-tap-min: 48px` / `--tap-min: 44px`, applied to the
sidebar rows). This is the other two thirds — plus one thing I deliberately
did **not** do.

## The tree role is withheld, and it is your call

`role="tree"` is the most keyboard-dependent role in ARIA. The APG pattern is
a composite widget: one tab stop, roving `tabindex`, Up/Down/Home/End. Ship
the roles without that contract and the widget lies — an AT is told arrows
work when they don't (no ARIA beats bad ARIA). Ship the contract and roving
`tabindex` takes every row's `×` **out of the Tab sequence**: each window row
has a `<CloseButton>` sibling inside its `<li>`, so a tree removes a keyboard
affordance operators have today, unless a second key binding is invented for
it. That reads to me as a redesign, not an a11y annotation, so I stopped.

Smaller second obstacle: the sidebar renders 3+ sibling `<ul>`s (home, admin,
one per network) inside an `<aside>` that also holds `NextActiveButton` and
`ResizeHandle`, so one `role="tree"` needs a new wrapper — and
`display: contents` on a role-bearing element has a history of dropping it out
of the a11y tree. A per-network tree with `aria-level` dodges that but not the
keyboard problem.

Precedent worth knowing before you decide: `AdminPane` already ships
`role="tablist"` + 10 `role="tab"` with no roving tabindex and no arrow keys.
So the codebase has accepted roles-without-the-keyboard-model before. A
tablist degrades gracefully; a tree is the role that degrades worst.

Also worth flagging: #96's stated motivation is **iOS VoiceOver**. The desktop
`Sidebar` is never mounted on mobile — `BottomBar` is. Whatever we do here does
not reach an iPhone.

## What did ship

**Hierarchy, without touching the DOM.** Each per-network `<ul>` is named
(`aria-label="freenode windows"`), so it announces as a named group of 6
instead of a bare "list, 6 items", and the network → window grouping stops
being pure indentation. Home and admin stay unnamed on purpose: single
identity-scoped rows, not groups.

**Two states that were pixels and nothing else.**
* `aria-current="true"` on the open window's `<button>` — selection was the
  `.selected` class alone.
* The not-live state now rides in the accessible name (`"#italia (parked)"`)
  via an `sr-only` span. This one isn't only an AT gap: `.parted` and
  `.sidebar-window-greyed` are BOTH muted + italic, so "you left this channel"
  and "the network is parked" are indistinguishable **for everyone**.
  `isGreyed/2` became `greyedState/2` returning the word, with the boolean
  derived from it — one derivation behind the class and the announcement, so
  they cannot disagree; the network-parked cascade keeps its precedence over a
  row's own state.

**Focus.** The whole 9,979-line stylesheet had exactly two `:focus-visible`
rules and *both* read `{ color: var(--fg); outline: none }` — they suppressed
the UA ring for keyboard users and left a `--muted → --fg` colour shift as the
only cue, an indicator with no area. A third surface
(`.service-modal-prompt-input`) has `border: none` + `outline: none` and puts
nothing back at all. The sidebar itself had no authored focus style, so the cue
was whatever the UA drew, on every theme including operator-authored ones.
Rings are `:focus-visible`, never `:focus` (a ring on a mouse click is noise —
presumably why the suppressions were written); inset on the edge-to-edge rows
so the scrolling aside can't clip them, outset on the padded chrome buttons.

Note the premise correction: there is **no global `outline: none` reset**, so
keyboard focus was not invisible app-wide — the UA ring survives everywhere
nobody killed it. The defect is narrower and sharper than "largely invisible":
three surfaces that removed it, and a primary nav surface that never authored
one.

## How it's tested

Every sidebar assertion queries by **role and accessible name** — the thing an
AT actually consumes. `getAttribute("aria-current")` would pass with the
attribute on an element the name computation ignores.

Mutations, each reverted after measuring:

| mutation | red |
|---|---|
| drop `aria-current` on the channel row | 1 |
| drop the per-network group name | 1 |
| drop the greyed state word | 4 |
| re-suppress the ring on `.shell-chrome-btn:focus-visible` | 2 |
| delete the sidebar ring rule | 5 |
| re-scope that ring `:focus-visible` → `:focus` | 1 |
| add a BRAND NEW suppressing rule on an unlisted surface | 1 |

That last one matters: the stylesheet invariant generalises past the surfaces
I enumerated. It is the durable half of this PR — no focus-state rule may drop
the outline without supplying its own indicator (text inputs legitimately swap
it for an accent border and pass; both original suppressions would have
failed).

**The first cut of that invariant was unfalsifiable and the mutation caught
it.** `/outline:\s*(?!none|0)/` backtracks its `\s*` to zero width, lands the
lookahead on the space before `none`, and matches `outline: none` itself — so
every suppressor was filtered out as a non-suppressor and the offenders list
was structurally always empty. Re-suppressing a ring left it green. Rewritten
as a declaration-level read; now that mutation turns it red. Separate commit.

**Bug found on the way:** `vi.clearAllMocks()` clears call history but keeps a
spy's implementation, so two `vi.spyOn(selMod, "selectedChannel")` calls in the
pseudo-row `×` block leaked "`#italia` is the open window" into every test that
ran after them in `Sidebar.test.tsx`. Nothing had ever read selection's output,
so nothing noticed — these tests failed in-suite while passing in isolation.

## What I refuse to claim

* **That any of this is audible.** No screen reader was run. VoiceOver, NVDA
  and JAWS each differ on when they insert a separator between adjacent text
  nodes, which is exactly the seam the state word sits in — the tests match
  `\s*` before it precisely because `dom-accessibility-api` concatenates
  without one where browsers add one.
* **That a ring is painted, has contrast, or isn't clipped.** jsdom has no
  layout and nwsapi doesn't match `:focus-visible`; there is no browser on this
  host to measure with. `--accent` is a per-theme token and operators author
  their own themes, so contrast isn't a property of this stylesheet at all. The
  inset-vs-outset geometry is **argued, not measured**.
* **That the greyed rows behave correctly on a real network park.** The
  cascade is exercised against the mocked `connection_state`; the derivation is
  unchanged from what shipped, only its return type moved from `boolean` to the
  word.

If you want the ring geometry or the announcements verified for real, that
needs an e2e (STACK lane) or a device check — say the word and I'll ask for
the lane rather than pretend jsdom covered it.

## Gates

`bun run check` rc=0 · `bun run test` rc=0 — 237 files / 4317 tests
(baseline at the merge-base: 236 / 4296; +21 = 12 sidebar + 9 focus).
Working tree clean before and after both runs. cic-only, no lane consumed.

### DIFF
--- a/cicchetto/src/Sidebar.tsx
+++ b/cicchetto/src/Sidebar.tsx
@@ -86,6 +86,17 @@ import NickText from "./NickText";
 const NOT_JOINED_STATES = new Set(["invited", "failed", "kicked", "parked"]);
 const NETWORK_GREYED_STATES = new Set(["parked", "failed"]);
 
+// #96 — a row's state, spoken. Every non-live sidebar row is rendered muted +
+// italic and NOTHING else: a screen reader gets no signal at all, and the
+// greyed (parked / failed / invited / kicked) and parted treatments are
+// pixel-identical, so a sighted operator can't tell them apart either. This
+// folds the word into the button's accessible name ("#italia (parked)")
+// without touching a single pixel. Renders nothing when there's no state to
+// report, so a live row's name is unchanged.
+const RowState: Component<{ state: string | null }> = (props) => (
+  <Show when={props.state}>{(state) => <span class="sr-only"> ({state()})</span>}</Show>
+);
+
 export type Props = Record<string, never>;
 
 const Sidebar: Component<Props> = () => {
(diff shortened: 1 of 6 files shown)


vjt:

Issue #888, opened 2026-08-05
Unread badge does not show that a window is far behind — a frozen count reads as a broken counter


vjt:
## Problem

When a window is *far behind* (#693), its read cursor is deliberately frozen: leaving the window does not advance it to the last rendered id, precisely so a region you never looked at does not get marked read across all your devices. The unread badge therefore keeps showing a large number (e.g. `1832` on a busy channel) that does not move while you read.

That behaviour is correct, but it is invisible. The badge looks identical to a normal unread badge, so a frozen count reads as "the counter is broken" rather than "you are 1832 messages behind". This came up on IRC today with two separate people reading it as a bug.

Related but distinct from #887 (stop hiding the badge on the focused window): #887 is about *suppression*, this is about *legibility of the frozen state*.

## Where it is today

- The far-behind state is already a module-level signal, exported and readable outside the pane: `farBehindByChannel` in `cicchetto/src/lib/scrollback.ts:251` (re-exported at `:1154`).
- The pane already renders an affordance for it — the far-behind bar with a jump control: `cicchetto/src/ScrollbackPane.tsx:3383-3431` (`data-testid="far-behind-bar"` / `far-behind-jump"`, which shows `"<n> unread"` and jumps to the boundary).
- The sidebar badge does **not** consult it: `cicchetto/src/Sidebar.tsx:352` (`.sidebar-msg-unread`, three occurrences — server window `:352`, and the two channel/query lists at `:481` and `:598`); same for the bottom-bar chips (`cicchetto/src/BottomBar.tsx:157`). Styling lives at `cicchetto/src/themes/default.css` (`.sidebar-events-unread` at `:1563` and neighbours).

So no new state or plumbing is needed — the signal the badge would need is already exported.

## Proposal

Give the badge a distinct treatment while the window is far behind, so the number reads as a *distance*, not a stuck counter:

1. In the sidebar (all three badge sites) and the bottom-bar chip, read `farBehindByChannel()[key]` and apply a modifier class when it is set.
2. Style that modifier differently (the exact visual is a design call — a muted/outlined variant, or a small "behind" affordance next to the number, rather than the normal solid unread pill).
3. Add a `title` / accessible label along the lines of "1832 messages behind — jump to the boundary to catch up", so the frozen state is explained where it is seen.
4. Optionally, make the badge a click target that does what the existing `far-behind-jump` control already does, instead of duplicating that logic.

[...]

Issue #887, opened 2026-08-05
Unread badge: stop hiding it on the focused window — keep it visible and let it fall as rows are marked read


vjt:
## Today

Two mechanisms combine into a badge that reads as broken:

1. **Focused-window suppression** — `cicchetto/src/lib/selection.ts:471-477`, the final overwrite in
   the `perChannelUnread` memo, zeroes the count for the selected window while the tab is visible.
   The read cursor is deliberately *not* advanced; the badge is only hidden.
2. **Far-behind freeze** — `setCursorIfAdvances` refuses to move the cursor while the window is far
   behind (#693), so the abandoned region is never marked read on leave. This is correct and must
   stay: taking the newest rendered id there would mark thousands of unseen messages read on every
   device.

Consequence, as observed by an operator on a channel he had just been reading: focus the window →
badge disappears; switch away → the badge comes back at **1832**. Nothing is wrong, but the display
says otherwise, because the two states "you are looking at this" and "you have read this" are drawn
identically (no badge).

## Wanted (owner's call, 2026-08-05)

> NON nascondiamo il badge ma lo aggiorniamo, e continuiamo ad aggiornarlo man mano che marchiamo read

So: **stop suppressing the badge for the focused window.** Keep it rendered and let it fall as rows
are actually marked read, so the number always answers "how much have you not read", in the focused
window exactly as in every other one. The count then decreases live while the operator scrolls
through the backlog instead of vanishing on select and reappearing whole on leave.

## Notes for the implementer

* The change is the removal of the suppression overwrite, but the counts must then be driven by
  something that actually advances while the window is focused. Today the cursor moves on settle
  events (scroll-settle, focus-leave, browser-blur, send) — with the badge visible, a cursor that
  only moves on leave would show a frozen number the whole time the operator reads, which is a
  different flavour of the same complaint.
* The far-behind freeze (#693) stays. A frozen cursor with a visible badge is exactly the state the
  UI should be able to express — consider making "far behind" legible on the badge itself
  (dimmed/marked, paired with the existing jump-to-bottom affordance) rather than letting a large
  static number look like a bug.
* The in-pane `── N unread ──` marker already survives selection and clears on its own settle
  events; badge and marker should agree after this change, not drift.
* `isDocumentVisible` gating exists so a selected-but-backgrounded tab keeps accruing. With
  suppression gone that special case may collapse — check it does not silently change the
  backgrounded-tab behaviour.
* Prior art to re-read before touching this: the design spec at
[...]

Issue #267, opened 2026-07-16
cicchetto: mention/activity counter must be computed server-side (currently client-side)


vjt:
The mention counter (the red badge) and the per-channel activity counters are currently computed **client-side**. This has to move **server-side**: grappa should be the authority for unread/mention counts, pushed to clients over the existing WS (Phoenix Channels), so the numbers stay consistent across tabs/devices/reconnects and reflect server truth rather than whatever a single client happened to observe while connected.

## Symptoms driving this (client-side artifacts)

- **Aggregate / overflow badge is always red** (reported on mobile, hidden tabs). The "other tabs" overflow indicator hardcodes red instead of inheriting the max real severity of the channels underneath. Per-channel counters correctly follow the colour convention (red = mention/highlight, blue = message, grey = join/part), but the aggregate ignores event type, so it stays permanently red and looks like there are always pending notifications.
- **Event-type scope** (#265): the activity-since-last-visit indicator counts JOIN/PART/MODE, not just messages (PRIVMSG/NOTICE). A client re-deriving counts from the raw event stream gets this wrong; a server-side count with a defined severity model would not.

## Desired

- Server computes, per credential/session: per-channel unread message count, per-channel mention/highlight count, and a severity level (mention > message > low-noise join/part).
- The aggregate / overflow badge derives its colour from the **max severity across the hidden channels**, not a constant.
- Counts + severities pushed over WS; the client renders, it does not compute.
- #265 (scope) and the overflow-colour bug both fall out of a single authoritative model.

## Notes

- Long-standing / design-first (roadmap), not queued — same footing as the counter/presence architecture work. The overflow-colour fix is a near-term client-only change that can land independently under #265 if wanted before the full server-side move.
- Related: #265 (event-type scope).

(overflow-badge symptom raised via chan report.)

vjt:
## #267 — Server-authoritative unread/mention counters: design proposal (for review, not yet built)

Design-first pass. Nothing implemented, no worktree cut. Below is the full build-plan; I'm holding for @vjt's verdict on scope + the 5 boundary-calls before writing any code.

---

### TL;DR — scope reframe (evidence from current `main`, not the issue text)

Investigation of `main` (`97dbf961`) turned up two things that shrink this issue substantially:

[...]

PR #894, opened 2026-08-05
cic: the unread badge on the window you are reading (#887) and the one that is frozen (#888)

vjt:
Two distinct defects, one PR: they land on the same four badge call sites
(`Sidebar` server/channel/query, plus the `BottomBar` chips), so separate
branches were guaranteed to conflict line-for-line.

## #887 — the badge vanished on the window you were reading

Focusing a window zeroed its unread count outright: a final overwrite in
`perChannelUnread` keyed on `selectedChannel` + `isDocumentVisible`, which
deliberately did NOT advance the read cursor. "You are looking at this" and
"you have read this" were drawn identically — as no badge — so a channel you
had just been reading lost its badge on select and got it back whole (1832)
on leave.

Removing the overwrite alone trades a vanishing badge for a stuck one: every
pre-existing cursor writer fires when the operator STOPS looking (leave /
blur / unmount) or when they SCROLL (settle, gated on a real input event). An
operator parked at the tail of a quiet window generates neither. So the
removal ships with a **read-at-the-tail** arm in `ScrollbackPane`: while the
tab is visible AND the pane is at the tail, advance the cursor to the last
fully visible row.

Both gates are load-bearing — a backgrounded tab keeps accruing, and a reader
parked above the fold (or a cold activation that jumped to a far unread
divider) must not have the rows below marked read for them. Everything
dangerous past them is already refused downstream by the single door
`setCursorIfAdvances`: the #693 far-behind freeze and the #233 forward-only
clamp. The in-pane divider is untouched — it reads the frozen
`markerCursorId`, not the live cursor.

Five writers shared the same three-line body verbatim; they now share one
helper.

## #888 — a frozen badge did not say it was frozen

A far-behind window's cursor is pinned on purpose, so its count does not move
while you read. Drawn as the ordinary solid pill it is indistinguishable from
a live counter. Now: an outlined variant (same size, same position, same
number — it just stops claiming to be a live count), plus a title and an
accessible label naming the distance and the way out. Both unread badges take
it, because the count memo serves the frozen server seed for both kinds while
far behind; the mention badge does not, being server-authoritative and
counted a different way (#267).

The badge triad sat inline in six verbatim copies. Adding one predicate to
six pasted blocks is copy-paste-with-tweaks, so the triad moved into
`WindowBadges` with the class prefix resolved through a closed table.

## What I measured (not just "tests pass")

Each claim was checked by mutating production and watching the specs go red:

| mutation | result |
| --- | --- |
| restore the suppression overwrite | 6 red across `selection.test.ts` + `unreadBadgeFocused.test.ts` |
| disable the read-at-the-tail arm | `advances to the arriving row with NO operator input event` red |
| drop its `isDocumentVisible` gate | `stands down while the tab is hidden` red — **and** the reworked BUGHUNT-2 input-gate test red, so its new "hide the tab to displace the confound" setup is doing real work |
| `farBehind` predicate → `false` | 4 red in `WindowBadges.test.tsx` |
| **drop the `atBottomNow` gate** | **all 171 ScrollbackPane tests still GREEN** |

That last row is the honest one: jsdom's zero-geometry pane reads as
at-the-tail unconditionally, so no unit test can tell the at-tail branch from
its opposite. The e2e is where that gate is actually pinned — it plants a
cursor 120 rows back, asserts the badge holds while scrolled up past the
debounce, then watches it fall under a real wheel, then watches a message
that arrives while the operator watches the tail clear itself with no input
(waiting out `INPUT_EVENT_RECENCY_MS` first, so the drain cannot be
attributed to the scroll-settle arm).

## What I will not claim

- **The e2e has not been run.** It needs the STACK lane. Everything below is
  what was actually run: `bun run check` rc=0, `bun run test` 4330/4330
  green, `bun run build` green, all on a clean tree.
- The intermediate badge blip (1, then gone) in the watched-arrival e2e is
  **not** asserted — the window is ~500ms and polling for it would be flaky.
  Only the settled end state is, after an explicit wait past the debounce so
  the assertion cannot pass against a broken arm on its first poll.
- The far-behind treatment is asserted at the DOM/class level, not visually.
  Nobody has looked at the outlined pill in a browser.
- Adding `aria-label` to the unread badges changes every window row's
  computed accessible name ("#italia 5" → "#italia 5 unread messages"). It is
  forced — a generic `<span>` cannot carry `aria-label` — and I think it is
  an improvement, but it is a change #96 territory may want an opinion on.
- `mentionCounts` keeps its own focus-zero overlay (#267). #887 did not ask
  to touch it and I did not, but the two badges now behave differently on a
  focused window: the unread one persists, the mention one still disappears.
  Flagging rather than deciding.

Branched from `c57d4dab`; `origin/main` moved (to `ffde30f3`) mid-work, so
this wants a rebase before merge.

### DIFF
--- a/cicchetto/e2e/tests/issue887-focused-unread-badge.spec.ts
+++ b/cicchetto/e2e/tests/issue887-focused-unread-badge.spec.ts
@@ -1,0 +1,194 @@
+// #887 — the unread badge on the window you are LOOKING AT.
+//
+// The report (vjt, 2026-08-05, on a channel he had just been reading): focus
+// the window → the badge disappears; switch away → it comes back at 1832.
+// Nothing was wrong — the read cursor was where it should be — but the two
+// states "you are looking at this" and "you have read this" were drawn
+// identically (no badge), so the display lied by omission.
+//
+// The fix removes the suppression overwrite from `perChannelUnread` and pairs
+// it with a read-at-the-tail cursor advance in ScrollbackPane, so the number
+// answers one question everywhere: how much have you not read.
+//
+// Why this has to be a browser test, not jsdom: BOTH halves of the contract
+// are geometry. "Holds while scrolled up" and "drains at the tail" are the
+// same code path taking opposite branches on `atBottomNow`, and jsdom's
+// zero-geometry pane reads as at-the-tail unconditionally — a jsdom test
+// cannot tell the branches apart (measured: deleting the at-tail gate leaves
+// the entire 171-test ScrollbackPane suite green). This spec is where that
+// gate is actually pinned.
+//
+// Per `feedback_e2e_user_class_parity_matrix`: this pins the unread-badge
+// CONTRACT, not a verb across user classes — one seeded registered user is
+// the right shape.
+
+import type { Page } from "@playwright/test";
+import { expect, test } from "../fixtures/test";
+import { loginAs, selectChannel, sidebarMessageBadge } from "../fixtures/cicchettoPage";
+import {
+  fetchAllMessagesAsc,
+  restoreReadCursorToTail,
+  setReadCursorToId,
+} from "../fixtures/grappaApi";
+import { IrcPeer } from "../fixtures/ircClient";
+import { AUTOJOIN_CHANNELS, getSeededVjt, NETWORK_NICK, NETWORK_SLUG } from "../fixtures/seedData";
+
+const CHANNEL = AUTOJOIN_CHANNELS[0];
+const PEER_NICK = "badge887-buddy";
+const RUN_ID = crypto.randomUUID().slice(0, 8);
+
+// How far back to plant the cursor. Two hard bounds, and the window between
+// them is where this scenario exists at all:
+//   * FAR ENOUGH that the divider the pane activates on is off-screen, so the
+//     pane lands NOT at the tail — the state whose badge must hold.
+//   * UNDER 200, or `isFarBehind` (scrollback.ts, #693) trips, the cursor
+//     freezes by design and the badge could never drain. That is #888's
+//     subject, not this one.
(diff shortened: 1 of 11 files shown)


vjt:
## e2e run + mutation round (STACK lane, chromium, `--grep '#887'`)

| run | rc | result |
| --- | --- | --- |
| baseline | 0 | **2 passed** (31.1s) |
| neuter the `atBottomNow` gate | 1 | **`holds while scrolled up` FAILED** — `Expected: 118, Received: 90` |
| disable the read-at-the-tail arm | 1 | **`watched arrival clears itself` FAILED** — `Expected: 0, Received: 1` |

Both mutations are caught, each by the test that owns the property, and in
each round **the other test stayed green** — so the two are not covering for
one another.

The first row is the one that matters: the `atBottomNow` gate is invisible to
all 171 jsdom ScrollbackPane tests (deleting it leaves them green), and here
neutering it drains 28 rows the operator never looked at, out from under a
pane scrolled 120 rows above the tail. That is the exact failure the gate
exists to prevent, and this spec is the only place it is observable.

### A correction to my own method

My first attempt at the gate mutation deleted the `!atTail` term outright.
That left `atTail` unread, `tsc --noEmit` failed, and the run died at
`cicchetto-build-test … exit 1` — rc=1 with **zero tests executed**. Had I
read only the exit code I would have logged a red as proof and been wrong:
an unbuildable bundle is not a caught mutation. Redone as
`atBottomNow() || true` (signal still read, gate always open), verified with
a local `bun run build` BEFORE burning a stack round, and only then does the
red above mean anything.

Full `scripts/integration.sh` is running now — the change alters badge
behaviour in every spec that asserts one, so the cascade risk is worth a
measurement rather than an assurance. Reporting the result either way.

Issue #407, opened 2026-07-25
cic modal-chrome CSS base extraction (.modal-backdrop / .modal-close)


vjt:
Split out of the #112 backlog item:

> Modal-chrome CSS dedup: `.media-viewer-close` duplicates `.archive-modal-close`, and `.media-viewer-backdrop` is a 4th `inset:0` dim backdrop — extract a shared `.modal-backdrop`/`.modal-close` base. `cicchetto/src/themes/default.css`

**Why its own issue (batch-4 scoping, 2026-07-26):** the shared pattern is far larger than the two classes named. An inventory of `cicchetto/src/themes/default.css` shows **~30 `*-close` classes** (sidebar, topic-modal, audio-mini-player, settings-drawer, delete-account, admin-pane, bottom-bar, archive-modal, names/who-modal, mode-modal, theme-editor, server-reply-modal, service-modal, registration-wizard, whois/whowas/lusers/peer-away cards, banlist-modal, directory, media-viewer, share-modal) and **~18 `*-backdrop` classes**. Some close buttons are ALREADY grouped (e.g. `default.css:6552` groups whois/whowas/lusers/peer-away).

Per CLAUDE.md **total-consistency**, converting only the 2 named classes forks the pattern — a real fix is all ~48 sites or none. That is a cross-cutting refactor touching ~25 TSX files (if a base *class* is used) and requiring **per-modal browser smoke** (visual-regression risk on every modal/card), which is disproportionate to bundle with unrelated one-line nits.

**Shared skeletons observed:**
- close buttons: `background:transparent; color:var(--muted); border:none; font-family:var(--font-mono); line-height:1; cursor:pointer;` + `:hover { color:var(--fg) }`; deltas are only `font-size` + `padding`.
- backdrops: `position:fixed; inset:0;` + dark scrim; deltas are opacity, `z-index`, flex-centering (and `.media-viewer-backdrop` is a `<button>` reset, `.settings-drawer-backdrop` a slide-in drawer — the two structural outliers).

**Design decision needed when picked up:** base *class* (`class="modal-close <delta>"`, self-documenting, per-element opt-in, touches every JSX element) vs grouped *selector* (CSS-only, no JSX, but drift-prone as new modals must be added to the group). The existing `default.css:6552` grouped-close is the in-tree precedent.

Batch 4 shipped the other two #112 cic items (ApiError/ChannelPushError info dedup; SubscriptionId brand).

vjt:
[...]

Issue #360, opened 2026-07-20
cicchetto: scroll-to-bottom button — mention-count badge + tap-to-next-mention


vjt:
## Summary

The scroll-to-bottom floating button (C7.4, `cicchetto/src/ScrollbackPane.tsx`) should become mention-aware. When the current channel has mentions below the current scroll position, the button shows a **badge with the mention count**. Tapping it jumps to the **next mention below** instead of straight to the bottom; each tap advances to the following mention and decrements the badge. When no mentions remain, a tap behaves as today (snap to bottom / newest).

## Desired behaviour

- Button gains a numeric **badge** = number of unseen mentions below the current viewport.
- Tap when badge > 0 → smooth-scroll to the **next mention below** the viewport (nearest first), badge decrements.
- Tap when badge == 0 → current behaviour: snap to bottom, advance read cursor to newest.
- Badge is per-channel (only mentions in the active window count) and clears as mentions are scrolled past / consumed.

## Grounding (origin/main)

- Button + snap-to-bottom gesture: **C7.4** in `ScrollbackPane.tsx`, dispatched via `lib/scrollToBottomCommand.ts` (`scrollToBottomRequest`). Today it advances the read cursor to the newest message.
- Mention detection already exists: `mentionsUser(body, ownNick)` from `lib/mentionMatch.ts`, rendered as `.scrollback-mention` on privmsg lines (P4-1); `isMention()` around ScrollbackPane.tsx:756.
- Related highlight track: `matchesWatchlist(body, ownNick)` → `.scrollback-highlight` (C7.7). Scope for this issue = **mentions** (own-nick); whether watchlist highlights also feed the badge is a follow-up decision, keep them separate for now per the existing split.
- Unread cursor / "── N unread messages ──" divider machinery is already in the pane and can be reused to bound the "below current position" walk.

## Open sub-decisions (implementer)

- "Next mention" ordering: nearest-below-viewport first (recommended), cycling downward.
- Whether the badge counts only mentions below the viewport, or all unread mentions in the window.
- Visual: reuse the existing mention accent colour for the badge.

Requested by vjt via #grappa.

PR #900, opened 2026-08-05
#866: per-conversation notification mute, keyed on the conversation

vjt:
Per-conversation notification mute (#866), on top of the #868 refactor.

Every list in `notification_prefs` was an allow-list, so a noisy channel
could not be silenced at all: even with `channel_messages_all: false` and
the channel absent from `channel_messages_only`, the third disjunct still
fired, because `channel_mentions` is a GLOBAL boolean. The only way to
quieten one room was to turn mentions off in every room.

Implements vjt's Q1–Q7 rulings verbatim; no product decision was invented.

## The decision the feature turns on

**The mute is keyed on the CONVERSATION, not on the row's `channel`
field.** An inbound DM is persisted with `channel = own_nick`, so keying
on `channel` would have made a single "mute vjt" entry silence every DM
the subject ever receives, while "mute alice" silenced nothing. So: the
folded channel for a channel row, the folded PEER for a DM. A truth-table
row expects TRUE for a mute on own_nick — a port that keys on `channel`
turns it red.

## The rest

- **The mute always wins (Q2).** It sits in the `cond` above both
  branches, so a direct mention does not pierce it and the next disjunct
  someone adds cannot forget to exclude it.
- **One map, `until` from day one (Q1).** `%{folded_target => %{"until"
  => unix_seconds | nil}}`. No UI writes an integer yet; two deny-lists
  would have needed a parallel structure the first time anyone wanted
  "for 8 hours". No network in the key — `user_settings` is per-SUBJECT,
  like `channel_messages_only` beside it.
- **Expiry is a read projection (Q3).** `get_notification_prefs/1` drops
  elapsed entries; no sweeper, no clock in the predicate, no `now` column
  in the shared truth-table. cic mirrors the rule at its own read, because
  the signal only re-hydrates on a user-topic (re)join and a one-hour
  snooze must stop silencing in a tab nobody reloaded.
- **Push / beep / badge only (Q4).** `WindowCounts` mention counts and the
  sidebar dot are untouched — a muted conversation still COUNTS. The badge
  follows because it is derived from this same predicate.
- **UI (Q5).** A picker over the conversations you actually have, not a
  text input; each mute is its own row with an ×, reusing the
  `.watchlists-list` idiom. The notifications section is regrouped under
  three headings rather than having the control bolted onto the end.
- **One departure from full-replace:** an ABSENT `muted_targets` on PUT
  means "unchanged". cic ships independently of the BEAM and an installed
  PWA can run a cached bundle for weeks; reading its silence as "clear
  them" would delete mutes the first time such a client ticked any other
  checkbox. An explicit `{}` still clears.

## Measured, not assumed

- 11 new shared truth-table rows. **7 measured red** in vitest before the
  cic port; the other 4 expect TRUE on purpose, so `-> false` cannot
  satisfy the set. With the server port reverted the SAME fixture reddens
  **7 of 41** in ExUnit — the drift gate works in both directions.
- Mutation testing on the drawer: neutering the add reddens 1, dropping
  the already-muted filter reddens 2. A dedupe guard I had "covered"
  turned out to be **unconstrained** (the Map dedupes on its own); the
  assertion was fixed so removing it now reddens 1.
- `scripts/check.sh` rc=0 — 8 doctests, 54 properties, 5266 tests, 0
  failures; 328 bats ok. Tree clean before and after.
- `scripts/bun.sh run test` 4365/4365, `run check` rc=0.
- Full `scripts/integration.sh`: **603 passed, 1 failed**. The failure is
  `issue270-peer-away-overlap` (the away banner never rendered), which
  runs BEFORE this spec, has no mechanism connecting it to notification
  prefs, and is green on main in CI at this branch's merge-base. **I did
  not establish a local full-suite baseline at the merge-base**, so I am
  not claiming it is unrelated — only that I found no link.

### DIFF
--- a/cicchetto/e2e/fixtures/grappaApi.ts
+++ b/cicchetto/e2e/fixtures/grappaApi.ts
@@ -525,6 +525,35 @@ export async function partChannel(
   }
 }
 
+// #866 — drop every per-conversation mute for `token`'s subject.
+//
+// notification_prefs live in `user_settings` and OUTLIVE the spec that wrote
+// them: the seeded vjt is shared by the whole suite, so a mute left behind
+// silences a channel for every later spec that expects a push or a beep. That
+// is the shared-stack poisoning class, and it is silent — the victim spec
+// fails on a missing notification with nothing pointing back here.
+//
+// Read-modify-write rather than PUT-the-defaults: the endpoint has no PATCH
+// semantics, so writing a fresh default map would also clobber whatever
+// whitelists / toggles another fixture had set.
+export async function clearMutedConversations(token: string): Promise<void> {
+  const url = `${GRAPPA_BASE_URL}/me/settings/notification-prefs`;
+  const headers = { authorization: `Bearer ${token}` };
+
+  const current = await fetch(url, { headers });
+  if (!current.ok) throw new Error(`clearMutedConversations: GET ${current.status}`);
+  const { notification_prefs: prefs } = (await current.json()) as {
+    notification_prefs: Record<string, unknown>;
+  };
+
+  const res = await fetch(url, {
+    method: "PUT",
+    headers: { ...headers, "content-type": "application/json" },
+    body: JSON.stringify({ ...prefs, muted_targets: {} }),
+  });
+  if (!res.ok) throw new Error(`clearMutedConversations: PUT ${res.status}`);
+}
+
 // JOIN a channel via REST POST (mirrors `cicchetto/src/lib/api.ts`'s
 // `postJoin`). Used by tests that PART a seeded channel and need to
 // restore it for subsequent specs (M9, in particular — without restore,
--- a/cicchetto/e2e/tests/push-866-conversation-mute.spec.ts
+++ b/cicchetto/e2e/tests/push-866-conversation-mute.spec.ts
@@ -1,0 +1,123 @@
+// push-866-conversation-mute — per-conversation notification mute (#866).
+//
+// Sibling of `push-prefs-whitelist.spec.ts`, and deliberately the MIRROR of
+// it: that spec proves the allow-list can let a channel through, this one
+// proves the deny-list can hold a channel back — including a channel the
+// operator is mentioned in, which nothing before #866 could silence without
+// turning mentions off everywhere.
+//
+// What it gates, and why an e2e is the right shape for it:
+//
+//   * the mute is created through the SettingsDrawer picker (`pref-mute-picker`),
+//     not by POSTing /me/settings/notification-prefs. So the folded key the
+//     picker emits, the server's normalization, and the key
+//     `Grappa.Push.Triggers` compares are all proven to be the SAME string. A
+//     unit test on either side cannot see a fold mismatch between them.
+//   * the assertion is on PUSH delivery, which is decided server-side. A
+//     client-only mute would silence the tab and still ring the phone —
+//     exactly the failure mode #866's own text calls out.
+//   * the negative arm carries a MENTION. That is vjt's Q2 ruling under test:
+//     the mute always wins, a highlight does not pierce it. Both truth-table
+//     ports assert this too; here it is asserted end to end, where a mention
+//     genuinely travels through `Mentions.mentioned?/3`.
+//
+// Removing the feature turns the negative arm red: with no mute, a mention in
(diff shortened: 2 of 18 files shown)


vjt:
## Verdict on the red `issue360-scroll-to-bottom-mention-badge.spec.ts:245`

**Not caused by this PR.** It is a known, previously-observed defect of that
spec, at the one site the earlier mitigation deliberately left uncovered.
Filed separately as #903.

### What actually fails

Line 245 is `await expect(page.locator(SCROLL_TO_BOTTOM)).toBeVisible();`
immediately after **tap 2**, under the comment *"Trailing content is still
below → the button remains"*. So after the jump to `MENTION_2` the button was
gone, i.e. the pane ended within 50px of the bottom. That is geometry plus
timing, not counting — the badge assertion just above it (`toHaveCount(0)`) had
already passed.

### 1. The file that owns the button is not in this diff

The button and its badge live in `ScrollbackPane.tsx`. That file is not among
the ten I touched, and it imports **none** of the modules I changed (grepped
for `pushTriggers` / `notificationPrefs` / `activeWindows` / `userSettings`:
zero hits).

### 2. The only globally-scoped file I touch is entirely inside the drawer

`default.css`: three hunks, all prefixed `.settings-drawer` (two selectors
widened to cover `select`, plus `.prefs-hint`). Nothing there can change the
scrollback pane's height or scroll position.

### 3. `activeWindows` is the only bridge, and the change is a no-op

`Shell.tsx` imports it. My diff renames a **private** function
(`buildCandidates` → `windowCandidates`) and exports it. Body unchanged line
for line; no pre-existing export changes name, signature, or body. The exported
surface only grows, so every current consumer is untouched.

### 4. With zero mutes the predicate is identical to main

`isMuted` is `muted !== undefined && Object.hasOwn(muted, key)`, and the
default is `{}` — false for every row, so `shouldNotify` returns exactly what
it returned before. #360 creates no mute. This PR's own spec creates two, but
runs **after** #360 (#407 vs #223) and clears them over REST in teardown.

### 5. The same eleven commits pass #360 locally

Full local `scripts/integration.sh` on this branch:
`✓ 223 issue360-scroll-to-bottom-mention-badge … (32.1s)`.
A change that broke this spec would break it deterministically. (Honest note:
that local run failed a *different* single spec, `issue270-peer-away-overlap`.
Two full runs of the same tree, two different single failures.)

### 6. The failure mode is already written down in git

Commit `e98fc0fa`, *"test(#653): tolerate scroll-to-bottom unmount race at tap
3 (#639)"*:

> The floating button lives under `<Show when={!atBottomNow()}>` in
> ScrollbackPane, so it UNMOUNTS the instant the pane reaches bottom. … a
> settling smooth scroll from the prior jump can flip `atBottomNow` true …

> … flaked … under full-gate load (the same flake that reddened PR #662's
> integration); green on rerun.

and decisively:

> Applied ONLY to tap 3, where unmount is a valid outcome; taps 1/2 (button
> stays up) keep the plain click.

Line 245 is exactly the tap-2 site: same button, same unmount mechanism, at the
point that cure chose not to cover.

### Confounder I am naming rather than hiding

My green local run was based on `3bd7da92`; CI runs on `5a78aac4`. In between,
main took two commits (`b87998e9`, `5a78aac4` — the self-nick-change DM re-key)
touching `lib/grappa/scrollback.ex`, `session/event_router.ex` and
`session/server.ex`. So the difference between my local green and the CI red is
not only this diff. I see no mechanism from there to pane geometry, but it is
not nothing.

### What I am NOT claiming

- Not that it is *proven* unrelated. I proved the button's code is byte-identical
  to main and unreachable from this diff; I did not run the measurement that
  would settle it.
- Not that it is "just a flake" — it is a real defect: an assertion resting on a
  geometry the setup does not guarantee.
- Not that a global timing shift from this PR is impossible. Implausible (one
  `Object.hasOwn` per message), but reading cannot rule it out.

The settling measurement would be
`--grep "mention-aware scroll-to-bottom badge" --repeat-each 5` on this branch
**and** on the merge-base, same machine, same load. And if #903 confirms the
gap, the cure is to make the setup deterministic — guarantee `TRAILING` exceeds
the pane height after the `MENTION_2` jump — not to raise a timeout or weaken
the assert.

Issue #904, opened 2026-08-05
cicchetto: a slow SINGLE send eats what the operator types during it — #737's guard only covers the multi-line drain


vjt:
Reported by peluche on IRC (mobile, poor network), confirmed by vjt as "happens absolutely always". Verified against `origin/main` at `5a78aac4`.

**Files:** `cicchetto/src/lib/compose.ts:1528-1530` (the end-of-submit clear), `:305-306` (`setDraft`'s only refusal), `cicchetto/src/ComposeBox.tsx:518` (`readOnly`), `:408` (Enter no-op)
**Related:** #737 (shipped in v0.11.0), #666, #772

## Problem

#737 fixed exactly this class of bug — operator input and the send machinery sharing one draft buffer — but the guard it installed is keyed on `isDraining(key)`, which is raised **only** by the #666 paced multi-line drain:

* `compose.ts:305` — `setDraft` refuses the write `if (isDraining(key))`, and nothing else.
* `ComposeBox.tsx:518` — `readOnly={isDraining(key())}`, same condition.

A **single** message on a slow link never raises `drainingKeys`. So during the in-flight `await`:

1. the textarea stays writable and `setDraft` happily accepts what the operator types;
2. Enter is a no-op — `doSubmit` early-returns on `sending()` (`ComposeBox.tsx:408`);
3. when the send finally resolves, `compose.ts:1530` runs the unconditional
   `writeState(key, (s) => ({ ...s, draft: "", historyCursor: null }))`.

`state` was captured before the `await`, so `pushHistory` correctly stores the *sent* line — but the clear wipes the *current* draft, which by then holds the operator's next message. It is typed, un-sendable, and then silently destroyed.

## Failure scenario

Mobile client on a weak connection. Operator sends a line; the POST takes ~5s. Meanwhile they type the follow-up ("ok also bring the charger") and press Enter — nothing happens. The first send's 201 lands, the composer empties, and the follow-up is gone with no error and no way to recall it (it never reached history either).

## What is being asked for

Ideal (peluche's request): a **send queue** — the in-flight message leaves the composer immediately and the operator can compose the next one, which dispatches when the first acks. This also makes Enter meaningful again during a slow send.

[...]

Issue #737, opened 2026-08-03
cic: the composer stays editable during a paced drain and loses what the operator types


vjt:
Found by the codebase-review campaign on `main` (`v0.10.0..388f883b`), `cicchetto/src` slice. Verified against the files at `388f883b`.

**File:** `cicchetto/src/lib/compose.ts` (the paced drain), `cicchetto/src/ComposeBox.tsx:494` (the textarea), `:522` (the only `disabled` in the file — the submit button)
**Severity:** MEDIUM
**Related:** #666

## Problem

#666 turned a multi-line paste into a drain that can hold `sending()` for minutes: per line, a 429 plus a `retry_after` sleep, up to 5 retries, clamped at 60s each. Only the submit BUTTON is disabled during that window. The textarea stays editable, and it is bound to `getDraft(key())` — which every `onProgress` overwrites with the current residue, and which the end-of-submit clear then wipes.

## Failure scenario

Paste 30 lines into a throttled channel (0.5 tokens/s, so roughly a 60s drain). Mid-drain the operator starts typing a reply. On the next acked line their text vanishes, replaced by the residue. Enter is a no-op because `doSubmit` early-returns while `sending()` is true. This repeats every couple of seconds until the drain ends, and then the box is cleared.

## Suggested fix

Make the textarea read-only (or disabled) while `sending()`, or render the residue as a separate "queued" affordance instead of writing it into the draft the operator is typing in. The drain and operator input must not share one buffer.

vjt:
Reopened: PR #780's body carried `Fixes #737`, which auto-closed this at merge. Issues close at the RELEASE CUT, not at merge — the code is on main (`3df70a09`) but there is no tag to pull yet, and self-hosters ship from tags. Label restored to `status:soon`.

vjt:
Shipped in **v0.11.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.11.0

The tag carries `.deb`, `.rpm` and `pkg.tar.zst` assets, so self-hosters have something to pull.

Note this is a **release, not a deploy**: the hosted service is unchanged and still runs the previous build until the code-review queue closes.

Issue #666, opened 2026-08-02
Multi-line paste loses every line after the first throttled one, and cannot be resumed


vjt:
## Summary

Pasting a multi-line body into the composer loses every line after the first
one that hits the send throttle, and leaves the composer in a state where
re-sending duplicates the lines that already went out.

Reported by Sonic and confirmed by Mezmerize: "if I paste more than n lines it
errors and then I cannot even resume with the lines after it".

## Where it breaks

`cicchetto/src/lib/compose.ts:126` — `sendBodyLines` fans a body out to one
PRIVMSG per line:

```ts
for (const line of splitMessageLines(body)) {
  await sendPrivmsg(slug, target, action ? ctcpFrame("ACTION", line) : line);
}
```

There is no `try`/`catch` inside the loop. The first line that gets refused
rejects the promise, the loop unwinds, and **every remaining line is dropped
without ever being attempted**. Nothing retries them and nothing records that
they existed.

The refusal comes from the send door's per-`(subject, network)` bucket:
`config/config.exs:143` sets `capacity: 5, refill_per_sec: 0.5`. In a paste
burst the first 5 lines pass, the 6th gets a 429, and lines 7..n vanish.

## Why "I cannot resume"

`submit` catches the rejection and returns `{ error: friendlyError(e) }`
(`compose.ts`, catch arm around line 1199). The draft is only cleared on the
success path (`compose.ts:1222`), so the composer still holds the **entire**
original body — including the lines that were already delivered to the
channel. Re-sending therefore:

1. duplicates the lines that already went out, and
2. immediately hits another 429, because the bucket is still empty.

From the operator's seat this reads as "the paste is stuck".

## Server side gap

The send door's 429 renders as `json(%{error: "rate_limited"})`
(`lib/grappa_web/controllers/fallback_controller.ex:221`) with **no
`retry-after` header**. The client already knows how to read that header —
`cicchetto/src/lib/api.ts:1383-1386` parses `retry-after` into
`body.retry_after` — and `Grappa.RateLimit.RequestBudget.retry_after_ms/0`
already computes the value for the coarse budget. The send door just never
emits it, so the client has no principled wait interval to honour.

## Proposed scope

Client:

- Make the multi-line fan-out resumable: on refusal, stop sending, keep the
  **unsent remainder** in the draft (not the whole body), and surface how many
  lines went out.
- Honour the retry interval instead of giving up: pace the remaining lines at
  the bucket's refill rate rather than firing them back-to-back.
[...]

Issue #254, opened 2026-07-15
cicchetto: own outbound message not rendered live (query AND channels) on iOS — echo lost when WS subscription not ready


vjt:
## Symptom

A user sends a **private message** (query/DM) and the message **never appears in their own live view** — no error, nothing. The message *is* delivered (the recipient receives it) and **reappears when the app is quit and reopened**, so it was persisted server-side; only the sender's live render is lost.

Reported during iOS dogfooding (mobile PWA, private message — not a channel). The channel path is unaffected.

## Root cause — query-window topic subscription race

The client's own outbound `/msg <target>` echo is delivered via a Phoenix PubSub broadcast on the per-target `(slug, target)` topic. For **channels**, the client already holds a persistent subscription to the channel topic (installed with a self-JOIN line), so the echo arrives and renders. For **query windows**, the `(slug, target)` topic is joined **lazily and reactively** — and the join races the first outbound send:

1. User sends `/msg <target>` → client POSTs the message.
2. Server persists it (`Scrollback.persist_event`) **and** broadcasts the echo on `Topic.channel(subject_label, network_slug, target)` — `lib/grappa/session/server.ex:2497-2502`.
3. The client's query-window join is a `createEffect` that may **not have joined that topic yet** — `cicchetto/src/lib/subscribe.ts:771-826`.
4. Phoenix PubSub fastlane **does not replay** to a not-yet-subscribed socket, so the echo is dropped.
5. The on-join `refreshScrollback(slug, target)` already ran (or runs before the persist is observable), so the own line never lands in the live scrollback.
6. The row is in the DB → it shows up on the next backlog fetch (reload), which is exactly what the user sees.

This exact race is **already documented in-source** at `cicchetto/src/lib/subscribe.ts:799-811`:

> "…the server broadcasts the operator's OWN outbound `/msg <target>` echo on THIS (slug, target) topic … so a spec that opens a DM via /query and immediately composeSends an own line races this subscribe: the echo fastlanes past the not-yet-joined socket and the on-join refreshScrollback already ran, so the own line never renders (marker-target T2, ~2/4 in suite)."

It was previously seen only as a test-flake ("T2, ~2/4 in suite") and worked around for the seam test — but it is a **real production bug**, most likely to bite on the **first message to a freshly-opened query window** and on slower mobile (iOS) connections where the join lag vs. broadcast timing is widest.

## Candidate fix directions (worker to choose)

[...]

PR #906, opened 2026-08-05
cic(#904): a one-deep send queue for the composer

vjt:
Fixes the half of #737's class it never covered: a **single** slow send.

## The defect

`#737`'s guard is keyed on `isDraining`, raised ONLY by the #666 paced
multi-line drain. A lone message on a bad link never raised it, so the
composer and the send machinery shared one buffer: the textarea stayed
writable, Enter was a no-op (ComposeBox's own `sending()` early-returned),
and the end-of-submit clear then wiped the follow-up. Typed, un-sendable,
silently destroyed — and it never even reached history.

## The shape

The composer buffer has exactly one owner at a time, and while a send is in
flight that owner is the OPERATOR.

* `submit` becomes a **pump** around a `dispatchDraft` that never touches the
  buffer. The pump takes the text out **at dispatch** (not on the 201) and
  hands it back on any failure — which is why ~40 dispatch arms can keep
  returning a bare `{error}` and still preserve the draft.
* A second Enter parks its line in a slot **exactly one deep**; the pump
  sends it on the ack. Enter means something again during a slow send.
* A third is **refused visibly**: the textarea goes `readOnly` (never
  `disabled` — that blurs and collapses the on-screen keyboard, #59).
* Hand-back is a **prepend, never a clobber**, and a failed first send does
  NOT fire the queued second — order beats delivery. The pump hands back from
  a `finally`, so even a throw out of the parse returns the text.
* The multi-line drain keeps the buffer it claimed (#737) and its own residue
  mirroring (#666/#723); it returns `keptBuffer` so the pump does not drop the
  whole paste on top of the remainder.
* In-flight state lives in the **store, keyed on the window** — a
  component-local flag dies on unmount and follows the operator to the wrong
  composer. The #241 spinner is keyed on it now too.

## Deliberately not built

No pending rows, no correlation token. cic never originates state and the
server echo stays the sole render path (#254/#251). Stated cost: between
dispatch and echo the queued line is visible only in history (ArrowUp), with
the send-button spinner as the in-flight affordance.

## Test plan

- [x] `scripts/bun.sh run test` — 4388 passing (10 new store tests, 3 new
      ComposeBox tests)
- [x] mutation-tested, 6 mutations, 6 killed: never-queue (2 red),
      clear-at-ack (5 red), clobbering hand-back (1 red), fire-queued-after-
      failure (1 red), reinstated component-local gate (1 red), readOnly
      without queue-full (1 red)
- [x] biome + `tsc --noEmit` clean
- [ ] `e2e/tests/issue904-one-deep-send-queue.spec.ts` — holds the first send
      POST open via `page.route` and pins all four properties end to end.
      **Not yet run: needs the STACK lane.**

### DIFF
--- a/cicchetto/e2e/tests/issue904-one-deep-send-queue.spec.ts
+++ b/cicchetto/e2e/tests/issue904-one-deep-send-queue.spec.ts
@@ -1,0 +1,110 @@
+// #904 — the composer's one-deep send queue, in a real browser.
+//
+// The defect: on a slow link a SINGLE in-flight send shared one buffer with
+// the operator. The textarea stayed writable, Enter was a no-op (ComposeBox's
+// own `sending()` early-returned), and when the 201 finally landed the
+// end-of-submit clear wiped the follow-up the operator had typed meanwhile —
+// silently, and without it ever reaching history.
+//
+// The fix moved the in-flight truth into the store, keyed on the window: the
+// message leaves the composer AT DISPATCH, a second Enter parks its line in a
+// slot exactly one deep, the pump sends it on the ack, and a third is refused
+// visibly (readOnly) instead of being eaten.
+//
+// Why an e2e and not only jsdom: three of the four properties below are
+// browser-rendered state on the real send path — the textarea emptying at
+// dispatch, the readOnly refusal, and the queued line actually reaching the
+// server in order once the first ack lands (`feedback_cicchetto_browser_smoke`
+// — jsdom can observe none of that against a live grappa + leaf).
+//
+// The slow link is synthesized with `page.route`: the FIRST send POST is held
+// open until the spec releases it, which is exactly the ~5s mobile ack the
+// issue was reported on, made deterministic. Everything else — the queueing,
+// the dispatch, the ordering — is the real client and the real server.
+//
+// Subject-/platform-agnostic (store state + one `readOnly` binding), so one
+// desktop chromium run with the registered seed is sufficient.
+
+import { composeTextarea, loginAs, scrollbackLine, selectChannel } from "../fixtures/cicchettoPage";
+import { AUTOJOIN_CHANNELS, getSeededVjt, NETWORK_NICK, NETWORK_SLUG } from "../fixtures/seedData";
+import { expect, test } from "../fixtures/test";
+
+const CHANNEL = AUTOJOIN_CHANNELS[0];
+
+// POST /networks/{slug}/channels/{channel}/messages (no query string); the GET
+// pagination variant carries `?before=`/`?after=` and is let through untouched.
+const SEND_POST_RE = /\/channels\/[^/]+\/messages(\?|$)/;
+
+test.setTimeout(60_000);
+
+test("#904 — a slow send frees the composer, queues ONE more, refuses the third, and delivers both in order", async ({
+  page,
+}) => {
+  if (!CHANNEL) throw new Error("AUTOJOIN_CHANNELS empty");
+  const vjt = getSeededVjt();
+  await loginAs(page, vjt);
+  await selectChannel(page, NETWORK_SLUG, CHANNEL, { ownNick: NETWORK_NICK });
+
+  const tag = crypto.randomUUID().slice(0, 8);
+  const first = `q904 ${tag} first`;
+  const second = `q904 ${tag} second`;
+  const third = `q904 ${tag} third`;
+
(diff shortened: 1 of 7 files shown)


vjt:
## What I refuse to claim about this change

The negative space, not a summary. Everything here is either *measured and
absent* or *asserted from code structure* — I say which.

### Asserted from structure, NOT measured

* **Slash commands through the queue.** Every store test queues a PRIVMSG. The
  pump is kind-agnostic by construction (it queues the raw draft and re-enters
  the same `dispatchDraft`), so `/join` behind a slow `/whois` *should* work —
  no test says so.
* **The identity-change reset of the outbox.** Added for symmetry with #737's
  `drainingKeys` reset, on the same reasoning (a send that never settles would
  otherwise leave the composer refusing writes across a logout/login).
  Untested. #737's own reset has no test either; I did not fix that.
* **The refusal BANNER.** I pinned the return value (`{error}`) and the
  `readOnly` textarea. Nothing asserts the red banner actually renders that
  copy in the DOM.
* **Window-switch mid-flight.** The store tests prove the keys are independent;
  no test mounts a ComposeBox, switches windows during a send and checks the
  spinner followed the right one. That is precisely the bug the store-keying
  exists to prevent, and it is argued, not demonstrated.

### A real window I found by inspection and did NOT close

`sendPacedBody` claims the drain lock **synchronously**, so for the plain
privmsg/`/me` path there is no moment between the queue slot emptying and the
lock being taken — the operator cannot type into the gap. **The `/msg` arm
breaks that**: it `await`s `ensureQueryTopicJoined` (`compose.ts:884`) *before*
`sendPacedBody`. So for a QUEUED multi-line `/msg` whose target query window
already holds a draft (`home` falls back to the source window, #723), the
operator can type into the source during the join await and the first residue
write clobbers it. Narrow (queued + multi-line + `/msg` + busy query + typing
during the join) but it is the same class this issue is about, and the
invariant that saves the other paths — *claim before you await* — is not
protected by any test.

### The empty-Enter divergence I introduced

With the queue full the pump refuses **before** the empty check, so Enter on an
(unwritable, therefore empty) textarea now surfaces a sticky red error instead
of the documented "an empty submit is a no-op regardless of session state".
Informative, but it is a divergence, it was a choice, and it is unmeasured.

### What the new e2e does NOT cover

Chromium desktop, one registered subject, one channel, happy path only:
**no** mobile/WebKit, **no** reconnect or WS drop mid-queue, **no** multi-window
(send in `#a`, switch to `#b` while it flies), **no** failure/hand-back in a
real browser, **no** interaction with the #666 multi-line drain, and **no** real
race — the "slow link" is a deterministically held route, so what is pinned is
ordering, not concurrency. Browser-side proof that a failed single send returns
its text to the composer rests on the PRE-EXISTING #342 spec, not on anything I
wrote.

### Things I did not exercise against #666 / #737

`isDraining` and its refusals are untouched, and the #737 + #723 pins all pass
unchanged — but the *combination* (a paced 429 ladder running as the QUEUED
submission) is exercised by no test in either suite. The multi-line drain path
was verified only through the existing suite; I added one regression guard that
a failed drain leaves only its residue, and that guard **passes on the pre-fix
code too** — it constrains nothing about the queue, it only stops the queue from
breaking the drain.

### Where I think the framing is wrong

* **The reconciliation constraint is vacuous for this cut.** "Pending rows in
  their own zone make a body match sufficient; a correlation token is only
  needed if pending rows interleave" presupposes pending rows exist. They
  cannot: cic never originates state and the server echo is the sole render
  path (#254/#251). So the token was not "not needed yet" — the question does
  not arise until an invariant is deliberately lifted. Cost of that, stated
  plainly: between dispatch and echo a queued line is visible **nowhere** but
  history.
* **"At least the block" is not the weaker option it looks like.** The issue's
  minimum fallback (widen `readOnly` to any in-flight send) is what the queue
  does anyway once the slot is full — the queue is that fallback plus one slot,
  not a different design.
* Everything else in the issue's diagnosis I checked held up, including the
  subtle part: `state` captured before the `await` is why `pushHistory` stored
  the *sent* line while the clear wiped the *current* draft.

### One thing I got wrong myself

My first version of the pump evaporated the draft on a throw out of
`parseSlash` (it sits outside `dispatchDraft`'s catch) — this issue's own
defect, wearing a stack trace, introduced by the fix for it. Caught by asking
what happens on the exit nobody plans for, reddened with a test, then fixed in
a `finally`.

Issue #908, opened 2026-08-05
TRACE reply numerics (200-209, 261) misrouted to per-token query windows instead of $server


vjt:
`/trace` replies are param-scanned into per-token windows instead of `$server`: the TRACE numeric family (200–209, 261) is missing from `@active_numerics`.

Reported from a live `/trace <server>` against Azzurra (bahamut). Measured on the prod DB and against `origin/main` (`c5426216`) — not inferred.

## Symptom

A single `/trace nightwish.azzurra.chat` produced **three separate query windows** named `Operator`, `Server` and `Class`, each holding one `:notice` row whose body is a bare number:

```
query_windows: (8628, "Operator"), (8629, "Server"), (8630, "Class")  — all opened_at 2026-08-05T23:31:18Z

messages (meta.raw_params verbatim):
  204  channel=operator  body="0"               ["vjt","Operator","1","vjt[~user@host]","0"]
  206  channel=server    body="94086917687820"  ["vjt","Server","0","8S","172C","raptor.azzurra.chat[unknown@0::0]","*!*@nightwish.azzurra.chat","94086917687820"]
  209  channel=class     body="2"               ["vjt","Class","0","2"]
  209  channel=class     body="31"              ["vjt","Class","1","31"]
  262  channel=$server   body="End of TRACE"    ["vjt","nightwish.azzurra.chat","End of TRACE"]
```

262 lands correctly only by accident: its `params[1]` is a server name, and `query_candidate?/2` excludes tokens containing a `.`.

## Mechanism

`NumericRouter.numeric_class/1` puts every TRACE numeric in `:scan`, so `scan_params/2` (`lib/grappa/session/numeric_router.ex`) walks the middles and takes the first nick-shaped, non-own, dotless token as a query destination. For the TRACE family that token is the **reply type**, never a target:

| numeric | bahamut format (`src/s_err.c`) | `params[1]` |
|---|---|---|
| 200 RPL_TRACELINK | `Link %s%s %s %s` | `Link` |
| 201 RPL_TRACECONNECTING | `Attempt %d %s` | `Attempt` |
| 202 RPL_TRACEHANDSHAKE | `Handshaking %d %s` | `Handshaking` |
| 203 RPL_TRACEUNKNOWN | `???? %d %s %d` | `????` |
| 204 RPL_TRACEOPERATOR | `Operator %d %s %ld` | `Operator` |
| 205 RPL_TRACEUSER | `User %d %s %ld` | `User` |
[...]

Issue #640, opened 2026-08-02
cicchetto: /ping self-echo opens a query window for the target while the RTT line stays in the source window


vjt:
`/ping <nick>` typed in a channel splits across two windows: the outbound CTCP self-echo lands in (and **creates**) a query window with the target, while the RTT line lands in the source window. Pinging a nick that does not exist leaves a phantom query window behind.

### Measured (staging `grappa.bad.ass`, `0.9.0-654f158f`, browser)

| command, typed in | outbound echo rendered in | reply rendered in |
|---|---|---|
| `/ping <bot>` from `$server` | new query `<bot>` | `$server` — `CTCP PING reply from <bot>: 1082 ms` |
| `/ping <bot>` from `#ccping-test` | same query `<bot>` | `#ccping-test` — `... 1090 ms` |
| `/ping <ownnick>` from `#ccping-test` | new query `<ownnick>` | `#ccping-test` — `... 51 ms` |
| `/ping #ccping-test` from `#ccping-test` | `#ccping-test` | (none — expected) |
| `/ping <nonexistent>` from `#ccping-test` | new query `<nonexistent>` | same query — `No such nick/channel` |

After five pings the window list held three query windows nobody opened, each containing a single `→ CTCP PING <token> to <nick>` line.

### Why

`compose.ts` `case "ping"` sends the frame with `sendPrivmsg(networkSlug, cmd.target, ctcpFrame("PING", token))`. The server echoes it back as a normal own-PRIVMSG addressed to `cmd.target`, so ordinary DM routing files it under the target's window and opens one if absent. The RTT line is synthesized separately by `subscribe.ts` `maybeConsumePingReply`, keyed on `hit.sourceKey` — the window `/ping` was typed in. The two halves are routed by two different rules.

The intent is already written down and says the echo belongs in the source window — `ScrollbackPane.test.tsx` (the `ctcpSelfEcho` case): *"The render layer shows `→ CTCP VERB [args] to <target>` instead of the raw `\x01` the operator would otherwise see **in the window they typed in**"*. The fixture pins `channel: "#grappa"` while the target is also `#grappa`, so target-window and source-window coincide and the mismatch stays invisible in the test.

### Expected

The outbound CTCP self-echo renders in the window the command was typed in, next to the RTT line it belongs to, and never opens a window for the target. Same for `/ctcp <target> <VERB>`, which has the identical shape.

### Notes

[...]

Issue #424, opened 2026-07-26
STATS replies lose every middle param: generic numeric persist stores only the trailing param


vjt:
Reported from chan by an oper using `/stats` from cicchetto: the command reaches the network fine, but the reply shows **only the last argument of each numeric** — every middle param is gone.

That is not a rendering gap. The data never reaches the client.

## Root cause (verified on `origin/main` @ `425a6b60`, present in the prod tag `v0.5.0` too)

The generic numeric persist path in `lib/grappa/session/server.ex` (~:2585) is:

```elixir
numeric_code = numeric_code(msg)
trailing = List.last(msg.params)
...
attrs = %{..., body: trailing, meta: %{numeric: numeric_code, severity: severity}}
apply_effects([{:persist, :notice, attrs}], state_with_labels)
```

`body` is the trailing param; `meta` carries only `numeric` + `severity`. **The middle params are not persisted anywhere** — unlike the unhandled-*verb* path (`event_router.ex` `persist_raw_event/3`), which does keep `raw_params` in meta for `renderRawEvent` to use.

STATS is exactly the family whose payload lives in the middle params, e.g.:

- `218 <nick> Y <class> <pingfreq> <confreq> <maxlinks> <sendq>` → only the sendq survives
- `213 <nick> C <host> <*> <server> <port> <class>` → only the class survives
- `219 <nick> <letter> :End of /STATS report` → fine by luck (payload is trailing)

`lib/grappa/session/numeric_router.ex:140` deliberately routes the whole `211..219 ++ 240..250` range to `{:server, nil}` (#184, so the stats letter stops being read as a DM target), but STATS has **no delegated bundle handler** — unlike LUSERS (251–255, 265, 266), MOTD (375/372/376/422/402), INFO (371/374), VERSION (351) and BANLIST (367/368), which are in `@delegated_numerics` and get accumulated by `EventRouter` into a `{:server_reply, …}` / bundle effect.

So STATS falls through to the generic persist above and loses its payload on the way to the DB.

## Impact

`/stats y`, `/stats c`, `/stats i`, `/stats k`, `/stats o`, `/stats l` are unusable from the client — which is the difference between "an oper can run the network from cicchetto" and "an oper still has to keep irssi open". #155 added the verb and #184 fixed its routing; this is the remaining leg.

## Suggested direction (implementer's call)

Two shapes, matching precedents already in the codebase:

[...]

Issue #221, opened 2026-07-12
grappa: Libera upstream query gaps — WHOIS numerics, on-connect usermodes, and /who <mask> returns no feedback


vjt:
Observed against a **Libera.Chat** upstream (post multi-network / phase-7): grappa mis-parses several server numerics coming from the upstream network, and some query commands return nothing at all.

## Three distinct gaps

**1. WHOIS numerics not parsed correctly**
The `RPL_WHOIS*` reply block from Libera is not being parsed/relayed correctly. Libera emits a richer/differently-ordered WHOIS numeric set than our testnet bahamut (e.g. account/`RPL_WHOISACCOUNT` 330, `RPL_WHOISSECURE` 671, bot flag, extended idle/signon, cloak lines), and our parser does not handle them faithfully.

**2. Allowed usermodes on connect not handled**
The usermodes the upstream advertises/allows at registration time (the mode set/echoed on connect) are not parsed correctly, so grappa does not track the client's effective/permitted usermodes against a Libera upstream.

**3. `/who <hostmask>` returns no feedback at all**
`WHO *hostmask*` produces **no response whatsoever** — no `RPL_WHOREPLY` (352), no `RPL_ENDOFWHO` (315), nothing. This looks broader than Libera-specific parsing: the WHO query with a mask is either not forwarded upstream or its reply/terminator numerics are dropped, leaving the client with zero feedback. Not just a mis-parse — a total silence.

## Why it matters
Multi-network shipped today; Libera is a real-world upstream with a different numeric/usermode surface than our bahamut testnet. Parsing has been developed/tested primarily against bahamut, so Libera-specific numerics slip through. WHO-by-mask silence affects any network.

## Suggested direction
- Audit the numeric handler against Libera's actual `RPL_WHOIS*` set and make unknown/extra WHOIS numerics pass through instead of being dropped/misrouted.
- Handle the on-connect usermode advertisement generically (do not assume bahamut's mode letters/ordering).
- Trace a `WHO <mask>` end-to-end: confirm it reaches the upstream and that `352`/`315` are relayed back to the client.

Reported via channel request against the live Libera upstream. No repro transcript captured yet — capturing a raw Libera WHOIS + connect trace + a `WHO <mask>` round-trip is the first useful step.

---

## Extension — validate against solanum source + real-network CI

Deliberately widening scope to do this correctly rather than guessing numerics.

[...]

PR #909, opened 2026-08-06
irc(#908): TRACE's reply TYPE is not a routing destination

vjt:
Fixes the misrouting reported in #908. Refs #184, #640, #424.

## What was wrong

One `/trace nightwish.azzurra.chat` against Azzurra opened three query
windows named `Operator`, `Server` and `Class`. Every TRACE reply carries
its reply KIND in `params[1]` — `Link`, `Attempt`, `Handshaking`, `????`,
`Operator`, `User`, `Server`, `<newtype>`, `Class`, `File` — and
`scan_params/2` takes the first nick-shaped, dotless, non-own middle param
as a routing destination. It routed the kind.

TRACE is server-directed by definition, so 200–210 + 261–262 join
`@active_numerics`: the third family after #184's STATS letter and UX-4
bucket I's connect-storm metadata.

## Decisions worth arguing with

**262 is covered even though it already worked.** Its `params[1]` is the
traced server's name, saved only by the `.`-exclusion in
`query_candidate?/2` — the correct outcome rested on that server happening
to be spelled with a dot. The test pins the DOTLESS spelling instead, so
the decision no longer depends on a naming accident.

**207/210 fix nothing today** — `NULL` in bahamut's table. Covered because
every reading of those slots is still server-directed (RPL_TRACESERVICE,
RPL_TRACERECONNECT, ircu's RPL_STATSHELP), and a contiguous range leaves no
hole for a network that does emit them.

**Not delegated.** TRACE has no cic UI, and delegating without an
EventRouter handler is the phantom-delegation silent drop already paid for
once with LIST (no-silent-drops B6.1 HIGH-3). `$server` notice + #424's
`meta.raw_params` is the whole display story. The bahamut replies carry no
trailing param, so the persisted body is a bare `0` / `31` /
`94086917687820` — a display concern #424/#569 already own, neither caused
nor repaired here.

**A moduledoc claim was false and is corrected.** It said the param scan is
safe by default — "at worst a row lands on `$server` instead of a more
specific window". That holds for the channel branch. The query branch's
failure mode is a row landing in the wrong CONVERSATION, and before #640 it
MINTED a ghost window that leaked into Archive via `list_archive`'s
`COALESCE(dm_with, channel)`. #640 is a backstop, not a correct decision: a
`/trace` issued while a window named `Operator` is open still misroutes.

## Why this is a third patch and not the root-cause fix

Recorded in the moduledoc and DESIGN_NOTES. The scan asks a syntactic
question ("could this token be a nick?") of a semantic slot, and there is no
syntactic discriminant: `params[1]` is a target for the error class, a
channel for the channel class, and DATA for every report class. The table
can only be keyed by the CODE — which both lists already are. A deny list
and an allow list differ ONLY in which side the unknown falls on, and today
it falls on the guessing side. The deny list now stands at 43 codes against
the two or three the query branch genuinely serves (401, plus the legacy
2-param shape); that ratio is the argument for inverting the default.

Left undone deliberately: inverting changes behaviour for every numeric in
neither list, and #221's WHOIS-leg guard is *defined* as "a `:scan`-class
numeric" and would have to be re-sited first. That is a design change, not a
bug fix.

## Verification

Red observed before green, with the fix reverted and the tests in place:
7 failures, each for the right reason — `{:query, "Operator"}`,
`{:query, "Server"}`, `{:query, "Class"}`, `{:query, "Link"}`,
`{:query, "File"}`, `{:query, "services"}`, plus the deny-list property.

`scripts/check.sh` rc=0 on the full tree: Credo `5119 mods/funs, found no
issues`; `8 doctests, 54 properties, 5272 tests, 0 failures`; Dialyzer
`Total errors: 0` / `done (passed successfully)`; Sobelow, doctor, bats all
green.

The six named tests use the `meta.raw_params` measured on prod verbatim.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -30464,3 +30464,48 @@ the drain lock: a ComposeBox-local flag dies on unmount (home / mentions /
 $list, the desktop↔mobile swap) and follows the operator to the wrong
 composer. The #241 send spinner is keyed on it too now, so it stops lying
 after a window switch.
+
+## 2026-08-06 — #908: TRACE is a report family, and the deny list is a patch on the wrong question
+
+`/trace nightwish.azzurra.chat` against Azzurra opened three query windows
+named `Operator`, `Server` and `Class`. `NumericRouter.scan_params/2` walks a
+numeric's middle params and takes the first nick-shaped, dotless, non-own
+token as a destination; every TRACE reply puts its **reply TYPE** in
+`params[1]` (`Link`, `Attempt`, `Operator`, `Server`, `Class`, `File`, …), so
+the scan routed the reply KIND. 200–210 + 261–262 join `@active_numerics`.
+
+**262 is covered even though it already worked.** Its `params[1]` is the
+traced server's name and `query_candidate?/2` excludes tokens containing a
+`.` — so the correct outcome rested on that server happening to be spelled
+with a dot. A family-wide rule cannot depend on a naming accident; the
+dotless spelling is pinned by test. 207/210 are `NULL` in bahamut and fix
+nothing today, but every reading of those slots (RPL_TRACESERVICE,
+RPL_TRACERECONNECT, ircu's RPL_STATSHELP) is server-directed, so the
+contiguous range costs nothing and leaves no hole for a network that emits
+them.
+
+**The moduledoc was lying about the failure mode, and that was the real
+finding.** It claimed the scan is a safe default — "at worst a row lands on
+`$server` instead of a more specific window". True of the channel branch;
+false of the query branch, whose failure mode is a row landing in the wrong
+CONVERSATION, and — before #640's `resolve_numeric_query_window/2` — MINTING
+a ghost window that then leaked into Archive via `list_archive`'s
+`COALESCE(dm_with, channel)`. #640 is a backstop, not a correct decision: a
+`/trace` issued while a query window named `Operator` is open still misroutes
+today.
+
+**Why a third patch instead of the root-cause fix.** The scan asks a
+syntactic question ("could this token be a nick?") of a semantic slot.
+Whether `params[1]` is a destination is a per-numeric FACT with no syntactic
+discriminant — a target for the error class, a channel for the channel class,
+DATA for every report class. So the table can only be keyed by the CODE,
+which both lists already are: a deny list and an allow list differ ONLY in
+which side the unknown falls on, and today the unknown falls on the guessing
+side. Three families have now been patched in (#184 STATS, UX-4 bucket I
+connect-storm, #908 TRACE) and the deny list stands at 43 codes against the
+two or three the query branch genuinely serves (401, plus the legacy 2-param
+shape). That ratio is the argument for inverting the default. Not done here:
+it changes behaviour for every numeric in neither list, and #221's WHOIS-leg
+guard is defined as "a `:scan`-class numeric" and would have to be re-sited
+first. Deny-list entries stay the right size of cut for one reported family;
+they are not a substitute for the inversion.
--- a/lib/grappa/session/numeric_router.ex
+++ b/lib/grappa/session/numeric_router.ex
@@ -23,13 +23,15 @@ defmodule Grappa.Session.NumericRouter do
      the recorded `origin_window` wins over the param scan and active-deny —
      perfect-correlation path (but NOT over delegation, per #276 above).
 
-  3. **Active deny list** (`@active_numerics`): a small set of numerics
-     whose params look nick-shaped but the "nick" is not a routing
-     destination — it's the rejected nick (433/432), the unknown command
-     name (421), the offending command's argument list (461), or an
-     ack (437). These ALWAYS go to `{:server, nil}`. Without this
-     deny list, the param-scan below would happily route 433's
-     "BLEH-as-nick" to a query window.
+  3. **Active deny list** (`@active_numerics`): numerics whose params look
+     nick-shaped but the "nick" is not a routing destination — it's the
+     rejected nick (433/432), the unknown command name (421), the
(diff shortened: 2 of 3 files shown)


vjt:
## What I refused to claim

### Asserted from code structure, never measured

- **Every param shape in the sweep below except TRACE's.** The TRACE shapes
  are vjt's verbatim `meta.raw_params` off the prod DB. Everything else is
  RFC 1459/2812 + IRCv3 spec knowledge. I have no copy of bahamut's
  `s_err.c` — I did not read the source the issue quotes, and I did not
  capture a live reply. If a shape below is wrong, the finding evaporates.
- **207/210 "cost nothing".** That rests on my claim that every ircd reading
  of those slots is server-directed. Unverified. If some ircd puts a nick
  there, I have just deny-listed a real destination.
- **"43 denied codes against the two or three the query branch genuinely
  serves."** The 43 is counted exactly (20 STATS + 13 TRACE + 10 literals).
  The "two or three" is NOT — I read the tests, I did not enumerate every
  numeric that currently reaches the query branch. The ratio's numerator is
  measured; its denominator is an estimate, and the argument leans on it.
- **900 RPL_LOGGEDIN.** I verified in code that the client forwards *every*
  parsed numeric (`client.ex` `process_line/2`) and that 900 sits in neither
  list with no dedicated `handle_info`. I did NOT verify that Azzurra emits
  900 at all, nor that the SASL account ever differs from the connected nick
  — and when they match, `nick_eq?/2` skips it and there is no bug.

### What the tests do not cover

- **No end-to-end binding.** `NumericRouter` is pure, so these tests pin the
  DECISION, not the outcome. Nothing here proves no row is persisted into an
  open query window named `Operator`. I asserted #640 degrades to `$server`;
  I did not exercise it for TRACE.
- **The test's deny-list is a hand-copied mirror** of the module attribute,
  and my change perpetuates that. The drift is silent and one-directional: a
  family added to the module without touching the test is simply unguarded.
  Closing it needs the set exposed from the module — a production change made
  for a test, which is why I left it.
- **207/210 are exercised only by the property test's synthetic params.**
  There is no realistic reply shape to write, because bahamut emits none.

### Where the framing — mine, the issue's, and the dispatch's — was wrong

- **"Three instances is a class" undercounts it.** The sweep found two more
  already broken, and both contradict an explicit moduledoc claim that
  321/322/323 "route via the default `{:server, nil}` path":
  - **321 RPL_LISTSTART** `[nick, "Channel", "Users Name"]` → the single
    candidate is `Channel` → `{:query, "Channel"}`.
  - **322 RPL_LIST** `[nick, "#chan", "5", "topic"]` → the **channel** branch
    wins → one notice row in *every listed channel's* window. Shape verified
    against `parse_list_entry/1` in `server.ex`, not guessed.

  These are normally consumed by the `directory_refresh` handler, but that
  handler nils itself on `:directory_refresh_timeout`, after which late
  frames fall straight into the scan. So the class is at least five, and the
  moduledoc is wrong about a family it names explicitly.
- **The dispatch's axis was deny-list-vs-something-else. The real axis is
  which side the UNKNOWN falls on.** Deny list and allow list are both
  code-keyed tables; they are the same mechanism. Nothing about the deny
  list's *shape* is the smell — its *default* is.
- **The issue's reason for including 262 is weak.** "Family completeness"
  does not explain anything; the actual reason is that the existing save
  depends on the traced server being spelled with a dot. That reframing is
  what produced the only 262 test that fails without the fix — the dotted
  spelling the issue quotes passes either way and would have been a test that
  constrains nothing.
- **The issue's own proposed comment says "200–209" while its code says
  `200..210`.** I used the range and said why.
- **#640 makes the reported symptom unreproducible.** The three ghost windows
  cannot be observed on `main` today. These unit tests are the only oracle
  for this bug; no e2e or integration spec would have caught it, and none is
  added here.

### Rest of the sweep — reported, not fixed, all reachable

`/quote <line>` is a documented escape hatch (`send_raw`, `server.ex:1803`)
and `/trace` has no dedicated handler, so it reached prod that way. Every
family below is therefore operator-reachable, not theoretical. Predicate:
`^[A-Za-z\[\]\\`_^{|}][\w\[\]\\`_^{|}-]{0,29}$` — never a leading digit,
never a dot.

| numeric | why `params[1]` is nick-shaped | confidence |
|---|---|---|
| 321 / 322 | above | code-verified |
| 900 RPL_LOGGEDIN | `params[2]` is the account name; hit only when it differs from the nick | forwarding verified, reply not |
| 436 ERR_NICKCOLLISION | `[nick, "collided", "…KILL"]` — twin of 432/433/437, which *are* denied | RFC shape |
| 472 ERR_UNKNOWNMODE | `[nick, "z", "is unknown mode char"]` — a bare letter, exact #184 twin | RFC shape |
| 257/258 RPL_ADMINLOC1/2 | 2-param free text; hit iff the admin line is one dotless word | conditional |
| 410 ERR_INVALIDCAPCMD | the CAP subcommand token | spec shape |
| 303, 732, 704–706, 908 | single-target blob is a bare nick; grappa never sends these, but `/quote` does | dormant |

Not re-counted: STATS 220–239, already flagged as a known gap in the
moduledoc.

### Out of band

`scripts/check.sh` exits 0 while printing four advisories including a HIGH
(`cowlib 2.18.0`, EEF-CVE-2026-59248) and `Found packages with security
advisories`. CLAUDE.md says CVE-flagged deps fail the build immediately;
they do not. Blast radius is small — `cowboy`/`cowlib` arrive via
`{:bypass, only: :test}` → `plug_cowboy`, and prod runs Bandit — so this is
a policy/gate discrepancy, not a prod exposure. Not mine to fix in a routing
PR.

### Time

Not meaningfully blocked: I wrote the fix, the tests, the docs and the whole
sweep before asking for COMPILE, so the wait cost a turn boundary and no
work.

Issue #913, opened 2026-08-06
cicchetto/iOS: rail actions menu opens into the safe area — top row hidden behind the status bar, no scroll (residue of #588)


vjt:
## Symptom

On iPhone 17 Pro (PWA, keyboard up), opening the rail **actions** menu from the burger menu puts the topmost row (`home`) **behind the status bar** — it overlaps the battery/clock indicators — and the menu **does not scroll**, so that row cannot be brought into view.

Reported on IRC with a screenshot: the `home` row sits flush at the physical top of the screen, with the battery pill drawn over it.

## Why it happens

This is the residue of #588, not a regression of it. #588 fixed the sibling half of the bug (the upward-opening menu grew off the top of the screen because CSS capped it at `var(--viewport-height)`, the *whole* viewport, so `overflow-y: auto` never engaged). The JS cap it introduced is correct in principle but measures against the wrong origin.

`cicchetto/src/RailActions.tsx`:

```ts
const RAIL_MENU_TOP_GAP = 8;
...
setMenuMaxHeight(spaceAbove(rootRef.getBoundingClientRect().top, RAIL_MENU_TOP_GAP));
```

with `spaceAbove` in `cicchetto/src/lib/menuPosition.ts`:

```ts
export function spaceAbove(anchorTop: number, gap: number): number {
  return Math.max(0, anchorTop - gap);
}
```

`getBoundingClientRect().top` is measured from the **layout viewport** origin. `cicchetto/index.html` sets

```
content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no, viewport-fit=cover, interactive-widget=resizes-content"
```

so with `viewport-fit=cover` that origin is the **physical top edge of the display**, underneath the status bar / notch — not the top of the safe area. `RAIL_MENU_TOP_GAP` is a bare `8`, which is the "breathing room" constant, not a safe-area allowance (the comment above it does mention notch/status bar, but 8px does not clear one: `env(safe-area-inset-top)` is ≈ 59px on a 17 Pro).

Both reported symptoms follow from that single off-by-an-inset:

[...]

PR #916, opened 2026-08-06
cic(#913): cap the rail actions menu against the SAFE top edge

vjt:
Fixes the rail actions menu opening into the iOS safe area (#913): the topmost
row rendered behind the status bar, and the menu would not scroll to bring it
into reach.

## What was wrong

Residue of #588, not a regression of it. #588 replaced a cap of the WHOLE
viewport with a cap of the space above the launcher — the right cap, taken
from the wrong origin. `getBoundingClientRect().top` is relative to the layout
viewport, whose origin under `viewport-fit=cover` is the PHYSICAL top of the
display. `RAIL_MENU_TOP_GAP = 8` was silently being asked to clear a ~59px
notch.

## The fix, and why it is not the one the issue proposed

The issue proposed reading the inset in JS via
`getComputedStyle().getPropertyValue("--safe-area-inset-top")`. I did not do
that. An unregistered custom property's computed value is a token stream, not
a length — `getPropertyValue` can return the literal `env(safe-area-inset-top)`,
whose `parseFloat` is `NaN`, which the proposal's own `|| 0` fallback would
turn into a fix that looks applied and does nothing.

Split by who can know what instead:

* **JS** publishes the one number CSS cannot measure — the anchor's distance to
  the viewport top, gap deducted — as `--rail-menu-space-above`.
* **CSS** subtracts the one value JS cannot reliably resolve,
  `var(--safe-area-inset-top)`, and clamps with `max(0px, …)`.

A pure-CSS cap was considered and is not available: no CSS construct yields an
element's distance to the viewport top, the viewport-unit forms are exactly the
whole-viewport cap #588 removed, and the anchor is not viewport-pinned on
desktop. The JS measurement re-runs on `resize` + `visualViewport` resize,
which is where rotation lands too — now load-bearing, since the inset itself
changes with orientation.

The `0px` fallbacks are load-bearing rather than defensive: an `env()` whose
variable the engine does not know invalidates the whole declaration, which
would drop the cap entirely and hand back the #588 overflow.

## Second symptom, second cause — and what is NOT measured about it

vjt reported TWO symptoms. The occlusion is the inset, straightforwardly. The
no-scroll half is a consequence in the reported configuration (a cap one inset
too generous means the rows fit, so `overflow-y: auto` never engages) — but it
has an independent blocker the issue does not mention: `.rail-actions-menu`
carried no `touch-action` carve-out, while it holds the overlay lock and
therefore sits under the `touch-action: none` blanket, and its action rows are
the hit-test target across most of its area. Capping correctly makes the menu
overflow; the carve-out makes the overflow pannable.

**Stated plainly: this half is pattern-derived, and its necessity is NOT
measured.** Scroller-level `pan-y` is near-universal in that stylesheet
(`.settings-drawer`, `.archive-modal`, `.image-upload-modal`, `.admin-pane`,
`.who-modal-body`, `.home-pane` — five of them under a comment naming this
exact trap). The DESCENDANT form, which is what I added, has exactly **one**
precedent: `.members-pane *`. That one is the instructive case — UX-6 bucket A
v2 added it *after* the scroller-only version had shipped and failed. An
earlier draft of this PR claimed three surfaces carried the descendant pair;
that was wrong, and the correcting commit says so. I cannot claim the carve-out
was necessary here, nor that it is sufficient. It ships because the symptom was
reported and a cap-only fix may leave it standing.

## Test plan

- [x] `scripts/bun.sh run test` — 4395 passed / 241 files
- [x] `scripts/bun.sh run check` — rc 0; standalone `tsc --noEmit` rc 0
- [x] 7-way mutation harness against the new guards — all 7 killed, none survived
- [ ] e2e — NOT run locally, by design: this PR's `integration` job runs the full
      suite, so `issue913-rail-menu-safe-area.spec.ts` and the #588 guard in
      `issue500-rail-launcher-overflow.spec.ts` both execute there. Expect the
      e2e count to be +1 and the new spec to appear by name.
- [ ] on-device verification on a notched iPhone — not reproducible in any gate
      we run; the felt behaviour remains vjt's to confirm

### DIFF
--- a/cicchetto/e2e/tests/issue913-rail-menu-safe-area.spec.ts
+++ b/cicchetto/e2e/tests/issue913-rail-menu-safe-area.spec.ts
@@ -1,0 +1,107 @@
+// #913 — the rail actions menu must cap itself against the SAFE top edge, not
+// the physical one.
+//
+// Sibling of the #588 guard in issue500-rail-launcher-overflow.spec.ts. #588
+// proved the menu stops growing off the top of the screen; it asserts the
+// topmost row lands at y >= 0. #913 is the residue: under `viewport-fit=cover`
+// y = 0 is the PHYSICAL top of the display, behind the status bar, so a row at
+// y = 8 satisfied #588 while sitting under the clock — and the cap being one
+// inset too generous, the rows still fit, `overflow-y: auto` never engaged, and
+// the occluded row could not be scrolled to.
+//
+// THE MEASUREMENT PROBLEM. Playwright does not synthesize
+// `env(safe-area-inset-*)` on any engine we run: it resolves to 0, so the real
+// on-device geometry is NOT observable here and no assertion about a 59px
+// notch can be honest. What IS observable is the WIRING — whether the inset
+// reaches the cap at all, and at what rate. So this spec measures the SAME menu
+// twice, once with the inset at its Playwright value and once with a stubbed
+// non-zero one, and asserts the topmost row moved down by exactly that much.
+//
+// That differential is what makes it non-hollow. Against the pre-fix build the
+// cap ignores the variable entirely, so the delta is 0 and this fails. An
+// absolute `y >= inset` assertion would NOT fail there — it is already true at
+// inset 0, which is precisely how #588 passed while #913 was live.
+//
+// The stub goes through `--safe-area-inset-top` on `:root`, the indirection the
+// fix introduced (JS cannot reliably read `env()` back out of a custom
+// property, so the arithmetic lives in CSS). Being a variable is what makes it
+// stubbable at all; the felt behaviour on a notched device is still vjt's to
+// confirm, and this spec does not claim it.
+
+import { expect, test } from "../fixtures/test";
+import { loginAs, openRailMenu, selectChannel } from "../fixtures/cicchettoPage";
+import { AUTOJOIN_CHANNELS, getSeededVjt, NETWORK_NICK, NETWORK_SLUG } from "../fixtures/seedData";
+
+const CHANNEL = AUTOJOIN_CHANNELS[0]; // #bofh — vjt's seeded autojoin channel
+
+// Short enough that the action rows cannot all fit above the bottom-pinned
+// launcher, so the menu is pinned to its cap rather than to its content height.
+// A tall viewport would let the menu fit, the cap would stop being the binding
+// constraint, and the delta below would be 0 on a FIXED build too — the #588
+// trap, restated. Mirrors the sibling spec's viewport for the same reason.
+const VIEWPORT = { width: 1280, height: 200 };
+
+// A plausible notch: iPhone 17 Pro reports ~59px. The exact number does not
+// matter to the assertion (it is a delta), only that it is non-zero and large
+// enough to dwarf sub-pixel layout noise.
+const STUB_INSET = 59;
+
+// No peers needed: menu overflow is a function of row count vs the space above
+// the launcher, independent of the member list (the #588 spec's own note).
+test.setTimeout(90_000);
+
+test("#913 — the safe-area inset shrinks the rail menu cap, pushing its top row clear", async ({
+  page,
+}) => {
+  const vjt = getSeededVjt();
+  await loginAs(page, vjt);
+  await selectChannel(page, NETWORK_SLUG, CHANNEL, { ownNick: NETWORK_NICK });
+
+  await page.setViewportSize(VIEWPORT);
+  await openRailMenu(page);
+
+  const menu = page.locator(".rail-actions-menu");
+  await expect(menu).toBeVisible();
+  const topRow = menu.locator("[data-testid='mobile-panel-home']");
+  await expect(topRow).toBeVisible();
+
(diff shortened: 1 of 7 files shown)


vjt:
## What I refuse to claim

**I did not verify this fixes what vjt saw.** No gate in this repo can. Playwright
resolves `env(safe-area-inset-*)` to 0 on every engine in the suite, and jsdom
resolves neither `env()` nor `calc()`. There is no local browser on this host to
measure in either. Every green below is about WIRING; the felt behaviour on a
notched iPhone is unmeasured and stays vjt's to confirm.

**The e2e in this PR has never been run.** It needs a lane and I did not take one
(w1 holds STACK). It typechecks only as far as the bun container allows —
`tsc -p e2e --noEmit` dies on a missing `@playwright/test` type entry point before
reaching my file, so what I have is "no `.js` emitted and no error attributable to
this spec", which is not the same as "it runs". Same for the #588 guard in
`issue500-rail-launcher-overflow.spec.ts`: I did NOT run it.

**So my answer to "don't regress #588" is an argument, not a measurement.** What I
actually measured: `spaceAbove` is unchanged and its 14 unit tests are green
untouched; no test or component anywhere reads the rail menu's inline `max-height`
(grepped), so moving it to a custom property breaks no reader; and mutation M6
(publish `max-height` again, pre-#913 shape) is killed, so the JS half still emits
the same number. What I did NOT measure: that the #588 e2e is still green. The
reasoning is that at inset 0 the new cap arithmetically reduces to the old value,
so the layout is identical — but that is arithmetic, not a run. Please point a lane
at both specs.

## Two claims in the issue I checked, and one I am contradicting

**Contradicting the proposed mechanism.** The issue says to read the inset with
`getComputedStyle(document.documentElement).getPropertyValue(...)`, `|| 0` on
non-notched devices. An unregistered custom property's computed value is a token
stream, not a length, so that call can return the literal string
`env(safe-area-inset-top)` — `parseFloat` → `NaN` → the `|| 0` swallows it → a
fix that is applied, green, and does nothing. **I could not settle empirically
which way it goes here** (no browser, and jsdom has no `env()`), which is exactly
why I chose the shape that does not depend on the answer: the engine substitutes
`env()` in a plain CSS rule, as it already does for the ~40 other insets in that
stylesheet. If vjt has a device to check the `getComputedStyle` behaviour on, the
answer is worth writing down — but the fix should not rest on it either way.

**"One fix closes both symptoms" — half right, and the other half matters.** The
issue's arithmetic is sound for the reported state: with the cap one inset too
generous the rows fit, so `overflow-y: auto` never engages. But that only holds
when the content already fills the cap; the no-scroll symptom is a CONSEQUENCE in
that configuration, not a shared cause. And it has a **second, independent
blocker the issue does not mention**: `.rail-actions-menu` carries no
`touch-action` carve-out. Scroller-level `pan-y` is near-universal in that
stylesheet (`.settings-drawer`, `.archive-modal`, `.image-upload-modal`,
`.admin-pane`, `.who-modal-body`, `.home-pane` — five of them under a comment
naming this exact trap). The DESCENDANT form, which is what the menu needs, has
exactly ONE precedent: `.members-pane *`, added by UX-6 bucket A v2 *after* the
scroller-only version had shipped and failed — touch-action does not inherit, and
iOS elects the gesture consumer from the hit-test target's own value. This menu
holds the overlay lock while open, so it sits under the `touch-action: none`
blanket, and its action rows are the hit-test target across most of its area. I
added the pair. **I cannot claim it was necessary, or that it is sufficient** — it
is pattern-derived from a single precedent, not measured. If vjt would rather it
not ride in this PR, it is three lines and one test to drop.

> **Corrected after posting.** This paragraph originally said three surfaces carry
> the descendant pair “from six documented instances”. Grep says one. The argument
> survives — that one precedent exists *because* the scroller-only form shipped and
> did not work — but the number was inflated, and a reviewer weighing whether the
> carve-out belongs here would have been weighing it. Fixed in the source comments,
> the design note and the PR body too (commit `53da4e5f`).

## Reported, not fixed (dispatch point 4)

`spaceAbove` has exactly ONE consumer (`RailActions`), so #913 is contained. But
`placeAxis` / `computeMenuPosition` (`UserContextMenu`) has the same origin bug and
worse: it clamps `top` against `viewportHeight` from origin 0 with no inset
awareness on EITHER edge, so a context menu opened near the top can land under the
status bar and one near the bottom under the home indicator. The issue flags the
top edge as "worth checking, out of scope"; the bottom edge is not mentioned
anywhere and is the same defect. Not touched here — vjt files it.

## What the guards actually constrain

Seven mutations, all killed, none survived: drop the inset subtraction; drop the
`max(0px, …)` clamp; drop the `:root` variable; drop the touch-action carve-outs;
drop the `resize` re-measure; publish `max-height` again instead of the custom
property; set the gap to 0. The e2e is a DIFFERENTIAL — same menu measured at
inset 0 and at a stubbed 59px, asserting the top row descended by exactly that
much, bounded from above as well so a double-counted inset fails too. An absolute
`y >= inset` assertion would not have caught this: it is already true at inset 0,
which is precisely how #588's guard stayed green for the entire life of #913.

Issue #902, opened 2026-08-05
cicchetto: surface inbound INVITEs as a banner with [Join], drop the greyed :invited pseudo-window


vjt:
Inbound INVITEs are easy to miss entirely. Reported via chat: several people were invited to a channel and nobody noticed.

## Decided shape (owner's call, via chat)

Surface the invite as a **top banner in the existing stacked banner region** — the same surface as the bundle-refresh prompt — reading e.g. `<nick> is inviting you to #chan`, with a `[Join]` action and the standard `×` dismiss. **And drop both of the current surfaces**: the greyed `:invited` pseudo-window (the "ghost channel") AND the INVITE line injected into the `$server` scrollback. The banner replaces them.

Note this **deliberately reverts #482**, which restored that `$server` row on purpose. The reasoning has changed: #482 existed because an invite left no trace anywhere: with a banner carrying the `[Join]`, the `$server` copy is a second place to look for something already in front of the operator. The INVITE row persisted in the *channel's* own buffer stays — that one is history, not notification.

Concretely, that is one new member of the closed set in `cicchetto/src/lib/errorBanners.ts`:

* add `"invite"` to `BANNER_SOURCES` (runtime guards `isBannerSource` / `sanitizeBanners` follow automatically);
* one `activeBanners()` push, `severity: "info"`, `actionHint: { label: "Join", onAction: … }` reusing the same join verb the scrollback `[Join]` CTA calls (`ScrollbackPane.tsx` `onJoinChannel`);
* the state owner stays the source signal — the `window_invited` arm in `lib/userTopic.ts` — per the registry's "derive, don't duplicate" rule. `#120` (sw-registration) is the precedent for how cheap this seam is: one enum member + one gated push.
* stacking order matters: an invite is an offer, not a fault, so it belongs low in the stack, next to `push-optin`.
* multiple concurrent invites need a decision the current registry does not make for us (one entry per channel vs one aggregate entry) — the registry projects a list, so N entries is possible, but N banners stacked is a wall.

Removing the pseudo-window touches `lib/pseudoChannels.ts` (the `:invited` slice), `Sidebar.tsx`, `BottomBar.tsx` (#71 INC-3 renders ONLY the `:invited` slice on mobile — with the pseudo-row gone that mobile surface has nothing left, so the banner must be the mobile answer too), `lib/windowClose.ts` (`dismissPseudoWindow` + the re-emit-on-reload behaviour) and `lib/archive.ts`, plus their specs.

[...]

Issue #511, opened 2026-07-28
cicchetto: dismissing an :invited pseudo-row is client-only — the tab returns after a reload


vjt:
## Symptom

1. Receive an INVITE to a channel you do not want to join — the greyed `:invited` pseudo-row appears in the sidebar / bottom bar.
2. Dismiss it with its ×. The row disappears, as expected.
3. Reload the page. The row is back in the active tabs.

Expected: dismissing an unwanted invite is durable — the row stays gone (or moves to archive), and a page reload does not resurrect it.

Reported on IRC against the production instance.

## Root cause — the × on an `:invited` row never reaches the server

`cicchetto/src/lib/windowClose.ts` has two different closes:

- `closeChannelWindow/2` (× on a **joined** channel) → `postPart(...)` **and** `forceParted(...)`. The server learns about it.
- `dismissPseudoWindow/2` (× on a **pseudo** row: invited / failed / kicked / parked) → `forceParted(...)` only. Purely local:

```ts
export function dismissPseudoWindow(networkSlug: string, name: string): void {
  const sel = selectedChannel();
  if (sel !== null && sel.networkSlug === networkSlug && sel.channelName === name) {
    setSelectedChannel({ networkSlug, channelName: SERVER_WINDOW_NAME, kind: "server" });
  }
  forceParted(channelKey(networkSlug, name));
}
```

Server-side, `window_states[channel]` stays `:invited`. The only code path that clears it is the PART path — `WindowState.set_parted/2`, reached from `PartCleanup.cleanup_local/2` (eager, on the `{:send_part, _}` cast) or from the `{:parted, channel}` upstream-echo effect. A dismissal never goes through either.

The resurrection is then by design of #482: `Session.Server`'s snapshot includes
`invited_windows: WindowState.invited_windows(state.window_state, state.network_slug)`, which emits a `window_invited` payload for **every** channel still in `:invited`. On reload, cic subscribes cold, gets the backfill, and `userTopic.ts` `case "window_invited"` writes the state straight back. `:invited` is the one non-terminal state with a dedicated cold-subscribe backfill, which is why this is visible for invites and not for `:pending`.

So the × is not broken — it simply has no server-side counterpart, and #482 faithfully restores the state the server still believes in.

## Suggested fix

Give the dismissal a server-side effect for the `invited` case. Shortest path: have the `invited` row's × go through the same DELETE as the normal close.

[...]

Issue #482, opened 2026-07-27
INVITE no longer leaves a trace in the status window: restore the $server row + [Join now] alongside the channel one


vjt:
Ruling by @vjt (per chan request), after the symptom came up again on IRC: *"non appare il canale nella bottom bar […] e non appare manco niente nella status window […] credo dovremmo ripristinare il testo che avevamo in status window col bottone 'join now'"* → *"trova dove lo abbiam tolto e ripristiniamolo"*.

Found. This is not a regression — it was a deliberate move, and this issue is about putting the `$server` copy **back alongside** the channel one, not about reverting it.

## Where it was removed

Commit **`834204b7`** — *"fix(session): route channel-scoped traffic by channel, not sender (#78)"*, 2026-06-28 — case (c) in its own message: *"Case (c) — inbound non-awaiting INVITE opens an `:invited` window (folds #128). Previously such an INVITE fell through to the `:server_event` catch-all on `$server`."*

The diff, in `lib/grappa/session/event_router.ex`:

```diff
       if MapSet.member?(awaiting, normalize_channel(channel)) do
         {:cont, state, [{:rejoin_invited, channel}]}
       else
-        route_unhandled_command(msg, state)
+        {state, persist_eff} = persist_raw_event(msg, state, channel)
+        {:cont, state, [persist_eff, {:invited, channel}]}
       end
```

and the persist target itself became a parameter:

```diff
-    {state, eff} = build_persist(state, :server_event, "$server", sender, body, meta)
-    {:cont, state, [eff]}
+    build_persist(state, :server_event, channel, sender, body, meta)
```

So pre-#78 an unrequested INVITE persisted a `:server_event` row on **`$server`** — that is the status-window text @vjt remembers. Post-#78 the identical row is persisted at **the invited channel**, plus an `{:invited, channel}` effect that greys the tab.

## Why the invite is invisible today

The channel row alone is not enough, for a reason that is independent of which window it lands in: `Grappa.Session.WindowState.to_wire/3` (`lib/grappa/session/window_state.ex:287-309`) returns `{:ok, payload}` only for `:joined`, `:failed` and `:kicked`. **`:invited` falls into the `_ -> {:error, :not_tracked}` catch-all**, exactly as its own `@doc` states.

[...]

Issue #206, opened 2026-07-10
cicchetto: upload TTL default shown as raw seconds (86400) instead of human-readable


vjt:
## Report
Reported on IRC (#it-opers).

In the options / settings UI, the **upload TTL** values are rendered human-readable (e.g. "1 hour", "1 week") — **except the default**, which shows the raw seconds value `86400`.

## Expected
The default should render in the same human-readable form as the other choices — `86400` → **"1 day"** (or "1 giorno" per locale).

## Notes
Minor UX polish. Likely the default option bypasses the same seconds→human formatter the other entries pass through. Apply the same formatting helper to the default value.

_Filed per channel request._

vjt:
**Merged to `main` (7c3e575), green — awaiting HOT `--cic` deploy.**

The **"use site default"** option in the upload-retention selector rendered the raw seconds token (`use site default (86400)`) because it emitted `activeHost().defaultTtl` directly, while every other option renders its `label` from the ttlOptions ladder.

- Fix: resolve the default through that **same** ladder — `86400` → **"24 hours"** — no new formatter. Falls back to the raw token only if the default isn't in the ladder (host misconfig). `cicchetto/src/SettingsDrawer.tsx`.
- Note: the human label is **"24 hours"** (the existing ttlOptions entry), not "1 day" as the issue phrased it — reusing the existing labels keeps the selector internally consistent rather than inventing a second wording.
- Guard: vitest asserts the site-default option renders "24 hours" and never "86400".
- Gates green: `bun run check` (biome + tsc), `bun run build`, full vitest **2422 tests pass**.

cic-only → HOT. Not deployed by me; orchestrator batches #206 with #208 into one `--cic` ship.

vjt:
Shipped to prod in the 9758397 cold-deploy bundle (02:57 UTC). Upload-TTL default option now renders the human label via the shared seconds→human formatter instead of raw 86400.

PR #917, opened 2026-08-06
#902: an invite is a notification, not a ghost window

vjt:
An inbound INVITE stops being a greyed ghost tab in the sidebar and becomes a
top banner that **names who invited you** and carries `[Join]`.

Five commits: a server leg (the inviter becomes window state), the cic leg (the
banner), the e2e leg (five specs moved off the deleted attribute), a replacement
witness for the one contract that lost its only e2e, and a fix for the one spec
that first execution caught.

## The server leg is not incidental

The issue frames #902 as essentially cic-only — "one new member of the closed
set in `errorBanners.ts`". It is not. The decided copy names the inviter, and
that nick is in **neither** the `window_invited` payload **nor** `window_states`,
so the cold-subscribe backfill could never have reproduced it.
`Session.WindowState` gains an `invited_by` sibling map; `Session.Wire` gains an
additive `inviter` field. Reverting #482's `$server` INVITE copy is likewise
server-side, in `event_router.ex`.

## What replaced `data-window-state="invited"` as #30's synchronisation seam

#30's e2e used the greyed `:invited` pseudo-row as its barrier because it was
the ONLY DOM projection of `windowStateByChannel` — the map the channel
completion filters — so once it read `invited` the key was provably in that map.
A sidebar row for a *joined* channel comes from `channelsBySlug` on the user
topic instead, a different topic with no cross-topic ordering guarantee (stated
at `Sidebar.tsx`), so a stale entry has no DOM footprint to wait on; the gap was
previously measured at up to 4.9s.

#902 deletes that row, so the seam was **REPLACED, not dropped**:
`BannerSlot.tsx` now emits `data-banner-id="invite:<slug>:<channel>"`, and the
invite banners are derived straight off the same `windowStateByChannel` map (via
`windowState.invitedWindows`).
`cicchetto/e2e/tests/issue30-channel-tab-completion.spec.ts` **was updated** to
wait on the shared `inviteBanner()` fixture locator instead.

The new seam is strictly tighter than the old one: it names the exact (network,
channel), where `data-window-state` only named a state. **Measured, not argued:
that spec passes on this branch** (chromium #206).

## Per repurposed spec — does the OLD issue's behaviour still exist?

Five specs leaned on the deleted attribute. A repurposed spec is deleted
coverage unless the old behaviour genuinely no longer exists, so that judgement
is stated per file rather than left implicit in a rename.

| spec | old behaviour still exists? | still covered? |
|---|---|---|
| `issue30-channel-tab-completion` (re-pointed) | **yes** — completion still pins to live joinedness | **yes**, same subject; only the barrier locator changed |
| `b2-inbound-invite-cta` (modified) | **yes** — an inbound invite still gets a CTA | **yes**, same journey on the new surface, plus two assertions the new design earns (the banner names the inviter; no sidebar row exists at all) |
| `issue482-invite-survives-reload` (modified) | **split**: the cold-snapshot backfill **exists**; the second `$server` INVITE copy is **deliberately reverted** | **yes for the half that exists** — the backfill is asserted explicitly, and the restored banner must still name the inviter, a nick that on that path can only have come from server-side window metadata |
| `issue511-invited-dismiss-durable` → `issue902-invite-banner-dismiss-ephemeral` | **reversed for `:invited`** (the × writes nothing, the invite returns on reload) — but #511's **mechanism stays live for `:failed`/`:kicked`** | **yes, via a new spec** — `issue511-failed-autojoin-dismiss-durable.spec.ts`; see below, the obvious version of it would have been vacuous |
| `issue71-inc3-bottombar-invite` → `issue902-invite-banner-mobile` | INC-3's subject (the `:invited` slice in the mobile BottomBar) **no longer exists in any form** | INC-3's **negative** half is preserved verbatim — the spec still asserts a JOIN-failed channel raises no bar entry, after its `:failed` state genuinely materialises |

## The finding: the obvious replacement spec would have been vacuous

#511's mechanism (× on a non-joined row → `partAndForget` → server `set_parted`
→ the cold load cannot resurrect it) is still live for `:failed`/`:kicked`, and
the only e2e that exercised it was the `:invited` one whose contract #902
reverses. That is deleted coverage of a live contract, so it needed a
replacement.

**The obvious replacement — × a `:failed` *pseudo-row*, reload, assert it stays
gone — passes identically with the fix reverted.** After a reload a `:failed`
state is unreachable by the client:

- `GrappaChannel.push_session_snapshot/2` pushes only `umode_changed`,
  `supported_umodes_changed` and `invited_windows` on the user topic — no
  `:failed`, no `:kicked`;
- `WindowState.to_wire/3` *does* re-assert `:failed`, but only on the
  per-channel topic, which requires a per-channel subscription;
- `subscribe.ts` derives those subscriptions from `channelsBySlug`, and its
  pre-subscribe loop skips `:failed`/`:kicked` outright.

A pseudo-row exists *precisely because* its channel is absent from
`channelsBySlug` — present, the Sidebar collapses it into the LIVE branch (#38's
own dedup analysis). No entry, no subscription, no snapshot, nothing to
resurrect. So `ux-5-bk-join-fail-dupe` is **not** one `page.reload()` away from
being this spec; that reload would assert nothing.

The shape where the durability *is* observable is #38's production scenario: a
`+k` channel in `autojoin_channels` that 475s on connect. There the × runs
`confirmLeaveChannel` → `closeChannelWindow` → the same `partAndForget`; under
the client-only mutation the autojoin entry survives and the next cold load
repaints the row. Two channels, one dismissed and one kept — the kept row
reappearing after the reload is a condition-based co-witness that the cold-load
fetch was dispatched and rendered, so the absence check cannot pass by racing it.

Stated in the file header rather than left for a reader to infer: the spec
**cannot** separate `set_parted`'s contribution from the de-autojoin's, because
one DELETE does both and the de-autojoin alone already explains the absence. It
constrains the composition (× → DELETE → server → cold load), which is the link
that had no witness left.

Both halves also remain unit-constrained independently — client at
`cicchetto/src/__tests__/windowClose.test.ts:232`, server at
`test/grappa/session/part_cleanup_test.exs:159` and
`test/grappa/session/window_state_test.exs:208-231`. Measured rather than
assumed: removing `postPart` from `partAndForget` turns **2 of 10** windowClose
tests red — two, not one, because `closeChannelWindow`'s twin falls with it.

## What first execution caught, that writing-by-construction could not

`issue482` was half-migrated. Its reload assertions were updated to the banner,
but the step *after* them still navigated by the pseudo-row — `selectChannel`
sat 30s on a `li[data-window-name]` that #902 deletes, and the spec timed out. A
red produced entirely by the spec, not by the code under test.

Worth reading precisely: the timeout was at the buffer-inspection step, which
means **everything before it passed** — the banner appears live, survives the
reload, and still names the inviter on the cold-snapshot path. The #482 contract
re-pointed to the banner is measured, not asserted by construction.

The buffer assertion was right to exist; only its door was stale. It now goes
through the **archive**, which is the door #902 leaves open by construction:
`visibleArchiveForNetwork` subtracts what the nav DRAWS, and the nav draws no
invite, so the invited channel's buffer surfaces there. That is the documented
answer to "where did the invite go once the banner is gone" — routing the
assertion through it turns the answer from a comment into something the suite
measures.

An assertion was **added** rather than negotiated away: after the reload, the
sidebar row count must be **zero**. The backfill is a second door onto the same
state, and #902's contract is that both doors produce a banner. Without it the
pseudo-row could re-enter from the snapshot after being removed from the live
path and nothing would have caught it.

The cleanup's join-then-part dance went with the row it clicked — it rested on
the belief that a PART cannot clear a never-joined channel, and #511 settled the
opposite. `issue902-invite-banner-mobile`'s cleanup also moved into `afterEach`:
it ran at the end of the test body, and under #902 a stranded `:invited` window
RETURNS on the next cold load, making it a cascade poisoner for every later spec
whose layout the fixed banner region shifts.

## Gates

- `scripts/check.sh` — **rc=0**: `8 doctests, 54 properties, 5278 tests, 0
  failures`; Credo `5120 mods/funs, found no issues`; Sobelow; `Doctor
  validation has passed!`; `cicchetto/src/lib/wireTypes.ts is in sync.`; bats
  `1..328`, zero `not ok`. Working tree clean before and after.
- `scripts/bun.sh run check` / `run test` / `run build` — **rc=0**, 240 files /
  4414 tests.
- `scripts/integration.sh` (full suite) — 604 passed, 3 failed. Every
  #902-touched spec green except `issue482`, whose failure was the
  half-migration above and is fixed here. Of the other two:
  `issue173-compose-recall-caret-visible` **fails on a merge-base baseline run
  too** (exonerated by measurement); `issue535-visibility-return-preserve-scroll`
  is **unresolved at PR time** — it was already in tonight's rotating red set and
  the baseline had not reached it, and one run per side does not attribute.

## Deploy: this ships COLD

`Session.WindowState` is in `HotReload.LongLivedModules` `@state_helpers`, so
`Deploy.Preflight` reads the changed `defstruct` and refuses a hot deploy.

## Design decisions (ruled, for the record)

- **One banner PER CHANNEL** with a per-entry dismiss `id`. The aggregate banner
  was a correctness defect, not a taste call: it breaks `rearmDismissed` and
  swallows the next invite.
- **The banner names the inviter** — server-side `invited_by` map plus an
  additive `inviter` wire field.
- **`×` writes NOTHING**, session-scoped; a dismissed invite RETURNS after a
  reload. A deliberate reversal of #511 for `:invited`: a banner is a
  notification, not a window, so nothing accumulates and re-announcing is cheap.
- **Both old surfaces dropped**: the greyed pseudo-row AND the `$server` INVITE
  copy. The channel's own INVITE row STAYS.
- Invites sit **above** push-optin in the stack, so "push-optin is LAST" stays an
  unconditional invariant.
- Out of scope: push/badge parity for invites, auto-join-on-invite.

Durable rationale is in `docs/DESIGN_NOTES.md` (2026-08-06).

A separate comment records what I refused to claim, and asks one open question
about `:failed`/`:kicked` after a reload.


vjt:
## What I refuse to claim

A measured negative is a result. Tonight the full suite ran, so several items
that started here died with a number instead of an argument — those are marked
**discharged**, because a refusal list that only ever grows is theatre.

**Discharged by measurement**

1. ~~No e2e has been run.~~ The full `scripts/integration.sh` ran: **604 passed,
   3 failed**. Every #902-touched spec is green except `issue482`, whose red was
   the spec's own half-migration and is fixed in this PR.
2. ~~The banner has never been rendered in a browser.~~
   `issue902-invite-banner-mobile` is green on `webkit-iphone-15`: the banner
   renders at that viewport, the invited channel raises **no** BottomBar entry, a
   JOIN-failed channel raises no banner, and `[Join]` responds to a real `tap()`.
   (Per `feedback_playwright_webkit_not_ios_scroll`, webkit-on-desktop proves DOM
   and layout wiring at that viewport, not iOS gesture behaviour.)
3. ~~I wrote into `archive.ts` that an `:invited` channel appears in the archive
   without checking.~~ I read `list_archive/3` + `build_active_keyset` to confirm
   it, and this PR now **asserts** it: `issue482` reaches the invited channel's
   buffer through the archive. It was a conjecture when written; it is a test now.

**Still standing**

4. **The #30 seam replacement is measured to work ONCE, not measured to displace
   the race.** `issue30-channel-tab-completion` passes on this branch. It was not
   run with `--repeat-each`. The old barrier existed to close a gap measured at up
   to 4.9s; that the new one closes it *as reliably* remains an argument from
   construction (same source map, same single-shot property), not a measurement.
   A single green cannot distinguish "the race is displaced" from "the race did
   not fire this time".

5. **My own new spec's non-vacuity is argued, not measured — the same standard I
   used to reject a different one.** I demonstrated by code reading that the
   pseudo-row version of `issue511-failed-autojoin-dismiss-durable` would pass
   with the fix reverted, and rejected it for that. I then asserted, also by code
   reading, that the autojoin version *would* go red under the client-only
   mutation. I did not run it against a mutated build. It is a stronger argument
   (the autojoin entry demonstrably survives a client-only drop, and
   `channelsBySlug` demonstrably repaints from it) but it is the same kind of
   argument. If the e2e mutation is wanted, it is one line and one suite run.

6. **One of my own unit tests does not constrain what it appears to.** In
   `errorBanners.test.ts`, "re-inviting a channel whose banner was dismissed shows
   it again" stays GREEN under the source-keyed mutation — with a single invite
   the two keyings coincide. It covers episode-scope, NOT the per-entry `id`
   widening. The tests that DO constrain it are "dismissing ONE invite leaves the
   other invites visible" and "an invite arriving AFTER a dismiss is shown, not
   swallowed by it" (2 red and 1 red respectively under the two mutations).

7. **`issue535-visibility-return-preserve-scroll` is unresolved at PR time.** It
   failed on this branch (unread divider count 0, expected 1). It was already in
   tonight's rotating red set; the merge-base baseline run had not reached it when
   this PR opened. One run per side does not attribute, so I am not claiming it is
   a flake and not claiming it is mine. A plausible mechanism does exist and I
   would rather name it than hide it: #902 adds a fixed top banner region, and a
   stranded `:invited` window from a sibling spec would render one and shift the
   scrollback geometry. That is exactly why this PR moves
   `issue902-invite-banner-mobile`'s cleanup into `afterEach`. Plausible mechanism,
   not a diagnosis.

8. **Three things in the brief/issue were wrong, one of them mine.**
   - The issue frames #902 as cic-only. It is not: the decided copy needs a
     SERVER leg (the inviter is in neither the payload nor `window_states`), and
     reverting #482 is server-side in `event_router.ex`.
   - The cross-dependency note said it was unchecked whether specs other than #30
     leaned on the pseudo-row. **Four more did**, one of them (`issue511`)
     asserting a contract this issue reverses.
   - `window_state.ex`'s own docstring claimed "the inviter is conveyed by the
     persisted `:server_event` scrollback row … so no sibling map is needed". True
     for a greyed tab, false for a banner. Corrected in place.
   - And mine: I first described the `:failed`/`:kicked` coverage loss as "no e2e
     covers it any more", which overstated it. Both halves *are* unit-covered;
     what was lost is their composition. Measuring narrowed my own claim.

## One open question I am not answering unilaterally

**A `:failed`/`:kicked` window does not survive a client reload. Intended, or an
unexamined gap?** My reading says: **intended at the surface, with a residue that
is a gap.**

Intended: the channel does not vanish, it **moves to the archive**.
`visibleArchiveForNetwork` subtracts only what the nav draws; with
`windowStateByChannel` empty after a reload it subtracts nothing, so the buffer
holding the 475's persisted `:notice` row surfaces as an archive entry. That is
the same "subtract what the nav DRAWS" rule (#402) the file states in words for
#902's invite.

The gap is server-side. After that reload, `window_states[ch] = :failed` and its
`failure_reasons` / `failure_numerics` siblings become **unreachable and
unclearable from any UI** — the only verb that clears them is the DELETE the ×
sends, and the × is gone with the row. They persist for the lifetime of the
`Session.Server`. The reachable failure mode: an operator adds that channel to
autojoin (an autojoin-only edit is `:left_alone`, no session restart) →
`channelsBySlug` picks it up → cic subscribes per-channel → `to_wire/3` serves a
`join_failed` carrying a **stale** reason and numeric, from a failure that may be
hours old and no longer true, before a single JOIN has been attempted in that
session. That is precisely the leak `window_state.ex`'s "must NOT leak the old
reason / numeric into the next snapshot push" guards against elsewhere.

Low severity, real, and orthogonal to #902. I would file it rather than fix it
here.

vjt:
### Follow-up: the merge-base baseline finished — correcting my own item 7

When this PR opened I wrote that the baseline "had not reached"
`issue535-visibility-return-preserve-scroll`. It has now finished, so that
sentence is stale and this is the measured picture.

**Baseline** (merge-base `f14425e4`, full suite, 23.3m): **4 failed / 602 passed**
- `cp15-b6-part-archive-rejoin`
- `issue168-scroll-authority`
- `issue173-compose-recall-caret-visible`
- `p0b-peer-away`

**This branch** (full suite, 29.5m): **3 failed / 604 passed**
- `issue173-compose-recall-caret-visible`
- `issue482-invite-survives-reload` — mine, real, fixed in this PR
- `issue535-visibility-return-preserve-scroll`

Intersection: `issue173` only. **Three specs are red on main that are green
here**, and one is red here that is green on main.

Readings, kept separate from the wishful one:

- `issue173` fails on **both** sides. Pre-existing, exonerated by measurement.
- `issue535` is **still not attributed.** It passed on the baseline, which on its
  own looks like it points at this branch — but the sets **rotate**: three
  baseline reds went green here in the same pair of runs. One run per side does
  not attribute an e2e red, and I am not going to let the direction that happens
  to favour this PR be the one I treat as signal while treating the other three
  as noise. Same instrument, same night, same host.
- Two of the three baseline-only reds (`issue168-scroll-authority`,
  `cp15-b6-part-archive-rejoin`) are scroll/archive-shaped, like `issue535`
  itself. That cluster looks like tonight's host, not like any of these changes —
  data for the flake epic, not a diagnosis.

The baseline ran at `f14425e4`; main has since moved to `1fb71383` (#908). This
PR's own `integration` run is now the better instrument, since it exercises the
union with a settled main.

Issue #915, opened 2026-08-06
denoise: raise the auto-activation threshold from 50 to 200 members


vjt:
vjt, in conversation: *"denoise va attivato in automatico dai chan con 200 persone in su. non 50 son troppo pochi."*

## The change

`cicchetto/src/lib/presenceFilter.ts:63`

```ts
export const LARGE_CHANNEL_THRESHOLD = 50;   // → 200
```

That constant is the whole auto-activation rule: with the per-channel pref **unset**, `resolvePresenceVisible` follows the size default, and a channel whose live member count crosses the line auto-flips. 50 is too low — a 50-member channel does not drown in J/P/Q, and denoising it hides traffic the operator wants.

## What follows the constant automatically (verified, not assumed)

Every current reference reads the exported constant rather than a literal:
- `cicchetto/src/__tests__/presenceFilter.test.ts:28,33,41-45`
- `cicchetto/src/__tests__/ScrollbackPane.test.tsx:286,5170,5179,5186` (`membersOfSize(LARGE_CHANNEL_THRESHOLD ± 1)`)

So the boundary tests keep testing *the boundary*, wherever it is.

## What does NOT follow — must be fixed in the same change

1. **`presenceFilter.test.ts:40` is named `"boundary is exactly LARGE_CHANNEL_THRESHOLD (49 shown, 50 hidden)"`.** Its assertions use the constant, so it will pass — with a name that lies. Same class as #912. Rename it (or drop the parenthetical) so the name states the property, not a stale instance.
2. **The moduledoc above the constant says `"50+ member channels drown in J/P/Q"`.** Update the prose; a comment that contradicts the code is the alibi that stops the next reader from checking (that is literally how #910's moduledoc bug survived).

## The sweep I have NOT done — do it before claiming done

Raising the line means channels of **50–199 members now show J/P/Q by default**, where they previously auto-hid. Any e2e that seeds a member count to *trigger* the auto-hide, rather than reading the constant, will silently change meaning — it will keep passing while testing the opposite branch.

[...]

Issue #912, opened 2026-08-06
windowClose.test.ts:232 is named "survives a reload (#511)" and never reloads anything


vjt:
Found by w1 while measuring the #902 coverage gap. Small, but it is the same class as the shottino test-name work (`347f71c7`, *"assert what these test names claim, not what passes anyway"*).

## The defect

`cicchetto/src/__tests__/windowClose.test.ts:232` is named `"... survives a reload (#511)"`.

It **reloads nothing**. What it actually constrains is the CALL:

```
expect(api.postPart).toHaveBeenCalledWith("utok", "freenode", "#inv")
```

That is a real and useful assertion — removing `postPart` from `partAndForget` does redden it, so the "client-only dismiss" mutation dies here. But *surviving a reload* is a property of the client half **composed with** the server half (`PartCleanup.cleanup_local` → `WindowState.set_parted` → the key leaves every map → the cold-subscribe snapshot omits it). A unit test with the network mocked cannot witness that, and this one does not try to.

## Why the name matters

The name is what the next reader greps for when they ask "is the durability of #511 covered?". It answers yes. The honest answer is "the client half is pinned; the composition is covered only end-to-end". That gap is what let the composition's only e2e witness be deleted without anyone noticing — see #902 and the replacement spec being added there.

## Fix shape

Either rename it to what it pins (`calls postPart when dismissing a non-joined pseudo-row (#511)`), or leave the name and make it earn it — which it cannot do at unit level, so realistically: rename.

**Not a behaviour change, not urgent.** Filed so it is not lost.

Refs: #511, #902.

vjt:
Rename only, which is what the issue's own "Fix shape" asked for: the case is now `"calls postPart, so
the dismissal reaches the server and not just the client (#511)"`, with a comment saying durability is
witnessed by `e2e/tests/issue511-failed-autojoin-dismiss-durable.spec.ts` (`7270a595`). Keeping the
assertion was measured, not assumed: dropping `void postPart(...)` reddens the test. The rename also
drops the now-wrong "invited" framing, which #902 made non-durable.


Closed at the **`v0.13.0`** release cut — that is the tag to pull.

Issue #910, opened 2026-08-06
LIST numerics 321/322 are param-scanned too — and the moduledoc explicitly claims they are not


vjt:
Found by w2 during the #908 sweep, **code-verified** (not RFC guesswork): the shapes were checked against `parse_list_entry/1` in `lib/grappa/session/server.ex`. Reported separately from #908 because it is a distinct defect with its own blast radius.

## The defect

`lib/grappa/session/numeric_router.ex` puts these in `:scan`, and `scan_params/2` takes the first nick-shaped, non-own, dotless token:

- **321 RPL_LISTSTART** — `[nick, "Channel", "Users Name"]` → the only candidate is the literal `Channel` → `{:query, "Channel"}`, i.e. a query window named **Channel**.
- **322 RPL_LIST** — `[nick, "#chan", "5", "topic"]` → the **channel** branch wins → one `:notice` row in **every listed channel's window**.

## Why it is reachable, not theoretical

Both are normally consumed by the `directory_refresh` handler — but that handler **nils itself on `:directory_refresh_timeout`**, and any late frame after that falls straight into the scan. `/quote LIST` (`send_raw`, `server.ex:1803`) reaches it too; that escape hatch is exactly how #908 reached prod.

## The part that makes it worth its own issue

`numeric_router.ex`'s **moduledoc explicitly claims 321/322/323 "route via the default `{:server, nil}` path"**. They do not. A wrong moduledoc about a family it names by number is worse than silence — it is the alibi that stops the next reader from checking.

## Relationship to the class

#184 (STATS), UX-4 bucket I (connect-storm), #908 (TRACE), and now this: **at least five families**, all the same disease — a middle param that is data or a type label but happens to satisfy `Identifier.valid_nick?/1`.

w2's sharper framing of the root cause, which I think is right: *deny-list and allow-list are the same mechanism — both are code-keyed tables. Nothing about the deny list's shape is the smell; its **default** is.* The real question is which side an UNKNOWN numeric falls on. **That default is a design decision and is deliberately NOT being taken here** — this issue is the concrete 321/322 defect plus the false moduledoc.

Refs: #908 / PR #909 and its "what I refused to claim" comment.

vjt:
`v0.13.0` fixes both halves: `@list_numerics Enum.to_list(321..323)` (`numeric_router.ex:278`, unioned
at `:402`), and the two docs that denied the bug are corrected in place — moduledoc item 3 at `:47`,
[...]

Issue #458, opened 2026-07-26
Filter join/part/quit server-side in the scrollback REST fetch (blocked by #449) — one page-up should be one screenful


vjt:
Ask from vjt (#grappa, 2026-07-26): now that the hide-join/part/quit preference
is server-side, drop the client-side filter and let the server omit the rows —
no more scrolling through pages that render empty, and less work on the client.

### First, a correction to the premise

The preference is **not** server-side yet. On `origin/main` the filter is still
`cicchetto/src/lib/presenceFilter.ts` — per-channel tri-state persisted to
`localStorage` under `cicchetto.presenceFilter`, explicitly documented as
"localStorage only […] no server-side persistence, no wire change". Moving it
server-side is **#449**, which is open and queued (and #443 carries the drawer
section it needs a home in). So this issue is **blocked by #449** — it is the
step *after* it, not a parallel one. Sequence: #449 makes the pref server-owned,
then this issue makes the server *act* on it.

### The part of the ask that is real, and worth doing

The empty-scroll symptom is genuine and is a pagination bug, not a rendering
one. `Grappa.Scrollback.fetch/6` (`lib/grappa/scrollback.ex:464`) has no kind
predicate: it orders by `server_time DESC, id DESC` and applies `limit` to the
raw row set. The client then hides `join|part|quit|nick_change`
(`SUPPRESSED_PRESENCE_KINDS`). On a busy channel a 100-row page can therefore
yield a handful of visible lines — or none — and the user pages again, and
again. Filtering in SQL makes `limit` count *visible* rows, so one page-up is
one screenful. That alone justifies the change.

### Hard constraint: filter HISTORY, never the live stream

The live WS presence events are load-bearing control flow in cic, not just
render fodder. Dropping them server-side breaks, at minimum:

* `cicchetto/src/lib/members.ts` — `applyPresenceEvent(key, msg)` is the ONLY
  live updater of the members store (bootstrap comes from the `members_seeded`
  push, "no REST fetch path remains"). No join/part/quit ⇒ the nicklist freezes
  until the next 366 RPL_ENDOFNAMES.
* `cicchetto/src/lib/subscribe.ts:461` — own-PART → windowState projection.
* `cicchetto/src/lib/subscribe.ts:482` — non-self `nick_change` → query-window
  re-key (#372 canonical peer re-key).
* `cicchetto/src/ScrollbackPane.tsx:1274`, `:1336` — own-JOIN auto-focus.

[...]

PR #918, opened 2026-08-06
#915: raise the denoise auto-activation cutoff from 50 to 200

vjt:
## The decision

vjt: *"denoise va attivato in automatico dai chan con 200 persone in su. non 50 son troppo pochi."*
`LARGE_CHANNEL_THRESHOLD` 50 → 200. A 50-member channel does not drown in
join/part/quit; auto-denoising it hid traffic the operator wanted.

**200 is a judgement call, not a measurement.** I have no evidence that 200 is the right
number and did not look for any: no member-count distribution from prod, no J/P/Q-rate-
versus-channel-size curve. vjt picked it from operating the thing, the same way 50 was
picked in #222. Nothing in this PR should be read as implying the threshold was derived.
That is precisely why the drift guard below pins EQUALITY and not the value — it survives
him changing his mind again.

## It was not a cic-only change

The cutoff exists **twice**. #458 gave the render-layer rule a server twin —
`Grappa.PresenceFilter.@large_channel_threshold` (`lib/grappa/presence_filter.ex`)
applies the same size default to the REST history fetch so `limit` counts VISIBLE
rows, while cic keeps its filter for the live WS tail. Both moduledocs say the two
MUST stay equal, and **nothing in the repo failed if they didn't**.

Landing this cic-only would have made every channel sized 50–199 show J/P/Q on the
live tail and lose it on page-up, silently, with a fully green suite. Both constants
move here.

## Finding: an assertion that stopped distinguishing the override from the default

The issue states every reference already read the exported constant. Measured, it did
not — `presenceFilter.test.ts` carried `49`, `50`, `80`, `100`. Two go red at 200 and
would have forced the edit anyway. The dangerous one is the one that stays green:

```ts
it("explicit 'show' overrides the size default even on a huge channel", ...)
  expect(resolvePresenceVisible("show", 100)).toBe(true);
```

At 200 that count is **small**, so an unset pref returns `true` as well: the assertion
no longer distinguishes the OVERRIDE from the DEFAULT. Green, under a name for a
property it has stopped testing — the same class as #912.

Every count now derives from the constant, and each explicit-pin case is placed where
the size default answers the OPPOSITE, so the pin is what the assertion pins. The
Elixir side was already written this way throughout
(`PresenceFilter.large_channel_threshold()`) and needed no test edits.

## Finding: the suites constrain the RULE, not the VALUE

With everything derived, I put cic's constant back to 50 and re-ran the cic suite:
**4395 tests passed, 241 files, zero reds.** Nothing in either language would have
caught a wrong threshold. That measurement is the entire reason the drift guard exists.

So the equality is now code, in the shape the repo already uses for once-per-language
rules (`RouterSwDenylistTest` parses `service-worker.ts`;
`should_notify_parity_test.exs` shares a JSON truth table). It pins EQUALITY, not the
number, so no gate in this repo asserts the cutoff is 200 — that is the design, not a
hole.

Verified by displacement, not by its own green: with cic at 50 and the server at 200 the
guard fails alone, `6 tests, 1 failure`, printing `cicchetto/src/lib/presenceFilter.ts:
50` vs `Grappa.PresenceFilter: 200`.

## The sweep

The four flagged e2e loops (`issue211-phase6-matrix:95`,
`issue211-phase7-multinet-reconnect:176,197`, `issue260-sticky-network-tab:162`,
`marker-target-window-regression:132`) seed poll ATTEMPTS or pad MESSAGES, never
members — none change meaning. The only member-count sources in the stack are
`membersByChannel` (unit-mocked) and a live NAMES burst (`session_with_members`), both
already derived. Stale prose naming "50" in four places now names the constant.

**The size default is unreachable from the e2e, before and after.** #222's own moduledoc
records why: that many nicks from one IP trips the bahamut same-host autokill and the
harness exposes no member-count seam. The boundary was gated by vitest alone at 50 and
still is at 200 — no coverage lost, none gained.

## The decision log was corrected at the source

#222 (~10630) said `LARGE_CHANNEL_THRESHOLD = 50` and #458 (~21472) said
`@large_channel_threshold 50` / "≥50 members → hide". I had left both, reasoning that a
chronological log is superseded by its newest entry. Wrong: an old entry asserting a
stale value as live is a false statement the next reader cites instead of checking —
how `numeric_router.ex`'s moduledoc hid #910 for months. Both now name the constant,
keep the value chosen THEN as the decision it was, and point at the #915 entry for the
value in force. #458's "held equal because both moduledocs cross-reference" is corrected
too: that WAS the mechanism, until this PR measured that nothing failed on drift.

## Gates — rc files on the worker host, check them yourself

| gate | rc file | rc | log |
|---|---|---|---|
| `scripts/check.sh` (COMPILE lane) | `/tmp/w2-check.rc` | **0** | `/tmp/w2-check.log` |
| mutation: cic 50 / server 200 | `/tmp/w2-mut2.rc` | **2** (expected red) | `/tmp/w2-mut2.log` |
| restore + 3 threshold files | `/tmp/w2-restore.rc` | **0** | `/tmp/w2-restore.log` |
| `mix docs` (DESIGN_NOTES is an ExDoc extra) | `/tmp/w2-docs.rc` | **0** | `/tmp/w2-docs.log` |

`check.sh` detail: `8 doctests, 54 properties, 5273 tests, 0 failures`; Dialyzer
`Total errors: 0`; credo strict, sobelow, doctor, deps.audit, wire-types check clean;
bats `1..328`. Working tree porcelain-clean before AND after. Run detached with an
explicit `cd` into the worktree — the log's first line is `CWD=` for that reason.

cic gates: `bun run check` exit 0, `tsc --noEmit` standalone exit 0 clean,
`bun run test` 241 files / 4395 tests green.

### DIFF
--- a/cicchetto/e2e/tests/issue222-presence-filter.spec.ts
+++ b/cicchetto/e2e/tests/issue222-presence-filter.spec.ts
@@ -5,13 +5,14 @@
 //
 // This e2e is the interactive witness for the per-channel toggle + the
 // render-layer filter + localStorage persistence. It exercises REAL presence
-// events (a peer joins + parts #bofh) — NOT 50 spawned peers: 50 nicks from
-// one IP risks bahamut flood/autokill (feedback_e2e_multinet_live_needs_
-// distinct_nicks), and there is no window-exposed member-count seam in the e2e
-// harness to inflate membership. The size-default MATH (49 shown / 50 hidden +
-// the precedence truth table) is proven authoritatively in the vitest boundary
-// test (src/__tests__/presenceFilter.test.ts); this spec owns the interactive
-// toggle/persistence path.
+// events (a peer joins + parts #bofh) — NOT LARGE_CHANNEL_THRESHOLD spawned
+// peers: that many nicks from one IP risks bahamut flood/autokill
+// (feedback_e2e_multinet_live_needs_distinct_nicks), and there is no
+// window-exposed member-count seam in the e2e harness to inflate membership.
+// So the size-default MATH (the boundary pair + the precedence truth table) is
+// UNREACHABLE from here at any threshold value and is proven authoritatively
+// in the vitest boundary test (src/__tests__/presenceFilter.test.ts); this
+// spec owns the interactive toggle/persistence path.
 //
 // Assertions (all VISIBLE outcomes, CLAUDE.md "assert outcomes not calls"):
 //   1. small channel, default shown → a peer join + part row IS visible
@@ -65,7 +66,8 @@ test("#222 — per-channel toggle hides join/part rows, persists across reload, 
       .locator('[data-testid="scrollback-line"][data-kind="privmsg"]')
       .filter({ hasText: privmsgBody });
 
-    // 1. small channel (< 50 members), pref unset → presence shown by default.
+    // 1. small channel (a handful of members, far below
+    //    LARGE_CHANNEL_THRESHOLD), pref unset → presence shown by default.
     await expect(joinRow).toHaveCount(1, { timeout: 10_000 });
     await expect(partRow).toHaveCount(1, { timeout: 10_000 });
     await expect(privmsgRow).toHaveCount(1, { timeout: 10_000 });
--- a/cicchetto/e2e/tests/issue239-hidden-msg-unread.spec.ts
+++ b/cicchetto/e2e/tests/issue239-hidden-msg-unread.spec.ts
@@ -69,9 +69,10 @@ test("#239 — hidden control message does NOT bump the unread badge; reading cl
   // for the auto-joined self-JOIN line, proving REST + WS both landed.
   await selectChannel(page, NETWORK_SLUG, CHANNEL, { ownNick: NETWORK_NICK });
 
-  // Pin #bofh to HIDE presence via the production toggle (no need to seed 50
-  // members — the size-default math is unit-tested; this is the interactive
-  // hide path). #500 folded the denoise toggle behind the rail launcher menu;
+  // Pin #bofh to HIDE presence via the production toggle (no need to seed
+  // LARGE_CHANNEL_THRESHOLD members — the size-default math is unit-tested;
+  // this is the interactive hide path). #500 folded the denoise toggle behind
+  // the rail launcher menu;
   // open it (desktop: taps the launcher) to reach the toggle.
   await openRailMenu(page);
   const toggle = page.locator('.rail-actions-menu [data-testid="presence-toggle"]');
--- a/cicchetto/src/__tests__/presenceFilter.test.ts
+++ b/cicchetto/src/__tests__/presenceFilter.test.ts
@@ -13,8 +13,12 @@ import { channelKey } from "../lib/channelKey";
 // The "tough" part the issue flagged: an automatic size-based default and a
 // manual per-channel override need a clear precedence — explicit choice WINS
 // over the size default. That precedence is the pure `resolvePresenceVisible`
-// truth table tested here (the size-default math the e2e deliberately does NOT
-// exercise with 50 real peers — flood/autokill risk).
+// truth table tested here — and ONLY here: the size default is not reachable
+// from the e2e at any threshold value (spawning LARGE_CHANNEL_THRESHOLD real
+// peers trips the bahamut same-host autokill, and the harness exposes no
+// member-count seam), so these cases are the sole gate on the cutoff. Every
+// count is derived from the constant, never a literal: a literal keeps passing
+// on the OPPOSITE branch after a tune (#915 raised 50 → 200).
 
 describe("presenceFilter module", () => {
   beforeEach(() => {
@@ -26,18 +30,22 @@ describe("presenceFilter module", () => {
 
   describe("resolvePresenceVisible() — pure precedence truth table", () => {
     it("unset pref: visible below the LARGE_CHANNEL_THRESHOLD", async () => {
-      const { resolvePresenceVisible } = await import("../lib/presenceFilter");
-      expect(resolvePresenceVisible(undefined, 49)).toBe(true);
+      const { resolvePresenceVisible, LARGE_CHANNEL_THRESHOLD } = await import(
+        "../lib/presenceFilter"
+      );
+      // Well below the cutoff, not the boundary (that pair has its own case).
+      expect(resolvePresenceVisible(undefined, Math.floor(LARGE_CHANNEL_THRESHOLD / 2))).toBe(true);
     });
 
     it("unset pref: hidden at-or-above the LARGE_CHANNEL_THRESHOLD", async () => {
-      const { resolvePresenceVisible } = await import("../lib/presenceFilter");
-      // 50 is the boundary: >= threshold hides.
-      expect(resolvePresenceVisible(undefined, 50)).toBe(false);
-      expect(resolvePresenceVisible(undefined, 500)).toBe(false);
+      const { resolvePresenceVisible, LARGE_CHANNEL_THRESHOLD } = await import(
+        "../lib/presenceFilter"
+      );
+      expect(resolvePresenceVisible(undefined, LARGE_CHANNEL_THRESHOLD)).toBe(false);
+      expect(resolvePresenceVisible(undefined, LARGE_CHANNEL_THRESHOLD * 10)).toBe(false);
     });
 
-    it("boundary is exactly LARGE_CHANNEL_THRESHOLD (49 shown, 50 hidden)", async () => {
+    it("boundary is exactly LARGE_CHANNEL_THRESHOLD: one below shown, at it hidden", async () => {
       const { resolvePresenceVisible, LARGE_CHANNEL_THRESHOLD } = await import(
         "../lib/presenceFilter"
       );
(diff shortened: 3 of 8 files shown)


vjt:
## What I refuse to claim

**That the Elixir half is verified.** `lib/grappa/presence_filter.ex` (50 → 200) and
the new drift guard in `test/grappa/presence_filter_test.exs` have **not been run
once**. The dispatch scoped this as cic-only and told me to take no lane; that premise
was false, and `scripts/check.sh` / `scripts/test.sh` are the orchestrator-allocated
COMPILE lane, which I will not self-serve. CI is the first execution of that code. Two
specific things it could catch that I could not: whether `mix format` reflows the new
test block, and whether `File.read!("cicchetto/src/lib/presenceFilter.ts")` resolves in
the test container — `RouterSwDenylistTest` reads a sibling path the same way and
`themes/builtin_backgrounds_test.exs` records that the oneshot test container mounts
`cicchetto/src` (my path is under it), so I expect it to work. Expect, not verified.

**That the cic suite constrains the new value.** It does not, and I measured it rather
than assuming: constant back to 50, 4395 tests still pass. Any "the tests prove 200 is
live" reading of the green tick below is wrong. The only thing that now fails on a wrong
value is the cross-language guard, and only for *disagreement* — set both to 50 and
everything is green again. That is deliberate (the number is vjt's tunable, not a
contract), but it means **no gate in this repo asserts the cutoff is 200**.

**That the size-default branch is exercised end to end at all.** It never was. #222's
moduledoc already recorded why (bahamut same-host autokill, no member-count seam in the
harness), so at 50 the entire auto-activation rule — the actual feature in the issue
title — was gated by one vitest truth table and the server unit tests, and it still is.
Raising to 200 does not make this worse. It does not make it better either, and I did
not build a seam to fix it: that is a harness change, not a threshold change, and it
wants its own issue if vjt wants the coverage.

**That the server tests still pass with 200 members seeded.** `session_with_members`
(`resolver_test.exs:40`) feeds all members in ONE `353` line; at 200 nicks that line is
roughly 1.5 KB, against a client socket opened `packet: :line` with no `packet_size`.
I read the code and believe it delivers — the 512-byte `@rfc_line_limit` in
`irc/client.ex` is applied to OUTBOUND framing only, never to inbound — but "I read it"
is not "I ran it". A real ircd chunks NAMES across several 353s; if CI reds here, the
fix is to chunk the seed, not to shrink the threshold.

**That the historical DESIGN_NOTES entries are now correct.** The #222 entry (line
~10630) still reads `LARGE_CHANNEL_THRESHOLD = 50` and the #458 entry (~21469) still
says `@large_channel_threshold 50`. I left both untouched on purpose: that file is a
chronological record of decisions as made, and a new dated entry supersedes them. A
reader who greps for `= 50` and stops at the old entry will be misled. If the project
would rather annotate superseded entries in place, say so and I will — it is a
convention question, not an oversight.

**That 200 is right.** I have no evidence either way and did not look for any: no
member-count distribution from prod, no measurement of J/P/Q rate versus channel size.
It is vjt's call and the issue records it as made in conversation. What I can say is
that the number is now cheap to change — one line per language, and a test that fails
if you only do one of them.

vjt:
## Retracting three of the six refusals — the COMPILE lane was granted and I measured them

Not a correction of the reasoning above, an update to what is now evidence rather than
expectation. The other three refusals stand unchanged.

**Retracted — the Elixir half is verified.** `scripts/check.sh` ran on this branch, rc
**0**, working tree porcelain-clean before and after: `8 doctests, 54 properties, 5273
tests, 0 failures`, Dialyzer `Total errors: 0`, bats `1..328`. `mix format` had nothing
to reflow in the new test block — and I proved the container reads THIS worktree rather
than main before trusting that green, by injecting a bad indent and watching
`format.sh --check` return 1 pointing at my line 50.

**Retracted — `File.read!("cicchetto/src/lib/presenceFilter.ts")` resolves in the test
container.** It does. The guard read `200` out of the real file.

**Retracted — the 200-member seed works.** `session_with_members` at 200 and the
outbound controller test at 201 nicks both pass: 42 tests, 0 failures across the three
threshold-touching files. The single ~1.5 KB `353` line is delivered intact, which is
what I argued from reading `packet: :line` with no `packet_size` — the `@rfc_line_limit`
512 in `irc/client.ex` is outbound-only. Argued, and now run. No chunking needed.

**And the guard is not dead green.** Displacement, not its own pass: with cic at 50 and
the server left at 200, `test/grappa/presence_filter_test.exs` fails **alone** — rc 2,
`6 tests, 1 failure` — printing the two values and which page-up/live-tail disagreement
they cause. Restored to 200, the same set is green again.

## Still refused, unchanged

- **No gate in this repo asserts the cutoff is 200.** Set both languages to 50 and
  everything is green. The guard pins *disagreement*, deliberately; the number is vjt's
  tunable, not a contract.
- **The size-default branch is not exercised end to end, and never was.** Unchanged by
  this PR. Building a member-count seam in the harness is a harness change and wants its
  own issue.
- **The historical DESIGN_NOTES entries still say 50** (#222 at ~10630, #458 at ~21469).
  Left deliberately — that file is chronological and the new dated entry supersedes — but
  a reader who greps `= 50` and stops there will be misled. Convention question, happy to
  annotate in place if preferred.
- **I have no evidence that 200 is the right number** and did not look for any. vjt's
  call, recorded as made in conversation.

vjt:
## Retracting the fourth refusal: the log now says what is true

I had refused to claim the historical DESIGN_NOTES entries were correct, and left them
alone on the grounds that a chronological log is superseded by its newest entry. That
reasoning does not survive contact with how the entries actually get used: #222 asserted
`LARGE_CHANNEL_THRESHOLD = 50` and #458 asserted `@large_channel_threshold 50` and
"unset + ≥50 members → hide" — three statements about live behaviour that stopped being
true the moment this branch landed, in the file a reader is pointed at for authority.

That is not chronological honesty, it is the `numeric_router.ex` failure mode: a
moduledoc asserting 321/322/323 take the default path was cited rather than checked, and
hid #910 for months. Corrected at the source, in the shape the order specified — the
decisions ("we chose 50 in #222", "we mirrored it server-side in #458") stay intact as
history, the VALUES now name the constant and point at the #915 entry for what is in
force. #458's line about the equality being held by "both moduledocs cross-reference"
got the same treatment: that WAS the mechanism, and the entry now records where it
stopped being.

`docs/DESIGN_NOTES.md` is an ExDoc extra (`mix.exs:50`), so the edits are parsed by a
gate rather than merely proofread: `mix docs` rc **0**, `/tmp/w2-docs.rc`. The two
warnings in that log are pre-existing and in unrelated modules.

## Two refusals stand, unchanged

- **The size-default branch is not exercised end to end, and never was.** Untouched by
  this PR. A member-count seam is a harness change and wants its own issue.
- **I have no evidence that 200 is right.** Now stated in the PR body rather than only
  here, since burying it in a comment would let the body imply the number was derived.
  It is a judgement call from operating the thing. The guard pins equality precisely so
  the next judgement call costs two lines and cannot be half-applied.

vjt:
## Rebased onto #902 and re-gated on the real union

`origin/main` is `843dbabe`; branch head `3b49830a`, `git merge-base --is-ancestor
origin/main HEAD` exits 0. The rebase was clean — **DESIGN_NOTES did not conflict**,
which surprised me enough to check rather than assume: both entries are present and
ordered, `#902` at ~30600 and `#915` at ~30693. Force-pushed with a lease derived from
`@{u}`, not typed.

Your reason for making me do this was right and the previous green was worth less than
it looked: it attested to my branch merged with a main WITHOUT #902.

### Union gates (rc files on the worker host)

| gate | rc file | rc | result |
|---|---|---|---|
| `scripts/check.sh` | `/tmp/w2-check2.rc` | **0** | `8 doctests, 54 properties, 5285 tests, 0 failures`; Dialyzer `Total errors: 0`; bats `1..328` |
| `bun run test` | `/tmp/w2-cic2.rc` | **0** | 241 files, **4421** tests |

Logs `/tmp/w2-check2.log` (first line `CWD=` — run detached with an explicit cd) and
`/tmp/w2-cic2.log`. Porcelain-clean before and after. The counts moved with #902 in
them: Elixir 5273 → 5285, cic 4395 → 4421.

## The `:invited` seam: reported, not fixed — neither spec leaned on it

Your worry was well aimed at the right file and lands on nothing. Concretely:

- **Neither `issue222-presence-filter.spec.ts` nor `issue239-hidden-msg-unread.spec.ts`
  mentions `invited`, `pseudo`, `window-state`, `windowState`, `banner` or `ghost`** —
  zero matches in either. No pseudo-row barrier, no `data-window-state="invited"`.
- **The overlap in `ScrollbackPane.tsx` is textual only.** #902 touched it, but not one
  presence line: the diff has no `presenceRowVisible` / `channelPresenceVisible` /
  `memberCount` / `membersByChannel` hunk. The call sites only slid (1355→1342,
  2525→2512).
- **The helpers my specs enter through did not move.** #902 only APPENDED to
  `cicchettoPage.ts` (`inviteBanner`, `inviteBannerJoin`, `inviteBannerDismiss`);
  `selectChannel` and `openRailMenu` are untouched.
- **`selection.ts` — the other half of the #239 predicate, where the unread badge calls
  `presenceRowVisible` — is not in #902's file list at all.** The badge derivation and
  the render filter still reconcile to the one predicate.
- **One thing that looked like a real hit and is not.** #902 rewrote
  `pseudoChannels.ts` and dropped `invited` from the projected set, and `issue239`
  navigates to a `"Server"` window — so I checked whether that window comes from the
  rewritten projection. It does not: `cicchettoPage.ts:204` resolves `"Server"` to the
  `$server` network console, a separate concept; `pseudoChannels.ts` has no `Server`
  and now covers only pending/failed/kicked/parked.

**What I will not claim:** I established this by reading the diffs and greping the
specs, not by running them. The e2e needs the STACK lane, which I do not hold and am not
asking for while the deploy is frozen. If you want the two specs actually executed
against the union before you merge, that is a lane request, not something I can infer.

Not merged. Still holding for #917.

Issue #914, opened 2026-08-06
cicchetto: toggle to hide the »N jump-to-next-active button (#235)


vjt:
## Request

Add a user-facing toggle to hide the **jump-to-next-active-window** button (`NextActiveButton`, #235) — the `»N` affordance.

Requested on IRC by a mobile user: on a phone the button is a viewport-fixed overlay that sits over the content and cannot be dismissed, and for someone who navigates by tapping the sidebar it is permanent visual noise rather than an affordance.

## Current behaviour

`cicchetto/src/NextActiveButton.tsx` renders unconditionally whenever any window has unread activity:

```tsx
<Show when={hasActiveWindows()}>
  <button class={`next-active-btn next-active-btn-${props.variant}`} ...>
    <span class="next-active-glyph">»</span>
    <span class="next-active-count" ...>{activeWindowCount()}</span>
  </button>
</Show>
```

`hasActiveWindows()` is the only gate. There is no preference, and on the `mobile` variant the button is a viewport-fixed overlay (per the #235/#280 module doc), so it cannot be scrolled away either.

## Proposed change

A boolean display preference — off by default, i.e. current behaviour unchanged — that ANDs into the existing gate:

```tsx
<Show when={hasActiveWindows() && !hideNextActive()}>
```

Notes on scope:

- The natural home is `lib/displayPrefs` (the same synced-pref mechanism `syncedSetChannelPresencePref` uses for denoise), with a row in the Settings panel next to the other display toggles.
- **Hiding the button must not disable the verb.** `jumpToNextActiveWindow()` is shared with the Alt+A keybinding and Ctrl+N (#235 is explicit that there is no second code path); the preference is presentational only, and the keyboard route must keep working when the button is hidden. On mobile that leaves the user without a jump affordance, which is the point of the request.
- Both `variant` placements (`desktop` sidebar, `mobile` overlay) should honour one preference — unless the report is specifically about the mobile overlay, in which case a per-variant gate is worth discussing first. The requester's complaint is the mobile overlay.
- `activeWindowCount()` / `nextActiveKind()` keep feeding the same single source; nothing about unread accounting changes.

## Open question for triage

Whether this should be a plain on/off toggle or "hide on mobile only". Filing as the general toggle since that is what was asked for; narrowing is cheap if the reviewer prefers.

vjt:
`v0.13.0` adds the toggle exactly where the issue proposed: `NextActiveButton.tsx:38` becomes
[...]

Issue #449, opened 2026-07-26
cicchetto: display prefs are localStorage-only, so they do not sync across devices (presence filter + timestamp format)


vjt:
## Report

Hypnotize on #it-opers: disabled join/part/quit on N channels from desktop browser, the change did not appear on the iOS PWA. Same grappa account, two devices.

## Why it happens

Working as built, not a bug in the toggle. Display preferences are client-only:

- Per-channel presence filter (#222) — `cicchetto/src/lib/presenceFilter.ts`, localStorage, tri-state (`show` / `hide` / unset-follows-size-default).
- Timestamp format (#217) — `cicchetto/src/lib/timeFormat.ts`, `STORAGE_KEY = "cicchetto.timeFormat"`.

localStorage is per-origin *per browser*, so an installed iOS PWA keeps a store separate from desktop Chrome/Firefox. Nothing converges.

Notification prefs already do the opposite and it works: `cicchetto/src/lib/userSettings.ts` posts the full map to `/me/settings/notification-prefs`, server is authoritative, re-read on drawer open, explicitly so that "multi-device prefs converge after a single round-trip".

## Ask

Move display prefs to the same server-backed path, so one account sees one UI configuration everywhere.

## Scope

- Server-side storage on the existing `/me/settings` surface for: per-channel presence filter, timestamp format, and whatever #443 lands (text size, colored-nicklist toggle).
- Full-map PUT + read-on-open, mirroring the notification-prefs shape — no PATCH/diff semantics.
- **Keep the tri-state**: `show` / `hide` / unset must survive the round-trip. Unset is not a boolean and must not be flattened server-side; it means "follow the live member-count default" (`LARGE_CHANNEL_THRESHOLD = 50`).
- localStorage stays as the offline/pre-auth fallback and as the write-through cache; the server wins on conflict.

## Coordination

- **#443** ("display options" settings section) is `status:queued` and adds the UI surface these prefs live in. Whoever picks these up should land #443's section first, or land both together — otherwise the sync work has no home in the drawer.
- The moduledocs of both `presenceFilter.ts` and `timeFormat.ts` state "localStorage only — cic owns UI/display preferences client-side, no server-side persistence". That was a deliberate rule; this issue supersedes it for cross-device prefs. **Update those moduledocs** rather than leaving them contradicting the code.

Reported by Hypnotize per chan request on #it-opers; filed after he confirmed the two devices are the same account.

vjt:
[...]

Issue #235, opened 2026-07-13
cicchetto: 'jump to next active window' button — replicate irssi's Alt+A


vjt:
### Summary
Reported via a #grappa channel request.

The bottom tab bar gets cramped with many channels, and manually hunting for which one has new activity is tedious. Replicate irssi's `Alt+A` as a button: tap it and it jumps to the next window that has unread activity.

### Behavior
- Tapping the button moves focus to the next channel/query with unread activity, in **chronological order** of that activity (cycling through all of them).
- **Priority**: windows where you were **mentioned/highlighted** and **queries (PMs)** come first — surfaced ahead of ordinary channel traffic.
- Repeated taps cycle until no window with unread activity remains.

### Rationale
- The bottom bar is cramped with many channels; a single "next active" affordance beats scanning the bar.
- Mirrors well-worn irssi muscle memory (`Alt+A` = jump to next activity window).
- Chosen over the alternatives discussed in the same thread — a multi-level tab bar in portrait, and auto-reordering tabs by recent activity — both rejected on design grounds.

### Notes
- Per-window activity/unread state is presumably already tracked for the tab unread badges; reuse it for the ordering and the mention/query priority tiers.

vjt:
Design constraint (raised in the same #grappa thread): direct `Alt+<digit>` tab-access shortcuts are a **non-goal** for the browser/PWA — Chrome intercepts `Alt+1`..`Alt+9` for its own tab switching, so they never reach the app. This is part of the rationale for a dedicated "next active" button (Alt+A semantics) rather than numbered shortcuts. If a keyboard binding is added later, pick one that browsers don't already claim.

vjt:
Design confirmed (per channel decision). Placement guidance (soft — no strong opinion on the exact spot):

- **Mobile:** bottom, likely an overlay on the bottom bar, right side.
- **Desktop:** bottom-left, in the left sidebar.
- Put it somewhere sensible and easily accessible, and make it **auto-hiding**: when there are no windows with unread activity, the button disappears (nothing to jump to → no affordance).

Spec is otherwise complete; ready to build.

vjt:
Shipped in batch-0716 (cic bundle B_9zgxMI). Alt+A / Ctrl+N jump-to-next-active-window (tier-ordered, DMs included), auto-hide at 0 unread. Real RED->GREEN e2e both platforms.

Issue #484, opened 2026-07-27
e2e specs statically unvalidated: 19 pre-existing type errors + always-false comparisons; extend `check` to cover e2e/


vjt:
## Summary

The Playwright e2e suite (`cicchetto/e2e/`) is **statically unvalidated**. Neither
`scripts/check.sh` nor `cicchetto`'s `bun run check` typechecks or lints the e2e
specs:

- Biome's `files.includes` (cicchetto config) excludes `e2e/`.
- The root `tsconfig.json` `include` is `["src"]`.
- Playwright's runner only `esbuild`-strips types at runtime — it never
  typechecks.

So the 19 e2e spec files have been able to accumulate type errors that no gate
catches. Running the e2e tsconfig through `tsc --noEmit` (inside the
playwright-runner container, which is where `@playwright/test` lives) surfaces
**19 pre-existing type errors across 14 files**.

Discovered while adding an e2e spec for #473; #473 itself stays scoped to the cic
rail-actions rework and does not fix this debt (the new #473 spec typechecks
clean — zero errors).

### Reproduce

```
cd cicchetto/e2e
docker compose build playwright-runner
docker compose run --rm --no-deps playwright-runner npx tsc -p tsconfig.json --noEmit
```

## ⚠️ The 4 `=== "warn"` errors are NOT cosmetic — always-false comparisons (dead branches)

`TS2367` fires 4 times on `msg.type() === "warn"`. Playwright's
`ConsoleMessage.type()` **never returns `"warn"`** — the value is `"warning"`.
So the `|| msg.type() === "warn"` branch is **statically dead: it can never be
true.**

This is the **same bug class** as the dead regression detector found while
working on #473 (a guard/comparison that can never fire). In all 4 current sites
the dead branch sits inside a `page.on("console", …)` diagnostic forwarder, so
the concrete effect today is that cic **`console.warn` output is silently
dropped from the run log** — a debuggability hole, not (yet) a broken assertion.

The reason this must be called out explicitly: a comparison that can never be
true is exactly how a test assertion becomes vacuous ("an assert that cannot
fail"). Fixing these 4 must include an **audit** for whether the same
always-false pattern (`=== "warn"`, or any other no-overlap comparison) gates a
real assertion anywhere in the suite, silently making it pass. Do not treat
these as a mechanical find-and-replace.

Sites:

- `tests/issue200-ws-sub-teardown.spec.ts:67`
- `tests/issue200-ws-sub-teardown.spec.ts:98`
[...]

PR #919, opened 2026-08-06
cic(#914): a toggle to hide the »N jump-to-next-active button

vjt:
Issue #914. Adds an off-by-default display preference that hides the #235
`»N` jump-to-next-active button, in BOTH placements, without touching the
verb behind it.

## What changed

- `cicchetto/src/lib/hideNextActive.ts` — new boolean pref, module-singleton
  Solid signal + localStorage write-through.
- `NextActiveButton.tsx` — the pref ANDs into the existing gate:
  `<Show when={hasActiveWindows() && !getHideNextActive()}>`.
- `SettingsDrawer.tsx` — a fourth fieldset in the existing display-options
  section.
- Unit tests (7 + 3) and an e2e spec (desktop + `@webkit` mobile).

`jumpToNextActiveWindow()`, `activeWindowCount()` and `nextActiveKind()` are
untouched: the preference is presentational and unread accounting does not
change. Alt+A and Ctrl+N keep jumping with the button hidden. On mobile that
deliberately leaves no jump affordance — that is the request.

## Deliberate divergence from the issue: the pref is LOCAL, not synced

The issue proposes `lib/displayPrefs` (the #449 server-backed mechanism). **I
did not do that**, and this is the one decision worth a review comment.

`Grappa.UserSettings`'s `display_prefs` typedoc already draws the line, and
attributes it: per-DEVICE prefs stay client-local (`fontSize.ts` is excluded
from the sync on exactly that ground), device-neutral prefs converge
(timestamp format, colored nicklist, presence filter).

This one has a device dimension. The report is about the viewport-fixed
**mobile overlay**; nobody objected to the desktop sidebar control. #449 was
filed for the opposite complaint — a desktop toggle that stayed invisible on
the iOS PWA, i.e. a convergence gap on a pref with no device axis. Syncing one
preference across both variants here would blank the desktop button on a
device the user never complained about: a bigger behaviour change than the one
requested, made silently.

Reversal is cheap in one direction only, which is part of why it went this
way: local → synced is additive later, synced → local is a migration. Flip it
in one review comment if you disagree.

Two secondary consequences, both in favour: no Elixir leg (the server's
`validate_and_normalize_display_prefs/2` builds an explicit 3-key whitelist, so
a fourth key would need the server deployed *first* — a cic ships independently
via `--cic`, and an undeployed server silently drops the unknown key), and the
e2e toggles synchronously instead of racing a fire-and-forget PUT.

## Shape notes

- `colorNicklist.ts`'s shape, not `fontSize.ts`'s: the flag is read at RENDER
  time inside the gate, so a bare `localStorage.getItem` would leave a mounted
  button stale until reload.
- The Settings checkbox binds straight to the module signal. The nicklist row
  above it keeps a drawer-local mirror signal; that is a second copy of state
  that has to be kept in sync, so the new row does not copy it.
- No CSS: the three sibling fieldset classes carry no rules either, they are
  semantic hooks.

## Placement

No rearrangement. The display sub-page is one `settings-section-display` block
with three fieldsets; this is a uniform fourth append. The four-flat-controls
rearrangement mentioned in triage is the **push** sub-page — a different page.
Worth noting that the section already mixes local (`text size`) and synced
(`timestamp`, `nicklist`) prefs and the user cannot tell them apart.

## Test plan

- [x] `bun run test` — 242 files / 4431 tests, exit 0 (baseline 241 / 4421).
- [x] `tsc --noEmit` on `src` — exit 0.
- [x] `biome check src` — exit 0.
- [ ] e2e `issue914-hide-next-active.spec.ts` — **not executed**, see the
      comment below. Needs the exclusive stack lane.

### DIFF
--- a/cicchetto/e2e/tests/issue914-hide-next-active.spec.ts
+++ b/cicchetto/e2e/tests/issue914-hide-next-active.spec.ts
@@ -1,0 +1,118 @@
+// GH #914 — a Settings toggle that hides the #235 "jump to next active
+// window" (»N) button. Off by default; ONE preference governing BOTH
+// placements (desktop sidebar + mobile overlay).
+//
+// The preference is PRESENTATIONAL. `jumpToNextActiveWindow()` is shared with
+// the Alt+A keybinding and Ctrl+N — #235 keeps exactly one code path — so
+// hiding the button must not disable the verb. THAT is what the desktop test
+// exists to catch: after the button is gone, Alt+A must still jump. An
+// assertion that merely proves the button is absent would pass against a
+// broken build that also killed the keybinding.
+//
+// The mobile test cannot make that claim: there is no physical Alt+A chord on
+// a phone, so hiding the overlay deliberately leaves NO jump affordance —
+// which is precisely what the reporter asked for. It earns its place for a
+// different reason: the mobile button is a SEPARATE mount site (a viewport-
+// fixed overlay / ScrollbackPane float stack, #280), so it proves the one
+// preference reaches the other placement, not just the sidebar one.
+//
+// Seeding mirrors issue235-next-active-window.spec.ts: clear the autojoin
+// channel's baseline unread by focusing it, park focus on the neutral $server
+// window (outside the channel/query cycle), then have a peer DM the operator
+// so exactly ONE window — a tier-0 query — is unread. Waiting on the count
+// "1" also removes the seed race.
+
+import {
+  closeSettings,
+  loginAs,
+  openSettingsSection,
+  selectChannel,
+  sidebarWindow,
+  waitForDmListenerReady,
+} from "../fixtures/cicchettoPage";
+import { IrcPeer } from "../fixtures/ircClient";
+import { AUTOJOIN_CHANNELS, getSeededVjt, NETWORK_NICK, NETWORK_SLUG } from "../fixtures/seedData";
+import { expect, test } from "../fixtures/test";
+
+const CHANNEL = AUTOJOIN_CHANNELS[0];
+const DM_LINE = "914 direct message";
+
+const NEXT_ACTIVE_BTN = '[data-testid="next-active-btn"]';
+const NEXT_ACTIVE_COUNT = '[data-testid="next-active-btn"] .next-active-count';
+
+test.setTimeout(90_000);
+
+// Produce exactly one unread window (a tier-0 DM) with focus parked on
+// $server. Returns the connected peer so the caller can disconnect it.
+async function seedOneUnreadDm(
+  page: Parameters<typeof loginAs>[0],
+  peerNick: string,
+): Promise<IrcPeer> {
+  if (!CHANNEL) throw new Error("AUTOJOIN_CHANNELS empty");
+  await selectChannel(page, NETWORK_SLUG, CHANNEL, { ownNick: NETWORK_NICK });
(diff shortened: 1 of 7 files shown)


vjt:
## What I refuse to claim

**I have not run the e2e. Not once.** `scripts/integration.sh` calls
`testnet.sh up` unconditionally, so even a `--list` probe takes the exclusive
stack lane, and `@playwright/test` is declared in `e2e/package.json` but
installed only inside the runner image — `tsc -p e2e --noEmit` dies at
`TS2688: Cannot find type definition file for '@playwright/test'` before
reading a single spec. So the e2e here has had **no static gate and no
execution**. What I did instead is check every fixture symbol and signature it
uses by hand against `fixtures/cicchettoPage.ts` and `fixtures/ircClient.ts`.
That kills the dumb failures; it does not make the spec green, and I am not
going to write "e2e added" as though it were evidence.

**Six of my seven unit assertions do not constrain anything.** I measured this
rather than assuming it: I first implemented `hideNextActive.ts` as a plain
`localStorage.getItem` getter with no signal — the wrong shape, the one that
leaves a mounted `<Show>` stale until reload — and ran the file. Six passed,
including the set-then-get round-trip. One failed:

```
× re-runs a tracked read, so the mounted gate re-renders on toggle
  AssertionError: expected [ false ] to deeply equal [ false, true ]
```

Only the tracked-read assertion carries the reactivity claim. The other six are
localStorage plumbing.

**That measurement indicts an existing test, and I left it alone.**
`colorNicklist.test.ts:54` — "updates the reactive getter so subsequent reads
reflect the change" — is a set-then-get round-trip, which is exactly one of the
six that passed against the non-reactive probe. It does not constrain
`colorNicklist`'s signal shape; that module's reactivity is currently pinned
only by `issue443-colored-nicklist.spec.ts` in a real browser. I am not fixing
it in this PR — out of scope, and I would rather report it than quietly widen
the diff.

**I did not verify the mobile overlay claim by looking at a phone.** The
`@webkit` test asserts the button is absent, which is the degraded assertion,
and I said so in the spec comment rather than dressing it up. It earns its
place only because the mobile button is a separate mount site (#280 float
stack / viewport-fixed overlay), so it proves one preference reaches the other
placement. It proves nothing about the verb — there is no Alt+A chord on a
phone, and hiding the affordance there is the whole point of the request.

**The deploy-skew argument in the PR body is reasoned, not observed.** I read
`validate_and_normalize_display_prefs/2` and saw it build an explicit 3-key map,
so an unknown fourth key is dropped rather than rejected. I did not stand up an
old server and watch a pref evaporate. It is one input to a design call that
was already made on other grounds.

vjt:
Addendum, so the claim above does not go stale: when I wrote "no static gate
and no execution" that was true of every environment. It is no longer — CI is
running `issue914-hide-next-active.spec.ts` now, and that run is the spec's
FIRST execution anywhere. I am not watching it and I have taken no stack lane;
the reviewer reads the result.

What has not changed is the TS2688 gap that made this necessary (tracked as
#484): `@playwright/test` resolves only inside the runner image, so
`tsc -p e2e --noEmit` dies before reading a spec and no e2e file in this repo
gets a static gate. For this PR that means the CI summary alone is not enough
to settle it — the run log has to show the e2e count go up by one and name this
spec, because a spec that fails to load can pass by not running.

Issue #276, opened 2026-07-16
away: stop emitting redundant 'marked away'/'no longer away' server-window lines; use 💤 indicator


vjt:
## Problem
The **"marked away"** and **"no longer away"** messages in the **server window** are redundant noise. The away-state is already conveyed by the state signal cic consumes — the extra scrollback lines add nothing.

## Proposal
- **The server should not emit these lines at all** — not merely hide them cic-side. Keep only the **away-state confirmation signal** (the typed away event cic already renders as an indicator).
- **Root-cause the emission source first.** Per the CLAUDE.md invariant *cic never originates state* — these are almost certainly **server-emitted** system messages. If so, suppress at the server (do NOT band-aid by filtering in cic). Verify with data before changing code.
- Replace the **"away" text label with the 💤 (zzz) emoji** in the away indicator.

## Scope
- **grappa (server):** stop writing the "marked away"/"no longer away" lines to the server window; retain the away-state event/signal.
- **cicchetto:** away indicator uses **💤** instead of the "away" text.

## Open at build
- Confirm exactly where 💤 replaces "away" (window/nick away badge vs status text) — pick the one place the label currently reads "away".

## e2e (mandatory, real assertion)
- Setting away no longer writes a "marked away" line to the server window; clearing away no longer writes "no longer away".
- The away-state is still reflected (indicator shows **💤**).

_Filed on vjt's direct request (orchestrator scribe). Primary fix is server-side (suppress emission); cicchetto for the 💤 swap — hence both component labels._

vjt:
Shipped on main **5cca7bfeeab88daf7fdbde4e2bdb1ca2177a3f59** (merge of `f/276-away-noise-emoji`).

## Root cause (server, confirmed with data)
The "marked away"/"no longer away" lines are the trailing text of the upstream ack numerics **305 RPL_UNAWAY** / **306 RPL_NOWAWAY**. `Grappa.Session.NumericRouter` had them in `@active_numerics`, routing to `{:server, nil}` → `Session.Server` persisted a `:notice` row on `$server` (the RED test surfaced the exact row: `body: "You have been marked as being away", meta: %{numeric: 306}`). The away **STATE** rides a separate, already-correct path: EventRouter's 305/306 clauses fire the typed `away_confirmed` effect → broadcast on `Topic.user` → cic's `awayStatus.ts`. The two are cleanly separable.

## Fix (server-side, not a cic band-aid — honors "cic never originates state")
[...]

Issue #184, opened 2026-07-04
cic: /stats output is routed to a query window named after the stats letter instead of the server window


vjt:
Reported on-channel (device dogfood).

## Symptom
Running `/stats <letter>` (e.g. `/stats o`) opens a **query window named after the stats letter** (a query named `o`) that receives the STATS output — and it even lands **in archive**. The `/stats` numeric replies should stay in the **server/status window**.

## Repro
1. `/stats o` (or any stats letter).
2. Observe a new query buffer named `o` gets created (archived), containing the stats output.
3. Expected: output rendered in the server window; no query created.

## Suspected root cause
The client is deriving the target window from a parameter of the STATS numerics. Note `RPL_ENDOFSTATS` (219) is `<client> <letter> :End of /STATS report` and several `RPL_STATS*` numerics carry the stats letter as a param — the router is very likely treating that letter as a query/nick target and bucketing the whole reply set into a query named after it.

## Fix direction
- STATS numerics (211/212/213/214/215/216/218/219/241/242/243/244/245/246/247/249/250 …) are **server responses** → route to the server/status window unconditionally.
- Do **not** create/select a query buffer from a stats-letter param.
- Verify nothing else keys a query target off a bare single-letter numeric param.

Standalone client-routing bug; not push-related.

vjt:
Fixed on main — merged **9d6cd2f** (2 commits: `4be122d` fix + `9d6cd2f` e2e).

**Root cause: SERVER-side** (`Grappa.Session.NumericRouter`), not cic. The STATS reply family (211–219 RPL_STATS* + RPL_ENDOFSTATS, 240–250) was in neither the `@active_numerics` deny-list nor `@delegated_numerics`, so it fell through to the `scan_params/2` param-scan. Every `/stats` query terminates with **219 RPL_ENDOFSTATS `[nick, <letter>, "End of /STATS report"]`**; the bare stats letter is nick-shaped (`Identifier.valid_nick?("o")` is true), so the scan resolved `{:query, "o"}` and grappa persisted a `:notice` on `channel="o"` — surfacing as a bogus query tab AND leaking into Archive via `list_archive`'s `COALESCE(dm_with, channel)`. Same disease as the 004/042 connect-storm ghost. cic was a faithful mirror of the server's (wrong) routing target.

[...]

Issue #84, opened 2026-06-26
grappa: implement /list (channel enumeration, REST + command) + homepage link


vjt:
Per chan request (#sniffo): grappa has no `/list` yet. Needed as the prerequisite for the theme gallery (#83) and for channel discovery in general.

## Scope
- **Server:** enumerate channels. Expose both as an IRC `/list` command and a REST endpoint (REST-first, the command can wrap it).
- **Homepage:** consume the list and surface it on the home page (folded into this issue on purpose — a separate issue just to add a link is ceremony).

## Notes
- Output should carry at least channel name + topic + user count.
- This unblocks the theme gallery (#83) listing and the "configured channel links / easy join" work (separate issue).

vjt:
Shipped to prod 2026-06-26 (merged FF in e5f41d6, cold-deployed m42).

Delivered: server-side `/list` executor + populating-snapshot model (`Grappa.ChannelDirectory`, REST index/refresh, user-topic progress pings), the cic `DirectoryPane` + Shell pane router, the per-network 📇 channels row, and the HomePane browse link. e2e covers browse + join.

One minor v1 follow-up (directory-search abort-and-replace) is tracked separately in todo (Low priority) — not blocking closure.

PR #920, opened 2026-08-06
irc(#910): route the LIST family by code, and stop the doc denying it

vjt:
## What

321/322/323 were `:scan`-class, so `scan_params/2` read their middle params as routing
destinations. They are now a fourth deny-list family, `@list_numerics` — same shape of cut
as #184 STATS and #908 TRACE.

The three are **three different defects**, and only one of them was user-visible. That
distinction is the substance of this PR, so it is spelled out rather than averaged:

| numeric | shape | pre-fix decision | actually observable? |
|---|---|---|---|
| 321 RPL_LISTSTART | `[nick, "Channel", "Users  Name"]` | `{:query, "Channel"}` | **No — latent.** |
| 322 RPL_LIST | `[nick, "#chan", "5", "topic"]` | `{:channel, "#chan"}` | **Yes.** |
| 323 RPL_LISTEND | `[nick, "End of /LIST"]` | `{:server, nil}` | Already correct. |

**322 is the real one.** The channel branch wins, and unlike the query branch it passes
#640's window-existence gate untouched, so an unsolicited LIST persisted one `:notice` row
into *every listed channel's* scrollback — including channels the user has never joined.

**321 is latent, and the issue I filed overstates it.** #910 says it opens "a query window
literally named `Channel`". It does not: `Server.resolve_numeric_query_window/2` (#640)
collapses a `{:query, _}` decision back to `$server` when no such window is open. The pure
classifier is still wrong and still worth fixing — the router is documented as a syntactic
classifier and #640 is a downstream net, not a licence — but the blast radius is "the
operator has a DM open with a peer nicked `Channel`", not "a ghost window on every /LIST".

**Reachability is the `directory_refresh` watchdog.** The family is normally consumed by
`Session.Server`'s `%{directory_refresh: %{}}` clause, but the #84 watchdog NILS that
tracker on `:directory_refresh_timeout`; every late frame of a stalled refresh then falls
into the scan. `/quote LIST` never arms the tracker at all. That timeout is injectable
(`directory_refresh_timeout_ms`), so this is covered by a real integration test, not only
by a unit test of the decision.

## The other half: the doc was wrong, and so was a test

`NumericRouter`'s moduledoc asserted **by number** that 321/322/323 "route via the default
`{:server, nil}` path".

Asked to check whether a doc wrong about a family it names by number was wrong only once —
it was not:

1. **`server.ex:3277` repeated the same claim in its own words**: with a nil tracker the
   numerics "route to `$server` scrollback as before (manual /LIST)". False for 321 and 322,
   and stated in the very clause whose fall-through creates the exposure. Corrected.
2. **A test asserted the defect.** `"322 RPL_LIST is no longer delegated; routes to channel
   param"` pinned `assert {:channel, "#chan"}` and framed that outcome as B6.1 HIGH-3's fix.
   That is why this survived: not just a doc nobody re-read, but a green assertion ratifying
   it. Inverted here.

Both docs are corrected **in place** rather than deleted — the wrong claim is kept as the
record of how the family hid.

The design note's count goes 43 → 46 and "three families" → four, because a stale number one
paragraph from where that exact disease is diagnosed is the disease.

## Audit findings — reported, NOT fixed (out of scope)

- **Moduledoc priority item 1 names 5 delegated families (~20 codes). The set holds 71.**
  Missing from the prose: LUSERS, WHOWAS, LINKS, BANLIST, INFO/VERSION, UMODEIS,
  channel-state, join-failures, presence/WATCH/MONITOR, and both WHOIS-leg blocks. It is
  not false line-by-line, but it reads as an enumeration and is not one.
- **Item 3 calls 437 ERR_UNAVAILRESOURCE "an ack".** It is an error numeric; the inline
  comment on the list itself says "nick temporarily unavailable".
- **Verified TRUE, since I was checking anyway:** "`labels_pending` is populated SOLELY by
  the away command" — all three `prepare_label/2` call sites are `set_explicit_away` /
  `unset_explicit_away`. #276's "delegation wins over the label override" argument still
  rests on solid ground. The pre-fix deny-list count of 43 also checks out (20 STATS + 13
  TRACE + 10 explicit).

## On the default, for @vjt — not decided here

The sharper framing is that deny-list and allow-list are the *same mechanism*: both are
code-keyed tables. Nothing about the deny list's shape is the smell; its **default** is. The
real question is which side an UNKNOWN numeric falls on.

#910 makes the case sharper in a specific way, and it is not the 43→46 ratio. It is that
**this family was not found by a bug report.** It was found by a sweep, two years after the
module's own doc said it was fine. The scan's default does not merely guess wrong — it
guesses wrong *silently*, so families surface one accidental read at a time and the count is
a floor, not an inventory. That is an argument about discoverability, which a ratio cannot
make. Your call, not taken in this PR.

## Verification

- **RED first:** `/tmp/w2-910-red.rc` = **2** — 4 failures: 322 real shape → `{:channel, "#chan"}`,
  321 real shape → `{:query, "Channel"}`, the deny-list property (sampled 321) →
  `{:query, "looks_like_a_nick"}`, and the integration test (no `$server` row).
- **GREEN after:** `/tmp/w2-910-green.rc` = **0** (3 properties, 72 tests).
- **Full gate:** `/tmp/w2-910-check.rc` = **0** — `8 doctests, 54 properties, 5289 tests, 0
  failures`; Dialyzer `Total errors: 0, Skipped: 0` / `done (passed successfully)`; doctor
  passed both envs; bats 328 ok. Tree porcelain-clean before and after.
- **Mutation measured, not assumed:** `/tmp/w2-910-mut323.rc` = **2** with 323 removed from
  `@list_numerics` — and only the membership property fails. See the comment below.

### DIFF
--- a/lib/grappa/session/numeric_router.ex
+++ b/lib/grappa/session/numeric_router.ex
@@ -27,9 +27,10 @@ defmodule Grappa.Session.NumericRouter do
      nick-shaped but the "nick" is not a routing destination — it's the
      rejected nick (433/432), the unknown command name (421), the
      offending command's argument list (461), an ack (437), or a whole
-     server-directed REPORT family whose middles are data and type labels
-     (`@stats_numerics` #184, `@trace_numerics` #908, the connect-storm
-     tokens). These ALWAYS go to `{:server, nil}`. Without this deny
+     server-directed REPORT family whose middles are data, type labels or
+     table headers (`@stats_numerics` #184, `@trace_numerics` #908,
+     `@list_numerics` #910, the connect-storm tokens). These ALWAYS go to
+     `{:server, nil}`. Without this deny
      list, the param-scan below would happily route 433's "BLEH-as-nick"
      to a query window.
 
@@ -63,11 +64,17 @@ defmodule Grappa.Session.NumericRouter do
     class). So the routing table can only be keyed by the CODE, which is
     what both lists already do — the deny list and an allow list differ
     only in which side the UNKNOWN falls on, and we currently put it on
-    the guessing side. Three families have now been patched in
-    (#184 STATS, UX-4 bucket I connect-storm, #908 TRACE) and the deny
-    list stands at 43 codes against the two or three the scan's query
-    branch genuinely serves (401 and the legacy 2-param shape) — that
-    ratio IS the argument. The root-cause fix is to invert: enumerate the
+    the guessing side. Four families have now been patched in
+    (#184 STATS, UX-4 bucket I connect-storm, #908 TRACE, #910 LIST) and
+    the deny list stands at 46 codes against the two or three the scan's
+    query branch genuinely serves (401 and the legacy 2-param shape) —
+    that ratio IS the argument, and #910 sharpened it: that family was
+    not found by a bug report but by a sweep, two years after this very
+    moduledoc asserted BY NUMBER that it already routed correctly. The
+    scan's default does not merely guess wrong, it guesses wrong
+    SILENTLY — so the families surface one accidental read at a time,
+    and the count above is a floor, not an inventory. The root-cause fix
+    is to invert: enumerate the
     target-bearing numerics and default the rest to `$server`. It is not
     done here because it changes behaviour for every numeric in neither
     list, and because #221's WHOIS-leg guard is defined as "a `:scan`-class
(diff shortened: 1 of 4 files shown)


vjt:
## What I refused to claim

**1. That I fixed a ghost window for 321. I did not, and the issue saying so is mine.**

#910 — which I filed — says 321 yields "a query window literally named `Channel`". I went
looking for that window and it is not there. `resolve_numeric_query_window/2` (#640,
`server.ex:4420`) gates every `{:query, target}` on `QueryWindows.open?` and collapses it to
`$server` when no window exists. So the pre-fix user-visible destination for 321 was **the
correct one**. What was wrong was the decision, not the outcome.

I fixed it anyway, and the reason is not "it might bite someday" hand-waving: the arm is
reachable the moment a peer nicked `Channel` has a DM window open, and #640 is documented as
a window-existence *policy* in the server, not as a licence for the router to classify
wrong. But "latent classifier defect behind an existing net" is a different claim from
"ghost window on every LIST", and only the first one is true.

**2. That the 323 test constrains anything. I measured it; it does not.**

323's real wire form is `[nick, "End of /LIST"]`, and it reached `$server` pre-fix purely
because the trailing has spaces and so fails `valid_nick?/1`. Same accident that saved 262
RPL_ENDOFTRACE in #908 — but 262 had an escape: a dotless server name is a real spelling, so
#908 could write a test that was genuinely RED. **For 323 I know no real spelling that
routes wrong today**, and I am not inventing one to manufacture a red.

So I measured instead of asserting. With `@list_numerics` cut to `321..322`:

```
/tmp/w2-910-mut323.rc = 2
  1) property "all @active_numerics route to {:server, nil} regardless of params"
```

**One failure, and it is the membership property.** The test that asserts 323's true wire
shape passes before the fix, after the fix, and with 323 removed from the deny list. It
documents the wire; it constrains nothing. 323's membership is held by the property alone,
and the property pins class membership rather than behaviour on any real frame.

That is a weaker guarantee than 321 and 322 get, and it is on the record rather than
smoothed into "the LIST family is now covered".

**3. That the integration test proves the row previously landed in `#elixir`.**

It does not — it fails at the `assert_receive` for the `$server` broadcast, so the
`Scrollback.fetch(... "#elixir" ...) == []` line never executed on the red run. The
`{:channel, "#chan"}` decision is proven by the unit test; the integration test proves the
frame reaches the scan after the watchdog and now lands on `$server`. Two facts, two tests,
not one test doing both.

**4. That I audited the whole moduledoc.**

I checked what the doc asserts by NUMBER or by COUNT, because that is the class the issue is
about — and there I found a second false site (`server.ex:3277`), a stale count (43), a
5-families-for-71-codes enumeration, and one mislabel (437 as "an ack"). I also checked one
claim that turned out **true** (`labels_pending` is away-only — all three `prepare_label/2`
call sites are AWAY), because reporting only the hits would misrepresent the sweep.

I did **not** re-derive the CP13 priority ordering, the #221/#785 WHOIS-leg carve-out
reasoning, or the purity contract against the code. Those paragraphs are unverified by me
and I am not implying otherwise by having touched the file.

**5. That the deny-list default question is settled.** It is in the PR body as an argument
for @vjt, not as a decision. Nothing in this branch changes what happens to an unknown
numeric.

Issue #767, opened 2026-08-03
wireTypes drift gate never runs in CI — `gen_wire_types --check` lives only in scripts/check.sh


vjt:
## Problem

`mix grappa.gen_wire_types --check` is the cic↔server wire-shape drift gate: it
fails when a `lib/grappa/**/wire.ex` typespec was edited without regenerating
`cicchetto/src/lib/wireTypes.ts`. It is invoked from exactly ONE place:

* `scripts/check.sh:58` — `"$SRC_ROOT/scripts/mix.sh" --env=dev grappa.gen_wire_types --check`

No GitHub workflow runs it. `.github/workflows/ci.yml` does **not** call
`scripts/check.sh`; it re-lists the individual gates by hand (compile, format,
credo, sobelow, hex.audit, deps.audit, doctor, coveralls, bats, shellcheck,
dialyzer, docs) and the wire-types drift check is simply absent from that list.
`integration.yml` and `release.yml` do not run it either.

Verified by grep over `.github/workflows/*.yml` and over the repo for
`gen_wire_types`: the only executable reference outside the task itself, its
tests, and docs is `scripts/check.sh:58`.

## Impact

The gate only fires when a human remembers to run `scripts/check.sh` locally.
A worker who runs the cic-only gates (`bun run check` + `bun run build`, both
free lanes) or who has no COMPILE lane allocated cannot run it at all — and CI
will not catch it for them. This is not theoretical: a stale `wireTypes.ts`
mirror survived an entire merge with CI green.

The failure is silent in the worst way. A server-side Wire typespec change ships,
`wireTypes.ts` keeps the old shape, `tsc` is happy because the stale types are
internally consistent, and the divergence only surfaces at runtime when a
narrower drops a payload.

## Aggravating factor

CI duplicating `check.sh`'s gate list by hand is the root cause: the two lists
drift by construction, and this is one instance of that class. Any future gate
added to `check.sh` inherits the same hole.

## Possible fixes (not a decision)

1. Add a `mix grappa.gen_wire_types --check` step to the `test + lint + audit`
   job in `ci.yml` — smallest change, closes this instance only.
2. Fold the drift check into the `ci.check` alias in `mix.exs`, so both CI and
   `check.sh` inherit it from one list — closes the instance and shrinks the
   duplication.
3. Have CI invoke `scripts/check.sh` (or a CI-shaped subset) rather than
   restating its steps — closes the whole class, biggest change, needs thought
[...]

Issue #702, opened 2026-08-03
ci: the integration job uploads no server logs, so a CI-only failure cannot be post-mortemed


vjt:
Surfaced while investigating #623 (the `recover-identity-real` setup race that has now failed four times).

**The gap.** The `integration` job's failure artifacts are the Playwright HTML report and traces, and nothing else. There are no `grappa`, `bahamut`, or `services` logs. When a spec fails only in CI — which is exactly the case for #623, whose whole signature is load-sensitivity we cannot reproduce on a faster machine — the trace shows the browser side of a failure whose cause is on the server side, and the investigation stops there.

**Why it matters now.** #623 has been merged past twice in one night under the standing "sole red, assert 268" posture. Each of those merges was a judgement made without the one artifact that would settle whether the IDENTIFY ever completed upstream. The tolerated-red posture is only safe while we can still tell a known race from a new bug, and today we cannot do that from CI alone.

**What would close it.** Upload the testnet container logs on failure alongside the existing artifacts — at minimum grappa's, ideally bahamut + services too. Scope is the `integration` workflow's failure-artifact step; no product code.

Filed from the orchestrator during the #623 investigation; not triaged into the queue.

vjt:
`v0.13.0` captures the server half of an integration failure: `capture_container_logs()` in
`scripts/integration.sh:52-101` runs inside the EXIT trap **before** teardown (the ordering is the
property, pinned by `test/scripts/integration_log_capture_test.bats`), the service set is derived from
`docker compose ps --all --services`, and `.github/workflows/integration.yml` uploads it on failure
(`9226fd55`). Proven on a genuine red: run 31112257417 produced a `container-logs` artifact —
**15.4 MB**, against the ~29 KB the commit measured against stubs. No size cap exists; that is worth
a follow-up, not a reopen.

PR #926, opened 2026-08-06
ci: server-side artifacts for integration failures (#702) + wireTypes drift gate in CI (#767)

vjt:
Two CI-workflow gaps, one commit each.

## #702 — the integration job uploaded no server logs

Failure artifacts were the Playwright trace and the HTML report: the
browser's half of a failure whose cause is almost always upstream. A
CI-only red showed an unexplained absence in the trace and the
investigation stopped there.

The capture cannot live in a workflow step. `scripts/integration.sh`
tears the stack down in its EXIT trap, so by the time a step here could
run `docker compose logs` the containers are gone. It now dumps one file
per container to `cicchetto/e2e/container-logs/` from inside that trap,
**before** the tear-down, and the workflow uploads the directory as the
`container-logs` artifact on failure.

* **Failure-only.** A green run writes nothing.
* **Derived service set** (`docker compose ps --services`), not a hand
  list — same reasoning as the #441 shellcheck set. A `compose run --rm`
  one-shot (the Playwright runner) is already gone and self-excludes;
  its stdout is the job log anyway.
* **Sizes are printed** by the capture, so the per-failed-run cost is a
  number to read rather than one to guess. If it turns out fat, the cap
  gets set from that number.

`test/scripts/integration_log_capture_test.bats` pins the ordering,
which is the whole property. Measured: with the capture moved *after*
the tear-down the files still get written and only the ordering
assertion fails — so that assertion is load-bearing, not decoration.

## #767 — the wireTypes drift gate never ran in CI

`mix grappa.gen_wire_types --check` lived in `scripts/check.sh` alone. It
ran when a human remembered, and not at all for a worker whose available
gates are the cic ones. The failure it catches is silent: the typespec
moves, `wireTypes.ts` keeps the old shape, `tsc` is happy because the
stale types are internally consistent, and the divergence surfaces at
runtime when a narrower drops a payload.

Added to the `test` job right after compile — it reads typespecs off the
beams and needs nothing else, and the codegen has no `Mix.env()` branch,
so MIX_ENV=test produces the same file `check.sh` produces in `:dev`.

**Not fixed here, deliberately:** ci.yml re-lists check.sh's gates by
hand rather than invoking it, so the two drift by construction and this
was one instance. Closing the class means deciding the container-vs-host
split for a CI-shaped `check.sh` — a different piece of work, and the
issue says as much.

## Verification

Local: full bats set 331/331, `scripts/shellcheck.sh` clean over 72
derived scripts.

A CI change is only verified by running CI, so the two claims that
matter are pinned to this PR's own runs, not to reading the YAML:

* #767 — the step must **exist and be green** here, and must go **red**
  on a deliberate drift.
* #702 — the `container-logs` artifact must **appear** in a failed
  integration run.

Both need a deliberately failing run; see the PR discussion for the
throwaway proof commit.


vjt:
Tip commit `491b0ed3` is a **throwaway**, pushed on purpose to make this
PR's own CI answer the two questions reading the YAML cannot.

* **#767** — a comment the generator would never emit is appended to
  `cicchetto/src/lib/wireTypes.ts`. Expected: the `wireTypes drift gate
  (#767)` step goes red in `ci`, and it is the *only* red there (a
  comment keeps `tsc`/biome happy, so the `cicchetto` job stays green).
  That is the "it exists AND it bites" half.
* **#702** — `tests/proof-702.spec.ts` fails deliberately and the
  workflow's run step is scoped to it with `--grep "@proof-702"`.
  Expected: `integration` goes red after a normal stack boot, and the
  run carries a **`container-logs`** artifact next to
  `playwright-traces` / `playwright-report`, with one file per compose
  service plus `compose-ps.txt`. Scoping keeps the proof at a stack boot
  instead of the full ~25-minute suite.

The capture prints `ls -l` of what it wrote, so that run's job log also
gives the real per-failure byte cost — which is the number any future
size cap should be set from, rather than one I guessed today.

Once both runs have shown this, the throwaway commit gets dropped and
the branch goes back to the two real commits.

vjt:

Issue #922, opened 2026-08-06
NumericRouter moduledoc undercounts the delegated set: 5 families named, 71 codes present


vjt:
Found by the #910 sweep, reported and deliberately not touched in that PR.

## The discrepancy

`lib/grappa/session/numeric_router.ex`, `@moduledoc` item 1 ("Delegated"), names **five** families:

> away acks (305/306, #276), WHOIS (311–319), WHO (352/315), NAMES (353/366), MOTD (375/372/376)

That reads as roughly 20 codes. `@delegated_numerics` actually holds **71 distinct codes** (counted on `origin/main` @ `d5226bcf`).

## Families present in the set but absent from the doc

LUSERS, WHOWAS, LINKS, BANLIST, INFO/VERSION, channel-state, join-failure, presence, and **both** WHOIS legs.

## Why it is worth an issue rather than a two-line doc edit

This family was not found by a bug report. It was found by a sweep, two years after the moduledoc declared the routing sound — the same way #910's own LIST defect surfaced. A doc that under-describes the set does not merely mislead a reader: it invites the next contributor to conclude a numeric is unhandled and add a parallel path.

Note the shape of the risk, which is discoverability, not correctness: **the routing itself is not alleged to be wrong here.** The doc is.

## Verification status

- The 71 count and the 5-family text: **measured** on `origin/main` @ `d5226bcf`, after #910 merged. #910 corrected item 3 (the deny-list wording, adding `@list_numerics`); it did **not** touch item 1, so this stands post-merge.
- Whether any of the undocumented families is *also* mis-routed: **not measured.** Do not read this issue as alleging that.

## Related, already fixed in #910

A test asserted the bug: `"322 RPL_LIST … routes to channel param"` with `assert {:channel, "#chan"}`, written as if it were the B6.1 HIGH-3 correction. A green that ratified the defect. Inverted in #910 — noted here only so this issue is not read as the same finding.

vjt:
`v0.13.0` stops the moduledoc counting a set it cannot count: item 1 (`numeric_router.ex:22-34`) now
names no codes and points at the attribute (`fa345edf`). The issue's number is right — parsing the
`@delegated_numerics` block on main yields **71 entries, 71 distinct**. Bonus in the same commit: the
TEST-side mirror held only 44 of the 71, so the "all delegated numerics return `:delegated`" property
was covering three fifths of the set; it is exhaustive now. No routing bug was alleged, and none was
introduced.

Issue #911, opened 2026-08-06
Audit: the remaining nick-shaped middle params in :scan (RFC-shape, unmeasured)


vjt:
The low-confidence tail of w2's #908 sweep, filed so it is tracked rather than buried in a PR comment. **Read the confidence column before acting on any row** — w2 was explicit that only the 321/322 pair (now #910) and TRACE itself (#908) are code-verified. Everything below is RFC 1459/2812 + IRCv3 spec knowledge, with **no bahamut source read and no live capture**. If a shape is wrong, that row evaporates.

Predicate that makes a token a scan candidate: `^[A-Za-z\[\]\\`_^{|}][\w\[\]\\`_^{|}-]{0,29}$` — never a leading digit, never a dot.

| numeric | why `params[1]` is nick-shaped | confidence |
|---|---|---|
| 900 RPL_LOGGEDIN | `params[2]` is the account name; only bites when it differs from the nick | forwarding path verified in `client.ex` `process_line/2`; **the reply itself never observed** — unknown whether Azzurra emits it |
| 436 ERR_NICKCOLLISION | `[nick, "collided", "…KILL"]` — twin of 432/433/437, which **are** already denied | RFC shape |
| 472 ERR_UNKNOWNMODE | `[nick, "z", "is unknown mode char"]` — a bare letter, exact #184 twin | RFC shape |
| 257/258 RPL_ADMINLOC1/2 | 2-param free text; bites only if the admin line is one dotless word | conditional |
| 410 ERR_INVALIDCAPCMD | the CAP subcommand token | spec shape |
| 303, 732, 704–706, 908 | single-target blob is a bare nick; grappa never sends these, but `/quote` does | dormant |

Not re-counted here: STATS 220–239, already flagged as a known gap in the moduledoc.

**Every family above is operator-reachable** via `/quote <line>` (`send_raw`, `server.ex:1803`) — that is a documented escape hatch, and it is how #908 reached prod.

Suggested first step, if this is ever picked up: **capture real replies** (or read the ircd source) before changing routing for any row — the whole point of this list is that it is unmeasured. A confident deny-list entry built on a wrong shape silently denies a real destination, which is the same defect in the other direction.

Refs: #908, #910, PR #909's "what I refused to claim" comment.

vjt:
`v0.13.0` resolves every row of the issue's table against **both ircds' numeric tables**, not the RFC
[...]

PR #927, opened 2026-08-06
irc(#911,#922): measure the numeric router's two class tables

vjt:
Refs #911, #922. Both live in `lib/grappa/session/numeric_router.ex`; one commit for #922, four for #911.

## #922 — the moduledoc undercounted the delegated set

Item 1 named five families (away acks, WHOIS, WHO, NAMES, MOTD), reading as some twenty codes. `@delegated_numerics` held 71. LUSERS, WHOWAS, LINKS, BANLIST, INFO/VERSION, UMODEIS, INVITE-ack, channel-state, join-failure, presence and both WHOIS legs were absent, and the "311–319 = WHOIS" range quietly swallowed 314 RPL_WHOWASUSER and 315 RPL_ENDOFWHO — WHOWAS and WHO.

The fix is not a longer list; a longer list drifts the same way. **The note now names no codes at all** and points at the attribute, whose per-family comments already carry the WHY. A doc that does not enumerate cannot undercount.

The test-side mirror had the identical defect one layer down: 44 of 71, so the "all delegated numerics return `:delegated`" property covered three fifths of the set. UMODEIS, INVITE-ack, LUSERS, presence and the whole bahamut WHOIS-leg family could have been deleted from production with every test green. It is exhaustive now.

## #911 — the audit, done by reading the ircds instead of the RFC

#911 was filed as nine low-confidence rows of "params[1] looks nick-shaped", explicitly RFC recall with no source read and no capture, and said the first step should be to measure. So the measurement is the change: both bound ircds publish their entire numeric FORMAT TABLE — [azzurra/bahamut @ 3b6ccff `src/s_err.c`](https://github.com/azzurra/bahamut) (Azzurra, all of prod) and [solanum @ 115b1e2 `include/messages.h`](https://github.com/solanum-ircd/solanum) (Libera, and the e2e testnet's second node). Reading them end to end answers the shape question per code, with the emitting `file:line` beside it.

**Denied on that evidence:** STATS 220–239 (the gap #184's own note asked for — Azzurra emits 222/223/224/225/227/229 with the same `%c` class letter #184 was filed for, all with live emitters; the range is now 211–250 less delegated 221), ADMIN 256–259, HELP 704–706, SASL 900–908, MONITOR list 732/733, plus 410 ERR_INVALIDCAPCMD and 472 ERR_UNKNOWNMODE.

The worst bite is not exotic. 704 puts the help TOPIC in `params[1]`, so `/help join` on Libera resolves the reply to `{:query, "join"}` — an ordinary command, no `/quote`, no privilege. 257 is two-param, so B6.1 HIGH-4's 2-elem clause hands the operator's A-line text to the scan and a one-word `admin { }` block becomes the destination.

**Disproved and deliberately left alone.** 436 ERR_NICKCOLLISION is not the twin of 432/433/437: those name a nick that is not you, while every 436 emitter on both ircds passes the RECIPIENT's own name, so `nick_eq?/2` already lands it on `$server`. 303 RPL_ISON appends a trailing space after every name and the parser keeps a trailing verbatim, so even the single-hit reply is `"nick "`. Denying either would add an entry no test could redden. Both are written into the module, because an audit that lists only its hits gives no way to tell measurement from confident guessing.

**Reachability cuts against this PR and is stated as such:** bahamut defines no 900, 410, 704–706, 732 or 908 at all. Everything except STATS and ADMIN is Libera-only and cannot fire on Azzurra, where prod lives. STATS 220–239 went the other way — filed as a known gap and left uncounted, it is the one family the network we actually run emits today.

## A green that constrained nothing, and its retraction

I shipped a disjointness property and called it "not decoration" in the commit message. Mutation testing killed it: putting 221 back into production's `@active_numerics` left the file green, because the property compared the two mirrors in the test file against each other. It is also unrepairable behaviourally — `numeric_class/1` resolves delegated first, so a code in both sets is indistinguishable through `route/2` from one in delegated alone. Intent defect, not behaviour defect. Deleted, with the reasoning left where it stood. The commits are not squashed so the retraction stays readable.

Adding 410 and 472 also reddened the channel-prefix property, whose precondition was a THIRD hand-maintained mirror of these tables. It is derived from the other two now, and both were hoisted above every describe so it can be.

## Gates

`scripts/check.sh` whole, exit 0: 5300 tests / 54 properties / 8 doctests, 0 failures; 0 `not ok` across bats; Credo strict clean over 590 files; Dialyzer 0 errors; Sobelow clean; doctor passed. Tree clean before and after.

Mutation-checked: 8 mutations, each removing one deny-list family. Seven redden the specific test for the family removed (222/225, 256/257, 704, 900/908, 732, 410, 472). The eighth is the one described above.

## Not done

This does not invert the scan's default. It sharpens the argument — the deny list now covers essentially every server-directed reply family the two ircds define, against the two or three destinations the query branch genuinely serves — but the inversion still needs #221's WHOIS-leg guard re-sited first, per the existing moduledoc note. The audit is also not exhaustive: runtime-`%s` slots on bahamut-only families (271 SILELIST, 606 WATCHLIST, 618/620 DCC, 728/729 RESTRICTLIST, 334, 340, 342, 362, 382, 391) are classified but not judged.

### DIFF
--- a/lib/grappa/session/numeric_router.ex
+++ b/lib/grappa/session/numeric_router.ex
@@ -11,12 +11,27 @@ defmodule Grappa.Session.NumericRouter do
 
   Priority order (highest → lowest):
 
-  1. **Delegated**: away acks (305/306, #276), WHOIS (311–319),
-     WHO (352/315), NAMES (353/366), MOTD (375/372/376) — owned by
-     dedicated handlers in `EventRouter`. We return `:delegated`; the
-     caller skips the matrix. Delegation wins even over a matching label
-     (see `route/2` — `labels_pending` holds only away labels, so a
-     labeled away-ack must not resurrect the row #276 suppresses).
+  1. **Delegated**: numerics owned by dedicated handlers in
+     `EventRouter` (and, for a few terminal ones, `Session.Server`). We
+     return `:delegated`; the caller skips the matrix. Delegation wins
+     even over a matching label (see `route/2` — `labels_pending` holds
+     only away labels, so a labeled away-ack must not resurrect the row
+     #276 suppresses).
+
+     **The inventory is `@delegated_numerics` below, not this note
+     (#922).** An earlier revision named five families here — away acks,
+     WHOIS, WHO, NAMES, MOTD — reading as some twenty codes, against the
+     71 the attribute actually held. LUSERS, WHOWAS, LINKS, BANLIST,
+     INFO/VERSION, UMODEIS, INVITE-ack, channel-state, join-failure,
+     presence and BOTH WHOIS legs were all absent, and the "311–319"
+     range quietly swallowed 314 RPL_WHOWASUSER and 315 RPL_ENDOFWHO —
+     WHOWAS and WHO, not WHOIS. Under-describing the set is not merely
+     misleading: it invites the next reader to conclude a numeric is
+     unhandled and add a parallel path — the same alibi #910 found in
+     item 3 below, in this same moduledoc, about a family it named BY
+     NUMBER. So this note names no codes at all. The attribute's
+     per-family comment blocks are the authority; restating them is
+     precisely what drifted.
 
   2. **Label-based** (IRCv3 `labeled-response` cap): if the numeric carries
      a `label` message-tag AND the label is registered in `labels_pending`,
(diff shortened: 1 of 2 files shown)


vjt:

Issue #907, opened 2026-08-05
compose: the /msg arm awaits the query join BEFORE claiming the drain lock, leaving a clobber window


vjt:
Found by w2 **by inspection** while building #904, and deliberately NOT closed there — it is a distinct defect of the same class, and widening #904 to cover it would have grown a settled cut. **Not measured: this is a code-structure reading, no repro has been run.**

## The invariant that saves the other paths

`sendPacedBody` claims the drain lock **synchronously**. For the plain `privmsg` / `/me` path there is therefore no moment between the queue slot emptying and the lock being taken — the operator cannot type into a gap, because there is no gap.

## Where it breaks

The `/msg` arm `await`s `ensureQueryTopicJoined` (`cicchetto/src/lib/compose.ts:884`) **before** calling `sendPacedBody`. That await is the gap.

Trigger (narrow, all of them at once):
- a **queued** submission (i.e. #904's one-deep slot in play),
- **multi-line**, so the paced drain is involved,
- a **`/msg`** whose target query window already holds a draft — `home` falls back to the source window (#723),
- the operator types into the source window during the join await.

The first residue write then clobbers what they typed.

## Why it is worth a ticket rather than a note

The rule that makes every other path safe — **claim before you await** — is protected by no test. Nothing stops the next arm that grows an `await` from re-opening the same window silently.

Suggested shape (not a decision): claim the lock before the join await, or assert the ordering rule in a test that would fail if any arm awaits first.

Refs: #904 / PR #906, and the "what I refuse to claim" comment there (`issuecomment-5198629169`). Same class as #737 / #666.

vjt:
Internal; no production repro was ever observed. `v0.13.0` gives `sendPacedBody` a **required**
`prepare` callback that runs inside the claim (`eed7cd30`), so the `/msg` arm can no longer await the
query-topic join before claiming the drain lock — a future arm that grows an `await` has one place to
put it. **The issue's stated trigger is falsified:** a queued submission is not required, because
#904 empties the composer on the first Enter too. The oracle exists only mid-await, and the test takes
it there with the join promise held open (`compose.test.ts:945`); an after-the-fact assert would have
passed against the bug.

Issue #723, opened 2026-08-03
cic: /msg partial send leaves the full command in the source window — duplicate resend


vjt:
Found by the codebase-review campaign on `main` (`v0.10.0..388f883b`), `cicchetto/src` slice. Verified against the files at `388f883b`.

**File:** `cicchetto/src/lib/compose.ts:648` (residue write) and `:1380` (the clear that is skipped)
**Severity:** HIGH
**Related:** #666 (the paced/resumable paste this comes from), #254

## Problem

Every `sendPacedBody` call site passes the SOURCE window key — the same key the end-of-submit clear at line 1380 writes — so the residue and the clear agree on one owner. The `/msg` arm is the exception: it passes `channelKey(networkSlug, canonical)` (the query window) so the remainder co-locates with the operator's new focus, but the failure path then does `if ("error" in r) return r`, which returns before the shared clear at line 1380. The source window's draft is never touched.

Result: on a partial send the remainder exists in TWO places — as residue in the query window, and as the ORIGINAL, COMPLETE command still sitting in the source window's compose box.

A second leg: `sendPacedBody` overwrites the target key's draft unconditionally, and the query window is not a key it started from.

## Failure scenario

1. In `#chan`, paste `/msg bob l1⏎l2⏎l3`.
2. `l1` is delivered; `l2` fails fatally (WS drop, `invalid_line`).
3. Query window `bob` now holds the residue `l2\nl3` — correct, that is the #666 contract.
4. `#chan` still holds the full `/msg bob l1\nl2\nl3` — untouched.
5. The operator resends from `bob` (correct: `l2`, `l3` go out), later returns to `#chan`, sees the command still in the box and hits Enter.
6. `l1`, `l2` and `l3` are all delivered a second time — the duplicate resend #666 exists to prevent.

Separately: with a half-typed message already in `bob`'s compose box, a *successful* `/msg bob hi` from `#chan` silently erases it, because the final residue write for the target key is `""`.

## Suggested fix

Give the residue exactly one owner. When the `/msg` arm redirects the remainder to the query key, clear the source key's draft on the same path (or do not populate the source key in the first place), and make `sendPacedBody` refuse to clobber a draft in a window it did not start from.

vjt:
Shipped in **v0.11.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.11.0

The tag carries `.deb`, `.rpm` and `pkg.tar.zst` assets, so self-hosters have something to pull.

[...]

PR #928, opened 2026-08-06
cic(#907): claim the composer, THEN await the query-topic join

vjt:
Ref #907.

## The gap

The `/msg` arm awaited `ensureQueryTopicJoined` (the #254 subscribe-before-send
join ACK) and only then called `sendPacedBody`, which claims the #737 drain
lock. Across that await the composer is already empty — #904 takes the text out
AT DISPATCH — and unclaimed, so a keystroke landing there is overwritten by the
first residue write of the paced drain.

Every other paced-send arm is safe by the SHAPE of `sendPacedBody`: it resolves
the residue home and claims in one synchronous block, so there is no moment to
type into. Nothing protected that shape.

## The fix is the shape, not the arm

`sendPacedBody` takes a required `prepare: () => Promise<void>` and runs it
INSIDE the claim. `/msg` hands over its join; the other three call sites pass
`nothingToPrepare`. Required rather than optional is the point: the next arm
that grows an await has exactly one place to put it, and passing a no-op is the
question asked out loud.

Patching only the arm (claim → await → release → let `sendPacedBody` re-claim)
also closes the hole — there is no yield point between that release and the
re-claim — but it duplicates the `multi` rule and the key set in the arm, and
"safe because nothing can interleave between these two statements" is precisely
the reasoning that was already load-bearing and already unprotected.

Both candidate homes are claimed across `prepare`: the home choice reads
`preferred`'s draft (#723 — a busy query window refuses the redirect), and
reading it after an await it stayed writable during is deciding on stale
ground. The loser is released the instant the choice is made, so post-`prepare`
the locked set is byte-identical to the pre-fix one and a window the drain will
never write to is not held hostage — least of all the query window `/msg` just
moved the operator's eyes into.

## What I refuse to claim

- **The width of the window and how often it bites.** Nobody has timed a join
  ACK; no production repro was run. This stays what the issue said it was — a
  code-structure reading. Nothing in the fix or the test rests on a duration.
- **That the issue's trigger list is right.** It is not, and this is measured:
  #907 lists a QUEUED submission (#904's one-deep slot) among the conditions.
  Not required — #904 empties the composer on the FIRST Enter too. The new test
  queues nothing and still reds on the pre-fix code.

## The test, and why its oracle is mid-await

`#907 — /msg claims the composer BEFORE it awaits the query-topic join` fails
on the pre-fix code at `expect(isDraining(source)).toBe(true)` and passes after.

The oracle is deliberately the state DURING the await, not the settled state:
post-fix and pre-fix the final draft is IDENTICAL, because the operator's text
is destroyed rather than merged. A before/after assertion on the end state
would have passed against the bug. The test holds the join promise open itself
(`setEnsureQueryTopicJoined`, the #254 seam, is an injection point), types while
it is unresolved, and asserts the write was REFUSED — the #737 posture, refuse
rather than lose. No timing is guessed: the test resolves the promise.

A second test pins the counterweight — the window that loses the residue is
writable again for the length of the drain — so the fix cannot degrade into
"freeze everything". That one is green both before and after; it guards this
change's collateral, not the defect.

**No e2e, deliberately.** The observable is a textarea readOnly for the length
of one WS round trip. A browser spec racing that window would pass by luck,
which is worse than not having one.

## Gates

- `scripts/bun.sh run test` — 4433 passed / 242 files, green.
- `scripts/bun.sh run check` — exit 0 (biome + tsc); `tsc --noEmit` standalone
  exit 0.
- Server side untouched (cic-only diff plus a DESIGN_NOTES entry).

### DIFF
--- a/cicchetto/src/__tests__/compose.test.ts
+++ b/cicchetto/src/__tests__/compose.test.ts
@@ -833,6 +833,112 @@ describe("compose submit — slash command dispatch", () => {
     }
   });
 
+  // #907 — the claim must happen BEFORE the arm awaits. Every other paced-send
+  // arm is safe by the SHAPE of `sendPacedBody`: it claims synchronously, so
+  // there is no moment between the composer emptying (#904 takes the text out
+  // at dispatch) and the lock. `/msg` awaited the #254 query-topic join first,
+  // and that await was a gap the operator could type into — destroyed by the
+  // first residue write in the one configuration where the source window is
+  // also the residue home: a query window that is already busy (#723).
+  //
+  // The join is held open by the test, so nothing here depends on how long a
+  // real join ACK takes — the operator types while the promise is unresolved
+  // by construction, not by timing.
+  it("#907 — /msg claims the composer BEFORE it awaits the query-topic join", async () => {
+    localStorage.setItem("grappa-token", "tok");
+    const sb = await import("../lib/scrollback");
+    const api = await import("../lib/api");
+    const qtj = await import("../lib/queryTopicJoin");
+    const compose = await import("../lib/compose");
+    const source = channelKey("freenode", "#a");
+    const query = channelKey("freenode", "bob");
+
+    let ackJoin!: () => void;
+    let joinAwaited = false;
+    qtj.setEnsureQueryTopicJoined(() => {
+      joinAwaited = true;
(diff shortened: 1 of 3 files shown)


vjt:

Issue #925, opened 2026-08-06
cicchetto/iOS: send button drops long presses — per-button onPointerDown preventDefault violates keepKeyboard's mousedown rule (ComposeBox.tsx:554)


vjt:
## Symptom

On iPhone (PWA, keyboard up), tapping the compose **send** button sometimes does nothing: the button's `:active` styling fires, but the message is not sent and **the text stays in the field**. Reported on IRC by Scatolone_Fabbricone; the "text stays in the field" detail rules out a rendering problem — the send genuinely does not happen.

## Suspected cause

`cicchetto/src/ComposeBox.tsx:554` wires the submit button with:

```tsx
onPointerDown={(e) => e.preventDefault()}
```

This is the one place in the app that does what `cicchetto/src/lib/keepKeyboard.ts` explicitly forbids. That module's header comment (lines 17-25) is unambiguous:

> **CRITICAL: this hooks `mousedown`, NOT `pointerdown`.** iOS Safari dispatches BOTH events; `pointerdown` is also the gesture-start signal for scroll/pan […] `mousedown` is the legacy focus-shift carrier and does NOT participate in iOS's scroll-gesture dispatch.

The same comment records that a per-button `pointerdown` variant has already cost one bug (the archive modal list could not be scrolled, 2026-05-18).

The failure mode this predicts matches the report: on iOS a touch that the engine resolves as a pan synthesises neither `mousedown` nor `click`, while `:active` styling still applies — animation yes, send no. A long press is the easiest way to push a touch into that classification.

The global `keepKeyboard` capture listener already keeps the keyboard up for every tappable surface, so the per-button `preventDefault` added for #59 is redundant with it as well as harmful.

## Evidence

Standalone repro page with the send button wired **identically** to `ComposeBox` (disabled while empty, `pointerdown` + `preventDefault`), plus an A/B toggle for a proposed wiring.

Measured with the **proposed wiring on** (`pointerup` handler + `preventDefault` on `mousedown`), iPhone / iOS 18.7 / Safari 26.5.2, 402x874 dpr=3, **installed as PWA (standalone)**, safe-area top=62px bottom=34px — log supplied by Hypnotize:

| touch | press duration | chain | result |
|---|---|---|---|
| 1 | 648 ms | pointerdown → touchstart → pointerup | sent |
| 2 | 732 ms | pointerdown → touchstart → pointerup | sent |
| 3 | 898 ms | pointerdown → touchstart → pointerup | sent |
[...]

Issue #366, opened 2026-07-20
cicchetto: long-press selects the entire message text (keyboard-active fallback for native selection)


vjt:
## Context

Companion enhancement to #79 (*iOS text selection won't start while the keyboard is open*). Native char-level text selection is unreliable on mobile while the composer keyboard is active; #79 tracks the underlying native-selection failure.

## Enhancement

When the keyboard is active, a **long-press** on a message should **select that message's entire text** (select-all of the message body), instead of relying on native character-range selection.

This gives users a working "grab this whole message" path on mobile with the keyboard open, sidestepping the native-selection limitation in #79 rather than blocking on it.

## Notes / boundaries

- Must coexist with the touch-gesture work in #308 (edge-swipe drawers + center-swipe channel nav): long-press-to-select-all and the swipe gestures must not fight each other, and vertical scroll must stay intact.
- Scope is the whole-message select-all affordance; #79 stays the tracker for restoring native partial selection.
- Priority P0 per maintainer request.

Filed on the maintainer's behalf; queued for the build pipeline.

Issue #59, opened 2026-06-15
cicchetto: mobile keyboard collapses after sending via send button (textarea not refocused)


vjt:
## Summary

On mobile (reproduced on Android), the on-screen keyboard collapses every time a message is sent via the **send button**. The user has to tap the input box again to bring the keyboard back, which makes rapid messaging painful.

## Reproduction

- On a touch device, focus the compose textarea (keyboard appears).
- Type a message and tap the **send button** (the arrow glyph).
- The message is sent, but the textarea loses focus and the virtual keyboard collapses.

Sending the same message with **Enter** on the soft keyboard does **not** collapse the keyboard — focus stays on the textarea. So the issue is specific to the send-button tap path.

## Root cause

In `src/ComposeBox.tsx`:

- The compose `<textarea>` has no `ref`, and nothing restores focus to it after a send completes (`doSubmit` clears the draft but never refocuses).
- The send `<button>` is `type="submit"`. Tapping it moves focus from the textarea to the button, which dismisses the virtual keyboard. Since nothing refocuses the textarea afterward, the keyboard stays collapsed.

This is not a regression from the recent custom-keyboard work — the refocus was never present; only the button-tap path triggers it.

## Proposed fix

Either (or both):

- Add a `ref` to the textarea and call `.focus()` at the end of `doSubmit`.
- Add `onPointerDown={(e) => e.preventDefault()}` to the send button so the tap never steals focus from the textarea in the first place (the click/submit still fires).

Two-line change; happy to open a PR on the current branch if wanted.

PR #929, opened 2026-08-06
cic(#925): send on the event iOS actually delivers

vjt:
## What

The compose **send** button is a `type="submit"`, so it sends only when a `click` reaches the form. `lib/keepKeyboard.ts` already records — from a real-device measurement taken for #366 on 2026-07-21 — that iOS Safari synthesizes **no mouse events at all** for a press the OS routes into a long-press gesture; only taps get them. No mouse events → no click → no submit, while `:active` still lights up from the touch and the draft stays in the field. That is the reported symptom, and the mechanism was already written down in this repo.

Activation therefore moves to `pointerup`, which fires either way (the issue's device log shows it on all six long presses, 648–982 ms).

## The wiring

- `pointerup` submits, guarded by the `pointerId` that went down on the button.
- The synthetic `click` that follows a normal tap is swallowed so a tap cannot send twice — gated on `detail > 0`, because a keyboard activation reports `detail 0` and carries no pointerup to have sent for it; swallowing that one would drop the send outright.
- The release is hit-tested against the button's box (`releasedInside`): touch pointers are implicitly captured by the press target, so without the test, pressing send and sliding away to abort would still fire — and a message to a channel is not undoable.
- Focus retention (#59) stays, moved onto `mousedown` (the legacy focus-shift carrier) and off `pointerdown` (also iOS's gesture-start signal — the rule `keepKeyboard.ts` exists to enforce).

## One correction to the issue's proposal

Option 1 was "drop the per-button wiring, let the global `keepKeyboard` `mousedown` listener do its job". It cannot: that listener is `isIos()`-gated, and #59 was an Android report ("Android especially"). On iOS the button's own `mousedown` cancel is redundant with the global one and harmless; on Android it is the only thing there is. So the handler stays, on the right event.

## Scope of the evidence, stated plainly

**Not reproduced.** Playwright webkit does not carry iOS touch physics. The failing arm is field-reported (Scatolone_Fabbricone); the passing arm is Hypnotize's device log of *this* wiring, 6/6. What is **not** established is that the old `pointerdown` cancel was the thing losing the send — a control run with the old wiring on the same device is what would close that loop. This change is justified on the click's absence being a failure mode this repo had already measured, not on the suspicion the issue opened with. **vjt device-verifies on the phone.**

## Also

Two comments named the wrong event and were fixed: `main.tsx` described the global preserve listener as "preventDefault on pointerdown" (it hooks `mousedown`), and the send button claimed "same trick as the image-picker button" (that button carries a plain `onClick`). A comment that misnames the event is how this class of bug gets written again.

## Gates

- `bun run test` — 4439/4439 green (242 files), incl. 8 new `#925` cases: pointerdown not cancelled, mousedown cancelled, pointerup sends, synthetic click swallowed once, keyboard activation survives the armed swallow, slide-off aborts, re-arm across presses, orphan pointerup ignored.
- `bun run check` (biome + `tsc --noEmit`) — exit 0; `tsc --noEmit` standalone also exit 0.
- e2e not yet run (needs a lane) — six existing specs drive the send button in a real browser and are the meaningful regression gate for this change.

Refs #925.

### DIFF
--- a/cicchetto/src/ComposeBox.tsx
+++ b/cicchetto/src/ComposeBox.tsx
@@ -108,8 +108,23 @@ const NOTICE_DISMISS_MS = 3_000;
 //                strings built in compose.ts, previously computed + discarded).
 type Feedback = { text: string; severity: "error" | "notice" };
 
+// #925 — did the pointer come up inside the box it went down on? The one
+// piece of geometry in the send button's pointer-driven activation, kept
+// pure so the abort-by-sliding-off case is decidable without touch physics.
+// Inclusive on all four edges: a release exactly on the boundary is a
+// release on the control.
+function releasedInside(rect: DOMRect, x: number, y: number): boolean {
+  return x >= rect.left && x <= rect.right && y >= rect.top && y <= rect.bottom;
+}
+
 const ComposeBox: Component<Props> = (props) => {
   const key = () => channelKey(props.networkSlug, props.channelName);
+
+  // #925 — send-button press bookkeeping. Plain `let`s, not signals: nothing
+  // renders off them, they only carry state from one event to the next within
+  // a single press.
+  let pressedPointerId: number | null = null;
+  let sentFromPointer = false;
   const [feedback, setFeedback] = createSignal<Feedback | null>(null);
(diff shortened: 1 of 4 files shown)


vjt:

Issue #893, opened 2026-08-05
CI flake: Uploads.ReaperTest "sustained DB busy → best-effort drop" fails under load


vjt:
First observed 2026-08-05 19:43 on PR #891's `test + lint + audit` job ([run 31030931942](https://github.com/vjt/grappa-irc/actions/runs/31030931942)): `1 failure` out of 5185 tests, at `test/grappa/uploads/reaper_test.exs:194` —

    1) test sweep/2 — resilience sustained DB busy → best-effort drop:
       sweep survives, logs, leaves row un-soft-deleted (Grappa.Uploads.ReaperTest)

**Why it is not #891's doing, and not main's either.** #891 changes only `cicchetto/` files and one e2e fixture — zero Elixir — so it cannot reach this code path. And PRs #889 and #890 both passed the same `test + lint + audit` job against the same base within minutes of this failure, so main was not broken at that commit.

**Filed rather than waved through** because the tracked-flake list is how the second occurrence gets recognised as a real regression instead of being re-diagnosed as "just a flake" a sixth time. If this recurs, it is load-sensitive setup in the test (the classic cure is a deterministic setup, never a longer timeout or a weakened assert) — or the resilience path genuinely mishandles sustained BUSY, which is what the test claims to assert.

Not investigated further: I re-ran the failed job rather than reproducing locally. I do not claim to know which of the two it is.

vjt:
Root-caused, not re-rolled. Neither candidate cause in the issue was right: `BusyRetry` injects before
the op, so the test process never ran that UPDATE — the writer was the app's own reapers, which tick
every 60 s in every env while `DataCase` puts the Sandbox in SHARED mode for `async: false`. Proven by
displacement (one manual `:tick` reproduced the CI signature). `v0.13.0` sets
`config :grappa, :reaper_interval_ms` to 24 h in `config/test.exs` and injects it into all three
reapers from `application.ex` (**prod cadence unchanged at 60 s**), with
`application_ambient_reapers_test.exs` deriving the reaper set from `Supervisor.which_children/1` so
the filter cannot match nothing (`19e68c6b`). No assert weakened, no timeout raised.

PR #930, opened 2026-08-06
test(#893): take the ambient reapers out of the suite's write path

vjt:
## What the CI log actually says

The issue left two candidates open — load-sensitive setup, or a real
mishandling of sustained BUSY. It is neither, and the log settles it
([run 31030931942, attempt 1](https://github.com/vjt/grappa-irc/actions/runs/31030931942)):

```
1) test sweep/2 — resilience sustained DB busy → best-effort drop …
   Expected truthy, got false
   code: assert is_nil(reloaded.deleted_at)

   17:43:35.333 fault=busy_locked [warning] db write unavailable: SQLite pool
     saturated for the full 300ms retry budget (31 attempts) — returning :db_unavailable
   17:43:35.333 error=:db_unavailable [error] uploads reaper failure
```

Both degrade lines are present. The injected fault fired on all 31
attempts, the budget expired, the flip degraded, `{:ok, 0}` came back —
the resilience path did exactly what the test claims. And `BusyRetry`
injects *before* the op, so the test process never ran that `UPDATE`.
Something else soft-deleted the row.

## The writer nobody armed

`Grappa.Application` starts three periodic reapers (Visitors / Uploads /
Accounts) in **every** env, test included, and `Grappa.DataCase` puts
the Sandbox in SHARED mode for `async: false` tests. So an ambient 60s
tick issues its sweep on whatever test is live and mutates rows that
test just created. It does not even need the file: `File.rm` against
another test's temp root returns `:enoent`, and the enoent arm
soft-deletes anyway.

**Proven by displacement, not by re-rolling the dice.** A probe sent the
app-supervised reaper one `:tick` and reproduced the exact CI signature
first try (`deleted_at` set, row read back in the same test). What is
**not** claimed: the 1-in-5185 timing was never reproduced and no attempt
was made to — twenty green repeats prove nothing about a timer
coincidence.

## Fix

Cure the setup, not the assert. The cadence becomes a
`:reaper_interval_ms` knob read once at the application boot boundary,
pushed past any suite runtime in `config/test.exs`. The reapers still
boot (`Uploads.Reaper.init/1` mkdir_p's the global storage root) and
every reaper unit test drives `sweep/2` directly or starts its own
instance with a short explicit interval. **No assert weakened, no timeout
raised** — a nondeterministic cross-test writer removed. Production is
untouched (60s default; dev/prod configs set nothing).

All three reapers get the knob, not just Uploads: the exposure is any
`async: false` test whose rows match an expired-visitor, expired-upload
or idle-session predicate. Only one of the three had rolled a bad tick
yet.

## Oracle

`test/grappa/application_ambient_reapers_test.exs` derives its set from
`Supervisor.which_children/1` (a fourth reaper is covered the day it
lands) and asserts set-equality against the known three so the filter
cannot silently match nothing. **Verified RED against the pre-fix tree**:
with the knob at `60_000` the pin fails, naming `Grappa.Accounts.Reaper`
and its 60000ms.

## Gates

`scripts/check.sh` green end to end: `8 doctests, 54 properties, 5290
tests, 0 failures`, exit 0, working tree clean before and after.

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -95,6 +95,22 @@ config :grappa, :start_bootstrap, false
 # app-boot ServerSettings arm-read out of the test tree too.
 config :grappa, :start_source_alias_manager, false
 
+# #893 — the three ambient sweepers (Visitors / Uploads / Accounts
+# Reaper) are started by the application supervisor in EVERY env,
+# including test, and `Grappa.DataCase` puts the Sandbox in SHARED mode
+# for `async: false` tests. An ambient 60s tick therefore runs its
+# `delete_all` / soft-delete on the CURRENT test's connection: a
+# cross-test writer nobody armed, firing at wall-clock. It reaped the
+# row out from under `Uploads.ReaperTest`'s sustained-busy case (1 in
+# 5185 on CI) — that test's own sweep degraded exactly as designed while
+# the ambient reaper flipped `deleted_at` behind it. Push the cadence
+# past any suite runtime so NO ambient sweep fires during `mix test`;
+# the reapers still boot (Uploads.Reaper's `init/1` mkdir_p's the global
+# storage root) and every reaper unit test drives `sweep/2` directly or
+# starts its OWN instance with an explicit short interval. Not a
+# timeout bump: it REMOVES a nondeterministic writer from the suite.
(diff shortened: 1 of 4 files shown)


vjt:

Issue #816, opened 2026-08-04
cicchetto: handle multi-line paste in the composer (and keep Shift+Enter a no-op)


vjt:
## Problem

The composer currently has no defined behaviour for text that contains newlines.

A newline cannot travel inside a single message: `PRIVMSG` is terminated by CRLF, so a line break in the body is not representable in the protocol. The only way to send it is to split into N separate `PRIVMSG`s — and that is exactly where it gets dangerous, because the server's flood protection will close the connection without warning if the burst is too large. The limits differ per network, and users with the relevant usermode set are exempt, so we cannot assume a single safe value.

Historically no client made Shift+Enter insert a line break: mIRC's editbox is single-line and pops a "paste N lines?" dialog, hexchat splits paste line by line. There is no precedent to match, only a hazard to contain.

## Proposed behaviour

Agreed in channel with the people who raised it:

1. **Shift+Enter does nothing.** No line break inserted, no send. The composer stays single-line.
2. **Paste containing newlines asks for confirmation**, stating explicitly how many messages will be sent.
3. **Hard limit** on the number of lines: above it the paste is refused outright rather than confirmed.

Sending, when confirmed, should also be paced rather than dumped at once, for the same flood-protection reason.

## Open questions

- What the hard limit should be, and whether it is a constant or derived from anything the server tells us.
- Whether to offer a paste-to-service alternative when the limit is hit, or simply refuse.

Reported by emme\\k, with the flood-protection framing from morph.

vjt:
**Both open questions ANSWERED by vjt (2026-08-06) — this issue is unblocked and queued.**

> *"OK to make shift+enter display a 'IRC does not support multi-line messages'. for pastes we should add a button in the modal to make the paste a txt file upload."*

1. **Shift+Enter** — no longer a silent no-op. It shows **"IRC does not support multi-line messages"**. The user learns why nothing happened instead of thinking the key is broken.
2. **Paste with newlines** — the confirm modal gains a **button that turns the paste into a `.txt` file upload**. This supersedes the earlier "refuse outright vs offer a paste service" question: the answer is neither — we already have an upload path, so the paste becomes an attachment via the existing 📸-style flow (text on the wire, clickable link in cic).

The rest of the issue stands as written: paste confirms and states the message count; sending stays paced.

⚠️ The hard line-limit question is **moot** under (2) — a long paste becomes a file, so no constant needs inventing.

vjt:
[...]

Issue #80, opened 2026-06-25
cicchetto: confirm dialog before pasting multiple lines (flood guard)


vjt:
## Idea

To avoid accidental channel flood, when a user pastes content containing multiple lines, intercept it before sending and either:

- show a confirmation popup ("send N lines?") before proceeding, or
- buffer it into a small paste/clipboard area the user can review/edit first.

## Rationale

A multi-line paste straight into the compose area can flood a channel. Most terminal clients and webchats guard against this with a paste-confirmation threshold.

## Notes

- Trigger above a small line threshold (e.g. >2-3 lines) so single-line pastes stay frictionless.
- Optionally offer "paste as single message" vs "send line by line".

Suggested by a tester.

vjt:
Shipped: paste flood-guard confirm dialog. Merged to main 56609d00 (bundle DbpcrNXp), HOT --cic deploy, both nets health 200. Full integration 391 passed / 0 failed (incl issue80-paste-flood-guard.spec.ts), CI green on 56609d00 (run 29536833833). Reviewed (3 LOW addressed).

PR #936, opened 2026-08-06
cic(#816): make a multi-line body arrive only by explicit consent

vjt:
## What ships

A newline cannot travel inside a `PRIVMSG` (CRLF terminates the frame), so honouring one means splitting into N messages — and N at once is exactly what upstream flood protection closes the connection for. The composer had two routes onto that and neither asked.

**1. Shift+Enter is a no-op.** `preventDefault`, no send, no line break. It used to insert a newline. No client sets the precedent (mIRC's editbox is single-line and pops a "paste N lines?" dialog; hexchat splits paste), only the hazard exists. Closing this door is what makes the remaining one sufficient: **paste is now the only way a multi-line body reaches the box, and paste is guarded.**

**2. Any paste that becomes more than one message confirms**, and the dialog quotes **messages**, not lines — `Paste as N messages?` — with N computed by `splitMessageLines`, the function the send path itself uses.

**3. A hard cap of five messages, with two doors above it.** `PASTE_HARD_MESSAGE_LIMIT = 5`, a flat constant. Above it the block never enters the composer: the operator gets **"Upload as file"** or **cancel**. Never a bare refusal — a dead end only teaches them to paste it in two halves, which is the same burst with extra steps.

`classifyPaste/1` returns a closed `"insert" | "confirm" | "over-limit"` and the router switches on it once, so a future fourth arm is a compile error at every call site rather than a silent fall-through.

## The second door is the upload verb, not a new service

`text/plain` is **already** an accepted upload MIME in the `document` category (`uploadCategory.ts`, a 1:1 mirror of the server's `@mime_categories`, 10 MiB on the default host). So the overflow path needs **no new category, no new server surface, no new taxonomy**: the pasted text is wrapped in a `File` and handed to the existing `dropUpload`, and the orchestrator posts the resulting URL as one 📄-prefixed PRIVMSG. Same shape as the 📸 image path CLAUDE.md names as the model — media is a link, IRC stays text. CLAUDE.md's "reuse the verbs, not the nouns", applied literally; the 20% that didn't fit would have been the domain boundary, and there wasn't any.

The cap is a constant on purpose. The issue asked whether to derive it; there is nothing honest to derive from — flood limits differ per network and usermode-exempt users escape them entirely, so a "computed" ceiling would be a guess wearing a formula.

## The reversal, called out

#80's counter was deliberately **not** `splitMessageLines`, with a comment defending the difference: it counted lines as SEEN in the box (blank interior lines included), because "how big is this paste" was held to be a different question from the send-time fan-out. That was right while the dialog only warned about size. It stops being right once the dialog must state how many *messages* the paste becomes, because that number is a promise about what send will do. The two are now one computation, pinned by an equality assertion rather than a restated rule — so if the send-side notion of "blank" moves again (#863 moved it once), the dialog cannot silently start lying.

Visible effects: a lone trailing newline (the commonest copy artifact) no longer costs a dialog; a block with blank lines between paragraphs quotes the real frame count, and cannot be pushed over the cap by lines that produce no frame.

## Still open, deliberately

**The cap is per-paste.** Two consecutive within-limit pastes still accumulate in the draft, so a paste-time cap is not a send-time guarantee. Whether the ceiling also belongs at submit is a product question — referred to vjt, not settled by whoever happened to be writing the guard.

## Gates

- `bun run test` — **4451/4451 green** (242 files).
- `bun run check` (biome + `tsc --noEmit`) exit 0; `tsc --noEmit` standalone exit 0. (biome ignores `e2e/`, so the specs have no static gate — run, not linted.)
- Scoped e2e, chromium: `paste|multiline compose` → **6/6 green**, including the new over-cap spec and the rewritten #80 boundary arm. **A spot-check, not the ship gate** — the full `scripts/integration.sh` on this PR's CI is.

**The over-cap e2e was verified by mutation, not trusted for being green.** It passed in 1.2s, which is fast for a multipart POST → auto-send → IRC echo, so the upload MIME was temporarily changed to one `dropUpload` filters out: the spec then failed at the 📄 assertion after its full 20s poll. It is load-bearing. It also asserts the burst did NOT also happen — none of the pasted lines appear as their own message — so an implementation that uploaded *and* pasted could not pass it.

Three e2e specs had comments/assertions updated because they encoded the old rule: `issue80-paste-flood-guard` (the 3-line frictionless arm asserted behaviour the ruling removed; plus the new over-cap test), and `multiline-compose-fanout` (its header cited Shift+Enter as a multiline source).

Refs #816.

### DIFF
--- a/cicchetto/e2e/tests/issue80-paste-flood-guard.spec.ts
+++ b/cicchetto/e2e/tests/issue80-paste-flood-guard.spec.ts
@@ -29,6 +29,7 @@ import {
   confirmModalBody,
   confirmModalCancel,
   confirmModalYes,
+  scrollbackLine,
   loginAs,
   selectChannel,
 } from "../fixtures/cicchettoPage";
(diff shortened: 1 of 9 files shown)


vjt:

Issue #923, opened 2026-08-06
infra: bring the FreeBSD substrate to parity with Linux — dedupe, drop what Linux dropped, don't break the deploy


vjt:
Filed on vjt's direction (2026-08-06), as the companion to #901.

> *"in aggiunta a quella issue dovremmo portare il substrate freebsd on par con quello linux che ha visto più sviluppi di recente. vorrei deduplicare e assicurarmi che freebsd non contenga roba che su linux non c'è. ad esempio il doppio nginx. e vorrei assicurarmi anche che il fottuto deploy non vada in merda."*

## Why

`infra/linux/` has seen most of the recent development. `infra/freebsd/` — which is what **production actually runs** (the m42 bastille jail) — has drifted behind it and has accumulated things Linux no longer has.

## Scope

1. **Diff the two substrates** and enumerate the divergence honestly, both directions: what Linux gained that FreeBSD lacks, and what FreeBSD still carries that Linux dropped.
2. **Drop the leftovers.** Named example: **the double nginx.** #485 deleted the nginx container and made every substrate a dumb reverse proxy; if FreeBSD still stacks two, one of them is dead weight.
3. **Deduplicate** what both genuinely need, rather than maintaining two copies that drift again.

## Explicit constraints from vjt

- 🔴 **Do not break the deploy.** *"non che dobbiam fare cose estremamente complicate: siamo l'unico deployment freebsd quindi sti cazzi, ma ecco non sfasciamo neanche tutto in prod."*
- **We are the only FreeBSD deployment**, so backward-compatibility gymnastics for other operators are NOT required. Simplicity beats generality here.
- Correspondingly: this is **not** a licence to rewrite the deploy machinery. Smallest change that removes the divergence.

## Relationship to #901

#901 is the concrete instance (the source-alias wrapper ships only on COLD, so a wrapper change silently misses a HOT deploy — *"il wrapper va sempre spedito"*). **This issue is the general case.** Do #901's fix regardless; this one asks whether the substrate has more of the same shape.

## Verification

Deploy-class classification is part of the work: state for each change whether it is HOT or COLD, and prove the deploy still works before claiming done — a substrate change that breaks `deploy-m42.sh` is worse than the drift it fixed.

vjt:
## Enumeration — read-only audit of `origin/main` @ `d023b104`

No code shipped. Nothing deployed, no jail, no preflight run against a real host. Everything below is
either read from the tree or **measured** by running the classifier locally; each claim says which.

[...]

Issue #628, opened 2026-08-01
Deprecate the NDP keepalive (keep the Perl script, stop shipping it as a service)


vjt:
## Context

As of the 2026-08-02 cutover the bouncer runs in a VNET jail on a **routed** `/64`, with the per-session source addresses bound as `/128`s on `lo0`. There is no upstream neighbor that needs its NDP cache kept warm any more: the provider routes the block to us instead of resolving each address on the link.

`infra/freebsd/ndp_keepalive.pl` (with its `.sh` sibling and `rc.d/grappa_ndp_keepalive`) exists to ping the gateway every few seconds so the old proxy-NDP arrangement kept answering for the secondary addresses. On the routed setup it is dead weight: it was enabled only inside the old jail, which is now stopped and boot-disabled, so nothing runs it today.

## Ask (vjt, #grappa 00:30)

**Deprecate it — do not delete it.** The Perl script stays in the tree on purpose ("fa fico avere un sorcio in perl"). Concretely:

- mark `ndp_keepalive.pl` / `ndp_keepalive.sh` / `rc.d/grappa_ndp_keepalive` as deprecated in a header comment that says *why* (routed /64, no NDP proxying) and *when* it was retired;
- stop `jail_install_rcd.sh` from installing/enabling the keepalive service on new jails;
- leave the files in place and out of the boot path.

## Notes

- Nothing to migrate: no live process depends on it (verified on the host — no `ndp_keepalive` process, and the only `_enable="YES"` line is inside the stopped jail's `rc.conf.d`).
- The host-side `ndproxy` service is a **separate** thing and is out of scope here — it serves other jails; do not touch it as part of this.

vjt:
Shipped in **v0.10.0** (release commit `b4047f04`, tagged 2026-08-02). Pull that tag to get this — https://github.com/vjt/grappa-irc/releases/tag/v0.10.0

Issue #485, opened 2026-07-27
prod compose ships two proxies: drop the nginx container, let the BEAM serve the SPA (headers + body ceiling must move first)


vjt:
## Problem

The `prod` compose profile runs **two long-lived containers** — `grappa` (BEAM, published on `127.0.0.1:4000`) and `grappa-nginx` (`nginx:alpine`, published on `3000:80`) — plus the `cicchetto-build` oneshot. For anyone deploying grappa with Docker, that nginx is **a second reverse proxy behind the one they already have to run**: `INSTALL.md` Option B tells the operator to terminate TLS with their own nginx (`infra/nginx-tls-frontend.example.conf`) or Caddy. So the shipped topology is `operator TLS proxy → our nginx → BEAM`, where the middle hop serves files the BEAM can already serve itself.

## What the BEAM already does

This is not a porting job — the work landed in #399 and nobody removed nginx afterwards:

- `lib/grappa_web/endpoint.ex:131` — `plug :serve_cic_static` mounts `Plug.Static` on the runtime-resolved cicchetto dist root (`Grappa.Cic.Bundle.root/0`), placed before Telemetry/Parsers/Session/Router.
- `lib/grappa_web/router.ex:449-450` — `get "/service-worker.js"` (no-cache) and `get "/*path"` (SPA history fallback). The comment at `router.ex:437` states outright that these replicate nginx's `location = /service-worker.js` and `location / { try_files $uri /index.html; }` "so the BEAM self-serves the cicchetto SPA".
- `endpoint.ex:119` already documents nginx as **"RECOMMENDED for TLS + static fronting in production, not required"**.
- Websockets need no proxy help — Bandit handles the `/socket` upgrade natively; the nginx `location /socket` block exists only because the traffic is routed through it.

## What is genuinely nginx-only today

These must move into the app **before** the container is dropped, otherwise removing it is a regression:

[...]

PR #938, opened 2026-08-06
deploy(#923): scope the nginx class to the substrate that reads it

vjt:
A splinter off #923, not the issue: the parity audit turned up one defect that is
provable without a jail, and it is **not FreeBSD-only**. #923 stays open — everything
that needs the FreeBSD staging jail remains a proposal on the issue.

## The defect

`Grappa.Deploy.Preflight.classify_paths/2` scopes every boot-substrate class with
`filter_on/4` — `:image_substrate`→`:docker`, `:rc_d`→`:jail`, `:systemd_unit`→`:linux`.
`:nginx` was the only one that was not.

**Measured** before the change, by calling the classifier directly
(`mix run --no-start -e 'Grappa.Deploy.Preflight.classify_paths(...)'`):

```
infra/linux/nginx.conf              su jail   => COLD  [nginx: [...]]   <-- il jail non legge mai questo file
infra/linux/nginx.conf              su linux  => COLD  [nginx: [...]]
infra/freebsd/nginx.conf            su jail   => COLD  [nginx: [...]]
infra/freebsd/nginx.conf            su linux  => COLD  [nginx: [...]]
infra/freebsd/nginx.conf            su docker => COLD  [nginx: [...]]   <-- docker non ha piu nginx (#485)
infra/snippets/locations-api.conf   su jail   => COLD                   <-- corretto: e davvero condiviso
infra/linux/systemd/grappa.service  su jail   => hot                    <-- fratello correttamente scoped
infra/freebsd/rc.d/grappa           su linux  => hot                    <-- fratello correttamente scoped
Dockerfile                          su jail   => hot                    <-- fratello correttamente scoped
```

So **editing the Linux nginx config forced a COLD deploy on the m42 jail** — a restart
that drops every live IRC session in production, for a file the jail never opens. That
is the same failure as 2026-06-10, when a Dockerfile diff the jail never reads
cold-restarted prod. That incident is *why* `substrate` became an explicit required
argument; nginx simply never got the treatment.

The rule generalises, and the moduledoc now says so: a class that names a
substrate-specific path must be scoped, or the other substrates pay its restart.

## What changed

- **`nginx?/1` → `nginx?/2`.** `infra/freebsd/nginx.conf` is COLD on `:jail` only,
  `infra/linux/nginx.conf` on `:linux` only. `infra/snippets/*` stays
  substrate-independent — the jail nginx, the Linux nginx and the e2e conf all
  `include` it, so a change there really does concern everyone.
  A 2-arity predicate rather than `filter_on/4` because this class is **mixed** (two
  scoped literals + a shared prefix); routing it through `filter_on/4` would split one
  class across concatenated filters and reorder the reported file list for no gain.
- **Deleted the retired `grappa_ndp_keepalive` triple** (`ndp_keepalive.{pl,sh}` +
  `rc.d/grappa_ndp_keepalive`), retired 2026-08-02 by #628 when the VNET cutover removed
  the proxy-NDP neighbour cache it kept warm. Verified dead, not assumed: 40 grep hits
  across 9 files, every one the triple describing itself, a comment, or a doc; zero
  tests; `jail_install_rcd.sh` contains no ndp code. Its one live edge was the derived
  shellcheck gate, which still linted two of the three every CI run.
- **Corrected the claims that pointed at the deleted files** — `jail_install_rcd.sh`
  ("the scripts survive… re-add the install block"), `OPERATIONS.md`,
  `scripts/shellcheck.sh` ("the two `infra/freebsd/rc.d/` services", "the perl
  keepalive"), and `preflight.ex`'s `rc_d?/1` comment, which explained why the sibling
  service was excluded *and* claimed `jail_install_rcd.sh` refreshed its bytes — which
  that installer has not done since #628.
- **Two stale #485-era comments** in `jail_cic_build.sh` and `grappa.env.example` claimed
  nginx fronts the cic bundle via a `/usr/local/www/cic` symlink. There is no such
  symlink; `jail_install_nginx.sh` installs two `.conf` files and says so, and the BEAM
  has self-served the bundle via `Plug.Static` since #485.

## Proof — the mutation, not just the green

**Two existing tests asserted the defect verbatim** ("cold on both substrates"), so this
rewrites them rather than adding beside them: a test that encodes a bug stops anyone from
finding it. The six replacements mirror the `:rc_d` / `:systemd_unit` describe blocks.

RED first, against unfixed code — 4 failures, exactly the cross-substrate assertions:

```
1) infra/freebsd/nginx.conf → hot on docker
2) infra/linux/nginx.conf   → hot on docker
3) infra/freebsd/nginx.conf → hot on linux
4) infra/linux/nginx.conf   → hot on jail
104 tests, 4 failures
```

Then GREEN with the fix: `104 tests, 0 failures`.

Then the **discriminating mutation** — both clauses made substrate-blind again
(`nginx?("infra/linux/nginx.conf", _)`), which is exactly the defect:

```
not ok  infra/linux/nginx.conf   → hot on jail      <-- il caso che costa le sessioni
not ok  infra/freebsd/nginx.conf → hot on linux
not ok  infra/freebsd/nginx.conf → hot on docker
not ok  infra/linux/nginx.conf   → hot on docker
ok      infra/freebsd/nginx.conf → cold on jail     <-- resta verde
ok      infra/linux/nginx.conf   → cold on linux    <-- resta verde
104 tests, 4 failures
```

The two same-substrate assertions stay green under the mutation, so the guard
discriminates on the **scoping** axis, not merely on the class existing. Mutation
reverted; tree identical to the committed fix.

## Gates

`scripts/check.sh` **rc=0**, read from file, every stage confirmed to have run: credo
`found no issues`, sobelow `No vulnerabilities found`, doctor 98.1% doc coverage,
ExUnit `8 doctests, 54 properties, 5308 tests, 0 failures`, dialyzer
`Total errors: 0 … passed successfully`, `wireTypes.ts is in sync`, bats **331 ok / 0
not-ok**. No cic files touched.

## Deploy class of this PR

HOT. No cold class matches `lib/grappa/deploy/preflight.ex`, the deleted files, or the
touched docs — and the classifier change alters no *existing* verdict for the substrate
that actually reads each file.

## Deliberately NOT in this PR

The rest of #923 needs the FreeBSD staging jail that does not exist yet. In particular
the jail's nginx hop **stays untouched**: removing it would move `RemoteIpFromProxy` off
its loopback+XFF row onto the non-loopback row — the row `Plugs.LoopbackOnly` gates
`/admin/reload` on, which is the endpoint the hot deploy itself calls — and the m42 host
nginx config is not versioned in this repo. Also left alone: the FreeBSD nginx `listen`
templating (COLD, and the probe target changes), a FreeBSD bootstrap script, and the two
Linux-side defects the audit found. All enumerated on #923.

## Not claimed

- **Nothing was deployed and no jail was involved.** The classifier table above is a
  local `mix run --no-start`, not a deploy.
- I did not verify that a real `deploy-m42.sh` run now classifies HOT where it used to
  classify COLD — that needs the jail. What is proven is the classifier's verdict.
- The `infra/snippets/*`-on-`:docker` case is left COLD. Post-#485 that substrate has no
  nginx either, so it is arguably the same bug — but the snippet is genuinely shared and
  I have not measured a case where the extra COLD costs anything, so I did not touch it.
  Flagged on #923.
- The deleted keepalive files are proven unreferenced **in this repo**. I cannot prove
  nothing on the m42 host invokes them out-of-tree; the installed service was disabled by
  #628, but I did not inspect the running jail.


vjt:

Issue #793, opened 2026-08-03
cicchetto: shareable channel invite links — irc.sindro.me/<network>/<channel> → confirm → join


vjt:
Shareable link that opens grappa and offers to join a channel, Telegram-invite style: paste `https://irc.sindro.me/azzurra/sniffo` anywhere, the recipient clicks, gets a confirm, and lands in the channel.

Requested by Sonic (#it-opers, 2026-08-03 21:47: *"ma un link che ti fa joinare un chan? dopo conferma? tipo i gruppi di telegram? puo' essere utile per l'utenza media?"*). Accepted by vjt (23:05) with the shape constraint: **grappa is multi-network, so the link needs two params** — `irc.sindro.me/<network>/<channel>` (his example: `irc.sindro.me/azzurra/sniffo`).

**Queued deliberately**: vjt's instruction was *"accoda per il dopo-code-review"* — nobody starts this before the code-review finding queue is closed. Filed now only so it isn't lost.

## Behaviour

- `GET /<network>/<channel>` on the cicchetto host serves the app, and the app routes on it.
- Already authenticated **and** the network is bound → confirm modal *"Join #channel on <network>?"* → on confirm, join + switch to the window.
- Already in that channel → just switch, no modal (same rule as #648 — asking to join a window that is already open is noise).
- Not authenticated → login/register first, then the pending invite is applied once the session is up. The link must survive the auth round-trip or the whole "send it to a normal person" premise dies.
- **Confirm is mandatory** on the join path. A URL must never silently make someone join a channel.

## Where it goes (extend, don't stratify)

The deep-link machinery already exists — this is a second entry point into it, not a new subsystem:

- `cicchetto/src/lib/pushTarget.ts` already parses `?network=<slug>&channel=<name>` deep-links (the push-notification cold path), resolves via `networkBySlug`, defers the selection until `networks()` seeds, and then cleans the URL with `history.replaceState({}, "", "/")`. A path-shaped invite should normalise into **the same `PushTarget` and the same routing**, not a parallel reader. Note the current code reads `location.search` only — path parsing is the new part.
- Confirm modal: reuse `ConfirmModal.tsx` / `lib/confirmDialog.ts`.
- Joining: reuse the existing command path (`lib/slashCommands.ts` → `{kind:"join", channel, key}` consumed by `lib/compose.ts`). No new socket call.
- The `replaceState({}, "", "/")` cleanup already in `pushTarget.ts` applies here too: a refresh must not re-fire the invite.

[...]

Issue #648, opened 2026-08-02
cicchetto: clickable #channel in scrollback → confirm → join


vjt:
Channel names in the scrollback should be clickable: tapping one asks "do you want to join #channel?" and, on confirm, joins it.

Requested by vjt (#grappa, 2026-08-02 11:50).

## Behaviour

- A channel token in message text (`#sniffo`, `#it-opers`, …) renders as an affordance, not plain text.
- Left-click / tap → confirm modal: *"Join #channel?"* → on confirm, join and switch to that window.
- If the channel is one we're **already in**, skip the modal and just switch to that window — asking to join a window already open is noise.

## Where it goes (extend, don't stratify)

vjt's standing constraint on this one: *"non facciamo strati su quel che c'è già, cerchiamo di estendere e migliorare non stratificare che poi abbiamo consumer che vanno in conflitto"*. The seams already exist — all four of these are single-owner surfaces today and must stay single-owner:

1. **Tokenizing** — `cicchetto/src/lib/linkify.ts`. `LinkifySegment` is `{type:"text"} | {type:"url"}`; add a `channel` variant to that union so channel detection happens in the **same single pass** as URL detection. A second parser bolted onto `MircText` would race the URL branch (a `#` inside a URL fragment must stay part of the URL) and duplicate the trailing-punctuation cleanup that `stripTrailingPunctuation` already owns.
2. **Rendering** — `cicchetto/src/MircText.tsx:115` (`const segments = linkify(run.text)`) is the only consumer; add the branch there. Note the textual-emphasis layer at `MircText.tsx:46` runs over **text** segments — a channel segment must be excluded from it the same way a url segment is, or `#foo_bar_baz` gets eaten by the emphasis pass.
3. **Confirm** — reuse `ConfirmModal.tsx` / `lib/confirmDialog.ts`. No new modal component.
4. **Joining** — reuse the existing command path: `lib/slashCommands.ts:349` already produces `{kind:"join", channel, key}` and `lib/compose.ts:339` consumes it. Dispatch that same object; do **not** open a new socket call for this.

Visual affordance should follow the nick-clickable precedent (`ScrollbackPane.tsx:124-125`: rendered as an inline `<button>`, styled `.nick-clickable`) rather than inventing a new interaction style — same reason, and it keeps keyboard/a11y behaviour consistent.

## Edge cases worth pinning in tests

`cicchetto/src/__tests__/linkify.test.ts` is the pattern to extend:

[...]

Issue #399, opened 2026-07-25
Self-hosting: release cannot serve its own frontend; then ship .deb/.rpm


vjt:
## Problem

The release cannot serve its own frontend. `lib/grappa_web/endpoint.ex` has no `Plug.Static`, and `config/runtime.exs` documents that nginx terminates TLS and fronts the app in our deployment. That is fine for our production jail, but it means a self-hoster who installs the release gets a running BEAM node and a blank page: nothing publishes the cicchetto bundle or the uploads directory.

This surfaced while discussing distro packaging for self-hosting (`.deb` / `.rpm` in addition to Docker): you cannot ship an installable package of something that does not stand up on its own.

## Proposal

**Part 1 — make the release self-sufficient (prerequisite).**
Serve the built frontend and the uploads path from the endpoint, so `systemctl start grappa` gives a working instance on a plain HTTP port. nginx stays *recommended* for TLS and production fronting, not required. If we decide against this, the packaging in Part 2 must hard-depend on a web server and ship a documented site config — otherwise the package installs a broken instance.

**Part 2 — packaging.**
The release is already self-contained (bundled ERTS) and already runs natively in production without Docker, so the build side is nearly free. The work is the distro contour:

- system user, `systemd` unit, config under `/etc`, data under `/var`
- one descriptor producing both `.deb` and `.rpm` from the same release artifact
- packages attached to GitHub releases; no signed apt/yum repository until someone asks for one — that is the piece we would maintain forever

**Constraint (not negotiable):** the package must not run database migrations during install or upgrade. It installs a migrate command; the administrator runs it deliberately. An unattended `apt upgrade` must never touch a production database as a side effect.

## Notes

Filed per channel request. Not queued — scoping/priority to be decided before any work starts.

vjt:
Direction approved per channel request: go with Part 1 — the release serves its own frontend from the embedded web server, so a self-hosted install works from a plain service start. nginx remains the production front for TLS and static files in our own deployment; it is recommended, not required, for self-hosters.

Still not queued: packaging scope (Part 2) has not been prioritised.

vjt:
Scope decided per channel request — queueing this.

**Part 1 (prerequisite):** the release serves its own frontend and uploads from the embedded web server, so a plain service start yields a working instance. nginx stays the production front for TLS and static files in our deployment; recommended, not required, for self-hosters.

**Part 2:** build packages in CI from the same release artifact, one descriptor for all targets:
[...]

PR #939, opened 2026-08-06
cic(#793): shareable channel invite links — path → confirm → join

vjt:
Read side of #793: `irc.sindro.me/azzurra/sniffo` opens the app, asks, and joins.

## What it is

A second URL shape for the deep-link reader that already existed, not a second
subsystem. `pushTarget.ts` normalised `?network=&channel=` into a `PushTarget`
and deferred routing until the store it reads resolves; the invite path
normalises into the SAME `PushTarget`, uses the SAME reader and defer, and
hands the target to `channelJoin.ts`'s #648 verb `confirmJoinChannel` — which
already owns confirm → `postJoin` → switch, and already switches with no modal
when we are in the channel. No new socket call, no second confirm, no parallel
window state. `applyPushTargetFromUrl` is renamed `applyDeepLinkFromUrl`
because it is no longer about push alone.

## Two of the issue's premises were stale

- **Decision 4 (server-side routing) needed no work.** The SPA fallback for a
  two-segment path has existed since #399 (`SpaController`'s `GET /*path`), and
  since #485 every nginx substrate is a dumb `location / → BEAM` proxy with no
  allowlist to extend. Nothing server-side is touched in this PR.
- **The join path the issue points at is one verb short.** It names
  `slashCommands.ts` `{kind:"join"}` via `compose.ts`; the verb that actually
  matches — confirm, join, switch, and the already-in exception — is
  `confirmJoinChannel`.

## Decisions taken (2, 3) and not taken (1, 5)

- **Encoding.** A literal `#` is a fragment delimiter and never reaches the
  server, so the shareable spelling is bare and the parser implies the sigil;
  `%23` is accepted and not doubled, `& + !` pass through. A channel segment
  carrying a comma is REJECTED rather than escaped — JOIN takes a
  comma-separated list, so `/azzurra/a,b` would otherwise turn one invite into
  a two-channel join nobody wrote. Same for anything at or below 0x20 and DEL.
  `/share` is reserved: it is the only other two-segment client route.
- **Keyed channels stay out of scope**, as the issue defaults.
- **Decision 1 (cross-user network identity) is NOT settled here.** An invite
  naming a network this account has not bound joins nothing and says so in one
  toast — a visible dead end, because a silent one is indistinguishable from a
  broken link. Whether the segment becomes a globally resolvable identifier, or
  the flow grows "add this network, then join", is still open.
- **Decision 5:** read side only. No "copy invite link" affordance here.

## Two things worth review attention

- **The URL is cleaned BEFORE `render()`**, not after the routing lands (the
  push path's timing). The router has no route for a two-segment path, so
  leaving it in place mounts the app on nothing at all. Cleaning early still
  satisfies what the push path cleans for: a refresh must not re-fire the
  invite. The pending target survives the login round-trip on its own, since
  cic's login is a client-side navigate and the channel list cannot resolve
  before the session is up.
- **The invite waits on the CHANNEL LIST, not on `networks()`.** That is the
  source its already-in check reads, and `channelsBySlug` is keyed on
  `networks`, so it resolves strictly later. Gating on the earlier signal read
  an unresolved list, concluded "not in it", and asked for consent to join a
  channel the operator was already sitting in.

## Verification

- vitest: 4467/4467 on the rebased tree (16 new in `inviteLink.test.ts`).
- `bun run check` (biome + `tsc --noEmit`): rc=0.
- e2e chromium `--grep "#793"`: 3/3 — link → confirm → in the channel with the
  address bar clean; already-in → switch with no modal; unbound network → the
  notice, and no modal.
- The already-in gate is proven by MUTATION, not asserted: reverting it to the
  `networks()` gate (build kept compiling) turns that spec red, and restoring
  it turns it green.

## What is NOT claimed

- The unauthenticated → login → invite-applies round trip is covered by the
  deferral logic and unit tests, not by an e2e: a spec that fires a real login
  mints a live session on the shared testnet and cascade-poisons downstream
  specs.
- A hard refresh while sitting on `/login` with a pending invite loses it — the
  target lives in memory, and that corner was accepted rather than given a
  sessionStorage store of its own.
- Only chromium was run locally; the full integration suite is CI's call.

### DIFF
--- a/cicchetto/e2e/fixtures/grappaApi.ts
+++ b/cicchetto/e2e/fixtures/grappaApi.ts
@@ -602,6 +602,25 @@ export async function joinChannel(
   }
 }
 
+// GET /networks/:slug/channels — the channel list cic itself reads at boot
+// (`networks.ts` `channelsBySlug`). A spec whose PRECONDITION is "already in
+// this channel" must wait on THIS list rather than on `joinChannel` returning:
+// the POST only asks, and the row lands once the upstream JOIN completes. A
+// spec that races it is testing the boot ordering, not the behaviour it names
+// (#793, where the invite's already-in branch was scored against a channel the
+// server had not listed yet).
+export async function listChannelNames(token: string, networkSlug: string): Promise<string[]> {
+  const url = `${GRAPPA_BASE_URL}/networks/${encodeURIComponent(networkSlug)}/channels`;
+  const res = await fetch(url, { headers: { authorization: `Bearer ${token}` } });
+  if (!res.ok) {
+    throw new Error(`listChannelNames: unexpected status ${res.status}`);
+  }
+  // The response IS the array (no named wrapper) — same shape cic's
+  // `api.ts` `listChannels` reads.
+  const body = (await res.json()) as Array<{ name: string }>;
+  return body.map((c) => c.name);
+}
+
 // PATCH /networks/:slug — T32 connection_state transition. Mirrors
 // `cicchetto/src/lib/api.ts`'s `patchNetwork`. Used by the parked-
 // flow e2e: setting `connection_state: "parked"` triggers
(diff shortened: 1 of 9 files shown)


vjt:

Issue #149, opened 2026-06-29
Restore mix hex.audit as a blocking CI gate when cowlib ships a patch


vjt:
## Why this exists
In PR #147 (2026 EEF CVE dep bump) `mix hex.audit` was made **non-blocking**
(`continue-on-error` in ci.yml + `|| true` in the mix.exs `ci.check` alias)
because cowlib has two advisories that **cannot be fixed by a version bump**:

- `CVE-2026-43969` / `GHSA-g2wm-735q-3f56` (LOW)
- `CVE-2026-43966` (MEDIUM)

Both are `cow_cookie:cookie/1` request-cookie **composition** issues. osv.dev
reports `fixed: None` — every cowlib version ≥ 2.9.0 is affected, no patched
release exists. `mix hex.audit` has **no per-advisory ignore mechanism**, so it
stays red as long as cowlib is in the tree.

These advisories are **N/A to grappa**: we only *consume* Cookie headers via
Plug; we never *compose* outbound request cookies via cow_cookie, so the
vulnerable path is unreachable (same assessment that already justified ignoring
the LOW on `mix deps.audit`).

`mix deps.audit` REMAINS a hard, blocking CVE gate (it supports per-advisory
ignores and currently passes) — so CVE enforcement is not lost, only the
redundant tool that can't express the N/A assessment is demoted to advisory.

## To restore (when cowlib publishes a fix)
1. `mix deps.update cowlib` to the patched version.
2. Remove `continue-on-error: true` from the "Hex audit" step in `.github/workflows/ci.yml`.
3. Remove the `|| true` wrap around `mix hex.audit` in the `ci.check` alias in `mix.exs`.
4. Confirm `mix hex.audit` + `mix deps.audit` both green, drop this issue.

Priority: low (no functional impact; advisories are N/A). Tracking only.

vjt:
## Patch-status check (2026-07-10)

cowlib **2.18.0** shipped **2026-07-02** on hex.pm — a new release after this issue was filed (tree is currently on 2.17.1). The CVE fixes exist at the git-commit level (osv.dev lists fixed commits for CVE-2026-43966 + CVE-2026-43969).

[...]

PR #941, opened 2026-08-06
chore(#149): the hex.audit derogation now describes the four advisories it covers

vjt:
#149 asks to restore `mix hex.audit` as a blocking gate "when cowlib ships a
patch". This measures whether that is possible yet. It is not — and while
measuring, the derogation turned out to be covering twice what it says.

## The question, answered with numbers (OSV, 2026-08-06)

| advisory | package | what it is | fixed in |
|---|---|---|---|
| CVE-2026-43966 | cowlib 2.18.0 | response splitting via non-VCHAR bytes in `cow_http_struct_hd:escape_string/2` (CWE-113, SI:L) | **nothing** — 2.19.0 enumerated as affected |
| CVE-2026-43969 | cowlib 2.18.0 | request-cookie injection in `cow_cookie:cookie/1` (CWE-93, AV:L) | **nothing** — 2.19.0 enumerated as affected |
| CVE-2026-59248 | cowlib 2.18.0 | unbounded HPACK/QPACK prefixed-integer decode → memory exhaustion (CWE-770, **AV:N VA:H**) | cowlib 2.19.0 |
| CVE-2026-65624 | cowboy 2.17.0 | `max_headers` bypass via duplicate header names → memory exhaustion (CWE-770, AV:N VA:L) | cowboy 2.18.0 |

**No bump closes all four.** cowlib 2.19.0 + cowboy 2.18.0 (both released)
would close the bottom two; the top two have no fixed release at any version,
so `hex.audit` stays red and the hard gate cannot be restored. **#149 stays
open** — this is a comment change, not a lock file.

## What was wrong, beyond being out of date

- **Count.** The note describes TWO advisories. Four are live under the
  derogation; the two newest arrived after it was written and nobody re-read it.
- **Identity.** It records both unfixable ones as `cow_cookie:cookie/1`
  request-cookie *composition* issues, N/A because grappa only consumes Cookie
  headers. **43966 is a response-header path** in `cow_http_struct_hd` — that
  argument never covered it.
- **Where it lives.** The derogation was stated in three places, and CLAUDE.md's
  security section still called `mix hex.audit` a blocking gate, two months
  after #147 demoted it.

## What this changes

The justification moves from "N/A by argument" to **unreachable by build env**:
cowboy and cowlib enter the tree ONLY through `bypass` (`only: :test`, via
plug_cowboy), production serves on Bandit, and the release is built at
`MIX_ENV=prod` — so none of the four is present in a deployed grappa. That is
narrower, verifiable, and *falsifiable*: it dies if either package becomes a
runtime dep or grappa moves off Bandit, whatever gets patched. The comment says
so explicitly.

The full list now lives ONCE, on the `ci.check` alias in `mix.exs`; the ci.yml
step and CLAUDE.md state the posture and point at it.

`mix deps.audit` remains the hard, blocking CVE gate — untouched.

## Decision surfaced, not taken: the two available fixes

Bumping cowlib → 2.19.0 and cowboy → 2.18.0 closes 59248 + 65624. It is a
`mix.lock` change, which the deploy preflight reads as **COLD** — for packages
that are not in the prod release, so the COLD buys no production security.
Recommendation: fold it into the next COLD-forcing change rather than ship it
alone. Say the word and it becomes a second commit here.

## What is NOT claimed

- **No gate was run.** This branch touches `mix.exs`, and the COMPILE lane is
  held by another worker, so `mix format --check-formatted`, `mix hex.audit`
  and `mix deps.audit` were NOT executed against it. The change is comment-only
  (consistent 8-space indentation inside the existing alias list) plus YAML and
  Markdown comments, but CI is the first thing that will actually compile it.
- The advisory set comes from OSV, not from `mix hex.audit` stdout — I could not
  run the tool. The two sources agreeing on FOUR is a reported observation, not
  something I verified myself.
- `api.osv.dev`'s by-version query returns zero vulns for cowlib 2.19.0, which
  contradicts the advisories' own `affected.versions` lists (2.19.0 is in both).
  I trusted the advisory records, not the query endpoint — worth knowing before
  anyone re-measures with a one-line curl.
- `docs/TESTING.md` documents a local `mix deps.audit --ignore-advisory-ids
  GHSA-g2wm-735q-3f56` while CI runs `deps.audit` bare and is green. I did not
  reconcile that; it needs a run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,13 +81,21 @@ jobs:
         run: mix sobelow --config --exit Medium
 
       - name: Hex audit (CVE check)
-        # Advisory-only (non-blocking): cowlib has two UNFIXABLE cow_cookie
-        # advisories (CVE-2026-43969 LOW + CVE-2026-43966 MED, fixed=None,
-        # no patched release) that are N/A to grappa (we only consume Cookie
-        # headers via Plug, never compose request cookies via cow_cookie).
-        # hex.audit has no per-advisory ignore, so it can't express that.
-        # `mix deps.audit` below remains the HARD, blocking CVE gate.
-        # Restore this to blocking when cowlib ships a patch — see #149.
+        # Advisory-only (non-blocking). FOUR advisories sit under this
+        # derogation today, not the two it was written for: cowlib 2.18.0
+        # carries CVE-2026-43966 (response splitting in
+        # cow_http_struct_hd:escape_string/2) and CVE-2026-43969
+        # (cow_cookie:cookie/1 injection), neither of which has a fixed
+        # release at ANY version, plus CVE-2026-59248 (HPACK/QPACK
+        # memory-exhaustion DoS, AV:N VA:H, fixed in 2.19.0); cowboy 2.17.0
+        # carries CVE-2026-65624 (max_headers bypass, fixed in 2.18.0).
+        # It holds because cowboy/cowlib enter ONLY via `bypass`
+        # (`only: :test`) — prod serves on Bandit and ships neither — NOT
+        # because of the severities, and NOT because of the cookie-composition
+        # argument, which never covered 43966. hex.audit has no per-advisory
+        # ignore, so it cannot express that. `mix deps.audit` below remains the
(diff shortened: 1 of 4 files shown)


vjt:

Issue #943, opened 2026-08-06
cicchetto: AdminUsersTab swallows 422 field_errors — admin sees bare "create: validation_failed"


vjt:
Reported on IRC by a self-hoster: creating a user from the admin UI fails with

```
failed: create: validation_failed — click ↻ refresh to retry
```

and nothing else. The admin has no way to know *which* field is wrong.

## Cause (read on `origin/main`)

`cicchetto/src/AdminUsersTab.tsx:104-105` renders the raw wire token and drops the payload:

```ts
const code = err instanceof ApiError ? err.code : "create_failed";
setError(`create: ${code}`);
```

The server already ships the detail: the unified 422 envelope carries `field_errors`, mirrored into `err.info.field_errors` by `readError` (`cicchetto/src/lib/api.ts:129`). `cicchetto/src/lib/friendlyApiError.ts:190-205` exists precisely to turn that into a compact `field: msg` summary — `AdminUsersTab` is one of the few surfaces that never calls it (`AdminSettingsTab`, `Login`, `ThemeEditor`, `PasskeySettings`, `TotpSettings`, `DirectoryPane`, `HomePane`, `ThemeGallery`, `SettingsDrawer` all do).

Same pattern at the other four error sites in the file: line 81 (list), 120 (toggle admin), 149 (rotate password), 164 (delete).

## What the reporter was most likely hitting

`lib/grappa/accounts/user.ex:73-79`:

- `name` must match `^[a-zA-Z][a-zA-Z0-9_\-]*$` — starts with a letter, then alphanumeric/`_`/`-` only, length 1..64 (a dot, space, or accented char in the nick fails)
- `name` unique
- `password` length 8..256

Any of those produces a `field_errors` entry the UI currently throws away.

## Fix

Route the five `setError` sites through `friendlyApiError(err)`, keeping the operation prefix (`create: `, `delete (name): `, …). No server change needed.

## Not verified

Not reproduced against a live instance — this is a code read plus a field report. The check is cheap: submit the create form with a name containing a dot and see whether the banner names the field.

vjt:
Reporter confirmed on IRC: the failing field was the password (under 8 chars).

[...]

PR #945, opened 2026-08-06
cic(#943): append the rejected field to the admin users error banner

vjt:
Ref #943.

## The issue's premise was false; the symptom was real

#943 asked to route five `AdminUsersTab` sites through `friendlyApiError`,
because "the other admin tabs already do". They don't — `AdminSettingsTab.tsx`
lines 33-35 refuse it *in writing* ("operator console policy — operators want
the wire token for debugging"), and a sweep finds **39 raw-token sites across 8
admin tabs with zero calls** into the friendly map. `AdminUsersTab` is the rule.

Two of the five prescribed sites also can't reach the error class in question:
`index` is a GET, and `delete`'s `@spec` is `{:error, :not_found | :last_admin}`
(`users_controller.ex:53`, `:176`). A `field_errors` branch there is dead code.

What *is* real: `fallback_controller.ex` answers a bad changeset with
`%{error: "validation_failed", field_errors: …}` and `Accounts.User` writes the
one sentence that says what to type instead — "must start with a letter, then
alphanumeric/_/-". The banner rendered `err.code` alone and dropped it.

## What this does — additive, per vjt's ruling

The raw wire token **stays** (policy, not oversight); the per-field detail is
appended: `create: validation_failed — name: must start with a letter…`.

Only the three verbs whose controller `@spec` admits an `Ecto.Changeset.t()`
take the new door: **create**, **toggle admin**, **rotate password**.

The per-field walk already existed, inline in `friendlyApiError`'s
`validation_failed` arm. It's lifted out as `fieldErrorSummary` and the arm now
calls it — one implementation, no drift. `operatorApiError(value, fallback)`
sits beside `errorMessage` as its deliberate opposite: same "an `unknown` out of
a catch" contract, token-first policy for the operator console.

Absent / null / non-object / all-empty `field_errors` degrades to the **bare
token rendered today** — not `""`, not `undefined`.

## Declined

Widening to the other 36 raw-token sites. Each needs its own answer to "can this
endpoint even produce a changeset?", and guessing ships dead branches. Separate
issue, not a drive-by.

## Test plan

- [x] TDD: 3 new component tests RED first (`create: validation_failed — click ↻ refresh to retry` observed, field missing), then green.
- [x] Non-regression test pins the 422-**without**-`field_errors` banner to the byte-identical string it renders today.
- [x] `operatorApiError` unit matrix: every rung of the degradation ladder + the non-`ApiError` fallback.
- [x] `WRITABLE_CIC=1 bun run check` (biome + tsc) — rc=0, tsc confirmed run, 0 `error TS`.
- [x] `bun run test` full suite — rc=0, **243 files / 4486 tests passed**.

cic-only: no lane needed.

### DIFF
--- a/cicchetto/src/AdminUsersTab.tsx
+++ b/cicchetto/src/AdminUsersTab.tsx
@@ -10,6 +10,7 @@ import {
   adminUpdateUserPassword,
 } from "./lib/api";
 import { token } from "./lib/auth";
+import { operatorApiError } from "./lib/friendlyApiError";
 
 // Admin-panel bucket 5 — Users admin tab.
 //
@@ -40,6 +41,14 @@ import { token } from "./lib/auth";
 // Per CLAUDE.md "No localized strings server-side": error tokens come
 // from the server as snake_case strings ("last_admin",
 // "validation_failed"); cic owns human-readable rendering.
+//
+// #943 — the banner shows the RAW wire token, on purpose: operator-console
+// policy (AdminSettingsTab lines 33-35). The three verbs whose controller
+// `@spec` admits an `Ecto.Changeset.t()` (create / toggle admin / rotate
+// password) additionally append the 422's per-field detail via
+// `operatorApiError`, because the token alone never says which field is
+// wrong. `delete` and the list GET cannot produce a changeset, so they keep
+// the plain `err.code` narrowing.
 
 type CreateForm = {
   name: string;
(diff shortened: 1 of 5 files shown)


vjt:

Il cherry-pick ha riscritto le SHA, quindi GitHub non chiude questa PR da solo. Il contenuto è su main come: `f5fdd11e` + `661ae3d1`

Verificato per CONTENUTO, non per SHA: `git diff origin/main...<branch>`, ogni riga aggiunta cercata su `origin/main` — 164 righe controllate, **zero mancanti**.

Issue #934, opened 2026-08-06
e2e: the per-test reset tail costs ~4.8 min a run, in bursts nobody has explained


vjt:
## What was measured, and why this is not #808

#808 asked whether the per-test `_vjtReset` restart is worth making conditional. It was measured (#808 comment, 2026-08-06) and **declined**: the restart is ~159 ms, ~36% of a reset; the other ~64% is the DB/ETS drain plus the 200-row scrollback re-seed, which no reconnect condition touches. Measured bounds on the saving were 1.45–3.72 min out of a 32.1-min test phase (4.5–11.6%), against reopening the rotating-victim cascade class the fixture exists to kill. Not bought.

**But the measurement surfaced something else, with a mechanism of its own, and it is bigger than the thing #808 proposed to fix.**

Same run: `scripts/integration.sh` on `origin/main` `7270a595`, 610 tests all green, 32.1 min test phase, `workers: 1` so every millisecond is on the critical path. 549 reset samples (22 spec files use Playwright's base `test` and pay nothing).

| | ms |
|---|---|
| median | 595 |
| **mean** | **1124** |
| p95 | 2180 |
| **p99** | **12612** |
| **max** | **17635** |

**The mean/median gap alone is `(1124 − 595) × 549 = 290.7 s = 4.85 min`** — very nearly half of the 10.29 min the resets cost in total. Almost half the price is paid by the tail, not by the typical case.

| | resets | share of resets | time | share of all reset time |
|---|---|---|---|---|
| ≥ 1 s | 79 | 14.4% | 340.8 s | 55.2% |
| ≥ 2 s | 28 | 5.1% | 275.5 s | 44.6% |
| ≥ 5 s | 25 | 4.6% | 266.3 s | **43.1%** |
| ≥ 10 s | 15 | 2.7% | 185.8 s | 30.1% |

## The tail is two bursts, not 25 expensive specs

In run order, with the fast neighbours either side. Playwright runs files alphabetically, so consecutive indices are consecutive in *time*, and the affected spec files have nothing to do with one another.

```
 …
 328    601 ms
 329  12612  marker-target-window-regression | focused-window send+reply does NOT spawn unread marker
 330  13250  marker-target-window-regression | switching to tall window after focused send scrolls target to bottom
 331  12041  media-link-alias-modal        | upload link on a server-advertised host alias re-roots to page origin (#324)
 332  13410  media-link-modal-viewer       | 📸 upload link click opens the in-app viewer
[...]

Issue #808, opened 2026-08-04
e2e: the per-test session restart leaves the upstream budget at bahamut's ceiling


vjt:
## Narrowed before anyone acts on it

An earlier revision of this issue claimed the fake-lag bank was a plausible
substrate under the load-sensitive flake families (#552, #585, #769, #653).
**That claim is withdrawn — it was falsified, structurally, and the argument
is worth keeping because it kills the idea outright rather than merely
failing to support it.**

`resetSubject` restarts the `Session.Server`, which means a NEW upstream
connection, which means a NEW bank. The bank is loaded by the reconnect burst
and by nothing else, and it decays 1s per second. It is therefore **identical
in isolation and under full gate** — the reset is an `auto` teardown of every
test either way. A constant cannot be the discriminant of a signature defined
as "green in isolation, red under load", which is precisely the #653
signature. Fake lag does not produce it.

Measured scope, rather than assumed: of ~590 specs, **7** block on traffic
originated by grappa (`waitForLine` / `waitForPrivmsg` / `waitForTopic`).
Their budgets are 5000ms (`nick-follow-query`), 8000 (issue153, issue263),
15000 (issue536, issue386), 25000 (issue671), 30000 (issue152). Only the first
is tight enough for a few seconds of fake lag to matter. Of the 11 specs in
the flake set, none waits on upstream traffic. All eight #653 sub-issues have
a landed fix with an identified mechanism, and none of them is fake lag.

**So the honest statement is: one spec in the whole suite has an upstream
budget tight enough to feel the fake lag, and #613 ate the margin it had.**

The bank numbers below stand. The generalisation does not.

## The condition

`_vjtReset` (`cicchetto/e2e/fixtures/test.ts`) restarts vjt's `Session.Server`
in the teardown of every test. The reconnect burst leaves bahamut's fake-lag
bank nearly full, and the next spec inherits it.

Measured 2026-08-04 on a scoped `--repeat-each 3` of `nick-follow-query`, with
#807's per-send accounting on a live testnet. `bank_model_s` is bahamut's
published arithmetic (`since += 2 + len/120`, recvQ drain gated on
`since - now < 10`) evaluated locally — **modelled, not read off the ircd**.

The reconnect burst, five charged commands in ~140ms:

```
51.201  NICK          bank 2.00
51.201  USER          bank 4.00
51.206  JOIN          bank 6.00
51.206  MODE <nick>   bank 8.00      (#229, umode query at 001)
[...]

PR #946, opened 2026-08-06
test-support(#934): make the reset say where its time went

vjt:
Instrumentation for #934, then the measurement run. **The headline is a negative: with the instrument in place, the tail did not happen at all.**

## What was built

Two channels, because they answer different questions.

**1. Server-side phase timer.** `SubjectReset.reset!/2` returns `{:ok, phases}` — `drain` / `baseline` / `seed` / `respawn` / `total`, so total-minus-the-parts is the unattributed remainder and a slow reset always has somewhere to land. The controller forwards it on the 204 as **`x-grappa-reset-phases`** — a header, so the 204 stays a 204 and no fixture learns a new status code. It has to reach the **client**, not just the log: a retry means one test produces several server-side resets, so pairing client samples against server log lines **by ordinal cannot work** — and the retry is the thing being measured. `respawn_connected` additionally logs its four spans per credential (**stop / spawn / WELCOME / autojoin**) *including on the failure path* — the only trace an attempt that never returns to the caller leaves, and a 433 `nick_rejected` is exactly that attempt.

**2. The silent retry is silent no longer.** `nick_rejected` appeared **zero** times in a full green suite log and that zero proved nothing, because nothing ever wrote one. It now emits a line per retry, and the fixture carries the **attempt count** out beside its own measured elapsed. The question becomes a count instead of an inference.

The per-test `__RESETCOST__` line is unconditional on purpose: twice this was a throwaway local diff, and twice the evidence was gone before the question was answered.

**Boundary respected:** no timeout moved, no backoff touched, no product code. The retry's shape is left exactly as it was — this measures it, it does not fix it.

One thing the gate caught that is worth keeping: an **undeclared Logger metadata key is silently dropped by the formatter**. Without the `config.exs` allowlist entry the instrument would compile, run, cost time and print a line with no numbers on it.

## The measurement — 549 samples, same suite, same host

| | w2's run (`7270a595`) | this run (`8ed4f2e0`+instr) |
|---|---|---|
| samples | 549 | **549** |
| median | 595 ms | 538 ms |
| mean | 1124 ms | **582 ms** |
| p95 | 2180 ms | 972 ms |
| p99 | 12612 ms | **1328 ms** |
| max | 17635 ms | **1457 ms** |
| samples ≥ 3 s | 26 | **0** |
| retries (attempts > 1) | unknown | **0** |
| total reset budget | 617 s | **319 s** |
| test phase | 32.1 min | 25.4 min |

**There was no tail to attribute.** The specs that cost 12–17 s in w2's run (`marker-target-window-regression`, the `media-link-*` family, `ux-5-bc2-nick-render`) cost **415–588 ms** here. So, honestly, against the three questions asked:

- *where the 15 samples with no compatible retry depth spend their time* — **cannot say: they did not occur.**
- *whether the 12 proven-by-ceiling really retried* — **cannot say: zero retries occurred.** The instrument now counts them, so the next occurrence answers this without an argument.
- *the burst edges, the 14 sub-ceiling samples, the singleton* — **did not occur, so unexplained.**

That negative is itself the result: **the tail is not a property of the suite or of a spec's position in the run.** Two runs of the same 549 resets, one with a 26-sample tail and one with none.

## What the run DOES establish — where a normal reset's time goes

549 samples, server-side, and it is not where #808 was looking:

| phase | median | share of total |
|---|---|---|
| **`seed_ms`** (scrollback truncate + 200-row re-seed) | 370 ms | **72%** |
| `respawn_ms` (stop + spawn + WELCOME + autojoin) | 119 ms | 22% |
| `drain_ms` (DB/ETS surfaces) | 14 ms | 5% |
| `baseline_ms` (channel baseline rewrite) | 1 ms | 1% |

Client-observed total minus server total: **median 2 ms** — the client number *is* the server number, no HTTP mystery in between.

This sharpens #808's decline rather than reopening it: the restart it declined to make conditional is **22%** of a reset measured across a whole suite, and the dominant cost is the **200-row re-seed** nobody proposed touching. Not proposing anything here — just recording where the lever actually is.

## 🔴 Unrelated, but `main` is red

The run had one failure: `issue772-compose-draft-reload.spec.ts` › *"a SENT message does not come back on reload"* — the sent text is still in the textarea after the reload.

Attribution measured, not argued: **3/3 red in isolation on this branch, and 3/3 red in isolation on clean `origin/main` `8ed4f2e0`** with the instrumentation absent from the tree. Deterministic and **pre-existing** — not this change, and not a flake. Reported here because it means main is currently red on e2e; it is not touched by this PR.

## What I refuse to claim

- **I did not reproduce the phenomenon**, so nothing here explains it. Everything above about the tail is a negative.
- **The cause of the difference between the two runs is NOT established.** Host contention is a candidate and only a candidate: `/tmp` artifacts show other workers' gates finishing at 14:09:23 and 14:22:37 while w2's bursts sit at ≈14:12–14:16 and ≈14:19–14:21 — they **bracket** the bursts without pinning a job inside either. An mtime is a completion time, not an occupancy record. No CPU, IO or load measurement was taken during either run.
- **bahamut was not measured.** The per-IP throttle remains a candidate; zero retries this run is consistent with it and proves nothing about it.
- The two runs are on **different commits** (`7270a595` vs `8ed4f2e0`+instr). Same 549 resets and same spec set, but I did not verify that nothing between those commits touches reset cost.
- The instrument itself is **unvalidated against a slow reset** — every sample it saw was fast. Its behaviour under a retry is exercised by unit tests, not by a real 433.

## Gates

`scripts/check.sh` **rc=0**, every stage read from the log: credo `found no issues` · sobelow `No vulnerabilities found` · ExUnit `8 doctests, 54 properties, 5305 tests, 0 failures` · dialyzer `Total errors: 0 … passed successfully` · `Doctor validation has passed!` ×2 · `wireTypes.ts is in sync` · bats **331 ok / 0 not-ok**. Full e2e suite run (615 passed, the one pre-existing red above). `tsc -p e2e --noEmit`: 23 errors, **all pre-existing in specs this PR does not touch** (the known "e2e has no static gate" hole) — **zero** in the two fixture files changed here.

### DIFF
--- a/cicchetto/e2e/fixtures/grappaApi.ts
+++ b/cicchetto/e2e/fixtures/grappaApi.ts
@@ -719,12 +719,22 @@ export interface BaselineSeedChannel {
   seedSender?: string;
 }
 
+/**
+ * What one reset actually cost, for #934. `attempts` is how many times
+ * the 433 retry below fired (1 = no retry); `phases` is the server's own
+ * `k=v;k=v` breakdown, forwarded on the 204 as `x-grappa-reset-phases`.
+ */
+export type ResetOutcome = {
+  attempts: number;
+  phases: string;
+};
+
 export async function resetSubject(
   adminToken: string,
   userName: string,
   baselineAutojoin?: Record<string, string[]>,
   baselineSeed?: Record<string, BaselineSeedChannel[]>,
-): Promise<void> {
+): Promise<ResetOutcome> {
   const body: Record<string, unknown> = { user_name: userName };
   if (baselineAutojoin) body.baseline_autojoin = baselineAutojoin;
   if (baselineSeed) {
@@ -762,18 +772,30 @@ export async function resetSubject(
       },
       body: JSON.stringify(body),
     });
-    if (res.status === 204) return;
+    if (res.status === 204) {
+      return {
+        attempts: attempt,
+        phases: res.headers.get("x-grappa-reset-phases") ?? "",
+      };
+    }
 
     const text = await res.text().catch(() => "<no body>");
     const nickStillTaken = res.status === 500 && text.includes("nick_rejected");
     if (!nickStillTaken || attempt === maxAttempts) {
       throw new Error(`resetSubject(${userName}) failed: ${res.status} ${text}`);
     }
+    // #934 — this retry used to be SILENT, and that silence cost a whole
+    // investigation: `nick_rejected` appeared zero times in a full suite
+    // log and the zero proved nothing, because nothing ever wrote one.
+    // A loop that can multiply a 500 ms call by eight has to say so.
+    process.stderr.write(`__RESETRETRY__\t${attempt}\t${userName}\tnick_rejected\n`);
     // Observed 433 → the ghosted nick isn't released yet. Back off (capped)
     // then re-register; each reset attempt also reconnects, which itself
     // gives bahamut time to reap the prior connection's ghost.
     await sleep(Math.min(500 * attempt, 2_000));
   }
+  // Unreachable: the loop either returns on 204 or throws at maxAttempts.
+  throw new Error(`resetSubject(${userName}): retry loop fell through`);
 }
 
 // Overwrite a user's `credential.autojoin_channels` via the admin
--- a/cicchetto/e2e/fixtures/test.ts
+++ b/cicchetto/e2e/fixtures/test.ts
@@ -117,11 +117,19 @@ export const test = base.extend<{
     },
     { auto: true },
   ],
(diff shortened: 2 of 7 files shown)


vjt:

Il cherry-pick ha riscritto le SHA, quindi GitHub non chiude questa PR da solo. Il contenuto è su main come: `28e24239` + `4cd9ec67`

Verificato per CONTENUTO, non per SHA: `git diff origin/main...<branch>`, ogni riga aggiunta cercata su `origin/main` — 169 righe controllate, **zero mancanti**.

Issue #356, opened 2026-07-20
cicchetto: /notify + /watch|/highlight feedback is computed then discarded — rewire (hybrid inline/settings) + rename to classic IRC names


vjt:
## Problem

`/notify` (presence-watch, landed via #247) and `/watch` | `/highlight` (keyword watchlist) both execute correctly server-side, but their **success / list feedback is computed and then thrown away** — CP13 removed the surface that used to render it and never rewired a replacement. Errors still show (red box); success and list output silently vanish, so the commands *feel* dead even though they work.

### Evidence (verified on `origin/main` @ `15988d79`)

- `cicchetto/src/lib/compose.ts:96-99` — the result contract already documents the missing branch:
  - `ok: true` = silent success
  - `ok: string` = **"success with inline feedback (e.g. watchlist list output)"**
  - `error: string` = failure
- `cicchetto/src/lib/compose.ts:898` — watchlist assembles `ok: \`watchlist (N): …\``
- `cicchetto/src/lib/compose.ts:933` — notify assembles `ok: \`notify (N): …\`` (both do a server round-trip to build the real list)
- `cicchetto/src/lib/compose.ts:1005` — comment: *"feedback for client-side commands is intentionally not surfaced"* — the caller consumes only `result.error`, never the `ok: string` branch.
- `cicchetto/src/themes/default.css:2363` — only `.compose-box-error` exists (red, `role=alert`); there is no success/notice variant.

The data is fully built; there is simply no consumer. No `numericInline` needs resurrecting.

### 1. Feedback — hybrid by argument presence
- **With arg** (`/notify gigi`, `/hilight foo`): execute **and** show a green inline confirmation in the compose-box notice area (the "seam verde"). It **auto-dismisses** (~3 s timeout / fade) — the user must not have to type a command or send a message to clear it. Errors stay **sticky** as today (`role=alert`, you need to read them).
- **Bare** (`/notify`, `/hilight`): open the relevant settings section instead of printing anything inline.

### 2. Settings panel — one new section holding both watch and notify
A **single new settings section** contains **both** lists (watch/keyword **and** notify/presence). Each list: the full entries, an **×** button to remove each, and an **input** field to add one.
- **Presence / notify (per-network):** this is the list currently shown on the **home page as "watched"** — **move it into this section and remove it from home**. It must be the **same source of truth** (per-network, via `notifyWatch` / `watchByNetwork`, imported at `compose.ts:25`), **not a copy** that can drift.
[...]

PR #953, opened 2026-08-06
cic(#816): make the refusal speak and the second door a choice

vjt:
Completes the two conditions vjt set on #816 in the acceptance comment
(2026-08-06 10:12Z). Both were unshipped: the audit that marked the issue
`soon` had measured their absence, not their presence.

## 1. Shift+Enter speaks

It still inserts nothing and sends nothing — a newline cannot travel inside a
PRIVMSG, so honouring the key means splitting into N messages. What changes is
that the refusal is no longer silent. A key that does nothing is
indistinguishable from a broken key, and every other chat app the operator uses
honours that combination.

It now shows **"IRC does not support multi-line messages"** — the ruling's
wording verbatim — on the existing #356 feedback seam, as a NOTICE
(`role=status`, self-clearing) rather than an error: nothing failed.

The old vitest **pinned the silence**. It was rewritten, not supplemented — a
second assertion beside the old one would have left two contradictory truths in
the file.

## 2. The .txt door is a choice, not a punishment

It shipped as the affirmative on the **over-cap** dialog only, i.e. reachable
exclusively after the operator had already been refused. The under-cap confirm
now carries it as a third door beside Cancel and Paste.

The door lives in the confirm **store**, not the paste guard: `ConfirmRequest`
gains `alternative: {label, onSelect} | null` and a `chooseAlternative()` verb
exclusive with `acceptConfirm()`. **Required**, not optional — all five existing
call sites write `alternative: null` explicitly, so a two-button dialog is a
stated fact rather than an inferred absence. No new category, no new server
surface: `text/plain` is already an accepted `document` MIME and the existing
`dropUpload` does the work, so the wire still carries text + a clickable link
(the 📸 model). Both arms now share one label, `"Upload as .txt"` — the over-cap
affirmative said `"Upload as file"`, and two spellings of one action read as two
actions.

## The hard cap: measured, not removed

`PASTE_HARD_MESSAGE_LIMIT = 5` is untouched, deliberately — deleting a shipped
guard is not a drive-by. **The measurement:** that 5 is numerically the server's
own burst allowance (`config :grappa, :send_throttle, capacity: 5,
refill_per_sec: 0.5`, read from bahamut's flood ladder). So the ceiling is
exactly "how many messages leave before the composer drops into a paced drip" —
past it `sendBodyLines` (#666) paces on the server's 429 retry-after, ~1 line
per 2s, never dropping. A 40-line paste was never a disconnect; it was ~80s of
drip.

That makes the guard defensible and its own comment wrong: `pasteFlood.ts` calls
the number "deliberately not derived from anything the server tells us" while
silently duplicating a server constant that no wire field publishes. Retune
`:send_throttle` and the client ceiling diverges in silence.

**Recommendation, for vjt to rule on:** keep the cap. Either publish `capacity`
on `GET /api/config` (additive, no protocol bump) and derive the ceiling, or
keep the literal and document the coupling. Not done here.

## Tests

vitest red first on both behaviours, then green:

- `ComposeBox.test.tsx` — Shift+Enter says why (plus a plain-Enter control that
  the notice does NOT fire on an ordinary send); the under-cap alternative
  exists, uploads `text/plain` with the paste's own bytes, never touches the
  draft, and does not displace the Paste affirmative.
- `confirmDialog.test.ts` — `chooseAlternative` fires only its own closure, is
  exclusive with `acceptConfirm`, and is a safe no-op with no alternative.
- `ConfirmModal.test.tsx` — the button is absent on a two-button request,
  renders its label and closes on click.

e2e (written here, run by CI — the STACK lane is held elsewhere):

- `issue816-shift-enter-speaks.spec.ts` — the line RENDERS in a real browser,
  the draft keeps no line break, nothing reaches the channel; positive control
  that a plain Enter does send the same draft.
- `issue80-paste-flood-guard.spec.ts` — the under-cap dialog shows three doors,
  Cancel fires neither, the alternative drives the whole upload chain to a 📄
  link with no burst, and the Paste door still pastes. The over-cap arm asserts
  there is NO third button up there.

Local gates on a clean tree: `WRITABLE_CIC=1 bun run check` rc=0, `bun run test`
rc=0 (243 files, 4484 tests).

## What I refuse to assert

- **That the e2e specs pass.** They were written, never executed — the STACK
  lane belongs to another worker. CI is their first run.
- **That `tsc -p e2e` is clean.** It cannot resolve `@playwright/test` in this
  worktree (e2e deps not installed); the failure is environmental, and the
  specs' types are unverified locally.
- **That three buttons lay out well at every width.** `.confirm-modal-actions`
  got `flex-wrap`, but no browser rendered it here — darwin has no local
  browser measurement. Geometry is CI's word, not mine.
- **That the cap should go.** The ruling called it moot; the measurement says it
  still removes the Paste door above the server's burst. That is vjt's call.

### DIFF
--- a/cicchetto/e2e/fixtures/cicchettoPage.ts
+++ b/cicchetto/e2e/fixtures/cicchettoPage.ts
@@ -315,6 +315,14 @@ export async function confirmModalCancel(page: Page): Promise<void> {
   await page.locator('[data-testid="confirm-modal-cancel"]').click();
 }
 
+// #816 — the optional THIRD button: a different route to what the operator
+// wanted, offered beside Cancel and the affirmative (the paste guard's
+// "Upload as .txt"). A Locator, not a click helper, so a spec can also assert
+// its ABSENCE on a plain two-button dialog.
+export function confirmModalAlternative(page: Page): Locator {
+  return page.locator('[data-testid="confirm-modal-alternative"]');
+}
+
 // Click the window to focus it. Solid's reactive flush + the shell's
 // auto-close-sidebar effect happen synchronously; the channel becomes
 // selected before this resolves.
--- a/cicchetto/e2e/tests/issue80-paste-flood-guard.spec.ts
+++ b/cicchetto/e2e/tests/issue80-paste-flood-guard.spec.ts
@@ -26,6 +26,7 @@ import { expect, test } from "../fixtures/test";
 import {
   composeTextarea,
   confirmModal,
+  confirmModalAlternative,
   confirmModalBody,
   confirmModalCancel,
   confirmModalYes,
@@ -163,6 +164,9 @@ test("#816 — a paste over the hard cap uploads as a file instead of flooding",
   // Both numbers, so the operator knows what would fit.
   await expect(confirmModalBody(page)).toContainText("6");
   await expect(confirmModalBody(page)).toContainText("5");
+  // No third door up here: the paste door is exactly what the cap closed, so
+  // uploading IS the affirmative. Two buttons, not three.
+  await expect(confirmModalAlternative(page)).toHaveCount(0);
 
   // Cancel first: the safe default drops the paste with no upload and no
   // text in the box. Asserted before the affirmative so a handler that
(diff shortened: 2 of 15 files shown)


vjt:

Il cherry-pick ha riscritto le SHA, quindi GitHub non chiude questa PR da solo. Il contenuto è su main come: `91af62ef` + `f0451e30`

Verificato per CONTENUTO, non per SHA: `git diff origin/main...<branch>`, ogni riga aggiunta cercata su `origin/main` — 450 righe controllate, **zero mancanti**.

Issue #947, opened 2026-08-06
Unread divider label saturates at 200 (page size leaks into the count)


vjt:
## Symptom

Reported on IRC by a tester: the in-pane unread divider "── N unread messages ──" often reads exactly **200**, regardless of how many unread messages the window actually holds. The sidebar badge for the same window shows the real (larger) number, so the two disagree.

## Cause (traced in code, not reproduced locally)

The divider count is recomputed **from the rows currently loaded in the pane**, not from the authoritative server seed:

- `cicchetto/src/ScrollbackPane.tsx:1376` — `unreadCount = msgs.filter(m => m.id > cursor && m.id <= sessionTop && …).length`

For a window far behind, the anchored fetch (#156) loads only `[cursor .. cursor+200]`, and 200 is the server ceiling:

- `lib/grappa_web/controllers/messages_controller.ex:66` — `@max_http_limit 200`
- `cicchetto/src/lib/scrollback.ts:674` — comment confirms the client caps at the server max
- `cicchetto/src/ScrollbackPane.tsx:3218` — "a channel with > 200 unread loads only `[cursor .. cursor+200]`"

So for any window with ≥ 200 unread the filter saturates at exactly the page size. It is not a random number — it is the fetch limit leaking into a user-visible count.

The authoritative count already exists client-side and is unbounded server-side:

- `cicchetto/src/lib/selection.ts:480` — `messagesUnread()` memo, seeded by `/me`
- `lib/grappa/read_cursor.ex:303` — `bulk_unread_split/3`, a plain SQL `count(m.id)`, no cap

The sibling case is already handled: the #693 "far behind" branch suppresses the divider precisely because the loaded-row count would be "a confident wrong number", and hands the true count to the jump-back affordance (`ScrollbackPane.tsx:1390-1396`). The same reasoning applies here, one notch earlier — the divider is shown, with a saturated label.

## Suggested fix

Keep the loaded-row math for **placement** of the divider row (it must sit between the last read and first unread loaded message), but take the **label** from `messagesUnread()[key]`. If the two disagree because the pane is partially loaded, the server seed is the one to trust. Rendering `200+` would also be acceptable and is strictly better than a flat `200`.

## Severity

Cosmetic. As the reporter put it, what matters is that the divider line is *there*; the number is a nicety. Filed for correctness, not urgency.

vjt:
**The issue located the defect one notch earlier than it lives, and its suggested fix was a no-op.**
[...]

PR #955, opened 2026-08-06
cic(#947): let the divider count the conversation, not the page

vjt:
Refs #947.

## The defect

The in-pane `── N unread messages ──` divider reads exactly **200** for any window resumed by a #693 jump-back. 200 is the server's `@max_http_limit` and the client's `PAGE_LIMIT` — the divider was recounting the rows the pane happened to hold, so the fetch cap surfaced as a fact about the conversation.

#693 solved this one notch earlier and stopped there. While a pane is far behind the divider is *suppressed*, precisely because a loaded-row count would be "a confident wrong number in the one place the operator reads to decide where they left off", and the server-measured count goes to the pinned `N unread — jump back` bar instead.

Taking the jump was the case nobody followed through on. `jumpToUnread` refetches `after(resumeFrom, PAGE_LIMIT)` — one page out of a gap that is >200 *by the definition of far-behind* — replaces the pane, and clears the flag. Clearing the flag un-suppresses the divider, and `far.missed` is dropped on the floor at the exact moment the label needs it. You tap a bar reading "239 unread" and land on "── 200 unread messages ──".

## The issue's suggested fix is a no-op — measured, not argued

#947 proposes taking the label from `selection.ts`'s `messagesUnread()`, described as the authoritative unbounded seed. **It is not, once the pane hydrates.** `perChannelUnread`'s local-rows branch *overrides* the seed for any key holding rows.

Measured: with a **5000** seed and **200** loaded rows past the cursor, `messagesUnread()` returns **200**. Badge and divider saturate identically — reading one off the other would have replaced 200 with 200.

The issue's paired symptom ("the badge shows the real larger number, so the two disagree") only holds while far-behind is SET, and there the divider is *suppressed*, so there is no second number to disagree with. Pinned in `selection.test.ts` so the next reader does not re-derive it.

## What shipped

- `scrollback.ts` — `measuredUnreadByChannel`: `{at, count}` per key, written by `jumpToUnread` **only** when the after-page came back at the cap. A short page drained the region, so the rows ARE the count and a carried number could only go stale against them.
- `count` is deliberately the **same** `far.missed` the bar advertised, not a second measurement. It counts rows the divider's predicate excludes (own-presence, operator echoes) so it can run slightly high — but a third figure for one question is worse than a consistent one.
- **`at` is the whole design.** It stamps the cursor the count was measured after; the pane spends the record only while its frozen divider is still anchored there. Self-invalidating, not a cache somebody has to remember to sweep. It still joins the identity-transition reset list and the #373 `renameScrollbackKey` migration set (a count is about a conversation, not a spelling of the peer's nick).
- **Placement stays loaded-row math.** The divider must sit between the last read row and the first unread row actually in the pane; no server count knows where that is. Only the label moved.

## Deliberately out of scope

A cold open whose gap probe **failed** also shows a saturated divider — `loadInitialScrollback` then keeps the anchored resume and loads one page of an unknown gap. There is no honest number to show there: the failure of the probe is exactly the absence of the measurement. Left alone rather than guessed at.

## Test plan

- [x] RED observed before the fix: `Expected "3000 unread messages" / Received "2 unread messages"`.
- [x] **Mutation-measured**: neutering the `afterPage.length === PAGE_LIMIT` arm turns the carrier test red, so the test binds to the change rather than passing alongside it.
- [x] `scripts/bun.sh install` — rc 0
- [x] `WRITABLE_CIC=1 scripts/bun.sh run check` — rc 0; `tsc --noEmit` also run standalone (rc 0), since `check` short-circuits it on a biome failure
- [x] `scripts/bun.sh run test` — rc 0, **243 files / 4481 tests passed**
- [x] e2e spec `issue947-unread-divider-count.spec.ts` contributes **0** of the 26 pre-existing `tsc -p e2e` errors
- [ ] e2e **written, not run** — the STACK lane belongs to another worker; CI runs it. Asserts the user-visible contract: the number you tap is the number you land on, and the divider has not moved.

### DIFF
--- a/cicchetto/e2e/tests/issue947-unread-divider-count.spec.ts
+++ b/cicchetto/e2e/tests/issue947-unread-divider-count.spec.ts
@@ -1,0 +1,167 @@
+// #947 — the in-pane unread divider labels itself with the page size.
+//
+// This is the notch AFTER #693. That change suppressed the divider for a pane
+// sitting far behind, because a count taken from the loaded rows would have
+// described the tail page rather than the abandoned region — "a confident
+// wrong number in the one place the operator reads to decide where they left
+// off" — and handed the true, server-measured count to the pinned
+// "N unread — jump back" bar instead (see `issue161-forward-paging.spec.ts`,
+// which pins that state).
+//
+// Taking the jump was the case nobody followed through on. `jumpToUnread`
+// refetches `after(resumeFrom, PAGE_LIMIT)` — ONE page out of a gap that is
+// > 200 by the very definition of far-behind — REPLACES the pane with it, and
+// clears the far-behind flag. Clearing the flag un-suppresses the divider,
+// whose count is recomputed from the rows now loaded: exactly 200, every
+// time, for every window in this state. The operator taps a bar reading
+// "239 unread" and lands two hundred milliseconds later on
+// "── 200 unread messages ──". The fetch cap, surfacing as a fact about the
+// conversation.
+//
+// The fix keeps loaded-row math for the divider's PLACEMENT (it must sit
+// between the last read row and the first unread row actually in the pane —
+// no server count knows where that is) and takes the LABEL from the
+// measurement the jump already had in hand: the same `missed` number the bar
+// advertised, stamped with the cursor it was measured after so it expires by
+// itself once the frozen divider re-latches somewhere else.
+//
+// The discriminating assertion is therefore NOT "the divider is present" but
+// "the number you tapped is the number you land on". Against the unfixed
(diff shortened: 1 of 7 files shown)


vjt:

Il cherry-pick ha riscritto le SHA, quindi GitHub non chiude questa PR da solo. Il contenuto è su main come: `a0bf01d0`

Verificato per CONTENUTO, non per SHA: `git diff origin/main...<branch>`, ogni riga aggiunta cercata su `origin/main` — 370 righe controllate, **zero mancanti**.

Issue #963, opened 2026-08-06
theme editor: font <select> renders black text — invisible on dark themes


vjt:
## Symptom

In the theme editor (Settings → Themes → edit), the **font** `<select>` renders its
label text in black while closed. On a dark theme that is black-on-dark: unreadable.
Opening the picker is fine — the dropdown list is drawn by the OS, not by us.

Reported by Hypnotize on a dark theme.

## Cause

`<select>` never receives a `color` anywhere in the stylesheet, so it falls back to
the UA default (`color: fieldtext`, i.e. black) — and `fieldtext` does **not** inherit
from the page.

The global form-control base rule in `cicchetto/src/themes/default.css` (the block
ending at the `textarea {` selector, ~line 376–420) is deliberately scoped to
`input:not(<denylist>)` and `textarea`:

```css
input:not(:is([type="checkbox"], [type="radio"], …)),
textarea {
  background: var(--bg);
  color: var(--fg);
  …
}
```

`select` is not in that selector list. So every bare `<select>` in the app inherits
nothing and paints black.

The theme-editor rule only sets a background, which is exactly what turns "black text"
into "invisible text":

```css
/* default.css:6551 */
.theme-editor-font-select,
.theme-editor-bg-url-input {
  background: var(--bg-alt);
  padding: 0 0.5rem;
}
```

Verified against the bundle actually served by staging
(`https://irc.sindro.me/assets/index-CiM1iK6g.css`): the rule above is present verbatim,
with no `color`, and no other rule in the bundle sets a color on a `select`
(the only `color:` hit on a `select`-ish selector is `.theme-card-select`, which is a
`<button>`, not a control).

Sibling inputs are unaffected: `.theme-editor-name-input` and `.theme-editor-bg-url-input`
are `<input>` elements, so the global rule already gives them `color: var(--fg)`.
That is why only the select is reported.

## Not just the theme editor

Every other bare `<select>` has the same defect class — admin tabs
(`AdminSettingsTab.tsx:187`, `AdminCredentialsTab.tsx:274/289/315/550`,
`AdminVhostsTab.tsx:286`), `BanlistModal.tsx:212`, and the settings drawer
(`SettingsDrawer.tsx:1089` identity network, `:1182` upload TTL, `:1471`).

There *is* one rule that tries to cover the drawer selects:

```css
/* default.css:4454 */
.settings-drawer label.prefs-list input[type="text"],
.settings-drawer label.prefs-list select {
  color: var(--text);
[...]

PR #967, opened 2026-08-07
cic(#964): device row shows last-used + a 'this device' marker

### LINKED ISSUE #964: Settings: let push devices carry a name (and mark the current one)
## Request

From Hypnotize on IRC: the device list in Settings shows him two rows both reading
"Firefox on Linux" and he cannot tell which is which. Ask: let a device carry a
**name** — defaulting to the host, editable by the user. vjt agreed on hostname as
the default.

## Today

`SettingsDrawer.tsx:1518-1552` renders the list from `PushDeviceSummary`
(`cicchetto/src/lib/push.ts:166`, fed by `GET /push/subscriptions`). Each row is
exactly:

```
{deviceClassIcon(parsed.deviceClass)} {parsed.browser} on {parsed.os}   [remove]
```

`parseUserAgent` (`cicchetto/src/lib/userAgent.ts`) deliberately collapses the UA to
`Browser on OS`, so **two instances of the same browser on the same OS are byte-identical
rows**. The only escape hatch is the full UA in the `title=` attribute — hover-only, so
it does not exist on touch, which is where the drawer mostly lives.

Worth noting the payload already carries two disambiguating fields the row throws away:
`created_at` and `last_used_at` (`push.ts:169-170`).

## Proposal

1. **Optional user-set label** on the push subscription, shown in place of
   `Browser on OS` when present (keep the parsed string as the secondary line, it is
   still useful). Rename inline from the row; `PATCH /push/subscriptions/:id`.
2. **Default the label to a disambiguated `Browser on OS`** — append an ordinal when a
   subscription with the same parsed string already exists for that user, so the second
   Firefox on Linux registers as `Firefox on Linux #2` instead of a byte-identical twin.
3. **Cheap interim win, independent of the above:** render `last_used_at` in the row.
   Two same-browser devices are almost never last used at the same minute, so this alone
   removes most of the ambiguity and needs no schema change.

## Why the default is not the machine hostname

The original ask was "default it to the host". That is not implementable from a browser:
there is no web API exposing the client machine's hostname, and the omission is
deliberate — a hostname is a network fingerprint. The one historical leak, the `.local`
name in ICE candidates, was closed in 2019 by replacing it with a per-origin random UUID
(mDNS candidate obfuscation).

Deriving it server-side does not rescue the idea either. The only server-visible
name-like thing is the PTR of the peer address — machinery that already exists here
(`AdminSessionsTab.tsx:380-401` renders `live.peer_name`, reverse-DNS first, raw
`address:port` as authoritative fallback, `null` on a cold cache). On a residential or
NAT'd line the PTR is the ISP pool name, identical for every device behind it, which
reproduces exactly the collision this issue is about.

So the user-set label is the fix; the ordinal-suffixed default is only there so the list
is never ambiguous before anyone renames anything.

## Follow-up asked in the same conversation

Hypnotize also wants a badge / green dot marking **which row is the device you are
looking at right now** ("un badgettino o pallino verde che indichi la current"); vjt
said ok. That is independent of the label: cic knows its own subscription id from the
`PushSubscription` endpoint it registered, so the row whose `id` matches is the current
one — no server change needed for the marker itself.

vjt:
Slice A of #964 — the two pieces the issue itself calls independent of the
label work: **point 3** (render `last_used_at` in the row) and the
**follow-up** (mark the device you are looking at right now). Pure cic, hot,
no Elixir and no schema change.

## What changes

`parseUserAgent` collapses the UA to `Browser on OS`, so two instances of the
same browser on the same OS render byte-identical rows — Hypnotize's report.
The only escape hatch was the full UA in `title=`: hover-only, so it does not
exist on touch, which is where the drawer lives. Both fixes come from data the
row was already handed and threw away.

- **Activity line** — `formatDeviceActivity(device, nowMs)` renders
  `last used 4h 12m ago`, falling back to `added 3m ago` while `last_used_at`
  is still nil (it is nil until the Sender has pushed to the row once). It
  REUSES `duration.ts:formatDurationSince` — the WHOIS-idle / server-rail
  formatter — rather than growing a second relative-time shape, and returns
  null when neither instant parses, so the line is omitted instead of guessing.
- **`● this device` marker** — the row id is PROVEN by endpoint match
  (`subscriptionIdForEndpoint`, the recall seam `disablePush` already used to
  never delete a row it cannot prove ours). Deducing it from the UA would badge
  whichever twin came first, which is the bug. No stashed match ⇒ no row is
  marked.

Timestamps are stamped at render, not on a ticking clock: the list is only
refetched on drawer mount, so a live `3m → 4m` would be fresher than its data.

## Explicitly NOT in scope

**Points 1 and 2 stay out** — the renameable label and the ordinal-suffixed
default. They need a migration, a `name` column and `PATCH
/push/subscriptions/:id`, and the design fork on the ordinal (STORED vs
DERIVED) is still unresolved. Shipping half of that shape now would bake the
wrong answer into a migration. Nothing in this slice depends on them.

Deliberately no `Closes #964`: the issue is only half done.

## Tests

- vitest: 8 new cases over `formatDeviceActivity` (branch preference, skew
  clamp, unparseable → null) and `subscriptionIdForEndpoint` (endpoint match /
  mismatch / nothing stashed). Both were **mutation-measured**: preferring
  `created_at` and dropping the endpoint check turn exactly 4 of them red.
- e2e `push-964-device-row.spec.ts`: enables push, asserts the visible
  `added …` line + the marker (and that exactly ONE row claims it), then drives
  a REAL delivery (peer DM → Triggers → Sender → push-catcher) — the only thing
  that stamps `last_used_at` — and asserts the row switched to `last used …`.

  Two deliberate choices in that spec:
  - a REST barrier (`awaitDeviceLastUsed`) between the catcher delivery and the
    UI read: Sender bumps the row AFTER the vendor 200, so the catcher alone is
    not a barrier;
  - the last leg reads from a SECOND browser context, not a reload. On boot
    `installPushResubscribe` renews a "dropped" subscription, and the stub's
    `subscribed` flag is per-document — a reload would POST `supersedes` and
    replace the very row under assertion. The clean context has no opt-in
    intent, so it renews nothing, and it proves the other half of the contract:
    the marker is per-BROWSER, not per-user.

  No `@webkit` twin: this is data rendering, engine-independent. The twin
  pairing belongs to UA-dependent defects like #963's `fieldtext`.

## Gates run locally

| gate | rc |
|---|---|
| `bun.sh run check` (biome + tsc) | 0 |
| `bun.sh run test` (243 files / 4518 tests) | 0 |
| `tsc -p e2e --noEmit` | 26 errors — the pre-existing baseline, zero mine |

Integration/e2e deliberately NOT run locally — CI on this PR owns them.

## Residue

- The DESIGN_NOTES entry for this slice is written but held out of the branch
  to avoid colliding with sibling worktrees; it lands with the merge.

### DIFF
--- a/cicchetto/e2e/fixtures/push.ts
+++ b/cicchetto/e2e/fixtures/push.ts
@@ -255,6 +255,32 @@ export async function resetPushSubscriptions(token: string): Promise<void> {
   }
 }
 
+/**
+ * #964 — blocks until the server has stamped `last_used_at` on at least one
+ * of `token`'s push subscriptions.
+ *
+ * Push.Sender bumps the row AFTER the vendor's 200, so push-catcher recording
+ * a delivery does NOT mean the DB write has landed: a UI refetch fired off the
+ * catcher alone races it. This polls the same REST view the settings drawer
+ * reads, so when it returns, a reload is GUARANTEED to render the stamped
+ * value — a real barrier, not a sleep.
+ */
+export async function awaitDeviceLastUsed(token: string, timeoutMs = 5_000): Promise<void> {
+  const base = "http://grappa-test:4000";
+  const deadline = Date.now() + timeoutMs;
+  while (Date.now() < deadline) {
+    const res = await fetch(`${base}/push/subscriptions`, {
+      headers: { authorization: `Bearer ${token}` },
+    });
+    if (res.ok) {
+      const body = (await res.json()) as { subscriptions?: { last_used_at: string | null }[] };
+      if ((body.subscriptions ?? []).some((s) => s.last_used_at !== null)) return;
+    }
+    await sleep(100);
+  }
+  throw new Error(`awaitDeviceLastUsed: no subscription stamped within ${timeoutMs}ms`);
+}
+
 export type CaughtDelivery = {
   headers: Record<string, string>;
   body_b64: string;
--- a/cicchetto/e2e/tests/push-964-device-row.spec.ts
+++ b/cicchetto/e2e/tests/push-964-device-row.spec.ts
@@ -1,0 +1,104 @@
+// push-964 — the device row stops being byte-identical to its twin.
+//
+// `parseUserAgent` collapses the UA to `Browser on OS`, so two instances of
+// the same browser on the same OS render the SAME row; the only escape hatch
+// was the full UA in `title=`, which is hover-only and therefore absent on
+// touch — where the drawer mostly lives. #964 point 3 + its follow-up put two
+// always-visible disambiguators on the row, both from data already on the
+// wire:
+//
+//   * the activity instant — `last_used_at`, falling back to `created_at`
+//     while nothing has been pushed to the device yet;
+//   * a "this device" marker on the row THIS browser registered (matched by
+//     endpoint→id, the same proof `disablePush` uses to never delete a row it
+//     cannot prove ours).
+//
+// Both assertions are on rendered TEXT, so this spec is engine-independent —
+// no @webkit twin (that pairing belongs to UA-dependent defects like #963's
+// `fieldtext`, not to data rendering).
+//
+// Why an e2e and not only a unit test: the "last used" branch needs the
+// server to have stamped `last_used_at`, and the ONLY thing that stamps it is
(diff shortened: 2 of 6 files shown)


vjt:

PR #968, opened 2026-08-07
web: gate /admin on the transport peer, not the resolved client IP

vjt:
## What

`Plugs.LoopbackOnly` gated `POST /admin/reload` + `POST /admin/cic-bundle-changed`
by reading `conn.remote_ip`. By the time the gate runs, that field is
`RemoteIpFromProxy`'s best guess at **who the client is** — an attribution, not a
capability. It can be loopback because the operator curled us from inside the box,
or because a forwarded chain resolved to nothing and the transport peer happened to
be a same-host reverse proxy. Nothing downstream can tell those apart from the value
alone, so an authorization decision was riding on a resolution heuristic.

The gate now reads `RemoteIpFromProxy.direct_loopback_peer?/1`: **transport peer was
loopback AND the request carried no forwarded header**. That is row 1 of the existing
trust matrix expressed as a predicate, stamped as an assign at the one point in the
pipeline that still sees the peer. It is a property of the transport, so no header
produces it, and it **fails closed** when the plug has not run.

## Operator cost: zero, measured

Every shipped caller of these two routes hits `127.0.0.1:4000` **directly**:
`scripts/deploy.sh`, `infra/docker/deploy.sh`, `infra/freebsd/deploy.sh`,
`infra/freebsd/jail_deploy_cic.sh`, `scripts/deploy-cic.sh`. Both shipped nginx
snippets set `X-Real-IP` + `X-Forwarded-For` unconditionally
(`infra/snippets/locations-api.conf`). Clean separation: through a proxy → never
passes; from inside the box → always passes.

## Still open — this closes the GATE and nothing else

Named explicitly so this does not read as done:

- **Attribution collapse.** The resolved value can still fall back to the transport
  peer for callers the forwarded chain cannot identify, so those callers share one
  identity across throttle keys, the `sessions.ip` audit column, and the #543 mode-2
  source derivation. Different fix, different risk profile — deliberately not here.
  `trusted_client_ip/3`'s moduledoc says so at the call site.
- **Chain-entry selection.** Which entry of a forwarded chain wins is still
  influenced by the chain's own contents. Untouched here; changing it needs its own
  threat test written first.

## One flipped test, and why it mattered

`admin_controller_test.exs` had a test pinning 200 for a loopback peer carrying a
forwarded header, with a comment claiming the wrapper bypassed the rewrite for
loopback peers. That is not the mechanism: a present header takes the header row,
the wrapper does not bypass, and the 200 came out of the resolver's
fall-back-to-peer. It was a **green pin on the defect**, explained by a mechanism
that was not in play. Now 403, with a comment describing what actually happens.

## Dialyzer note

Extracting the predicate produced one `missing_range: nil` on `trusted_client_ip/3`:
with the `if` condition behind an opaque call, Dialyzer stops eliminating the `nil`
in `RemoteIp.from/2`'s spec through `x || peer_ip`. Isolated by displacement (inline
green / extracted red; `||` red / explicit `case` green) rather than explained — I am
not claiming to know Dialyzer's pruning heuristic. The explicit `case` is the better
shape anyway: the "chain resolved to nothing, fall back to the peer" line is
load-bearing and now reads as a branch instead of a trailing `||`.

## Test plan

- [x] `scripts/check.sh` from the worktree root — **rc=0**: 5323 tests / 0 failures,
      Credo `found no issues`, Dialyzer `Total errors: 0`, Sobelow, `deps.audit`
      (`No vulnerabilities found`), doctor.
- [x] Rebased onto `origin/main` c58ec3eb; `git merge-base --is-ancestor` verified.
- [ ] CI (integration + e2e not run locally).

### DIFF
--- a/lib/grappa_web/plugs/loopback_only.ex
+++ b/lib/grappa_web/plugs/loopback_only.ex
@@ -1,7 +1,7 @@
 defmodule GrappaWeb.Plugs.LoopbackOnly do
   @moduledoc """
-  Halts any request whose `remote_ip` isn't `127.0.0.1` (or `::1`) with
-  a uniform 403 JSON body.
+  Halts, with a uniform 403 JSON body, any request that did not reach the
+  BEAM directly from inside the container / jail.
 
   Used to gate the admin reload endpoint (`POST /admin/reload`) so it's
   only callable from inside the running container / jail — `docker exec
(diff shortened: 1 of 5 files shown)


vjt:

Issue #215, opened 2026-07-11
Structured session-lifecycle logging for the IRC connection layer


vjt:
## Problem

The application logs (jail `grappa`, `runtime/log/erlang.log.*`, via `run_erl`) contain **only Phoenix web-layer events**: HTTP requests, WebSocket connect/refuse, and channel JOINs keyed on the grappa user. There is **zero logging of IRC session lifecycle**.

When an upstream IRC session drops (observed tonight: a user disconnecting with `Read/Dead Error: Input/output error`), nothing is recorded. No entry for the connection, the nick, the disconnect reason, or session expiry. The event is invisible and impossible to correlate after the fact — grepping the logs for the nick returns nothing.

This makes any disconnect / session-expiry investigation blind.

## Proposal

Introduce proper structured logging for the IRC session/connection layer.

- **Structured JSON logging** (a real logging framework / formatter) for machine correlation.
- **Plain human-readable text log** alongside it — 1995-style, greppable by a human at 2am.

### Events to log (at minimum)

- **Connection open**: upstream IRC connect, network, nick, session id.
- **Registration / auth**: NickServ identify, `+r` acquired/lost, account name.
- **Disconnect**: reason (quit message, ping timeout, server error, I/O error), duration, was-clean vs error.
- **Session expiry**: when/why a session is reaped or expires.
- **Reconnect / respawn**: attempts, backoff state, success/failure.

Each entry should carry: timestamp, network, user/visitor id, nick, session id, and a stable event type.

## Context

Raised by vjt on #it-opers after a real disconnect (H\mob) could not be explained from the logs. Standing instinct confirmed: "ci servono log migliori di quelli che abbiamo."

vjt:
Scope note from vjt (~2026-07-14 23:05 UTC) — pairs the admin-visibility half with the structured-logging half:

Beyond emitting the structured IRC session-lifecycle events, **persist the system logs to disk and make them browsable/tailable through the admin interface**. The admin UI already has an "events" log but it appears **not persisted** (ephemeral / in-memory). Requirement: write the logs to disk and give admin a **read + tail** view of them. Explicitly **"nothing fancy"** — no search/analytics/filtering bells; just persist to disk and read/tail via admin.

So #215 has two halves:
1. **Emit** structured IRC session-lifecycle logs (the original issue body — connect, auth/+r, disconnect reason+duration, expiry, reconnect/backoff; JSON + human-readable).
[...]

PR #969, opened 2026-08-07
auth: bring two login doors onto the shared failure-window shape

vjt:
Three slices on the credential doors, all on `RateLimit.FailureWindow`. The
first two bring two doors onto the shared shape; the third makes every door
say so out loud. No new infrastructure, no schema change, no
`protocol_version` bump.

## What changes

**Visitor login door** (`AuthController.visitor_login/7`) gains a per-source-IP
window, bucket `:visitor_login`, 10 per 15 minutes. Only a credential-bearing
attempt is gated; only a wrong password charges it. It sits at the HTTP edge
rather than beside the compare because `Grappa.Visitors` depends on neither
`Grappa.RateLimit` nor `Grappa.AdminEvents` — the Boundary compiler said so,
and widening a domain context to hold a request-edge policy is the wrong trade
when all three sibling windows already live in controllers.

**TOTP verify** (`:totp_login`) gains the aggregate per-IP ceiling its siblings
have. It was keyed on `{ip, user_id}` alone — a key stricter than per-IP, which
means the fine limit bounds the pair while nothing bounds the address.
`:passkey_recovery` already carries both dimensions; this brings the second
factor to the same shape, same 30. The success path still clears the fine key
only, so one satisfiable account cannot reset the ceiling.

**Every window now emits, and names itself.** `GrappaWeb.LoginThrottle.charge/4`
records the failure AND raises the crossing signal in one verb; all eight charge
sites go through it. A new door cannot be added mute, because there is no longer
a way to record a credential failure without signalling. `FailureWindow.check/3`
deliberately stays at each call site — only the caller knows what work it gates
and what refusing must return. No deny logic moved.

## The count was wrong, and the correction is the point

The brief said three of four doors were silent. The unit that shuts is not a
door but a `FailureWindow` row, `(bucket, key)`, and `:totp_login` and
`:passkey_recovery` carry two rows each. **Seven windows guard a credential
door; two emitted; five were silent — and one of the five was added by the
second commit in this very PR.** A window that denies without a signal does not
fail loudly, it fails as an unexplained login problem, which is the one failure
mode an operator cannot tell from a bug.

Attribution is not decoration either: seven windows sharing one undifferentiated
event would report that some address hit thirty failures and leave the reader
guessing between a TOTP ceiling, a recovery ceiling and challenge-allocation
abuse. So the event carries `door` (which IS the bucket — derived, not tracked
beside it) and `scope` (`:ip` / `:ip_account`, derived from the shape of the
key). On the two-row doors that distinction is the whole read: the fine key says
one account is hammered from that address, the ceiling says that address sprays
across accounts.

The account id is deliberately absent from the payload. On the recovery door the
identifier is attacker-supplied, so carrying it would let a spray write names of
its own choosing into the admin stream.

## The key choice, stated

Every window is keyed on the source IP (optionally paired with an account) and
never on an account alone. A per-account dimension that DENIES is a remote
lockout on someone else's identity — the window renews, so a slow trickle would
hold an account shut indefinitely with the victim paying rather than the caller.
Nothing here adds a lockout vector that per-IP admission did not already carry
(DESIGN_NOTES 2026-07-03). Nothing here bounds the distributed case: N addresses
still get N windows. That is detection before prevention, and stays out of scope.

## Why the two new fields are `optional` on the wire

Not a hedge. The admin ring is mirrored to disk and reloaded at boot (#215
Option B), and `snapshot/0` serves those rows to a joining admin socket — so
after this ships the Events tab genuinely replays `login_throttled` rows minted
by a server that predates the fields. Absence is a real wire state, so cic
narrows both to `undefined` rather than nulling the event: the trip is
load-bearing, the attribution is detail, and a missing detail must not delete
the alert. Additive fields, so no version bump.

## Verification

- `scripts/check.sh` green end to end, rc read from a file: **5329 tests /
  0 failures**, Dialyzer `Total errors: 0`, Credo `found no issues`, doctor
  passed (dev + test), `wireTypes.ts is in sync`, bats 331/331. Sobelow's only
  finding is a pre-existing Low in `cic/bundle.ex`; the gate is Medium-or-above.
  Working tree clean before AND after.
- cic: `bun run check` rc 0 (biome *and* tsc — biome short-circuits, so the rc
  is what proves tsc ran), build clean.
- `wireTypes.ts` is codegen output under a drift gate. It was transcribed by
  hand first, then regenerated: byte-identical.
- Every claim proven by mutation rather than by a green run. Disabling the
  emitter reddens exactly seven tests and nothing else. Hardcoding the
  attribution to `:mode1_login, :ip` reddens six — mode-1 stays green *by
  construction*, because that is precisely what that door emits.

## Known gap, stated rather than buried

One cic vitest run out of three came back with 2 failures; the two later runs
were green on identical code. The failing run's log was not captured, so the
tests are not even named, and **this is not attributed** — not to this branch
and not to the host. A cold start is plausible (that run followed a 483-package
install with ~1400s of environment time) but plausible is not measured.

### DIFF
--- a/cicchetto/src/AdminEventsTab.tsx
+++ b/cicchetto/src/AdminEventsTab.tsx
@@ -1,6 +1,10 @@
 import { type Component, For } from "solid-js";
 import { adminEvents } from "./lib/adminEvents";
 import { assertNever, type WireAdminEvent } from "./lib/api";
+import type {
+  AdminEventsWireLoginThrottleDoor,
+  AdminEventsWireLoginThrottleScope,
+} from "./lib/wireTypes";
 
 // M-11 — Admin events tab. Renders the in-memory ring buffer from
 // `adminEvents()` newest-first. Per `feedback_no_localized_strings_server_side`,
@@ -98,8 +102,13 @@ function renderEvent(ev: WireAdminEvent): string {
     case "credential_unbound":
       return `${ev.user_name} unbound from ${ev.network_slug}${actorSuffix(ev.actor_user_name)}`;
     case "login_throttled":
-      // S6 — admin-login brute-force gate tripped for this source IP.
-      return `login throttled: ${ev.source_ip ?? "(unknown ip)"} hit ${ev.failures} failures in ${Math.round(ev.window_ms / 60000)}m`;
+      // S6 — a credential door's brute-force gate tripped for this source
+      // IP. Seven windows across five doors share this event, so the door
+      // and the key that crossed are what make it actionable: the fine key
+      // says one account is being hammered from that address, the ceiling
+      // says the address is spraying across accounts. Both are additive —
+      // a row minted before they existed reads as the bare trip.
+      return `login throttled: ${ev.source_ip ?? "(unknown ip)"} hit ${ev.failures} failures in ${Math.round(ev.window_ms / 60000)}m${throttleAttribution(ev.door, ev.scope)}`;
     case "web_session_severed":
       // #630 — a subject's web session was severed for sustained inbound
       // flooding (socket closed + bearer revoked; IRC session untouched).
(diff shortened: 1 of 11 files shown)


vjt:

PR #970, opened 2026-08-07
auth: make session death and socket teardown one fact

vjt:
## What

Establishes one invariant: **a subject's bearer sessions and its live
WebSockets die together.**

`GrappaWeb.UserSocket` authenticates once, at connect, and afterwards holds no
tie to the row it authenticated against — so the teardown has to be pushed from
wherever the row dies. That push was written out by hand at each call-site,
which makes it a rule, and a rule drifts. It had.

## How

The kill sites emit a domain event (`Grappa.Accounts.Revocations`); a supervised
web-layer listener translates it into the existing
`UserSocket.disconnect_user_name/1`. The indirection is not ceremony — calling
the socket from `Grappa.Accounts` is a context → web dependency `Boundary`
refuses, the same reason `Grappa.Operator` carries no web deps.

The point is that announcing is **not a new per-door rule** — that is the shape
that drifted. Every write to `accounts_sessions` passes one of seven
chokepoints: the four revoke functions in `Grappa.Accounts`, which own the
table, plus the three places a parent row's deletion cascades onto it
(`delete_user/1`, `delete_expired_sessions/0`, and Visitors' single hard-delete
mechanic). Both timer-driven reapers are covered **without being touched**,
because they funnel through those. A door added later inherits the teardown by
construction rather than by memory.

## Second commit: the two recovery verbs change lanes

`reset_totp` / `reset_passkeys` were mix tasks. A mix task runs a second
`:grappa` instance in another OS process: its DB writes land, its in-node
effects reach only itself. No in-node mechanism can close that from the
outside, so the lane was the fix. They are now `rpc` verbs (`bin/grappa
reset-totp` / `reset-passkeys`) and the mix tasks are deleted rather than kept
as aliases — a door left standing is a door someone uses.

General rule left behind: boot-lane verbs are legitimate when their entire
effect is rows in a file. The moment a verb's effect includes anything the
running node holds in memory, it is an `rpc` verb.

## Decisions worth review

- **Granularity is per-subject.** `UserSocket.id/1` is keyed by subject, so a
  revoke of the *other* sessions also closes the acting device's socket. Its
  bearer is still valid and `phoenix.js` reconnects on a 1001 close, so the cost
  is a blip. Per-session granularity would mean re-keying `UserSocket.id/1` and
  teaching `WSPresence` to carry `session_id` — left as an **open question for
  vjt, strictly additive**, not a closed door.
- **Over-firing is the safe direction.** In-transaction sites announce from
  inside the transaction, so a rollback or a `BusyRetry` replay can announce
  something that never committed. That costs a reconnect. Announcing only after
  commit would push the call back out to the call-sites — the drift, restored.
- **Re-checking on inbound frames was rejected**, on top of the cost: it ties
  the guarantee to the client *sending* something, the wrong half of a duplex
  connection to hang it on.

## Test plan

- [x] TDD, measured both ways. Red produced by reverting **only** the door
  call-sites (`accounts.ex`, `visitors.ex`) to the merge-base while leaving the
  mechanism in place, so the red measures "the doors do not use it", not "the
  component is missing": **18 tests, 13 failures** — exactly the positive
  door tests. Restored: **18 tests, 0 failures**.
- [x] `scripts/check.sh` green: `rc=0`, 0 `not ok`, 335 bats, **5333 ExUnit
  tests / 0 failures**. Working tree clean before and after.
- [x] Test harness itself proven live with a deliberate formatting violation
  (`rc=1` + diff) before trusting a silent `rc=0`.

## Not claimed

- Not verified reachable on the m42 jail substrate. `bin/grappa` is
  `docker compose exec`, exactly as `scripts/mix.sh` was, so the lane move does
  not regress substrate reach — but neither path was probed on prod.
- The `reset_totp` / `reset_passkeys` unit tests prove the chokepoint announces
  inside its transaction. They **cannot** reproduce the second-BEAM condition,
  which is unreproducible in ExUnit; the lane change is covered by the bats
  asserting `--rpc-eval` and the absence of `mix.sh`.
- `revoke_other_sessions_for_user!/2` is driven at the chokepoint, not through
  `confirm_totp_enrollment` (which needs a live TOTP code).
- Out of scope, noticed not fixed: `db-latency` / `db-latency-reset` are in
  `VERBS` but missing from `VERB_DISPLAY_ORDER`, so they do not appear in
  `bin/grappa help`. Pre-existing.


vjt:

Issue #447, opened 2026-07-26
protocol: version handshake, unauthenticated /api/config discovery, additive-only contract + client-authors doc


vjt:
## Why

grappa is REST-first and means to be consumed by clients we don't write (cicchetto and shottino already are two, a third-party one is the point of the whole design). Today there is nothing in the wire contract that lets such a client discover what it is talking to, or fail cleanly when the server moves ahead of it:

- **No version anywhere.** Neither the REST surface nor the WebSocket mount (`lib/grappa_web/endpoint.ex:77`, `socket "/socket", GrappaWeb.UserSocket`) carries a protocol version. A client written today against a server from six months from now finds out by misparsing a frame.
- **No unauthenticated discovery endpoint.** The only unauth route is `get "/healthz"` (`lib/grappa_web/router.ex:87`), which says the server is alive and nothing else. A client cannot ask "what are you, and what do you speak?" before authenticating.
- **No written compatibility rule.** We add frame kinds and fields as we go; nothing states whether a client may assume an unknown field is fatal or ignorable. Both sides are currently guessing.
- **No document for client authors.** The protocol is knowable only by reading the Elixir.

## Prior art — credit where it's due

This is lifted, deliberately and with attribution, from [**amiantos/lurker**](https://github.com/amiantos/lurker) (MPL-2.0), an independently developed project with the same architectural premise as grappa — server-side bouncer, client never speaks IRC — reached in parallel, not derived from us or we from it (different stacks entirely: TypeScript/Node/SQLite/Vue vs Elixir/BEAM/Postgres). Their `docs/CLIENT_PROTOCOL.md` solves exactly this problem and solves it well. We are copying the *contract shape*, not any code.

## What to build

1. **`protocolVersion` integer**, exposed on both the discovery endpoint and the initial WS payload, plus a `minProtocolVersion` the server refuses to go below.
2. **Version announced on the WS handshake** — client sends its version on connect; a server whose `minProtocolVersion` has moved past it rejects the upgrade with a clean **`426 Upgrade Required`** rather than accepting the socket and feeding frames the client will mangle. Note the existing token-in-subprotocol mechanism (see `cicchetto/e2e/tests/issue95-ws-token-subprotocol.spec.ts`) — the version signal must not collide with it; decide query param vs subprotocol as part of this work.
[...]

PR #980, opened 2026-08-07
#964: name a push device — stored label, derived ordinal

vjt:
Closes the two points of #964 that slice A (`3aefadab`) deliberately left
open: the renameable label and the ordinal-suffixed default. The
stored-vs-derived question is settled — **the label is stored, the ordinal
is derived.**

## Deploy classification: COLD

Verified, not assumed: `Grappa.Deploy.Preflight` classifies every
`priv/repo/migrations/*` path as `:migration`, substrate-independent, with
no exemption for an additive nullable column.

## The three commits

| SHA | What |
|---|---|
| `f43421f0` | server — migration, `label` column, `label_changeset/2`, `Push.update_label/2`, `PATCH /push/subscriptions/:id`, `label` on the JSON summary |
| `b03cd413` | cic — `deviceRows` (label → parsed → parsed + ordinal), inline rename from the row, CSS |
| `a9280925` | e2e — the rename and the ordinal driven through the real drawer |

## Why the ordinal is derived, and derived in cic

Two independent reasons, either sufficient on its own:

1. An ordinal is a function of *how many rows currently share a parsed
   name*. Stored, it goes stale the instant one of them is deleted — a
   `#2` with no `#1`, which nothing realigns. Derived, the list self-heals
   on the next read. The stored thing here is the LABEL.
2. The grouping key is the OUTPUT of `parseUserAgent`, which exists only
   in cic — `grep -rn user_agent lib/ --include='*.ex'` shows no
   server-side UA parser. Deriving it server-side would mean a SECOND
   parser in Elixir, free to classify differently, numbering a grouping
   the user cannot see.

Consequences, each with a test: a device alone in its group takes no
suffix; labelled rows leave the GROUPING (not just the suffix), so naming
one twin drops the other's number; ordering is `created_at` ascending with
the id as tiebreak, so a new twin appends `#3` instead of renumbering the
two already on screen.

## Corrections to the brief this was written from

1. **Point 4 (the "this device" marker) was already shipped.** `3aefadab`
   is on `origin/main` (`git branch -r --contains`) and carries
   `subscriptionIdForEndpoint`, the badge and its e2e. Nothing rebuilt.
2. **The ordinal is derived in cic, not on the server** — the second-parser
   argument above.
3. **No `Grappa.Push.Wire`, and no wire-type regeneration.** The push REST
   shapes come from `GrappaWeb.PushSubscriptionJSON`; CLAUDE.md's `*.Wire`
   rule governs PubSub/Channel payloads. `wireTypes.ts` carries no push
   shapes, and `mix grappa.gen_wire_types --check` passes untouched.
4. **COLD confirmed from `Preflight`**, rather than from the rule of thumb.

`label` is additive on an existing shape, so per the #447 additive-only
contract it costs no `protocol_version` bump.

## What I measured, and what I am NOT claiming

**Measured.** `scripts/check.sh` exit 0 end-to-end: 5377 tests / 0
failures, Dialyzer `Total errors: 0`, credo --strict, sobelow, doctor,
`gen_wire_types --check`, bats. The 13 new vitest cases are
mutation-measured — four mutations (number lonely groups, let labelled rows
re-enter the grouping, drop the `created_at` ordering, ignore the label)
each killed 1–4 of them, green on restore. The server side likewise: four
mutations (drop the normalisation, widen the cap, add `label` to the create
allowlist, widen the controller guard) each killed 1–3 of the 69.

A hand-written blank→NULL arm in `normalize_label/1` was **proven dead** —
Ecto's `cast/3` already folds `""` and whitespace-only to nil via its
default `:empty_values` — so it was deleted and the blank/whitespace tests
stayed green. The tests pin the contract; the implementation is minimal.

**NOT claiming.** The new e2e spec `push-964-device-label.spec.ts` has
**never been executed**. The STACK lane was held by other work, so it was
written from the sibling spec's shape and reasoned through, not run. The CI
`integration` job is its first execution — treat a red there as expected
cost, not as a surprise.

Also unverified by measurement: that the two rename buttons fit the row on
a narrow mobile viewport. The CSS puts the editor in the flex slot the name
occupied so the list should not reflow, but I read that off the stylesheet
rather than off a screenshot.

### DIFF
--- a/cicchetto/e2e/tests/push-964-device-label.spec.ts
+++ b/cicchetto/e2e/tests/push-964-device-label.spec.ts
@@ -1,0 +1,147 @@
+// push-964 — naming a device, and the ordinal that names the ones you have not.
+//
+// The sibling spec (`push-964-device-row.spec.ts`) covers the two
+// disambiguators derived from data the row already carried: the activity
+// instant and the "this device" marker. This one covers the two that needed a
+// schema change — the user's own label, and the ordinal that separates the
+// twins nobody has named yet.
+//
+// Four contracts, every one asserted on what a person would SEE:
+//
+//   1. Two devices that parse to the same `Browser on OS` render DIFFERENT
+//      names, numbered oldest-first.
+//   2. Renaming one from its row replaces that name, and demotes the parsed
+//      string to the secondary line instead of dropping it.
+//   3. Naming one twin drops the OTHER's ordinal. This is the assertion a
+//      STORED ordinal would fail: a column written while the pair collided
+//      keeps its `#2` forever, whereas the derived name re-reads the list and
+//      finds the survivor alone in its group.
+//   4. Clearing the label hands the row back to the derived default — the
+//      twins collide again and BOTH are numbered once more.
+//
+// ## Why the rows are POSTed rather than registered from two browsers
+//
+// The contract under test is about two rows whose `user_agent` parses to the
+// same string. The runner has one engine per project, so a second browser
+// would supply the same UA — but also a second push stub, a second SW
+// registration and a second opt-in dance for rows this spec only ever reads.
+// POSTing them is the same insert the server would have performed, and it lets
+// the spec pin the UA, so the expected names are literal (`Firefox on Linux
+// #1`) instead of "whatever this engine calls itself".
+//
+// ## Why there is no page.reload()
+//
+// On boot `installPushResubscribe` renews a subscription the per-document push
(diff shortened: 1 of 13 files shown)


vjt:
Closing: #964 was ruled DROP by vjt (2026-08-11) and closed as not planned. This PR has no issue to land against. The branch `964-device-label` is kept and the head is preserved at `a9280925` — reopen if the ruling changes.

Issue #988, opened 2026-08-07
rail actions menu: a tall menu must open with its FIRST row visible, not anchored at the bottom


vjt:
Ruled by vjt. Ship hot with 0.13.3. Follows from the actions menu getting three
more rows (`detach`, `quit`, `@ mentions`) in the sibling issue.

## The ask, verbatim

> quando apriamo il menu dovremmo sempre mostrarlo con il primo bottone visibile.
> il menu actions diventerà alto e non entra tutto nella view, attualmente viene
> mostrato al bottom, dovrebbe essere mostrato al top

A screenshot was attempted and did not go through, so the geometry below is read
off the code, not off a picture — the discriminator at the end says which of the
two readings the implementer should confirm on a device first.

## How it is anchored today

`default.css:1891` — `.rail-actions-menu { position: absolute; bottom: 100%; left: 0; right: 0 }`.

The menu hangs off the **top edge of the launcher**, and the launcher is pinned
to the **bottom** of the rail (`.rail-actions { margin-top: auto; flex-shrink: 0 }`,
`:1856`). So the menu grows **upward** from the bottom of the screen. Its height
is capped by `--rail-menu-space-above`, measured in `RailActions.tsx:171` via
`spaceAbove(rootRef.getBoundingClientRect().top, RAIL_MENU_TOP_GAP)` and published
as a custom property, with `overflow-y: auto` engaging only past that cap
(#588, and #913 for the safe-area correction).

That design was chosen when the menu was short: zero permanent vertical cost, no
displacement of the nick list. With six rows becoming nine, "short" stops being
true, and a bottom-anchored scrolling column puts the FIRST row — `home` — at the
far end of a scroll on the small viewport where the cap bites hardest.

## Two readings, and the discriminator

1. **Position.** The menu is simply drawn at the bottom of the screen and vjt
   wants a tall menu to start at the top instead — i.e. once it cannot fit above
   the launcher, stop anchoring to the launcher and anchor to the top of the rail,
   filling downward.
2. **Scroll offset.** The menu is top-anchored enough, but something leaves it
   scrolled so `home` sits above the fold.

Reading 2 is the one to falsify first, because it is cheap: a fresh
`overflow-y: auto` container has `scrollTop === 0`, so the first row should
already be visible unless something scrolls it. Open the menu on a short viewport
and read `scrollTop` on `.rail-actions-menu`. **If it is 0, reading 2 is dead and
[...]

Issue #987, opened 2026-08-07
delete-account modal is stretched to full viewport height — the backdrop is align-items: stretch


vjt:
Reported by vjt. Ship hot with 0.13.3.

`delete account` stays where it is (settings drawer, `SettingsDrawer.tsx:1043-1050`).
What is wrong is its modal: it takes the **whole screen height** regardless of how
little content it holds. It should sit on the screen like a dialog — centred,
sized to its content, capped — not stretched edge to edge.

## Cause

One property, and the sibling modal already gets it right:

- `default.css:3773-3781` — `.delete-account-backdrop { display: flex; align-items: **stretch** }`
- `default.css:3870-3878` — `.confirm-modal-backdrop { display: flex; align-items: **center** }`

`align-items: stretch` on a flex backdrop forces the child to the full cross-axis
extent, so `.delete-account-modal` (`:3783`) is stretched to the viewport height
whatever its content is. Its `max-height: var(--viewport-height, 100dvh)` then
reads like a cap but never bites — the box is already at that height because the
backdrop put it there. The modal also has **no margin**, so it runs edge to edge
horizontally up to its `max-width: 32rem`, where `.confirm-modal` carries
`margin: 1rem` (`:3888`).

## Fix

Bring the delete backdrop/modal to the `.confirm-modal` contract:

```css
.delete-account-backdrop { align-items: center; }
.delete-account-modal    { margin: 1rem; }   /* max-height stays: now it caps instead of describing */
```

With `center`, the box sizes to content and `max-height: var(--viewport-height)`
becomes a real ceiling — and the `overflow-y: auto` + `touch-action: pan-y`
already on the modal (`:3796-3798`) start doing their job for the tall case
(large text size, keyboard up) instead of being dead weight in the short case.

Keep everything else as-is: the red border is the destructive signal (#157) and
the type-your-name gate is the point of the dialog.

## Test

A geometry assertion, not a computed-style one: open it with short content and
assert the dialog's height is well under the viewport's and that scrim is visible
above and below. `getComputedStyle` under device emulation has already lied once
(#963).

vjt:
Shipped in **v0.13.3** — https://github.com/vjt/grappa-irc/releases/tag/v0.13.3

cic-only, so it is live in production now (hot bundle deploy, no restart). Self-hosters: pull the `v0.13.3` tag.

Issue #986, opened 2026-08-07
session lifecycle: move detach + quit (and the @ mentions button) into the rail actions menu, behind per-subject confirm modals


vjt:
Ruling by vjt. Ship hot with 0.13.3.

## What changes

`detach` and `quit` move **out of the settings drawer** and **into the rail actions
menu**, each behind a **confirm modal** that says, in words, what is about to
happen — with copy that differs per subject, because the consequence differs per
subject. `delete account` does **not** move: it stays in settings (its modal has
its own issue).

Third mover, same destination, ruled separately by vjt: the **`@` mentions
button** also moves into the actions menu. It lives today in
`ShellChrome.tsx:82-96`, gated on `isMobile() && mentionsOpenSlug()`, and it is
the only door back into a mentions bundle on a phone. Moving it here is what
lets #985 drop the whole `.shell-chrome` band and float a lone ☰ — so that issue
depends on this one landing the `@` first.

## Where they live today

- `SettingsDrawer.tsx:1002-1013` — `detach` button, gated on `showDetach()`
  (`:158` → `isPersistentIdentity(getSubject())`), fires `onDetach` (`:223`)
  with **no confirmation at all**.
- `SettingsDrawer.tsx:1020-1030` — `quit`, an `InlineConfirmButton` two-tap arm
  ("really quit IRC?"), fires `onQuit` (`:235`).
- `RailActions.tsx` — the target surface. Every entry is already
  `<button class="shell-chrome-btn rail-action rail-action-*">` with a
  `.rail-action-icon` emoji span + a `.rail-action-label` (see `home` at
  `:204-224`), so the two new entries are the same shape as the existing five.

## Copy — the consequence really is three different things

Read off `lib/lifecycle.ts` (the comment at `SettingsDrawer.tsx:230-234` states
the routing) — the modal body must say the true one, not a generic warning:

| subject | detach | quit |
|---|---|---|
| **registered user** | offered — leaves cic, **the bouncer keeps running**: networks stay connected, scrollback keeps filling, you come back to it | parks **every** network; the bouncer goes offline; the identity and its history survive |
| **registered visitor** (NickServ-identified) | offered — same promise as the user | drops the upstream, then detaches; **the row is kept**, so the identity survives |
| **ephemeral visitor** (anon) | **not offered** — `showDetach()` is false | detaches, and **the server purges the anon row**: the session and everything in it are gone for good, there is nothing to come back to |

[...]

Issue #985, opened 2026-08-07
mobile: non-channel windows spend a whole header band on the ☰ alone — float it top-right


vjt:
Reported by vjt on iPhone (PWA), grappa 0.13.2.

## Symptom

On mobile, a **query** window (and every other non-channel window) renders a
full-width header band across the top whose only content is the ☰ rail opener,
right-aligned. The rest of the band is empty. With a large text size the band is
tall enough to cost a meaningful slice of the scrollback — and it costs it in the
one state where vertical space is scarcest, keyboard up.

vjt's ask, verbatim: show **only the hamburger, top right**, and give the band
back to the conversation.

## Where it comes from

- `Shell.tsx:800` — the mobile branch renders `<ShellChrome>` under
  `<Show when={selectedChannel()?.kind !== "channel"}>`.
- `ShellChrome.tsx:79` — the component is a `<header class="shell-chrome">`
  holding a `.shell-chrome-spacer` (`flex: 1`) and the ☰ button.
- `default.css:2300` — `.shell-chrome { display: flex; padding: 0.5rem 1rem;
  border-bottom: 1px solid var(--border); flex-shrink: 0 }`, and the button
  carries `min-height: var(--chrome-tap-min)` (`:2329`), so the band's height is
  `tap-min + 1rem + border` and **scales with the operator's text size**.

Note this row is already suppressed for **channel** windows precisely for this
reason — the comment at `Shell.tsx:797` says so: "Mobile-channel still suppresses
this row so the scrollback reclaims the ~32px", because there the TopicBar
hamburger does the same job inline. The query case is the one that never got a
host for its ☰, so it kept the whole row.

## The constraint that must not be dropped

`.shell-chrome` is **not** ☰-only in every state. `ShellChrome.tsx:82-96`
conditionally renders a second button — `@`, the mentions re-open — gated on
`isMobile() && mentionsOpenSlug()`. On mobile there is no Sidebar, so per that
component's own moduledoc this `@` is **the only door back into a mentions
bundle**. Removing the band without rehoming `@` silently deletes that door on
phones. Whatever replaces the row has to carry both glyphs when both apply
(a floating cluster, not a floating single button), or `@` needs an explicit new
home and a note saying where.

Second constraint: the ☰ must stay one glyph opening one drawer. The
"ONE drawer, one ☰" ruling (Opt-A, referenced at `Shell.tsx:796` and
`ShellChrome.tsx:98`) is what keeps the channel hamburger and this one from
[...]

PR #991, opened 2026-08-07
cic(#986): lifecycle verbs move to the rail, behind a subject-true confirm

vjt:
Branch 2 of 3 in the cic 0.13.3 cluster. #985 and #988 depend on this landing.

## What moved

`detach` · `quit` · the `@` mentions door — three affordances, one
destination (the rail actions menu). The settings drawer's `admin console`
entry was deleted as an exact duplicate of the rail's 🔧.

## The point of the change

The control being replaced was an `InlineConfirmButton` two-tap reading
**"really quit IRC?"** — six words shown identically to a registered user,
whose account and scrollback survive a park-all untouched, and to an
anonymous visitor, whose row the server HARD-DELETES on the way out
(`Visitors.purge_if_anon` → `destroy_visitor`, CASCADE). Those are not the
same event.

So the shared #195 confirm modal carries a body chosen for the subject in
front of it — four sentences across the two verbs. The three-way `quit`
split ROUTES on `isPersistentIdentity` (the shared #477 predicate, reused
not re-derived, so `canDetach()` and `quit()` cannot answer the persistence
question differently); only inside the persistent arm does `kind` pick the
noun the operator owns. The load-bearing test asserts the three bodies
genuinely DIFFER — a "the modal opens" test would pass with one shared
string, which is the defect this closes.

**No typed re-entry gate here.** Per vjt's ruling, that friction is
exclusive to `delete account`, for BOTH registered subjects, because the
gate is about irreversibility rather than subject kind. `delete account`
did not move; its geometry is #987.

**`InlineConfirmButton` is not retired.** It keeps serving ~20 call sites
unchanged, including the vhost apply gate in this very drawer. Only its use
as a lifecycle-verb gate ends, in the two settings call sites the issue
names.

## Two judgment calls, flagged for veto

1. **The `@` mentions entry drops its `isMobile()` gate.** In
   `.shell-chrome` it was mobile-only so it would not duplicate the desktop
   Sidebar mentions row (#71 INC-2). In the rail I made it bundle-gated
   only: #473 documents the rail as capability-gated with *no form-factor
   gates*, and `home` is the standing precedent — a sidebar row AND a rail
   launcher, deliberately. Cheap to reverse if you want it mobile-only.
2. **`.logout` carried `margin-top: auto`**, which pushed the tail of the
   drawer's button stack to the floor. I deleted the rule with the button
   and gave the margin no successor: the index above overflows in every
   shipped viewport, where an auto margin in a flex column buys nothing.
   That is reasoning, not a screenshot.

## Blast radius the issue did not mention

Removing the drawer's admin entry would have reddened ~20 e2e specs that
hand-rolled `admin-console-entry.click()`. They now go through the
`openAdminConsole` primitive — the migration that helper's own comment
tracked as a follow-up. That also retires the flake it was written to
absorb: reaching admin via the drawer clicked an entry that CLOSED the
drawer, whose 200ms slide-out at z-index 100 could swallow a click aimed at
the 9th admin tab (~9% once #508's iOS font floor perturbed the timing). The
rail launcher opens no drawer, so the `not.toBeInViewport()` wait is gone.

`m7-admin-gate-settings-drawer.spec.ts` keeps its filename on purpose — six
sibling specs cross-reference it as THE M-7 gate spec, and the gate is
unchanged; only the surface moved. Its header records that.

## Measured

- `scripts/bun.sh run check` — exit 0 (biome + `tsc --noEmit`).
- `scripts/bun.sh run test` — exit 0, **244 files / 4545 tests**, on the
  final tree with a clean `git status --porcelain` before and after.

## NOT measured — do not read the green above as covering these

- **No e2e run at all.** The local stack lane was not requested, and this
  branch touches 27 spec files. The `e2e` job on this PR is the first real
  execution of every retargeted spec and of the new modal-copy assertions.
- **No `tsc` over `e2e/`** — there is no static gate there (biome ignores
  the directory, and `tsc -p e2e/` emits `.js`), and `e2e/node_modules` is
  not installed in this worktree. A type error in the 27 touched specs
  surfaces only when Playwright runs them.
- **No browser.** No screenshot, no rendered geometry: the two emoji rows
  and the drawer's lost bottom-pin are unverified visually.

### DIFF
--- a/cicchetto/e2e/fixtures/cicchettoPage.ts
+++ b/cicchetto/e2e/fixtures/cicchettoPage.ts
@@ -945,9 +945,10 @@ export async function openSettingsSection(
 
 // #500 — open the settings drawer ROOT (viewport-aware), returning the drawer
 // dialog locator. The sibling of openSettingsSection for specs that need the
-// drawer's main index itself — e.g. to reach the Admin Console entry
-// (admin-console-entry), which lives in the drawer, not in a #460 sub-section —
-// or to assert drawer chrome. The cog (aria-label "open settings" /
+// drawer's main index itself — to reach a main-page affordance (share session,
+// delete account) or to assert drawer chrome. (#986 — reaching the ADMIN
+// console is no longer one of those reasons: use openAdminConsole, which goes
+// through the rail.) The cog (aria-label "open settings" /
 // action-cluster-cog) now lives behind the RailActions launcher (#500):
 // openRailMenu reveals the menu (opening the rail drawer first on mobile), then
 // the cog is tapped. Idempotent — the open step is a no-op if the drawer is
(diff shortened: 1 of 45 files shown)


vjt:

A cherry-picked union rewrites the SHAs, so GitHub cannot close this PR itself. Landing verified BY CONTENT: `git log origin/main --grep '#986'` shows the commits on `main`. The work shipped; only this PR object is stale.

Issue #975, opened 2026-08-07
/mode on a channel you are not in returns nothing, or stale cached modes


vjt:
## Symptom

`/mode #channel-I-am-not-in` reports no modes at all. `/mode #channel-I-was-in-earlier` *does* report modes — stale ones, from the last time the operator was in it. Reported by Sonic on #it-opers while writing an IRC facade against grappa; his guess ("qualche stale state cachato") is exactly right.

## Cause

`/mode <#chan>` with no mode arguments **never sends anything upstream**. It is a pure local read.

`cicchetto/src/lib/compose.ts:1393-1403`:

```ts
case "mode-view": {
  const ch = cmd.channel ?? getActiveChannel();
  if (!ch) return { error: "/mode requires a channel — switch to one or use /mode #chan" };
  openModeModal(networkSlug, ch);
  result = { ok: true };
  break;
}
```

`openModeModal` (`cicchetto/src/lib/modeModal.ts:23-25`) only sets a signal — no fetch, no push. The modal then reads the cache: `cicchetto/src/ModeModal.tsx:132` / `:136` → `modesByChannel()[k]`.

That cache has exactly one filler and no emptier:

- **Filled** only at `cicchetto/src/lib/subscribe.ts:408` → `seedModes(key, payload.modes)`, on a `channel_modes` event delivered on the **per-channel WS topic**. Server-side the 324 RPL_CHANNELMODEIS handler (`lib/grappa/session/event_router.ex:1054-1068`) emits `{:channel_modes_changed, …}`, and `lib/grappa/session/server.ex:4745-4753` fans it out with `Broadcaster.to_channel` — i.e. onto `channel:<chan>`, which cic subscribes to only for channels it is in, pending, or invited to (plus query windows).
- **Never emptied**: `cicchetto/src/lib/channelTopic.ts:44-62` exposes `seedModes` and nothing else. There is no per-key delete on own-PART, and no `clearModes()` on logout / token rotation — unlike `scrollback.ts`, `selection.ts`, `members.ts` and `readCursor.ts`, which all have an identity-scoped reset.

So:

- **Never-joined channel** → no subscription ever existed → no cache entry → the modal renders empty, and nothing is ever requested to fill it.
- **Previously-joined channel** → the entry seeded while joined survives the PART forever → stale modes, presented as current.

`topicByChannel` in the same module has the identical shape and the identical gap, so a parted channel's topic is stale by the same mechanism.

## Two separate defects

[...]

Issue #495, opened 2026-07-27
cic: focus stuck on $server after re-joining a parted channel (selection close-watcher evicts a re-joining window)


vjt:
## Summary

The client-side selection **close-watcher** in `cicchetto/src/lib/selection.ts`
can permanently move focus to a network's `$server` console when a channel is
**re-joined shortly after being parted**. The user re-joins a channel they just
left, the channel rejoins successfully, but the active window stays on `$server`
and never returns to the channel.

## Repro (real user, desktop)

1. Be focused on a channel (e.g. `#chan`).
2. PART it (`/part`, or close-to-archive).
3. Open the archive and click the archived channel — it opens a read-only view.
4. `/join #chan` to rejoin.

**Expected:** focus lands on the rejoined `#chan` (members pane visible).
**Actual (intermittent, load-dependent):** focus lands on the network's
`$server` console; the channel rejoins (sidebar shows it joined) but is not the
active window, and focus never returns to it.

## Mechanism

The bucket-E close-watcher (`selection.ts`, ~L704–754) redirects on a
**live→dead transition** of the selected window, calling `resolveFallbackWindow`
(→ the network's `$server` window when no live MRU candidate exists, L752).

During the part→rejoin cycle the selected channel transiently reads **"dead"**
(its `windowState` entry is cleared and/or the `channelsBySlug` refetch
transiently omits it — cf. the mid-reconnect snapshot gap documented for #187).
The watcher treats this **transient** re-join dead-moment as a **permanent
close** and evicts to `$server`.

The redirect is **sticky**: once `selectedChannel.kind === "server"` the watcher
early-returns (L707), so focus never returns to the channel even after it lands
live again.

## Scope

Pre-existing and **independent of any archive-surface change** — `selection.ts`
is the defect site. Affects all users. It surfaces reliably as an e2e flake
(`cp15-b6-part-archive-rejoin`) but the defect is in **window selection**, not
the archive.

## Fix direction (to confirm)

The user-visible harm is the **stickiness** (never returning). The root cause is
the **false-positive live→dead classification** of a re-joining window. The
surgical fix is to *not evict on a transient dead* — a window with an in-flight
/ pending join, or one the user just explicitly navigated to — rather than a
blanket "return on re-live" (which would yank focus on netsplit/reconnect).
A genuine PART/close must still redirect and stay redirected.

## Verification requirement

The fix must ship with a detector **proven RED on the unfixed code and GREEN
[...]

Issue #216, opened 2026-07-11
cic: channel modes not visible on join + /mode viewer/editor modal


vjt:
## Bug (cosmetic, P1)

Channel modes are **not visible on JOIN**. Currently the chanmode display only appears *after* a mode has been set/changed during the session — join a channel and the existing modes are invisible until something toggles them.

## Feature

`/mode #channel` should surface the channel's current modes:

- At minimum, print them in the **scrollback**.
- Better: open a **modal** showing the current chanmodes with **checkboxes** (or an equivalent nice control) that also let you **edit** them.
- The same modal should be reachable by **clicking / tapping** the area where chanmodes are currently displayed (the area that today only shows up after modes are set).

So: one modal, two entry points (`/mode #channel` and tap-the-chanmode-area), read + edit chanmodes from it, and the modes must be populated/visible from the moment you join.

vjt:
Clarification from vjt: in-channel the chanmodes are not shown, but after issuing e.g. `/mode #chan +s` the `+s` then appears next to the topic. So the chanmode indicator near the topic is populated **only by a MODE-change event**, never from the channel's initial mode state on join. Likely root cause: the client renders modes from incoming MODE messages but never queries/renders the mode reply (RPL_CHANNELMODEIS / 324) on JOIN. The fix should fetch and render current modes at join time so the indicator is correct without needing a mode change to trigger it.

vjt:
Additional entry point (vjt): bare `/mode` with no channel argument should open the modal for the **currently-viewed channel**. So three entry points total: `/mode #channel` (explicit), bare `/mode` (current channel), and tap/click on the chanmode-indicator area.

vjt:
Further design detail (vjt):
- `/mode #chan` opens the modal for the indicated channel, but **editing must be gated on chanop** — a non-op can view the modes but not change them (read-only modal unless the user holds ops on that channel).
- The set of **allowed/available channel modes must come from the server's ISUPPORT numeric on connect** (RPL_ISUPPORT / 005 — the `CHANMODES=` token, and `PREFIX`). Parse it at connect time and keep it in memory, then drive the modal's available checkboxes from that per-network capability set rather than a hardcoded list.

vjt:
[...]

PR #993, opened 2026-08-07
cic(#975): drop the channel topic/modes cache when it stops being true

vjt:
Addresses **defect (2) only** of #975 — the cache that is never invalidated.
Defect (1), the refresh path, is deliberately untouched (see Scope below).

## The hole

`cicchetto/src/lib/channelTopic.ts` exposed `seedTopic` / `seedModes` and
nothing else. Modes arrive only via the 324 the server queries in its
`:joined` arm (#216 — ircds do not send 324 unsolicited), pushed on the
per-channel WS topic. Nothing ever removed a key, so an entry seeded while
joined outlived the PART for the tab's lifetime, and `/mode #chan-I-left`
— a pure local read — answered with it. Reported by Sonic on #it-opers.

It was the last per-channel signal map with no lifecycle at all:
`scrollback` / `selection` / `members` / `readCursor` all have an
identity-scoped reset.

## What changed

- **own-PART → `dropChannelTopicState(key)`**, in `subscribe.ts` beside
  the existing `setParted(key)` projection.
- **logout / rotation →** the store moves from `moduleRoot` to
  `identityScopedStore`, the shared factory the siblings already use.
- **`topicByChannel` goes with `modesByChannel`.** Same datum class, same
  hole. Splitting the lifecycle would leave twin structures behaving
  differently for no reason a later reader could recover.
- **`ModeModal` distinguishes entry-ABSENT from entry-EMPTY.** With the
  drop in place, "no entry" is the common case, and the old
  `?.modes ?? []` fed an all-off grid — which states "+" (no modes set)
  as fact. Absent now renders `modes unknown — not in this channel`
  instead of the grid; an empty `modes` array is a real 324 answer and
  still renders it.

### Why the drop reads `windowIsPresent` rather than re-testing state

`setParted` carries the #495 stale-echo guard: a part echo landing on a key
already back at `"pending"` is a no-op, because it is the echo of an
EARLIER part that lost the race to a re-join. Copying that condition into
the drop would create two guards free to drift. Gating on the OUTCOME —
the window still being present means the projection declined, so the
freshly-seeded modes must survive — keeps one. There is a test for exactly
that case.

## Measured

- `scripts/bun.sh run check` → exit 0 (biome + `tsc --noEmit`; biome does
  not read `e2e/`, so that was typechecked separately).
- `scripts/bun.sh run test` → **244 files / 4540 tests, all green**, tree
  clean before and after.
- `tsc -p e2e/tsconfig.json --noEmit` → 0 errors in the new spec (the 26
  it reports are pre-existing, in other files, none touched here).
- **Mutation-checked, not just green.** Each new assertion was proven to
  bite by breaking the thing it guards: dropping only `topicByChannel`
  (2 red), removing the `subscribe.ts` call (1 red), making the drop
  unconditional so the #495 case regresses (1 red), forcing
  `modesKnown()` true (1 red). All restored.

## NOT measured — do not read this as verified

- **The e2e has never been run.** The STACK lane is orchestrator-allocated
  and was not requested for this bucket. `issue975-mode-cache-invalidation`
  typechecks and follows the `issue216` shape, and that is the entire
  extent of the claim.
- No browser, no screenshot: the copy string and the muted style are
  reasoned from the existing `.mode-modal-readonly` treatment, not looked at.

## Scope, and a sibling instance left open

Defect (1) needs a server surface that issues `MODE #chan` for a channel
the session is not in AND a route home for the 324 (today it fans out on
the per-channel topic, which a not-joined channel has no subscription to).
Not here. **(2) alone converts a wrong answer into an absent one** — it
does not make `/mode` on a foreign channel work.

**A KICK is the same class and is NOT fixed.** After a kick you are equally
out of the channel and the cached modes go equally stale, but `setKicked`
keeps the window present (`:kicked` pseudo-row), so the gate declines and
the modal still shows the pre-kick modes. Flagged rather than widened
silently.

## Product call inside

The issue asks whether the modal should say "modes unknown" instead of
showing an empty toggle set. Implemented as unknown, on the grounds that an
empty set presented as current is a lie. It is one string if vjt disagrees;
the structural half (absence is not an answer) is what matters.

DESIGN_NOTES entry is staged at `/tmp/w1-designnotes-975.md` rather than
committed, so parallel workers do not conflict on the file.

### DIFF
--- a/cicchetto/e2e/tests/issue975-mode-cache-invalidation.spec.ts
+++ b/cicchetto/e2e/tests/issue975-mode-cache-invalidation.spec.ts
@@ -1,0 +1,100 @@
+// #975 — `/mode` on a channel you have LEFT must not answer with the modes
+// you saw while you were in it.
+//
+// Reported by Sonic on #it-opers: `/mode #channel-I-was-in-earlier` printed
+// modes, and they were the ones from his last visit. `modesByChannel` had
+// one filler (the 324 the server queries on JOIN, pushed on the per-channel
+// WS topic) and no emptier, so the entry outlived the PART forever and the
+// modal presented it as CURRENT.
+//
+// The witness needs the live upstream: the whole mechanism is a real JOIN
+// eliciting a real 324, then a real PART echo arriving on the per-channel
+// topic. jsdom can prove the store drops a key; only this can prove the key
+// was ever filled by the round-trip it is supposed to be filled by, and that
+// the operator-visible answer changes.
+//
+// Shape:
+//   1. a peer creates a per-run UNIQUE channel (→ op) and sets +m, a mode
+//      no bahamut channel carries by default, so its presence in the modal
+//      is unambiguously the cached 324 and not a default.
+//   2. vjt joins → the modal shows "moderated" PRESSED. This half also
+//      guards the #216 join-time query: if it ever stops firing, this fails
+//      here rather than making step 4 pass for the wrong reason.
+//   3. vjt PARTs.
+//   4. `/mode <channel>` again → the modal says the modes are unknown and
+//      renders NO toggles. Pre-fix it rendered the same grid as step 2,
+//      "moderated" still pressed — the reported symptom, exactly.
+//
+// Anti-pollution: unique channel per run, peer disconnected in `finally`
+// (which drops the channel once vjt has parted).
+
+import { composeSend, loginAs, selectChannel } from "../fixtures/cicchettoPage";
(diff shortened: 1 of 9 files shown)


vjt:

A cherry-picked union rewrites the SHAs, so GitHub cannot close this PR itself. Landing verified BY CONTENT: `git log origin/main --grep '#975'` shows the commits on `main`. The work shipped; only this PR object is stale.

Issue #974, opened 2026-08-07
Shift+Enter must just send: drop the multi-line refusal that swallows the message


vjt:
## Symptom

`IRC does not support multi-line messages` shows up in the compose box seemingly at random, on Enter presses the operator did not intend as Shift+Enter. Reported by vjt on 0.13.1. Worse than cosmetic: on that keystroke the message is **not sent** — the operator has to press Enter again.

## Decision (vjt, 2026-08-07) — Shift+Enter just sends

**Shift+Enter submits the message. No refusal, no notice.** This reverses vjt's 2026-08-06 ruling that "the refusal must SPEAK": there is nothing left to speak about, because nothing is refused any more. It is his ruling and his reversal.

It also supersedes the fix options this issue previously listed — (a) "send anyway, keep the notice" is **not** what ships. The notice goes.

Why this is the right shape and not a cop-out: the trigger that arms `shiftKey` on the reporter's device is **still unknown** (see below), and this fix does not need to know it. Whatever sets the modifier, Enter sends. The whole class of "the message got eaten" closes without a diagnosis.

## What ships

`cicchetto/src/ComposeBox.tsx:448-469` — the Enter branch collapses to:

```ts
if (e.key === "Enter") {
  e.preventDefault();
  void doSubmit();
  return;
}
```

**`e.preventDefault()` stays** — this is the one part of vjt's wording that must not be taken literally. It is what stops the textarea from inserting its own line break. Drop it and the newline lands in the box *after* the async `doSubmit()` has cleared the draft, leaving a stray `\n` in a freshly-emptied composer. The `#816` premise is untouched: a newline cannot travel inside a PRIVMSG (CRLF terminates the frame), so the box stays single-line. Paste remains the ONE route a multi-line body takes into the box, and paste is still guarded (`lib/pasteFlood`, `lib/pasteRoute`).

Consequence worth stating: after this, **every** Enter variant sends — Ctrl+Enter and Cmd+Enter already fell through to `doSubmit()`, so this makes the modifier handling uniform rather than adding a special case.

## Cleanup that is part of the change, not a follow-up

- `ComposeBox.tsx:106` — `SHIFT_ENTER_REFUSAL` becomes dead, remove it.
- `ComposeBox.tsx:450-465` — the comment block describes the OPPOSITE behaviour ("Shift+Enter inserts NOTHING and sends nothing", plus the 2026-08-06 ruling). Rewrite it; do not leave it contradicting the code.
[...]

PR #994, opened 2026-08-07
fix(#974): Shift+Enter just sends — delete the refusal that ate the message

vjt:
Fixes the defect in #974: the Enter branch gated the submit behind a bare
`e.shiftKey` while running `preventDefault()` unconditionally, so on any press
where the modifier was set the keystroke was consumed and **the message was not
sent**. Per vjt's 2026-08-07 ruling — which reverses his own 2026-08-06 one —
Shift+Enter now just sends, and the notice goes with the refusal.

## What ships

- `ComposeBox.tsx` — the Enter branch collapses to `preventDefault()` +
  `doSubmit()`. `SHIFT_ENTER_REFUSAL` deleted; the comment block that described
  the opposite behaviour rewritten.
- **`e.preventDefault()` stays.** It is what stops the textarea inserting its
  own line break, which would otherwise land in the box *after* the async
  `doSubmit()` cleared the draft. Dropping it is a second defect wearing the
  first one's clothes.
- Three specs inverted (two vitests + the `#816 shift-enter-speaks` e2e, renamed
  `issue974-shift-enter-sends.spec.ts`). The product intent moved by ruling, so
  the spec moves — nothing was weakened to reach green.
- `DESIGN_NOTES.md` — the two 2026-08-06 paragraphs corrected at the source, plus
  a 2026-08-07 entry for the reversal.
- Separate commit, explicitly **not** the fix: `autocapitalize="none"
  autocorrect="off"` on the composer.

The trigger that arms `shiftKey` on the reporter's device is **still unknown**
and deliberately not diagnosed. Both candidates are dead (vjt's real-iPhone
probe read `shiftKey` false in every box; the live prod bundle's minified gate
is identical to `origin/main`), and neither was implemented. Making the modifier
harmless closes the whole class without a diagnosis.

## Measured

- `scripts/bun.sh run test` — **244 files / 4530 tests green** (identical counts
  to the pre-change baseline run on this worktree at `origin/main`).
- `scripts/bun.sh run check` — **exit 0** (biome + tsc; warnings are the
  pre-existing CSS-specificity set).
- **Mutation, to prove the tests are not mirrors.** Deleting `e.preventDefault()`
  from the shipped handler turns both new vitests red — `defaultPrevented`
  false, and `setDraft` called once by the modelled UA line break. The
  "submits" assertion was independently measured red before the fix landed
  (`submit` called 0 times).

## Not measured — stated, not implied

- **The e2e has never been run.** The `integration` lane was denied for this
  branch (contended), so it is gated from this PR's CI. Nothing local proves it
  passes.
- **The e2e is not statically checked either.** `biome check src` does not cover
  `e2e/`, and `tsc -p e2e --noEmit` cannot run in the bun container (no
  `@playwright/test` types installed there). Its imports were verified by hand.
- The jsdom test measures `setDraft`, **not** the rendered textarea value:
  `getDraft` is a static mock, so the box reads empty either way. The
  real-browser proof of the newline is the e2e — which, per above, is unrun.


vjt:

A cherry-picked union rewrites the SHAs, so GitHub cannot close this PR itself. Landing verified BY CONTENT: `git log origin/main --grep '#974'` shows the commits on `main`. The work shipped; only this PR object is stale.

Issue #981, opened 2026-08-07
cic: unread badge flashes 1 on arrival while the pane is already at the bottom


vjt:
## Symptom

Reported by vjt (2026-08-07 09:45): with a channel selected and the pane **already at the bottom**, a new incoming message makes the unread badge appear as `1` and then disappear on its own. His ruling: **while the pane is at the bottom, the badge should not be shown at all.**

## Mechanism (read from the code, not guessed)

The badge is a pure derivation over `rows past the read cursor` — `perChannelUnread` in `cicchetto/src/lib/selection.ts:398-455` counts every row with `row.id > cursor` (minus the presence-hidden ones, #239, and minus your own content rows outside the self-window, #576). Nothing in that memo knows or cares whether the operator is looking at the window: **the focused-window suppression was deliberately deleted** (`selection.ts:459-478`, #887, 2026-06-02).

What replaced it is the read-at-the-tail arm in `cicchetto/src/ScrollbackPane.tsx:2481-2488`: while the tab is visible **and** the pane is geometrically at the tail, it advances the cursor over what is on screen. That arm is **debounced by `READ_AT_TAIL_SETTLE_MS = 500`** (`ScrollbackPane.tsx:172`).

So the sequence for an arrival while at the bottom is:

1. the row lands in the store → `row.id > cursor` → `perChannelUnread` recomputes → **badge shows 1** (synchronously, this is a memo);
2. 500 ms later the debounce fires `settleCursorToVisibleTail` (`ScrollbackPane.tsx:1021`), which reads `lastFullyVisibleRowId(listRef)` and calls `setCursorIfAdvances`;
3. `setReadCursor` advances the local cursor map **synchronously** before its POST (`readCursor.ts:218-222`, the optimistic advance) → the memo recomputes → **badge back to 0**.

That is the flash, and its duration is the 500 ms constant. The cursor round-trip is *not* involved — the optimistic local advance already collapses that.

Two consequences worth stating: the flash does **not** happen for your own messages (excluded by #576), which matches "arriva un msg"; and the 500 ms is not a layout requirement — the arm needs the new row *laid out* before `lastFullyVisibleRowId` can see it, which is a frame, not half a second. The debounce is there to coalesce arrival storms into one POST.

## Scope (vjt's ruling)

**No badge on a window the operator is at the bottom of.** Gate it on exactly the predicate the read-at-tail arm already uses — `isDocumentVisible() && atBottomNow()` — for the selected window.

[...]

PR #996, opened 2026-08-07
cic(#981): no unread badge on a window you are at the bottom of

vjt:
Fixes the badge that blinks `1` on a pane the operator is already at the bottom of (GH #981).

## The defect, read from the code

The badge is a pure derivation (`perChannelUnread`, `selection.ts`) counting rows past the read cursor. It knows nothing about where the operator is looking — the focused-window suppression was deliberately deleted by #887. What replaced it is the read-at-the-tail arm in `ScrollbackPane`, which advances the cursor over what is on screen but is **debounced at `READ_AT_TAIL_SETTLE_MS = 500`**. So: row lands → memo recomputes → badge `1` immediately; 500 ms later the arm advances the cursor (the local advance is synchronous) → badge `0`. **The blink window IS that constant**, not a network round-trip.

## The change

- **`lib/readingAtTail.ts`** — a shared signal carrying which window, if any, the pane is reading at its tail. One writer: the pane that owns the geometry; cleared on unmount.
- **`ScrollbackPane`** — the read-at-the-tail gate is now ONE expression (`isDocumentVisible() && atBottomNow() && rows > 0`), read by both the cursor arm and the publisher. Not forked: `atBottomNow` is the geometry, deliberately distinct from `followMode` (the intent), and the suppression must read the geometry exactly like the arm does.
- **`selection.ts`** — `perChannelUnread` zeroes the window the pane names.

### The extra term, and why it is not a second predicate

Every window switch re-arms `atBottomNow` **true as an intent default** (the 2026-06-01 scroll-contamination re-arm) before any layout is read; the real measurement lands in the activation's rAF×2. The cursor arm tolerates that — it re-reads the DOM 500 ms later and its write is forward-only — but a suppression acting on the default would blink the **arriving** window's badge to 0 and back within two frames, which is the "vanishes on select" complaint #887 was filed about. `tailGeometryMeasured` withholds the answer until a measurement lands. Every unmeasured path publishes `null`, so the failure direction is **show the badge**.

### Why this is allowed to reinstate part of #887 — and why that is written in the code

#887's objection was that suppression was DISHONEST: it hid the badge *without advancing the cursor*, so the count vanished on select and came back whole (1832) on leave. That objection no longer holds, because the cursor now advances while the operator reads. The paragraph saying so sits at the #887 note in `selection.ts`, where the next reader will look before deleting this again.

### Guards, one test each

- **A backgrounded tab keeps accruing** — visibility is folded into what the pane publishes; a hidden tab publishes `null`.
- **A far-behind window (#693) keeps its badge** — its cursor is frozen on purpose and the rows it sits at the bottom of are not the unread region.

## Tests

- `src/__tests__/unreadBadgeAtTail.test.ts` (new) — the memo: suppressed when reported, restored when not, only the reported window, far-behind stands.
- `src/__tests__/ScrollbackPane.test.tsx` — the publisher: visibility, empty window, unmount, and withhold-across-a-switch.
- `e2e/tests/issue981-no-badge-at-tail.spec.ts` (new) — the badge never appears, not even for an instant.

**On the e2e's non-vacuity.** A `toHaveCount(0)` across a 500 ms window passes against the broken build by sampling late. So the observation is CONTINUOUS (a MutationObserver armed before the send, recording every appearance) and outlives the debounce; the row-visible assertion proves something actually arrived; and a POSITIVE CONTROL — the same observer, same selector, made to record a real badge on a defocused window — proves the instrument was not blind.

## Measured / not measured

- Measured: full cic suite (245 files / 4540 tests) and `bun run check` green from the worktree root, clean tree before and after. Both new guards mutation-checked — deleting the far-behind term reddens its test, deleting the freshness term reddens the switch test.
- **Not measured**: the e2e has never been run (no lane; e2e also has no static gate — biome ignores `e2e/` and its `node_modules` are container-only). CI is its first execution.
- **Not measured**: nothing on a device. The 500 ms figure is the constant in the source, not a stopwatch reading — I make no claim about how long the blink looks.
- **Watch item for CI**: `issue887-focused-unread-badge.spec.ts` asserts an intermediate `afterReading > 0` after a 1200 px wheel down a ~120-row unread region. If that wheel now reaches the tail, suppression zeroes it instantly where the 500 ms debounce used to leave a window. That would be a genuine collision between #887's spec and #981's ruling, not a flake.

A DESIGN_NOTES entry is written and held out of this branch to avoid conflicting with the other open PRs; it is handed over at merge.

### DIFF
--- a/cicchetto/e2e/tests/issue981-no-badge-at-tail.spec.ts
+++ b/cicchetto/e2e/tests/issue981-no-badge-at-tail.spec.ts
@@ -1,0 +1,172 @@
+// #981 — no unread badge on a window whose tail the operator is watching.
+//
+// The report (vjt, 2026-08-07): a channel is selected, the pane is already at
+// the bottom, a message lands — the badge blinks `1` and clears itself. The
+// blink is the gap between the badge memo (synchronous, the row is past the
+// cursor the instant it lands) and the read-at-the-tail cursor arm (debounced
+// `READ_AT_TAIL_SETTLE_MS = 500`). The ruling is the wider one: while you are
+// at the bottom of a window, no badge at all.
+//
+// Why this cannot be a jsdom test: the suppression is gated on the pane's
+// GEOMETRY (`atBottomNow`), and jsdom's zero-height boxes read as at-the-tail
+// unconditionally — measured under #887, deleting that gate leaves the whole
+// ScrollbackPane suite green. jsdom can only pin the pieces either side of the
+// geometry; the geometry itself lives here.
+//
+// ── How this spec avoids being vacuous ──────────────────────────────────────
+// It asserts an ABSENCE across a window that is 500 ms wide, so two things
+// have to be true or it proves nothing:
+//
+//   1. The observation must be CONTINUOUS and must OUTLIVE the mechanism it
+//      denies. A `toHaveCount(0)` check — even a retrying one — samples, and
+//      every sample can legitimately land in a gap: it would pass against the
+//      broken build simply by arriving after the arm had cleared the badge.
+//      So a MutationObserver is armed BEFORE the message is sent and records
+//      every appearance of the badge for the whole run, and the wait after the
+//      row lands is longer than the debounce plus browser slop.
+//   2. The instrument must be provably able to SEE a badge. An empty
+//      recording is only meaningful if a non-empty one was reachable, so the
+//      same observer, with the same selector, is then made to record a real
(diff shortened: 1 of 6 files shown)


vjt:

A cherry-picked union rewrites the SHAs, so GitHub cannot close this PR itself. Landing verified BY CONTENT: `git log origin/main --grep '#981'` shows the commits on `main`. The work shipped; only this PR object is stale.

Issue #579, opened 2026-07-31
cic: /lusers discards its arguments — server accepts <mask> [<server>] since #571


vjt:
Follow-up to #571 (server side, merged in `1fabeaeb`).

## What

`Grappa.IRC.Client.send_lusers/3` now accepts the full RFC shape — bare, `<mask>`, and `<mask> <server>` — each token gated by `safe_oper_token?/1`, with `server`-without-`mask` rejected as `:invalid_line`. `Session.send_lusers/4` and the channel-side `validate_lusers_args/2` carry it end to end.

cic never sends those arguments. In `cicchetto/src/lib/slashCommands.ts`:

```ts
lusers: (_verb, _rest) => ({ kind: "lusers" }),
```

`_rest` is discarded, and the parsed command type is a bare `{ kind: "lusers" }` with no fields. So a user typing `/lusers *.azzurra.org` silently gets the unfiltered network-wide counts — the wrong answer with no error, exactly the failure mode #374 fixed for `/motd`.

## Scope

- Widen the parsed type to `{ kind: "lusers"; mask: string | null; server: string | null }`.
- Parse first/second token like `/whois` does (that command already splits `nick` + `server`); reject `server` without `mask` client-side too, mirroring the server's `:invalid_line`.
- Thread both through the send path to the existing wire verb.
- `LusersCard` renders the reply bundle and is keyed per network — a filtered snapshot replacing the unfiltered one is the existing `#248` solicited-gate behaviour and needs no redesign.

## Notes

Naming: the server settled on `server` (not `target`) to match `/whois` and the `:server` validator key — cic should use the same word.

The wire verb is unchanged and additive, so this is a pure client change; no `protocol_version` bump.

vjt:
## Correction to the premise: on bahamut a LUSERS mask is accepted and **ignored**

While implementing this I read the ircd that all of prod runs —
[`azzurra/bahamut`](https://github.com/azzurra/bahamut) `src/s_serv.c`, `master` at
`3b6ccff1263458314d2e5570aa58374434fadf91`. The issue body says a user typing
`/lusers *.azzurra.org` "silently gets the unfiltered network-wide counts — the wrong
answer with no error". The dropped-argument bug is real, but that sentence is not what
the server does, and the difference changes what #579 is worth.

### 1. `hunt_server` needs BOTH tokens (`src/s_serv.c:2081-2092`)

```c
int m_lusers(aClient *cptr, aClient *sptr, int parc, char *parv[])
{
[...]

Issue #571, opened 2026-07-31
grappa: /LUSERS discards client-supplied parameters (forced oper recount unreachable)


vjt:
The grappa server handler for `/LUSERS` accepts no parameters and emits a bare `LUSERS` upstream, silently dropping any mask/server args a client sends.

**Source (origin/main)**
- `lib/grappa_web/channels/grappa_channel.ex` — `handle_in("lusers", %{"network_id" => network_id}, socket)` pattern-matches only `network_id`; no `mask`/`server` args are read.
- `lib/grappa/session.ex` — `send_lusers/2` is documented as "Sends bare `LUSERS` upstream" — no parameters are forwarded.

**Why it matters**
Upstream bahamut gates the *live* recount of the "unknown connection(s)" figure behind `forced = IsAnOper(sptr) && parc > 3` in `send_lusers` (`s_serv.c`); otherwise the count is served from a 180s cache (`LUSERS_CACHE_TIME`). Because grappa strips the parameters, an operator on grappa can **never** trigger the forced live recount — they always receive the up-to-180s-stale cached value. `LUSERS <mask> <server>` also cannot be routed to a remote server.

**Repro**
Issue `/LUSERS <mask> <server>` from cic on a bahamut-backed network → the upstream connection receives a bare `LUSERS`, and an oper sees the cached figure rather than a fresh recount.

**Expected**
Forward client-supplied `LUSERS` parameters upstream (at minimum the 3-argument oper form that triggers the forced recount), rather than discarding them.

Reported on IRC during the SSL unknown-client leak investigation on the Azzurra testnet.

vjt:
Shipped to production in the 2026-08-01 cold deploy (prod now `0.8.0-e3cbe718`, `healthz` green). Announced on #grappa on both Azzurra and Libera.

PR #998, opened 2026-08-07
cic(#579): send /lusers' mask and target server instead of dropping them

vjt:
Refs #579. Client-only; the server has accepted the full RFC 2812 §3.4.2
shape since #571.

## What

`lusers: (_verb, _rest) => ({ kind: "lusers" })` discarded `rest`, so a
user's arguments never reached the wire. The parsed type is now
`{ kind: "lusers"; mask: string | null; server: string | null }`, split
positionally like `/stats`' `query` + `target`, and both tokens thread
through `pushLusers` to the existing `lusers` verb.

Naming follows the server's `server` (not `target`), per the issue.

## `server` without `mask`

That is the shape `validate_lusers_args/2` rejects as `:invalid_line`. A
positional parse cannot BUILD it, so there is no reachable input for a
client-side rejection to refuse — the mirror is an unconstructible state
rather than a runtime guard, and the guard would be dead code.

An invariant test pins the property instead, and it is not vacuous:
`/whois`'s two-arg form is `<server> <nick>` (server FIRST), so copying
that order here would make the one-token form yield exactly the mask-less
shape the server refuses. Measured — the test fails against that mutation.

## What the args actually do on bahamut

Read `azzurra/bahamut` `src/s_serv.c` while writing this: `m_lusers` calls
`hunt_server` only when BOTH tokens are present, and the answering server
reports its own local counts — the two-token routed form is the one with a
visible effect. `send_lusers` never `match()`es the mask, so a mask alone
is accepted and ignored there; its forced cache-bypass wants oper AND a
fourth parameter, which grappa's 2-slot path cannot produce. We still send
the mask: an RFC-honouring ircd may filter on it, and dropping user input
is the bug. The third commit corrects the earlier comments accordingly.

## Wire

Additive: the `lusers` verb is unchanged and a null arg omits its key
(pushMotd/pushLinks' shape), so a bare `/lusers` sends byte-identical
params to a pre-#579 client. No `protocol_version` bump.

## Gates

- `bun run check` (biome + tsc) — exit 0.
- `bun run test` — 244 files / 4538 tests, exit 0.
- Mutation-measured: the payload tests fail against a `pushLusers` that
  ignores its new args; the parser invariant fails against the `/whois`
  argument order.

## Not verified

No e2e. Per the source above, the single-token form is NOT observable on
bahamut by construction. The two-token form would be — the e2e testnet is
a real mesh (hub + leaf-v4 + leaf-v6) and `/lusers * hub.azzurra.chat`
should return the hub's local counts rather than leaf4's — but I could not
run the e2e lane in this pass, and I am not shipping a spec I never
executed. Flagged for the orchestrator rather than guessed at.

### DIFF
--- a/cicchetto/e2e/tests/issue579-lusers-routed-to-server.spec.ts
+++ b/cicchetto/e2e/tests/issue579-lusers-routed-to-server.spec.ts
@@ -1,0 +1,93 @@
+// #579 — `/lusers <mask> <server>` must actually route the query to the
+// named server.
+//
+// Bug: cic's parser was `lusers: (_verb, _rest) => ({ kind: "lusers" })`, so
+// both RFC 2812 §3.4.2 arguments were discarded and every /lusers reached the
+// wire as a bare LUSERS. The operator who named a server read the LOCAL
+// server's counts believing they had queried the one they typed — a wrong
+// answer with no error.
+//
+// Why this spec drives the TWO-token form and not `/lusers <mask>`:
+// bahamut's `send_lusers` never `match()`es the mask (azzurra/bahamut
+// src/s_serv.c:2100-2315 — `parv[1]` is never dereferenced and there is no
+// `match()` in the function), so a mask ALONE has no observable effect on
+// this ircd; a spec built on it would be green whether or not the fix is
+// present. Routing is the observable half: `m_lusers` calls `hunt_server`
+// only when BOTH tokens are present (src/s_serv.c:2085), and the server that
+// answers reports ITS OWN figures via RPL_LUSERME (`m_client` / `m_server`)
+// — the card's "this server" field.
+//
+// The assertion is a DIFFERENCE BETWEEN TWO ROUTED REPLIES, not a comparison
+// against the local one. The testnet is a real mesh (hub + leaf-v4 + leaf-v6
+// + services) and the bouncer's session lands on a leaf via the
+// `bahamut-test` alias — but which leaf is not pinned anywhere the client can
+// see, and an older spec (issue540) still describes the session as dialling
+// the hub. Comparing hub-vs-leaf4 sidesteps that entirely: whichever of them
+// happens to be the session's own server, the two are DIFFERENT servers, so
+// their per-server counts must differ (a hub carries the leaves' links and no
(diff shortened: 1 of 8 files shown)


vjt:
Follow-up on the two open threads from the first pass.

**1. bahamut correction is now on the issue, with citations** — #579's premise
that a dropped mask returns "the wrong answer" does not survive a read of
`azzurra/bahamut` `src/s_serv.c`: `send_lusers` never `match()`es `parv[1]`, so
a mask ALONE is accepted and ignored, and the routed two-token form is where
the real defect lives. Quotes and line numbers in
https://github.com/vjt/grappa-irc/issues/579#issuecomment-5216628826 so the
record carries the evidence, not a paraphrase. It is a source read, not an
observation of a live server — stated as such there.

**2. e2e added — WRITTEN, NOT RUN.** `cicchetto/e2e/tests/issue579-lusers-routed-to-server.spec.ts`
drives `/lusers * hub.azzurra.chat` then `/lusers * leaf4.azzurra.chat` and
asserts the card's per-server line DIFFERS between the two replies. It compares
two routed replies to each other rather than against the local one, because
which server the session sits on is not observable from the client and the tree
disagrees with itself about it (compose.yaml says clients always reach a leaf;
issue540's header still says the session dials the hub). Two named servers are
different servers either way, so their counts must differ — and pre-#579 both
commands reached the wire as the same bare LUSERS, making the two snapshots
identical. No spec on the single-token form: bahamut ignores a lone mask, so
that spec would be green with or without the fix.

A `data-testid` was added to the card's "this server" `<dd>` to anchor it (one
attribute; no layout, copy, or behaviour change).

**Gates, re-run after these commits:** `bun run check` exit 0, `bun run test`
244 files / 4538 tests exit 0. `tsc -p e2e --noEmit` clean for the new spec,
and I proved the file is actually in that program by making a deliberate type
error surface from it — the e2e tree has ~20 pre-existing errors of its own,
untouched.

**Still not verified:** the new spec has never been executed. The integration
lane was denied this pass (contended `_build`). It supersedes the "Not verified"
section of the PR body above — the spec exists, its result does not.

Issue #997, opened 2026-08-07
cic: the far-behind jump/dismiss sits at the top edge, out of thumb reach — the #280 stack is where users look


vjt:
Reported by a user on IRC after hitting the frozen `(199)` badge on a channel and not understanding why it would not fall.

The diagnosis was the intended one: the window was far behind, so the read cursor is deliberately frozen (#693) and the badge only falls once the bar is dismissed. That part works. The complaint is about **where the control lives**, quoted verbatim:

> it's a different UX from what I expected — usually the two little buttons for scrolling down or going to the next one are at the bottom right, within thumb reach; that one is at the top and I wasn't expecting it.

He was already at the tail and never looked up, so the badge read as a broken counter for as long as he used it — which is exactly the failure mode #888 set out to prevent by making the frozen state visible.

## What is where today

- The far-behind bar is **pinned to the top** of `.scrollback-pane`: `position: absolute; top: 0` (`cicchetto/src/themes/default.css:2838`), rendered at `ScrollbackPane.tsx:3464` (`data-testid="far-behind-bar"`). It holds two controls: `.scrollback-far-behind-jump` (go back to the first unread) and `.scrollback-far-behind-dismiss` (the `×`, which is what actually unfreezes the cursor). Both already carry the 44px tap floor in absolute px.
- The **floating action stack** (#280) is anchored at the **lower right** of the same pane: the scroll-to-bottom button (C7.4) and, on mobile, the jump-to-next-active-window affordance (#235). It is container-anchored on purpose, so it holds its position regardless of soft-keyboard state, and its box is `pointer-events: none` with taps falling through.

So the client already has a thumb-reach corner for scrollback navigation, and the far-behind controls are not in it.

## Why the original reasoning does not settle this

The comment above `.scrollback-far-behind` justifies pinning it *out of the scroll flow*: a tail-anchored pane opens at the bottom, so an **in-flow** row at the boundary would sit viewports above the fold and never be seen. That argument is sound and is an argument against an in-flow marker — it is **not** an argument for the top edge specifically. The lower-right stack is also out of flow, also container-anchored, and is where the user's thumb and attention already are.

## Not a duplicate of the scroll-to-bottom button

Worth stating so the scope does not collapse: jump-to-first-unread goes **backwards**, scroll-to-bottom goes **forwards**. They are different actions and both must survive. The overlap is only in *placement*.

## Scope to decide (not decided here)

[...]

Issue #289, opened 2026-07-17
cicchetto: mobile floating buttons (next-unread, scroll-to-bottom) obscure message text — add opacity


vjt:
On mobile, the floating **next-unread** and **scroll-to-bottom** buttons are fully opaque and sit on top of the message list, covering the text underneath them. When a message wraps behind a button the content becomes unreadable.

**Request:** give both floating buttons some transparency so the text behind stays legible — e.g. a reduced/translucent opacity on the button (and/or its background), tuned so the control is still clearly tappable but the message text shows through.

**Context:** fast-follow to #280 (which fixed the two buttons overlapping each other and aligned/sized them). This is the next polish pass on the same pair: now that they coexist cleanly, they should not hide message content.

Scope: cicchetto frontend, mobile floating-button styling only. No behaviour change.

vjt:
Field report from PWA iOS (new bundle, kill+relaunch confirmed) — the scroll-to-bottom floating control is not only obscuring text via overlap, it is **oversized** on the narrow iOS viewport and **clips the right edge of the last message lines** (text like "oscu[ra]", "decisa[mente]" is cut off under the button).

Implication: the queued "add opacity" fix is necessary but **not sufficient** — opacity alone still leaves the button covering/clipping glyphs. Needs a size reduction and/or reposition (and to sit outside the text column, or the message area to reserve gutter width) on narrow viewports.

Screenshot (iPhone, 1206×2622): https://sindro.me/static/~vjt/grappa-289-overflow-ios.png

Reporter: chan request on #grappa.

vjt:
Merged to main (28f8f152, PR #295) with GH ci + integration green. Deploy **HELD** for the batched COLD ship (cic bundle rebuild via `--cic` — pure CSS + e2e, no BEAM restart). Reopened + `status:soon` to track until the deploy window; close after deploy.

vjt:
Shipped to prod grappa.chat in the 4am batch COLD deploy (origin/main 2fa3d749): migrate (themes table + system_user) + cic bundle rebuild + BEAM restart + theme gallery seeded (13 built-ins). Public 200, dual-net announce posted.

Issue #280, opened 2026-07-17
cicchetto: "next" and scroll-to-bottom buttons must coexist cleanly (positioning, badges, sizing)


vjt:
Follow-up refinement to the mobile "next unread window" button (shipped in #264 / #278). With the on-screen keyboard open the placement is still wrong — the "next" button now overlaps the scroll-to-bottom ("go to bottom") button. Four items:

**1. "next" button placement (keyboard open)**
- Currently overlaps the scroll-to-bottom button when the keyboard is up.
- Move it up so it clears the scroll-to-bottom button.
- Its position **relative to the message container** must be constant whether the keyboard is open or closed (no jump on keyboard show/hide). Related keyboard-layout issue: #209.

**2. Coexistence of "next" + scroll-to-bottom**
- Both buttons must be able to show at the same time, well aligned and evenly distributed (a proper stacked/spaced pair, not overlapping).

**3. "next" badge color semantics**
- Show a **red** numeric badge when the next target is a **mention or a private message**.
- Show a **blue** numeric badge when it is a **regular channel message**.
- Counter source should be the server-side mention/activity counter — see #267.

**4. scroll-to-bottom button sizing**
- Currently much smaller than the "next" button. Make it the **same size** as "next" so the pair looks consistent.

Acceptance: keyboard open or closed, both buttons render in a fixed, non-overlapping, aligned position relative to the message container; "next" carries a red badge for mention/PM and blue for a normal message; both buttons are the same size.

vjt:
Shipped in c032488a (feature tip fe7f37ee) — cic-only, deployed `--cic` to both hosts (bundle `index-VE-juJ1n.js`, HTTP 200 on irc.sniffo.org + irc.sindro.me).

## Root cause (two independent anchors)
The two floating buttons were owned by different components and anchored to different reference frames: `scroll-to-bottom` was `position: absolute` inside `.scrollback-pane` (the message container, `bottom: 0.75rem`, 2rem square), while the mobile "next" circle was `position: fixed` to the visual viewport (`top: viewport-height − 3.5rem − --nab-lift`, 3.5rem) with a discrete `--nab-lift: 8rem` keyboard-open bump (`:has(textarea:focus)`). Keyboard-open shoved the viewport-fixed circle up into scroll-to-bottom's band on the shared right edge → overlap; the e2e even measured overlap keyboard-CLOSED (49×49 @ y498–547 vs 28×28 @ y528–556).

## Fix
[...]

PR #1002, opened 2026-08-07
cic(#997): put the far-behind dismiss where the thumb is

vjt:
Refs #997.

An operator on IRC sat under a frozen `(199)` badge and read it as a broken
counter. Everything worked as designed — the window was far behind, so #693
freezes the read cursor on purpose and #888 marks the badge as a DISTANCE
rather than a count — but the dismiss is the ONLY gesture that unfreezes that
cursor, and it lived pinned to the top edge of the pane. They were at the tail
and never looked up.

The comment above `.scrollback-far-behind` justifies pinning the bar OUT of
the scroll flow, which is an argument against an in-flow marker; it says
nothing in favour of the top edge. The #280 float stack is also out of flow,
also container-anchored, and is where the thumb already is.

## Scope — option 2, as decided

The bar stays as the status LABEL ("you are N behind"); the affordance also
lands in the corner stack. The bar is not moved (option 3) and the label is
not dropped (option 1).

## The three constraints

- **One owner, one state.** Both controls call one `dismissFarBehindGesture`;
  the corner button has no handler of its own. Two handlers that agree today
  are the #975 twin-structure shape.
- **The jump gets NO corner twin.** It goes backwards and does not unfreeze
  anything, so a reachable copy would leave the badge exactly as stuck. A
  backwards control one gap above the forwards scroll-to-bottom is also the
  collision #280 anchored this stack to prevent.
- **44px tap floor in absolute px** (root font-size is 14px), which makes the
  desktop control larger than its 2rem sibling — the physical constraint wins
  over the visual match.
- **Opaque**, opting out of the `#289/#302` mobile translucency its stack
  siblings carry: those are permanent chrome, this appears only while the
  state does and is the only way out of it.

## What was measured

- `scripts/bun.sh run check` — exit 0 (biome + tsc).
- `scripts/bun.sh run test` — 4535 passed / 244 files, on the committed tree.
- `tsc -p e2e/tsconfig.json --noEmit` — the new spec contributes none of the
  26 pre-existing errors, and emitted no `.js`.
- **Mutation, both legs red.** Corner control dismissing without re-latching
  the frozen marker → 1 test red. Corner control calling `jumpToUnread`
  instead → 2 tests red. Reverted, green again.

## What is NOT measured

The e2e (`issue997-far-behind-thumb-reach.spec.ts`) has **never been run** —
no integration lane, and there is no local browser here. Its geometry
assertions (the 44px box, the lower-right quadrant) and the badge-falls
outcome are gated by this PR's CI, not by anything I observed.

### DIFF
--- a/cicchetto/e2e/tests/issue997-far-behind-thumb-reach.spec.ts
+++ b/cicchetto/e2e/tests/issue997-far-behind-thumb-reach.spec.ts
@@ -1,0 +1,165 @@
+// #997 — the far-behind dismiss was out of thumb reach.
+//
+// Reported from IRC. An operator sat at the tail of a channel under a frozen
+// `(199)` badge and could not make it fall. The diagnosis was the intended
+// one — the window is far behind, so #693 freezes the read cursor on purpose
+// and #888 marks the badge as a DISTANCE rather than a counter — but the
+// dismiss is the only gesture that unfreezes that cursor, and it lived pinned
+// to the TOP edge of the pane. They never looked up, so for the whole session
+// the badge read as a broken counter.
+//
+// The bar keeps the label; the gesture now also sits in the #280 float stack
+// at the lower right, where the thumb and the attention already are. This
+// spec pins the OUTCOME the operator could not reach: with the window far
+// behind, a tap on the corner affordance drops the badge.
+//
+// Three things are measured here rather than asserted by class name:
+//   * the control's box is at least 44×44 CSS px (the tap floor; the root
+//     font-size is 14px, so a rem-based target silently comes out short),
+//   * its centre sits in the lower-right quadrant of the scrollback pane —
+//     the placement claim itself, and the thing that was wrong,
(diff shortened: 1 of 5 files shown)


vjt:
Superseded by the union PR #1007, merged to main as 5bb3bcea.

The union cherry-picked this branch's commits, so the SHAs were rewritten and GitHub cannot mark this PR merged by ancestry. Landing verified BY CONTENT instead: every line this branch adds is present on main. (For #1002 specifically, the eight `.shell-chrome` band lines main lacks are pre-existing lines this PR never touched — #985/#988 deliberately delete that band.)

Union gate at 5bb3bcea: ci green, integration 638 passed / 0 failed. The three branches were gated together rather than merged one by one, because #1002 and #1004 both touch cicchetto/src/themes/default.css and a per-PR green attests branch+base, never branch+the other branches landing beside it.

Issue #473, opened 2026-07-27
cicchetto: ONE labelled button drawer at the bottom of the members rail — same buttons on desktop and mobile (post-#71 cleanup)


vjt:
Ruling by @vjt (per chan request) after driving the post-#71 build on a real box: the button surface came out of the sidebar/rail rework **split in two**, and it should be **one**.

> "il reworking della side bar e dei bottoni ha fatto casino. adesso ci sono DUE punti bottone […] io voglio un solo cassetto per i bottoni, bottom members list. questo cassetto deve contenere tutti i bottoni che abbiamo, sia su desktop che su mobile. in modo che sia consistente."

## What is on `main` today

Two separate button groups, with different contents and different visibility:

| surface | where | contents | visible on |
| --- | --- | --- | --- |
| `ActionCluster` | **top** of `.shell-members` (`Shell.tsx:732` desktop, `Shell.tsx:964` mobile) | settings cog, presence/denoise toggle (channel-gated) | desktop **and** mobile |
| `.mobile-panel-actions` footer | **bottom** of the mobile members drawer (`Shell.tsx` ~1049 ff.) | home, rooms (`/list`), themes, admin, archive | **mobile only** |

Consequences, both visible on the staging box:

1. On mobile the user faces two button clusters at opposite ends of the same drawer, with no rule explaining which button lives where.
2. On desktop the second group **does not exist at all** — home / rooms / themes / archive / admin are simply not reachable from the rail. The desktop rail is a cog and a monkey.

The split is a straight consequence of how #71 INC-2 landed: the footer cog was deduped *up* into the `ActionCluster` (`Shell.tsx` comment: *"the footer settings cog was DEDUPED into the ActionCluster at the TOP of this drawer"*) while the window-nav launchers stayed *down*, and the desktop rail only ever got the cluster.

## Target

**One** button drawer, at the **bottom of the members rail**, carrying **every** button, **identical on desktop and mobile**:

`home` · `rooms` · `themes` · `archive` · `settings` · `admin` · `denoise`

- The top `ActionCluster` goes away as a separate surface; the cog and the presence/denoise toggle move into the single bottom drawer.
- The mobile-only gates on the launchers are dropped — desktop gets the same set. Per-button gating that is about **capability** rather than form factor stays: `admin` remains `isAdmin()`-gated, `denoise` remains channel-gated, `archive` keeps its existing selection gate.
- One component, one place, mounted by both branches of the `isMobile()` split in `Shell.tsx`.

[...]

PR #1004, opened 2026-08-07
cic(#985/#988): float the lone ☰, and measure #988 before patching it

vjt:
Two mobile-chrome P0s on one branch, cic-only. Built on `origin/main` after
#986 landed (union PR #999) — nothing from #986 is re-implemented here.

## #985 — the band a query window spent on one glyph

`.shell-chrome` stopped being a bar three issues ago: #473 took the archive
button out and #986 took `@` mentions into the rail. What was left was a
full-width `<header>` wrapped around ONE right-aligned ☰, still charging
`var(--chrome-tap-min) + 1rem + 1px` for it on every non-channel mobile
window, with the price SCALING with the operator's text size. The
mobile-CHANNEL branch had already reclaimed those pixels (its ☰ rides inline
in the TopicBar); the query case never got a host.

The row is now zero-height and its single child overflows over the pane's
top-right corner, so the scrollback starts at the top of `.shell-main`
exactly as it does on a channel window.

Verified by reading the new tree, not assumed: `ShellChrome.tsx` renders the
opener and nothing else — #986 already moved `@` into `RailActions`
(`rail-action-mentions`). So the float carries one glyph, and the "ONE
drawer, one ☰" mutex (`toggleMembersPanel`) is untouched.

Three decisions worth the review time:

* **The containing block is `.shell-chrome` itself, not `.shell-main`.**
  Positioning the pane would silently re-anchor every `position: absolute`
  descendant that today resolves past it — blast radius bought for nothing,
  since a zero-height positioned row anchors the opener just as well.
* **No `env(safe-area-inset-top)` at the float — the issue's shape-of-the-fix
  is overruled here.** `.shell-mobile` already carries the inset (UX-3 BIS)
  and `.shell-main` sits inside that padding box, so the pane's top edge is
  ALREADY below the island. Re-applying it would double-count — the #913 trap
  pointed the other way. Whether an inset is owed depends on WHICH BOX you
  anchor to: `.rail-actions-menu` does owe it (#913) because it measures an
  absolute viewport coordinate; this float does not.
* **The glyph brings its own opaque backing.** `.shell-chrome-btn` is
  `background: transparent`, free while a band sat behind it; floating puts
  it over arbitrary scrollback text and over a themed background image.

**Overlap, declared** (issue constraint 2): the ☰'s own 48px box overlaps the
top-right corner of the scrollback, so a line ending exactly there sits
behind it. The hit area is the button and nothing wider — no full-width strip
— and the alternative was the band.

## #988 — the premise is UNCONFIRMED and this branch does not patch it

Read this section as a report, not as a fix.

**Reading 2 is ARGUED, NOT FALSIFIED. The number is missing.** #988 asks the
implementer to open the menu on a short viewport and read `scrollTop` on
`.rail-actions-menu` before writing anything. That reading was never taken:
no browser is runnable on the authoring host, and the stack lane was not
held. What exists here is the argument that nothing in `RailActions` scrolls
a freshly-mounted `overflow-y: auto` container, so 0 is expected — an
argument, which is exactly the kind of claim #988 says to settle by
measurement. Until this spec runs, "reading 2 is dead" is an unproven
statement and should not be quoted as a result.

The rest of #988's premise is unconfirmed the same way, and the issue says so
itself: the screenshot never arrived and the geometry was read off the code.

**Reading 1 gets no code change, and that is a finding.** When the content
exceeds the cap, `max-height` is `spaceAbove - inset`, so the bottom-anchored
box already runs from `inset + RAIL_MENU_TOP_GAP` down to the launcher — the
identical rectangle a top-anchored menu with the same cap would occupy. The
two anchorings differ only when the menu FITS, and #988 explicitly keeps
`bottom: 100%` there. A re-anchor would be green, shipped, and change
nothing. This is STRUCTURE read off the code, not magnitude measured — and it
is offered as a reason not to write a no-op, not as proof the menu behaves.

So `issue988-rail-menu-first-row.spec.ts` measures the actions menu at four
viewport heights plus one XXL pass and reports `scrollTop`,
`scrollHeight/clientHeight`, the menu box, the first row box, the published
`--rail-menu-space-above` and the inset — then asserts the criterion vjt
stated: on open the first row is visible without scrolling. Nothing in the
#588 cap or the #913 subtraction is touched.

The readings leave the run by two exits, both verified against a real
Playwright run on this host: a NODE-side `console.log` (a passing test's
stdout does reach the `list` reporter) and a `testInfo.attach`. They are
emitted from a `finally` BEFORE any assertion runs, so a mid-sweep failure
still prints every height already collected — proven with a deliberately
failing probe. The numbers are wanted most on the red run.

**If CI comes back green, the premise is unconfirmed and the guard is the
right outcome for #988.** If it comes back red, the printed numbers name the
mechanism and the fix follows them.

## Gates

Run locally, from the worktree root:

* `scripts/bun.sh run test` — **246 files / 4572 tests green** (full suite).
* `scripts/bun.sh run check` — biome + tsc, **rc=0**.
* `tsc -p e2e --noEmit` — both new specs clean (the tree carries 26
  pre-existing errors; neither new file is among them).
* **The #985 opacity guard was proven IN CI by mutation.** No browser runs on
  this host, so the mutation was pushed instead: commit `8704ddc5` set the
  opener's backing to `rgba(255, 255, 255, 0.5)`, and run
  [31186523383](https://github.com/vjt/grappa-irc/actions/runs/31186523383)
  came back **1 failed / 636 passed** — the single red being this spec, at the
  opacity assertion, `Expected: 1 / Received: 0.5` off a computed
  `rgba(255, 255, 255, 0.5)`. The guard rejects a translucent backing, and it
  rejects it *for the alpha*, not for a colour channel. The mutation was
  reverted in `6fd4c8ba`; the tree is byte-identical to the pre-mutation
  commit. **This assertion is not a mirror.**
* **Mutation, 8/8**: every new assert in `shellChromeFloat.test.ts` was
  killed by a targeted mutation of the rule it guards (re-add the band
  padding; drop `height: 0`; drop `position: relative`; z-index to 40; z-index
  to 90; drop the backing; re-apply the inset; resurrect the spacer). Baseline
  green after each revert. No mirrors.

## NOT run, and not claimed

* **The e2e GEOMETRY assertions still have no mutation proof.** Both specs now
  execute in CI and pass, but "passes" is not "would catch the defect". Only
  the #985 opacity assertion has been shown to fail on a real defect (see
  Gates). The pixel assertions — pane top, zero-height chrome row, tap floor,
  corner containment — have never been seen red, so they are unverified as
  discriminators.
* **The first version of the #985 opacity assertion was itself the bug.** It
  scraped alpha with `/rgba?\([^)]*?(?:,\s*([\d.]+))?\)$/`, whose lazy
  `[^)]*?` handed the BLUE channel of an opaque `rgb(255, 255, 255)` to the
  alpha group and failed a correct product. Recorded because it is the same
  class of error the geometry asserts are still exposed to: an assertion that
  has only ever been seen green tells you nothing about why it is green.
* **`scrollTop` was never measured.** See above: reading 2 is argued only.
* **The notch is not covered by any gate.** Playwright synthesizes no
  `env(safe-area-inset-*)` in either project, so both the #985 island
  clearance and #988's inset arithmetic read 0px. Device verification on a
  real iPhone is owed after ship.

### DIFF
--- a/cicchetto/e2e/tests/issue985-mobile-floating-opener.spec.ts
+++ b/cicchetto/e2e/tests/issue985-mobile-floating-opener.spec.ts
@@ -65,20 +65,33 @@ test("@webkit #985 — a mobile query window spends no band on the ☰, and the 
     // absent and every assertion below is green for the wrong reason.
     await expect(page.locator(".topic-bar")).toHaveCount(0);
     const chrome = page.getByTestId("shell-chrome");
-    await expect(chrome).toBeVisible({ timeout: 10_000 });
+    const opener = page.getByTestId("shell-chrome-rail-opener");
+    // The host is MOUNTED, not VISIBLE — and this spec is the reason. #985
+    // collapses `.shell-chrome` to `height: 0` and lets its one child overflow,
+    // so the element Playwright is asked about has an empty bounding box and is
+    // `hidden` by definition. Asserting it visible here contradicted this same
+    // test's own outcome assertion ("the chrome row must cost zero height")
+    // twenty lines down: both could never hold at once. What the precondition
+    // actually needs to establish is that the non-channel branch mounted its
+    // door — so: the host exists, and the door itself is on screen.
+    await expect(chrome).toHaveCount(1);
+    await expect(opener).toBeVisible({ timeout: 10_000 });
 
     const main = page.locator(".shell-main");
     const pane = page.locator(".scrollback-pane");
-    const opener = page.getByTestId("shell-chrome-rail-opener");
     await expect(pane).toBeVisible();
 
     const mainBox = await main.boundingBox();
     const paneBox = await pane.boundingBox();
-    const chromeBox = await chrome.boundingBox();
     const openerBox = await opener.boundingBox();
-    if (!mainBox || !paneBox || !chromeBox || !openerBox) {
+    if (!mainBox || !paneBox || !openerBox) {
       throw new Error("#985 — a measured element has no bounding box");
     }
+    // Read through the DOM, not through `boundingBox()`: the latter is defined
+    // to return null for an element Playwright considers invisible, and a
+    // zero-height box is exactly that — so the very property under test here
+    // would have come back as "no bounding box" and thrown above.
+    const chromeHeight = await chrome.evaluate((el) => el.getBoundingClientRect().height);
 
     // THE OUTCOME. The conversation starts at the top of the pane. This is the
     // pixel vjt asked for back, and it is the one number that was wrong before:
@@ -89,7 +102,7 @@ test("@webkit #985 — a mobile query window spends no band on the ☰, and the 
     ).toBeLessThanOrEqual(1);
     // The mechanism, measured rather than read: the row still exists (the
     // window needs its door) and costs nothing.
-    expect(chromeBox.height, "#985 — the chrome row must cost zero height").toBeLessThanOrEqual(1);
+    expect(chromeHeight, "#985 — the chrome row must cost zero height").toBeLessThanOrEqual(1);
 
     // The float must not have shrunk the affordance on its way out of the
     // flow: still the 48px HIG floor `--chrome-tap-min` guarantees.
--- a/cicchetto/e2e/tests/issue985-mobile-floating-opener.spec.ts
+++ b/cicchetto/e2e/tests/issue985-mobile-floating-opener.spec.ts
@@ -1,0 +1,137 @@
+// @webkit — #985. On a mobile QUERY window the ☰ must float over the pane's
+// top-right corner instead of sitting in a band that costs the conversation a
+// row.
+//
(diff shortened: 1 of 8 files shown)


vjt:
Superseded by the union PR #1007, merged to main as 5bb3bcea.

The union cherry-picked this branch's commits, so the SHAs were rewritten and GitHub cannot mark this PR merged by ancestry. Landing verified BY CONTENT instead: every line this branch adds is present on main. (For #1002 specifically, the eight `.shell-chrome` band lines main lacks are pre-existing lines this PR never touched — #985/#988 deliberately delete that band.)

Union gate at 5bb3bcea: ci green, integration 638 passed / 0 failed. The three branches were gated together rather than merged one by one, because #1002 and #1004 both touch cicchetto/src/themes/default.css and a per-PR green attests branch+base, never branch+the other branches landing beside it.

Issue #1008, opened 2026-08-07
Composer: predictive text (QuickType) is gone on iOS — autocorrect="off" should not have been copied from the identifier inputs


vjt:
## Symptom

On iOS/WebKit the composer no longer offers the predictive-text bar (QuickType) and no
longer autocorrects. Reported from `#sniffo` by a user typing on a phone: *"non riesco a
scrivere"* — with fat fingers and no autocorrect the box is effectively unusable for
prose.

## Cause (measured, not guessed)

`cicchetto/src/ComposeBox.tsx:544-545` on `origin/main`:

```jsx
autocapitalize="none"
autocorrect="off"
```

Introduced by `6a12346e` (*"fix(cic): stop WebKit auto-capitalising the composer — not the
#974 fix"*, 2026-08-07 12:51). Shipped in **v0.13.3** only — this is a same-day regression,
nothing older is affected.

The commit is correct about the problem it set out to solve: with no attribute WebKit
applies `autocapitalize=sentences`, and this box eats nicks, channel names and `/commands`,
so the first character kept getting upper-cased. But the change set the whole trio-style
pair copied from the *identifier* inputs (`PerformSettings`, `Login`), and on WebKit
`autocorrect="off"` also suppresses the QuickType predictive bar — not just the silent
autocorrection. Identifier fields want that. A **message body is prose** and does not.

Note the source comment already draws exactly this line for `spellcheck`:

> `spellcheck` is left alone on purpose — message bodies are prose.

`autocorrect` belongs on the same side of that line.

## Proposed fix

Keep `autocapitalize="none"` (that one is right, and it is what actually fixed the
upper-cased nicks). Drop `autocorrect="off"` from the composer only:

```diff
           autocapitalize="none"
-          autocorrect="off"
```

Leave both attributes as they are on `PerformSettings`, `Login`, `AliasSettings` — those
are identifier inputs and genuinely want autocorrection off.

## Trade-off, stated explicitly

With autocorrection back on, WebKit will occasionally rewrite an unusual word mid-message —
including a bare nick typed without tab-completion. That is the cost of getting the
predictive bar back. The reporter's position is that being unable to type at all is worse
than the odd mangled nick; whoever rules on this should decide knowingly rather than by
omission.

If we want both, the only real answer is a richer composer (contenteditable with the
mention/command spans marked non-correctable), which is a much larger piece of work and
should be its own issue, not a rider on this one.

## Acceptance

[...]

PR #1009, opened 2026-08-07
fix(#1008): composer keeps its typing aids — drop autocorrect=off, keep autocapitalize

vjt:
Same-day regression of **v0.13.3**. One attribute removed from one element.

## What changed

`6a12346e` (today 12:51) put `autocapitalize="none"` and `autocorrect="off"`
together on the composer textarea. This keeps the first and drops the second.

`autocapitalize="none"` is right and stays — it is what actually fixed the
upper-cased nicks, and with no attribute WebKit applies `sentences` to a box
that holds nicks, channel names and `/commands`. `autocorrect="off"` was
copied wholesale from the **identifier** inputs, and on WebKit it suppresses
the QuickType predictive bar rather than merely the silent correction. A
reporter on `#sniffo` could no longer type on a phone.

**The root cause is not the attribute — it is copying a pattern across a
context boundary.** Those fields hold identifiers; this one holds prose. The
source comment already drew that line for `spellcheck` ("message bodies are
prose") and `autocorrect` had been placed on the wrong side of it. The comment
now says so explicitly, at the site, so the pair does not get re-copied.

## Scope, verified by grep rather than by memory

Only `ComposeBox.tsx` changed. `autocorrect="off"` is untouched on every
identifier input — PerformSettings (3), Login (3), AliasSettings (4),
ServiceModal (1), SettingsDrawer (3), WatchlistsSettings (2),
RegistrationWizardModal (3).

## The trade-off, ruled on knowingly

With autocorrection back, WebKit may occasionally rewrite an unusual word —
a bare nick typed without tab-completion among them. Per the issue: being
unable to type at all is worse than the odd mangled nick. Getting **both**
needs a contenteditable composer with mention/command spans marked
non-correctable; that is its own issue, not a rider on this one.

## Test

The composer had **no attribute guard at all** — the regressing commit touched
no test — so this adds the first one. It asserts `autocapitalize` present and
`none`, `autocorrect` **absent**, and `spellcheck` still absent. The
identifier half of the contrast stays pinned where it already lives, in
`Login.test.tsx`'s "keeps autocapitalize/autocorrect/spellcheck/autocomplete
on the nick field" — not duplicated here.

**Mutation, 3/3, every assert killed:**

| mutation on the source | result |
|---|---|
| re-add `autocorrect="off"` (the v0.13.3 state) | 1 failed / 86 passed |
| delete `autocapitalize="none"` | 1 failed / 86 passed |
| add `spellcheck={false}` | 1 failed / 86 passed |

Each named this test; baseline green (87/87) after every revert.

## Gates

* `scripts/bun.sh run test` — **245 files / 4575 tests green**. A measured
  `origin/main` baseline is 245 / 4574, so this adds exactly the one test.
* `scripts/bun.sh run check` — biome + tsc, **rc=0**, 48 warnings, 0 errors.

## 🔴 Not proven here, and it is the actual acceptance criterion

**No test in this repo can witness the QuickType bar.** jsdom has no text-input
UI, and Playwright's webkit is not iOS — neither synthesizes the predictive
bar. What is asserted is the DOM attribute that governs it, and nothing more.

**"The predictive bar is present while typing in the composer" remains a
device check on a real iPhone, owed to vjt by hand.** A green tick on this PR
must not be read as that acceptance criterion having been met. Likewise the
second criterion — that a first-token `nickname` or `/command` still is not
upper-cased — is argued from `autocapitalize="none"` being retained, not
measured on a device.

### DIFF
--- a/cicchetto/src/ComposeBox.tsx
+++ b/cicchetto/src/ComposeBox.tsx
@@ -538,11 +538,25 @@ const ComposeBox: Component<Props> = (props) => {
           // mechanism behind it (the iOS auto-capitalisation hypothesis was
           // falsified on a real device). Correct on its own merits: with no
           // attribute WebKit applies `sentences`, and this box eats nicks,
-          // channel names and `/commands`. Same trio spelling as the other
-          // identifier inputs (PerformSettings, Login). `spellcheck` is left
-          // alone on purpose — message bodies are prose.
+          // channel names and `/commands`.
+          //
+          // #1008 — and it stops there. `autocorrect="off"` was copied in
+          // alongside it from the IDENTIFIER inputs (PerformSettings, Login,
+          // AliasSettings) and shipped in v0.13.3; on WebKit that attribute
+          // also suppresses the QuickType predictive bar, not merely the
+          // silent correction, and a reporter on #sniffo could no longer type
+          // on a phone. **This box is prose, and prose keeps its typing
+          // aids** — the same line `spellcheck` already sits on, which is why
+          // `spellcheck` is likewise left alone. Do not re-add `autocorrect`
+          // here by copying the pair from an identifier field: those fields
+          // want it, a message body does not.
+          //
+          // Accepted cost, ruled on knowingly in #1008: autocorrection may
+          // occasionally rewrite an unusual word — a bare nick typed without
+          // tab-completion among them. Being unable to type at all is worse.
+          // Having BOTH needs a contenteditable composer with mention spans
+          // marked non-correctable; that is its own issue, not a rider here.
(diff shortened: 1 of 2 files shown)


vjt:
Superseded by the union PR #1012, which carried this branch's commits verbatim and landed on main as b1ff00e2 (rebased onto ea213c8f; patch-set proved identical, 898 lines both sides, before and after the rebase).

A cherry-picked/rebased union rewrites the SHAs, so GitHub cannot mark this PR merged by ancestry — closing it by CONTENT, not because the work was dropped. The commits are on main.

Issue #982, opened 2026-08-07
Admin-side share-token mint, so a locked-out visitor can be let back in


vjt:
## Problem

A visitor who loses access to their session has no way back in. Visitors have no
password — the browser's session *is* the identity — so if the device dies, the
profile is wiped, or the cookie goes away, the account is unreachable and the
operator's only tool today is `DELETE /admin/visitors/:id`, which throws the
session away rather than returning it.

The recovery mechanism already exists and works: `POST /me/share-token`
(`GrappaWeb.ShareTokenController.mint/2`) mints a Phoenix-signed token bound to
the visitor id, and `POST /auth/share/consume` redeems it, unauthenticated and
one-shot, into a fresh `accounts_sessions` row for the same visitor. The gap is
who can press the button: mint is visitor-only, and it authenticates the caller
as the visitor — precisely what a locked-out visitor cannot do.

## Ask

An admin-side mint, so an operator can generate the link and hand it to the
person who lost access. Requested by the project owner, who has explicitly
accepted that the capability is abusable by an admin and considers it worth it
because there is no alternative recovery path for a passwordless identity.

## Proposed shape

`POST /admin/visitors/:id/share-token`, in `GrappaWeb.Admin.VisitorsController`,
under the existing `:api + :authn + :admin_authn` scope that already carries
`GET /admin/visitors` and `DELETE /admin/visitors/:id` (`router.ex:118-119`).

It should mint **the same token the visitor-side endpoint mints** — same salt
(`"visitor-share-v1"`), same `Phoenix.Token.sign(Endpoint, salt, visitor_id)`
payload, same 600s TTL, same one-shot ETS ledger in
`Grappa.Visitors.ShareTokens`. No second redeem surface: `/auth/share/consume`
must keep working unchanged for both origins. The salt and TTL are already
exposed as `ShareTokenController.salt/0` and `max_age_seconds/0`, so the
signing logic is worth extracting into one function both controllers call
rather than duplicating the `sign` call — one feature, one code path, every
door, per this controller's own house rule.

Response mirrors the visitor-side one: `{token, expires_at}`. Wrapping it into
`https://<host>/#/share/<token>` stays the client's job, as it is today.

## The one thing not to get wrong

**Incognito visitors must still be refused.** `mint/2` returns 403 for
`%Visitor{incognito: true}` on purpose (#363): an incognito session is
deliberately non-portable, and cic hides the control as the client-side twin of
that gate. An admin door that skips the check would silently undo a product
[...]

Issue #363, opened 2026-07-20
cicchetto: incognito mode — auto delete-account on browser close, server-side 1h linger fallback


vjt:
## Summary

A per-session **"incognito mode"**: when active, closing the browser deletes the account/session and all its data. Because the browser's `onBeforeUnload`/`beforeunload` is unreliable (mobile Safari, force-kill, tab crash often skip it), the **server enforces a 1-hour linger** — if the client doesn't reconnect within 1h, the account and data are deleted server-side regardless.

Requested by gpiero (via vjt, #grappa).

## Requested behaviour

- [ ] **Incognito toggle** — expose it at login and/or in settings; marks the session ephemeral ("won't be remembered — closing means gone").
- [ ] **Best-effort client trigger** — on `beforeunload`/`pagehide`, signal the server to delete immediately. Treated as best-effort only.
- [ ] **Server-side 1h linger (authoritative)** — on disconnect with **no reconnect within 1 hour**, the server deletes the account + all data. This is the real guarantee; the client event is a fast-path, not the source of truth.
- [ ] **Deletion scope** = same as the existing delete-account (account + scrollback + session/data). Reuse the delete-account server path.

## Grounding (origin/main)

- Delete-account already exists: `cicchetto/src/DeleteAccountModal.tsx` + its server endpoint (gated to registered user/visitor, non-admin). Incognito reuses that deletion path but triggers it automatically (client best-effort + server linger) instead of via the manual modal.
- Session TTL: visitor sessions already carry a persistence TTL (~2 days) — incognito is a session variant with a **1h linger** and an **auto-delete-account-on-close** intent (it wipes the account/data, not merely expires the session).

## Notes / decisions for implementer

- **`onBeforeUnload` is unreliable** — do NOT make deletion depend on it. Client fires it best-effort; the 1h server linger is the guarantee.
- Interaction with "share session" (visitor multi-device): likely disable share-session while incognito (an ephemeral session shouldn't be portable). Confirm.
- UX: unambiguous incognito indicator + a warning that closing ends everything. Consider a confirm when enabling.
- Product tie-in: this addresses the privacy/ephemeral "not for you if" objection on the grappa.chat landing — once shipped, that copy can flip from "not for you" to "for you (incognito)".

Requested by gpiero via vjt (#grappa).

vjt:
**Scope decided (vjt, 2026-07-26): minimal — incognito WITH persistence.**

[...]

PR #1010, opened 2026-08-07
#982 — admin-side share-token mint, to let a locked-out visitor back in

vjt:
Admin-side mint of the visitor share token, so an operator can let a locked-out
visitor back in. Issue #982.

## The gap

A visitor has no password — the browser session IS the identity. Lose the
device, wipe the profile, drop the cookie, and the account is unreachable:
there is no credential to re-issue and nothing to reset. The recovery machinery
already existed (`POST /me/share-token` mints, `POST /auth/share/consume`
redeems), but the mint authenticates the caller *as the visitor* — precisely
what a locked-out visitor cannot do. The only operator tool was
`DELETE /admin/visitors/:id`, which throws the session away rather than
returning it.

## What lands

`POST /admin/visitors/:id/share-token`, in `GrappaWeb.Admin.VisitorsController`,
under the existing `:api + :authn + :admin_authn` scope. No per-controller
checks: `:admin_authn` is the gate. Response mirrors the visitor side —
`{token, expires_at}`; wrapping into `https://<host>/#/share/<token>` stays the
client's job.

It mints **the same token**: both doors now go through the new
`GrappaWeb.ShareToken` (salt, TTL, `mint/1`, `verify/1`), and
`/auth/share/consume` is untouched as the single redeem surface. A second
`Phoenix.Token.sign` site with its own copy of the salt fails in the worst
direction — the mint keeps answering 200 while consume rejects, so the operator
hands out a dead link and nobody sees an error. The controller-level `salt/0` /
`max_age_seconds/0` accessors this replaces had **zero callers**; the test file
re-declared both literals, which is the drift the shared module removes.

TTL (600s) and one-shot are unchanged.

## The accepted abuse, and what pays for it

An admin can mint a session grant for someone else's identity. vjt accepted
this explicitly — the alternative for a passwordless identity is *no* recovery
path. So the code owes visibility, not prevention:

- every mint records `:visitor_share_token_minted` naming the acting admin;
- telemetry is `[:grappa, :admin, :visitor, :share_token, :minted]`,
  deliberately **not** folded into the visitor-side event, or "did an operator
  do this?" stops being answerable;
- that event's actor fields are the **only non-nullable** ones in
  `AdminEvents.Wire`. Every other actor-bearing event has a real system path
  (reaper, `bin/grappa`); this verb is reachable only behind `:admin_authn`, so
  `nil` is not "the system did it" — it is an unattributed grant. The
  constructor's guard refuses it; cic's narrower drops the shape to `null`.

## The one not to get wrong

Incognito visitors are refused 403 at **both** doors (#363). The gate is
re-stated in the admin action rather than assumed — a check applied at one of
two callers is how a product decision gets undone through the back entrance.

## Carried in: the representative-nick derivation

The event carries `visitor_nick` so the console names a person, not a UUID.
That derivation was already inline in four places (`Operator` ×2, `Reaper`,
`Themes`); a fifth copy was not acceptable, so it is now
`Credentials.representative_visitor_nick/1`. It lives in `Credentials` for a
structural reason, not a stylistic one: `Grappa.Visitors` already deps on
`Grappa.Themes`, so a helper in `Visitors` could never be reached from `Themes`
without inverting that edge into a Boundary cycle. `Grappa.Networks` is the one
boundary all callers already share. Each caller keeps its own absent-nick
consequence — the events render `nil` as the id, `Themes` leaves the changeset
alone so the wire falls back to the guest label.

## Verification

- `scripts/check.sh` — **rc=0**, `8 doctests, 56 properties, 5455 tests, 0 failures`;
  `git status --porcelain` empty before and after.
- `scripts/bun.sh run check` — **rc=0** (biome + tsc; tsc is the real gate here,
  three exhaustive `assertNever` switches grew an arm).
- `scripts/bun.sh run test` — **4579 cic tests green**.
- **6 mutants injected, 6 killed** (injector aborts unless it matches exactly
  once; committed before the run):
  1. incognito gate neutered → 1 red
  2. wire constructor accepts a nil actor → 1 red
  3. admin door mints with a private salt → **3 red** (verify, consume,
     one-shot — the "one redeem surface" claim)
  4. audit record dropped → 1 red
  5. telemetry folded into the visitor-side event → 1 red
  6. cic narrower stops requiring an actor → 1 red

## Not claimed

- **No e2e, and no empty spec written instead.** There is no cic control that
  calls this endpoint, so the outcome is not client-visible and there is
  nothing for a browser to drive. Whether the admin console grows the button is
  a separate unit.
- The `AdminEventsTab` row is covered by cic unit tests only; it has not been
  seen rendered in a browser.

### DIFF
--- a/cicchetto/src/AdminEventsTab.tsx
+++ b/cicchetto/src/AdminEventsTab.tsx
@@ -52,6 +52,11 @@ function renderEvent(ev: WireAdminEvent): string {
       return `${visitorLabel(ev.visitor_nick, ev.visitor_id)} deleted${actorSuffix(ev.actor_user_name)}`;
     case "visitor_reaped":
       return `${visitorLabel(ev.visitor_nick, ev.visitor_id)} reaped (TTL expired)`;
+    case "visitor_share_token_minted":
+      // #982 — the actor is not optional here (see the server-side
+      // constructor), so it is named outright rather than through
+      // `actorSuffix`, which renders nothing when the name is absent.
+      return `share link minted for ${visitorLabel(ev.visitor_nick, ev.visitor_id)} by ${ev.actor_user_name}`;
     case "reaper_swept":
       return `reaper swept ${ev.count} visitor(s)`;
     case "upload_reaped":
--- a/cicchetto/src/__tests__/AdminEventsTab.test.tsx
+++ b/cicchetto/src/__tests__/AdminEventsTab.test.tsx
@@ -112,6 +112,37 @@ describe("AdminEventsTab — per-kind rendering (closed-union exhaustiveness)", 
     expect(row.textContent).toContain("S`grappa deleted by vjt");
   });
 
+  it("visitor_share_token_minted names the visitor and the acting admin (#982)", () => {
+    events = [
+      sample({
+        kind: "visitor_share_token_minted",
+        visitor_id: "v-uuid",
+        visitor_nick: "S`grappa",
+        actor_user_id: "u-uuid",
+        actor_user_name: "vjt",
+      } as WireAdminEvent),
+    ];
+    render(() => <AdminEventsTab />);
+    const row = screen.getByTestId("admin-event-visitor_share_token_minted");
+    expect(row.textContent).toContain("share link minted for S`grappa by vjt");
+  });
+
+  it("visitor_share_token_minted falls back to the id when the nick is absent", () => {
+    events = [
+      sample({
+        kind: "visitor_share_token_minted",
+        visitor_id: "v-uuid",
+        visitor_nick: null,
+        actor_user_id: "u-uuid",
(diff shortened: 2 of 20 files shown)


vjt:
Superseded by the union PR #1015, which carried this branch's commits and landed on main as f41952f3.

Gated as a union deliberately: #1013 and #1010 both edit `lib/grappa/networks/credentials.ex`, and #1014 changes the gate itself (`scripts/check.sh` + `mix.exs`), so three separate greens would never have covered the tree that actually merged. The union run: 640 tests, 640 passed, 25.1m.

Before opening it I checked, per PR, the count of lines this branch has that the union LACKS: 0. A cherry-picked union rewrites the SHAs, so GitHub cannot mark this MERGED by ancestry — closing by CONTENT, not because the work was dropped.

Issue #546, opened 2026-07-30
NOTICE from a peer should not open a query window (irssi/HexChat rule: open query → query, else $server)


vjt:
A NOTICE from a plain user nick currently **opens a query window** for the recipient. Azzurra staff (morph, Sonic, Mezmerize, Hypnotize — `#it-opers`, 2026-07-30) ask that it follow the irssi/HexChat convention instead: a notice lands in the query **only if that query window is already open**, otherwise it goes to the status/network window.

**This is not a bug that slipped through — it is a documented decision** (UX-6-L, 2026-05-20 + #422 Option B), which is why this issue carries no status label: reversing it is vjt's call.

## Requested rule

Agreed in-channel after some back and forth:

> Query window already open (and not archived) → the query. Otherwise → `$server`.

Sonic initially proposed a third branch — "no query open but a channel in common → route into the channels" — and **morph objected** ("nei canali no, irssi manda in status"). Sonic conceded ("status mi sta anche bene"). So the rule above is what both settled on; recording the disagreement because it is the part most likely to be re-litigated.

## What the code does today (verified against `origin/main` @ `46fe7b29`)

**(a) The routing decision — `lib/grappa/session/event_router.ex`, `route_non_channel_notice_non_chanserv/3` (~2470-2490).** For a sender that is a valid nick and not a service:

```elixir
Identifier.valid_nick?(sender) ->
  # Regular user nick → me; persist on channel = sender_nick so it
  # lands in the same query window a PRIVMSG-to-own-nick would.
  {sender, body}
```

Persisting on `channel = sender_nick` is what makes the window appear.

**(b) The requested behaviour already exists — but only for services.** `service_route_channel/2` (~2509-2516), introduced by #400:

```elixir
defp service_route_channel(sender, state) do
  open? = Map.get(state, :query_window_open?, fn _, _, _ -> false end)

  if open?.(state.subject, state.network_id, sender),
    do: sender,
    else: "$server"
end
```

That is *exactly* the rule being asked for, already implemented and already wired to a `query_window_open?` callback. The change is largely a matter of which senders reach it.

[...]

Issue #422, opened 2026-07-26
Inbound DM with no browser attached never opens a query window — lands silently in Archive


vjt:
Reported on IRC (per channel request): a DM arrived in a query that was in the Archive section. Days later it was still sitting in Archive — never appeared in the active window list, no unread, and the reporter only noticed by chance. A second, smaller complaint came with it: the message row showed `19:15` with no way to tell *which day*.

Both are verified on `origin/main` (`05c44c78`). Prod currently runs the `v0.5.0` tag (`6c27d422`), which contains the same code for every file cited below.

---

## Part 1 — an inbound DM never opens a query window unless a browser is connected

**Auto-open lives only in the client.** `cicchetto/src/lib/subscribe.ts` handles the own-nick topic and calls `openQueryWindowState` on inbound DM-shaped traffic — line 615 (PRIVMSG / ACTION) and line 661 (peer NOTICE). `openQueryWindowState` (`cicchetto/src/lib/queryWindows.ts:79`) does not mutate local state: it pushes `open_query_window` over the WS and waits for the server's `query_windows_list` broadcast.

**The server never opens one on its own.** `QueryWindows.open/4` has exactly one call site in `lib/`:

```
lib/grappa_web/channels/grappa_channel.ex:963
```

which is the handler for the client's `open_query_window` message. Every other `openQueryWindowState` caller in cicchetto is a user action (MembersPane, WhoModal, UserContextMenu, compose `/msg`+`/query`, ArchiveModal, push-notification tap in `lib/pushTarget.ts:59`).

**So the window creation is conditional on a live socket.** The session is a bouncer and stays connected with no browser attached. When a DM lands in that state:

1. `Session.Server` persists it and broadcasts (`lib/grappa/session/server.ex:3987`, `Persistor.persist_and_broadcast(..., push: true)`), row keyed `dm_with = <peer>`.
2. Nobody is listening on the own-nick topic, so no `open_query_window` is ever pushed. No `query_windows` row exists.
3. On the next login the peer is absent from `query_windows_list`, so there is no active row and no unread state.
[...]

Issue #400, opened 2026-07-25
Service NOTICE replies bypass an already-open query window


vjt:
## Problem

Sending to a service by hand — `/msg SeenServ ...` when no `/ss` alias exists — opens a query window with that service, but the service's replies never land in it. They are re-keyed onto the synthetic `$server` window, so the operator is left staring at an empty query window while the answer sits somewhere else.

Reported from a third-party instance. It matters most on mobile, where defining an alias before every ad-hoc service interaction is awkward: you type in a query window and expect the answer in that query window.

## Current behaviour (verified on `origin/main`, 6478ef72)

* `Grappa.Session.EventRouter.route_non_channel_notice_non_chanserv/2` — a NOTICE whose sender passes `Grappa.Irc.Identifier.services_sender?/1` (a closed allowlist: NickServ, ChanServ, MemoServ, OperServ, BotServ, HostServ, HelpServ, RootServ, SeenServ, StatServ, DebugServ) is persisted on `channel = "$server"`, unconditionally.
* The PRIVMSG arm does the same, gated on a non-channel target (`route_channel = if services_sender?(sender) and not channel_target?(channel), do: "$server"`).
* Outbound is asymmetric by design: `Session.Server`'s `service_target?/1` makes a PRIVMSG addressed to a service wire-only, with no scrollback row.

Both behaviours are deliberate and documented in place. The `$server` re-key exists so that a service NOTICE does not trip cicchetto's dm-listener auto-open path and spawn a query window per service; the outbound suppression exists so `IDENTIFY <password>` never persists in scrollback.

## Proposal

Keep both invariants, and make the `$server` re-key **conditional on there being no already-open query window** for that service nick:

* if a query window with the service exists for this subject on this network → persist on the service nick's window;
* otherwise → `$server`, exactly as today.

This does not weaken either reason the current rule exists. Auto-open is not triggered, because a window that already exists is not being opened; and `$server` stays the fallback for unsolicited service traffic, which is the common case (NickServ notices at connect, ChanServ access changes, and so on).

The state needed already exists — no new table:

* `Grappa.QueryWindows` is the server-side persisted record of open DM windows, subject-scoped (`user_id` XOR `visitor_id`), with `new_window_exists?/3` and rfc1459-folded uniqueness on `(subject, network_id, fold(target_nick))`.
* `EventRouter` already carries `state.subject`, so the lookup is available at the routing site.

[...]

PR #1011, opened 2026-08-07
#546: a peer NOTICE lands in the query only if it is already open

vjt:
Reverses the NOTICE arm of UX-6-L (2026-05-20) / #422 Option B, on a ruling from Azzurra staff (morph, Sonic, Mezmerize, Hypnotize — `#it-opers`, 2026-07-30) that vjt queued:

> **Query window already open → the query. Otherwise → `$server`.**

A NOTICE from a plain peer used to persist at `channel = sender`, which the server-side auto-open turned into a sidebar tab. Azzurra's welcome notice is sent from a **user**, not a service, so every operator collected a stray window on connect.

## The change is one deleted branch

The requested behaviour already existed. `service_route_channel/2` has implemented "open → the window, else `$server`" since #400, because services traffic needed it. All this does is widen which senders reach it: `route_non_channel_notice_non_chanserv/3` drops its `Identifier.services_sender?/1` arm, so every nick sender takes the same test.

Deleting that carve-out was morph's own argument for the change. The allowlist branch, the `$server` re-key and the `*Serv` handling existed **only** because the default was "notices open windows"; flipping the default makes them redundant. A parallel branch beside the old one would have been the double structure CLAUDE.md forbids. The helper is renamed **`open_query_or_server/2`** — it is no longer about services.

Sonic's third branch ("no query open but a channel in common → the channels") was refused by morph and conceded. Recorded in DESIGN_NOTES rather than left for the next reader to re-propose.

**The PRIVMSG door keeps the allowlist, deliberately.** A peer's DM still opens the conversation — only NOTICEs became non-opening. `services_sender?/1` retains exactly one job: telling a robot's reply apart from somebody talking to you.

## Two claims the issue asked us to assume, proven instead

**`query_window_open?/3` does not need an archive check — and must not have one.** `query_windows` has no `archived` column: `close/4` DELETEs the row, and `Scrollback.list_archive/3` derives the archive as "has scrollback MINUS the active keyset", whose query half is exactly those rows. Open and archived are complementary by construction; the ruling's conjunction is one condition here. A second check would duplicate derived state and could only disagree with itself.

**The CTCP VERSION visibility row does not go quiet** — the issue's one concrete regression. It never rides the NOTICE door: it is emitted by the inbound-**PRIVMSG** CTCP arm, persists at `channel = own_nick` with `dm_with = sender`, and the auto-open keys on `dm_with || channel`. Pinned by a test, because a claim about a path this change never visits is exactly the kind that rots unnoticed. Both corrections posted on the issue: #546 (comment).

## What the mutation matrix actually proved

| mutation | rc | result |
|---|---|---|
| source reverted to `9bb56fb3` | 2 | 4 failures — both reversals, the predicate call, the full-stack case |
| **M1** kill the CTCP short-circuit | 2 | **1 failure, only the new test** |
| **M2** drop `services_sender?` from the PRIVMSG door | 2 | 4 failures, incl. both new PRIVMSG asserts |
| **M3** route the CTCP VERSION row to `$server` | 2 | **1 failure, only the new test** |

Two things this turned up that reading had not:

- **M1 is red alone.** All three pre-existing CTCP tests use `base_state()`, so they land on `$server` regardless — without the new open-window assert, removing the short-circuit would have been silent. The short-circuit is kept on a *new* justification: post-#546 it stops a peer you have an open query with from getting raw `\x01PING …\x01` filed into that conversation.
- **M2 caught a duplicate of mine.** I added a Conserv PRIVMSG guard believing I was relocating the NOTICE one; the guard already existed from UX-4 bucket G. Removed in `af616b8f`.

The green run also surfaced a test my greps missed: `event_router_test.exs:533`, the plain-notice pin **#591 deliberately parked on this ruling** ("whether THAT is right is #546/#548's question"). Flipped, not deleted — with both cases now given an open window they still diverge, which is the only assertion that the CTCP carve-out *is* a carve-out.

## Verification

Measured locally, on the rebased tree:

```
scripts/check.sh   rc=0   8 doctests, 56 properties, 5448 tests, 0 failures
                          Credo: 0 errors · Dialyzer: passed · Sobelow: run
bun run check      rc=0   502 files
bun run test       rc=0   246 files, 4594 tests
```

The Elixir gate ran pre-rebase, but `git diff 9bb56fb3..origin/main -- lib/ test/ mix.*` is **empty** (the union was cic-only), and the branch's patch-set hash is byte-identical before and after the rebase (`f4f71a93…`, DESIGN_NOTES excluded), so its basis is unchanged. The cic gates above were re-run after the rebase.

⚠️ **The two e2e specs in `issue546-peer-notice-no-query-window.spec.ts` have NEVER been executed.** No STACK lane was allocated, and biome does not cover `e2e/` (it processed 0 files on that path), so they have had no static gate either. **This PR's CI is their first run.** They are written, not proven — do not read them as evidence until they are green.

## Out of scope, noted so it exists somewhere

`cicchetto/src/lib/compose.ts:1064-1066` tells the operator that a `/query` on a services nick would be "a dead window". That has been false since #400 (an open window does receive service traffic) and is doubly false now. Not touched here; worth its own issue if vjt wants it.

### DIFF
--- a/cicchetto/e2e/tests/issue546-peer-notice-no-query-window.spec.ts
+++ b/cicchetto/e2e/tests/issue546-peer-notice-no-query-window.spec.ts
@@ -1,0 +1,123 @@
+// #546 — a NOTICE from a plain peer must NOT open a query window.
+//
+// Azzurra staff (morph, Sonic, Mezmerize, Hypnotize — #it-opers 2026-07-30)
+// asked grappa to follow the irssi/HexChat convention:
+//
+//   query window already open → the query.  Otherwise → the status window.
+//
+// Until #546 a peer NOTICE persisted at `channel = sender`, which the
+// server-side auto-open (#422) then turned into a sidebar tab — Azzurra's
+// welcome notice is sent from a USER, not a service, so every operator got a
+// stray tab on connect. This reverses UX-6-L (2026-05-20) / #422 Option B for
+// the NOTICE arm only; an inbound PRIVMSG still opens the conversation.
+//
+// Why e2e and not just ExUnit: the VISIBLE outcome is a sidebar tab that must
+// not appear, and its appearance is a three-hop consequence (route → persist →
+// `query_windows_list` fan-out → sidebar projection) that jsdom cannot stage.
+// Per feedback_ux_e2e_mandatory.
+//
+// The negative assertion is GATED on a positive barrier: the notice must first
+// be observed on `$server`. Without that gate, "no tab appeared" would pass on
+// a session that never processed the line at all — a green that proves nothing.
+
+import { expect, test } from "../fixtures/test";
+import {
+  composeSend,
+  composeTextarea,
+  loginAs,
+  scrollbackLine,
+  selectChannel,
+  sidebarWindow,
+} from "../fixtures/cicchettoPage";
+import { assertMessagePersisted } from "../fixtures/grappaApi";
+import { IrcPeer } from "../fixtures/ircClient";
+import { AUTOJOIN_CHANNELS, getSeededVjt, NETWORK_NICK, NETWORK_SLUG } from "../fixtures/seedData";
+
+const CHANNEL = AUTOJOIN_CHANNELS[0];
+
+// Fresh nick per run/retry: a fixed one is a 433/ghost time bomb under CI
+// rerun + bahamut per-IP clone limits (the #600 red), and a leftover
+// `query_windows` row from a prior run would pre-open the very window test 1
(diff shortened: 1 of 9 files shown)


vjt:
Superseded by the union PR #1012, which carried this branch's commits verbatim and landed on main as b1ff00e2 (rebased onto ea213c8f; patch-set proved identical, 898 lines both sides, before and after the rebase).

A cherry-picked/rebased union rewrites the SHAs, so GitHub cannot mark this PR merged by ancestry — closing it by CONTENT, not because the work was dropped. The commits are on main.

Issue #1001, opened 2026-08-07
Repair the credentials already corrupted by #977 — nothing backfills them today


vjt:
## What

#977 stopped the corruption going forward (merged as #983 — the interceptor now takes the SECOND token of `SET PASSWD <old> <new>`), but **nothing repairs the credentials already corrupted in production**. #983 says so itself: "an editable password field remains the cure for a credential already corrupted." At least one live account is in this state today. Those rows will never identify again on their own.

## The detector is exact, not heuristic

A corrupted value is literally `"<old> <new>"` — the two passwords joined by the space `do_set_password` split on. **A valid Azzurra password cannot contain a space**: `do_set_password` (`azzurra/services/src/nickserv.c:2192`) and `do_resetpass` (`:3927`) reject `strchr(newpass, ' ')` outright. So:

- **contains a space ⇒ corrupted, with certainty.** No false positives by construction.
- **longer than `PASSMAX` (32, `azzurra/services/inc/config.h:96`) ⇒ unusable** even if space-free; the network will never accept it. Worth reporting in the same sweep even if not auto-repaired.

## Repair rule — take the LAST token, not the second

The obvious rule is "keep the second token", but a credential that went through **two** in-session rotations holds `"<a> <b> <c>"`, and the live password is `<c>`. Splitting on whitespace and keeping the **last** field is correct for one rotation and for N. State this in the code, with the two-rotation case as a test.

## Constraints

- `password_encrypted` is a `Grappa.EncryptedBinary` (Cloak AES-GCM, `lib/grappa/networks/credential.ex:189`, `redact: true`) ⇒ **the sweep must run inside the app** (a `mix` task / `Grappa.Release` eval on a short-lived BEAM, like `Grappa.Release.migrate/0`), never as SQL. There is no way to grep this from the DB.
- Write through `Credential.password_changeset/2` (`:486-497`) so the value is re-encrypted through the same path, not hand-rolled.
- **Dry-run first, and report counts**: how many credentials carry a space, how many exceed `PASSMAX`, how many were repaired. A silent data fix that touches N rows without saying which N is not reviewable.
- Do not touch `nickserv_pass_encrypted`: that field was never fed by the `SET PASSWD` capture path, so it is not part of this corruption.

## Ordering

This is a data fix on top of the already-merged #983. It does not need a schema change, so it does not gate on #41 — it is an eval, not a migration.

vjt:
[...]

Issue #977, opened 2026-08-07
SET PASSWD capture stores old+new concatenated — every in-session password rotation corrupts the credential


vjt:
## What is wrong

The in-session `SET PASSWD` capture (#131) stores the **old password concatenated with the new one** on Azzurra. Every operator who rotates their NickServ password from inside grappa ends up with a credential that can never identify again — the exact split-brain #124 describes, produced by grappa itself rather than by an out-of-band change.

Reported in the wild by vjt: a visitor rotated their password, ended up with "a password that was too long", then went through `RECOVER` + `RESETPASS` and grappa still auto-identified with the dead secret. The "too long" password is the concatenation.

## The truth about Azzurra's syntax (read off the services source, not inferred)

`azzurra/services` @ `23473ed`:

- `src/nickserv.c:2090` — `do_set` routes `PASSWD` and hands the handler `param = strtok(NULL, "")`, i.e. **everything after the verb**.
- `src/nickserv.c:2182` — `do_set_password` then splits that at the FIRST space: `newpass = strchr(param, ' '); *newpass++ = '\0';`. So **`param` is the OLD password and `newpass` is the new one** — the comment at `:2204` says it outright: *"param holds the old password"*.
- The command form is therefore `SET PASSWD <oldpass> <newpass>`, and the change is applied only when `str_equals(param, callerUser->ni->pass)` — it is an authenticated rotation, not a blind set.

Our interceptor's moduledoc asserts the opposite and is the origin of the bug:

> SET PASSWD (Azzurra parses the new password with `strtok(NULL,"")`, so it may contain spaces — never split on the first space)

That `strtok(NULL,"")` is `param` — old **and** new together. Azzurra passwords cannot contain spaces at all: `do_set_password` rejects them explicitly (`strchr(newpass, ' ')` → `CSNS_ERROR_PASSWORD_WITH_SPACES`), and `do_resetpass` does the same.

## The defect in our code

`lib/grappa/session/ns_interceptor.ex:78`:

```elixir
@set_passwd_re ~r/^(?:PRIVMSG\s+NickServ(?:@\S+)?\s+:|(?:NS|NICKSERV)\s+)?SET\s+PASSWD\s+(\S.*?)\s*$/i
```

[...]

Issue #124, opened 2026-06-28
grappa: handle split-brain when the NickServ password changes outside grappa's sync path


vjt:
## Problem

grappa keeps a per-identity stored password (`visitors.password_encrypted`) that it syncs from NickServ via the identify path: `NSInterceptor` captures the password from `IDENTIFY/ID/SIDENTIFY/GHOST/REGISTER`, and `Grappa.Visitors.commit_password/2` writes it on the `+r` MODE observation (see #32, shipped).

That sync only fires on **identify/register**. Any NickServ password change that doesn't go through that path leaves grappa's stored password **stale** — a split-brain between grappa's credential and the real NickServ secret. **Not supported today.**

Two ways in:

1. **`SET PASSWORD` via grappa.** `/msg NickServ SET PASSWORD <new>` is passed upstream untouched — `SET` is not in `NSInterceptor`'s verb list (`@verb_re`, ns_interceptor.ex:32), so the new password is never captured. No `+r` transition happens (already identified), so `commit_password` never re-fires.
2. **Password changed entirely outside grappa** (another client, services web panel, ChanServ/NickServ on a different session). grappa has no signal at all.

## Consequences (today)

- Stored grappa password stays the OLD value → diverges from the real NickServ secret.
- **Lockout:** next grappa login compares the user's NEW password against the OLD stored one (`check_password/2`, login.ex:319) → `:password_mismatch`. The user can only log into grappa with the now-dead old password.
- **Failed identify on reconnect:** `:nickserv_identify` sends IDENTIFY with the stale old password → NickServ rejects → no `+r`. Made worse by #90 (no fallback IDENTIFY retry) — the session may never identify.

## Design wrinkles (why it's not a one-liner)

- **No `+r` transition to key the commit on.** SET PASSWORD / external change don't toggle the umode, so the existing `+r` observer can't trigger a re-commit. A different commit trigger is needed.
- **NOTICE-confirmation parsing is the trap.** Keying off NickServ's "Password successfully changed" NOTICE is exactly the fragile per-network (Anope/Atheme) text parsing deliberately killed in #91. Avoid reintroducing it.
- Candidate approaches to evaluate:
  - Intercept `SET PASSWORD` (extend the verb list) and **optimistic commit on send**, or confirm-then-commit with a narrow, bounded correlation (not open-ended NOTICE scraping).
[...]

PR #1013, opened 2026-08-07
Repair the credentials #977 concatenated, and refuse to guess the rest

vjt:
## Summary

`mix grappa.repair_passwords` sweeps the credentials #977 corrupted. #983 stopped the capture concatenating `SET PASSWD <old> <new>` into one stored value and #124 gave the subject a field to retype the secret in, but neither backfills a row that is already wrong.

**The issue's premise had to be corrected, and the correction is the substance of this PR.** The body claims a stored space means corruption *with no false positives by construction*. That was true of `nickserv_pass_encrypted`, a NickServ-only column. It stopped being true when #124's fold ran `SET password_encrypted = nickserv_pass_encrypted`: `password_encrypted` is now the single home for the NickServ secret **and** the SASL one **and** the server password. SASL PLAIN base64-encodes its passphrase (`auth_fsm.ex:722`), so a space there is legitimate and identifies normally; `password_changeset/2` guards only CR/LF/NUL (`credential.ex:491`), and `update_credential_password/2` applies Azzurra's chain only for `:nickserv_identify` (`credentials.ex:664`) — the #124 field *deliberately* accepts a space-containing password on a `:sasl` row. Implemented literally, the sweep would have truncated `"my pass phrase"` to `"phrase"`: Cloak AES-GCM, no plaintext backup, nothing to reverse it from. Gating this issue behind #124 was right for the schema, and the right schema is exactly the one where the space alone no longer decides. Written up as [a comment on #1001](https://github.com/vjt/grappa-irc/issues/1001#issuecomment-5219853366).

**Classification** — the verb the issue already owed the >2-token case, applied to a second axis. No new mechanism:

| condition | outcome |
|---|---|
| `:nickserv_identify` + space + exactly 2 tokens | **repaired** (last token) |
| `:nickserv_identify` + space + >2 tokens | reported |
| `:sasl` / `:auto` / `:server_pass` / `:none` + space | **reported, never repaired** |
| `:nickserv_identify`, space-free, but services would refuse it | reported, never written |

**The rule everything follows from:** a missed repair costs a manual intervention, a wrong repair costs an unrecoverable secret, and the two do not weigh the same. Hence: dry run is the default; the split is a literal single space with **no `trim:`** (a run of spaces means services refused the rotation outright, so neither token is known to be live); a recovered password services would themselves refuse is reported rather than written; and no output line ever carries the secret or any token of it.

**The detector is borrowed, not re-derived.** `NSInterceptor.vet_password/2` is already public as the second door #124 needed, and its spaces arm is FIRST in the chain, so no later arm can shadow it — that is what makes the call exact rather than approximate. It supplies the `PASSMAX` verdict too, so `32` is not hard-coded where it could drift from the wire door.

**A second reader was needed.** `list_all_credentials/0` is scoped `user_id IS NOT NULL` for #211, correctly, for the subject-scoped admin surfaces it backs. But `Session.Server.rotate_stored_password/2` has a `{:visitor, visitor_id}` clause beside the `{:user, user_id}` one, so a visitor credential corrupts identically. A task whose only output is a count, printing "0 candidates" with a corrupted visitor row in the table, is not a coverage gap but a false statement. Added as a sibling with a doc explaining why both radii are deliberate, rather than by widening the existing one.

**The task stays in the tree** (per vjt's correction): idempotent, inert on a healthy database, and its standing value is the inverse reading — a run that finds anything after #983 and #124 is the signal of a NEW bug, which the output says.

## Test plan

- [x] TDD red first: 23 specs run before the module existed — `23 tests, 23 failures`, all `module is not available`
- [x] `23 tests, 0 failures` after
- [x] **10 mutants injected, 10 killed, 0 survived**, injector aborting unless a substitution matched exactly once, tree committed before each, revert verified clean after each
- [x] A mutation exposed a defect in the *specs*, not the code: removing the auth_method gate left `leaves a SASL passphrase untouched` green, because `"my pass phrase"` is three tokens and the token count was protecting it. Replaced with a two-token SASL passphrase — the shape indistinguishable from a #977 concatenation by count alone. That mutant now kills 5 specs instead of 4 (fixed in its own commit)
- [x] Dialyzer's two `extra_range` warnings treated as design signals, not noise: `verdict/0` split out so each branch carries the reasons it can actually produce
- [x] `scripts/check.sh` rc=0 on the pre-rebase head, tree clean before and after — Credo `no issues`, `deps.audit` `No vulnerabilities found.`, `8 doctests, 56 properties, 5483 tests, 0 failures`, Dialyzer `Total errors: 0`, Sobelow run, bats 336 ok / 0 not ok

## Not asserted

- **No e2e, and no empty spec written in its place.** There is no client surface — this is a mix task. One written for appearances would be worse than none.
- **The task has never been run against a real database**, in production or in dev. Everything known about it comes from the sandboxed suite.
- **The OPERATIONS recipe for running a mix task inside the jail is not verified against m42.** What is established from the repo: the jail has the full source tree and Mix (`deploy.sh` runs `git pull`, `mix deps.get`, `mix compile`, `mix release`, and `mix run --no-start` in the preflight), and the `:4000` collision the old bullet cited was fixed on 2026-07-23 when `Boot.start_app_silent/0` began suppressing the Endpoint. What is **not** established is whether a second BEAM writing the same SQLite file while the live node runs is safe. The doc says so in those terms rather than giving a clean all-clear.
- **That a repair takes effect on the session's next reconnect** is read off `SessionPlan.base_plan/6`, by the same route a #124 field edit takes — inferred from the code, not observed.
- **`scripts/check.sh` was green on `2376d38f`, the pre-rebase head.** The rebase onto `c841bf10` is proven content-identical (patch-set byte-identical excluding docs; docs additive, `80/0` and `33/2`, the two deletions being the stale OPERATIONS bullet replaced on purpose), but the gate has not been re-run on the rebased head. CI decides that.
- **A genuinely corrupted `:sasl` row is reported, not repaired** — one repair fewer for zero destroyed secrets. The operator override discussed for that case was deliberately **not** built, and no half-finished hook was left for it.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -32782,3 +32782,83 @@ about width still restart the pool at 5 and 10 themselves. The ceiling
 correspondingly stops binding here: 14 tests × ~80 migrations per run becomes
 14 × 3, and the file drops from 8.6s to 0.4s. The undiagnosed replay red is
 still undiagnosed — it simply no longer has a home in this test.
+
+## 2026-08-07 — #1001: the space is an exact signature, but only where Azzurra's rule reaches
+
+`mix grappa.repair_passwords` sweeps the credentials #977 corrupted. #983
+stopped the capture concatenating `SET PASSWD <old> <new>` into one stored
+value and #124 gave the subject a field to retype the secret in, but neither
+backfills a row that is already wrong, and at least one live account was left
+in that state.
+
+**Why the space is a signature and not a heuristic.** An Azzurra password
+cannot contain one: `do_set_password` cuts the parameter at the first space
+(`strchr(param, ' ')`) and then refuses a `newpass` that still holds one, and
+`do_resetpass` does the same. So a space in a secret Azzurra's NickServ
+governs is a concatenation artifact by construction — there is no legitimate
+value it could be. The detector does not re-derive this: it calls
+`Session.NSInterceptor.vet_password/2`, already public as the second door
+#124 needed, whose spaces arm is FIRST in the chain and therefore cannot be
+shadowed by a later one. The same call supplies the `PASSMAX` verdict, so 32
+is not hard-coded a second time where it could drift from the wire door.
+
+**Where that certainty stops, and it is #124 that stopped it.** The issue
+claimed no false positives *by construction*. That was true of
+`nickserv_pass_encrypted`, a NickServ-only column. It stopped being true the
+moment `20260807120000_fold_nickserv_pass_onto_password` ran
+`SET password_encrypted = nickserv_pass_encrypted`: `password_encrypted` is
+now the single home for the NickServ secret AND the SASL one AND the server
+password. SASL PLAIN base64-encodes its passphrase
+(`Grappa.IRC.AuthFSM.sasl_plain_payload/1`), so a space there survives and
+identifies normally; `Credential.password_changeset/2` guards only CR/LF/NUL,
+and `Credentials.update_credential_password/2` applies Azzurra's chain only
+when the method is `:nickserv_identify` — the #124 field *deliberately*
+accepts a space-containing password on a `:sasl` row. Gating #1001 behind
+#124 was right for the schema, and the right schema is precisely the one
+where the space alone no longer decides.
+
+So the sweep repairs `:nickserv_identify` and nothing else. Every other
+method with a space is reported. That is not a new mechanism: it is the verb
+the issue already owed the >2-token case ("report, do not guess"), applied to
+a second axis of ambiguity.
+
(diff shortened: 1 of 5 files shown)


vjt:
For the record, since this closed without the explanation: superseded by the union PR #1015, which carried this branch's commits and landed on main as f41952f3.

Gated as a union deliberately — this branch and #1010 both edit `lib/grappa/networks/credentials.ex`, so neither PR's own green ever covered the tree that actually merged. The union run: 640 tests, 640 passed, 25.1m. Lines this branch has that the union lacks: 0.

The work is on main; only the SHAs differ, because a cherry-picked union rewrites them.

Issue #75, opened 2026-06-22
cicchetto: customizable and shareable themes with a theme gallery


vjt:
Bring back rich theming as a first-class feature (this was one of irssi's main draws).

**Customizable themes**
- Editable colors.
- Customizable background, including user-uploadable background images (e.g. a Star Trek background with the Enterprise).
- **Font family** from a **curated font set** (see fork decisions below).

**Shareable themes**
- A theme can be published/shared so others can apply it.

**Theme gallery**
- Browse available themes by name and author.
- Each entry shows a screenshot/preview of the theme.
- Show simple usage metrics, e.g. how many people have applied it.

Requested via chat.

---

## Design & security model — session 2026-07-17

Design converged with vjt (+ alk) on #grappa. Guiding principle (vjt): **KISS.**

### Data model — dead simple (no CoW, no references)
- **The gallery is the gallery.** It's a catalog of published themes: the built-in curated set + anything users publish.
- **Pick a theme from the gallery → it's COPIED into your account.** Always a full, independent copy.
- **Edit a theme → you save your own copy.** Edits only ever touch your copy.
- **Publish → your copy is pushed into the gallery** for others to pick.
- Every theme is an **independent copy**. There is **no** copy-on-write, no shared storage, no reference/usage counter, no delete-in-use guard, no orphan logic — deleting your copy can't affect anyone, because everyone else already has their own copy. (vjt: "un user ha il proprio tema che è sempre una copia, niente copy on write… diventa troppo un casino altrimenti.")
- Each theme has an **owner/author**. Built-in themes are owned by a **"system"** author (read-only; you copy them like any other).
- **All themes are public** (no visibility flag).
- **Active theme is SERVER-persisted** (fork 1) — the currently-applied theme is stored server-side (active-theme column / apply endpoint), so it **follows the user across devices**, consistent with grappa's server-owned-state invariant. It is NOT a client-local localStorage per-device setting.
- **Rate limit:** ~5 theme creations/day/user (save + publish), server-side, anti-abuse.

### API / authz (minimal)
- Authenticated REST: copy-from-gallery, save-own, publish, delete-own, **apply/set-active**.
- Optional share link to a published entry: `/theme/<id>` (references the stored, sanitized theme **by id** — never raw payload/CSS in the querystring, which would be a malicious-theme phishing link).
- **admin** can edit/delete any gallery entry (moderation); **owner** edits/deletes their own; **anyone** browses + copies.

[...]

PR #1014, opened 2026-08-07
build(#621): run the gate's doctor in :test, where its count is honest

vjt:
One line of `mix.exs`: the `ci.check` alias runs `"cmd env MIX_ENV=test mix doctor"`
instead of `"cmd mix doctor"`, reusing the shell-out precedent three lines below it.
`scripts/check.sh`'s now-duplicate second doctor run is deleted, its #75 rationale folded
into the alias comment.

## The artifact

`mix doctor` derives a module's function **count from the source AST** but reads each
function's **doc/spec presence from the compiled beam of the env it runs in**. The alias is
pinned to `:dev` (credo / sobelow / ex_doc are `only: [:dev, :test]`), and in `:dev`
everything inside `if Mix.env() == :test do … end` is compiled out — counted, unscorable,
filed as "No Docs" AND "No Specs" even with `@doc false` plus a full `@spec`.

Put more sharply: **the dev run is not lenient about `test/support`, it is blind to it.** It
never compiles those files, so it cannot score them at all — 297 modules against `:test`'s
313. The `lib/` half is the same blindness one scope down. One defect at two radii.

Nine modules in `lib/` carry such a seam (the issue lists seven; it misses
`GrappaWeb.PasskeyOrigin` and `Grappa.RateLimit.RequestBudget`).

## ⚠️ The red that motivates the issue no longer exists

Measured on this tree, the `:dev` doctor scores `Repo.BusyRetry` **43% and PASSES**.

The issue's "43% → 38%, reddening the gate" was real and is **historical**. The 38% lived at
`bf74899e`, where `arm_faults` still had two arities — three documented of eight is 37.5%,
under doctor's 40% floor. The very next commit of the same #594, `2b42190a`, collapsed those
arities for an unrelated design reason (a default argument in disguise) and took it back to
43%. The surviving comment in `busy_retry.ex` says so outright: *"the 40%-doctor-floor
artifact #621 is a side effect, not the reason"*.

So **this change fixes no live red.** It removes a three-point margin and a scoring lie that
the test-env doctor never tells. That is the honest case for it; "the gate is red" would not
have been. Noted on the issue too, so the next reader who measures 43% does not conclude we
were confused.

## The issue's first option rested on a premise that does not hold

It proposed a `.doctor.exs` excluding test-gated functions from the *dev* count, reasoning
that "the dev run was added deliberately for a reason (`check.sh` comment)". The comment says
the opposite — it justifies the **other** run, the `MIX_ENV=test` one added by `2eed58ca`
(#75, a four-red-commits post-mortem): *"Add a second `mix doctor` run in MIX_ENV=test to
check.sh so local == GH"*. And `git log -S 'doctor' -- mix.exs` shows doctor entering with the
project scaffold and later being wrapped in `cmd` for exit propagation (`fc4575a0`) — **no
commit ever argues for doctor-in-dev.** It is a side effect of the alias's env pin.

The reason for the dev run did not exist, rather than having been superseded. Checking the
premise changed the cure instead of confirming it: not teaching the dev count to lie less, but
stopping it from gating.

## Proven by mutation, in four passes

Exit codes only, never a grep of the output.

| pass | result |
|---|---|
| **Baseline**, `:test`, clean tree | **exit 0** — 313 modules, 0 failed; `test/support` modules appear in the listing |
| **MUT-A weak** — one `@doc` off an always-compiled fn of `busy_retry.ex` | **exit 0**, module at 86% |
| **MUT-A strong** — five of its seven | **exit 1**, 29%, 1 failed module |
| **MUT-B** — the `@doc` off `test/support/themes_image_fetcher_test_resolver.ex` | **exit 1**, 0% |
| **Counterfactual** — the SAME MUT-B against doctor in `:dev` | **exit 0**, and it never names the module |
| **Restore** | **exit 0**, 313 / 0, both targets back at 100% |

MUT-B is the one that could have condemned the change: that module is the exact one named in
the #75 post-mortem, and a green there would have meant I had switched off the guard #75
bought rather than the dishonest run. It is red.

The counterfactual is what makes MUT-B non-circular — of the two doctor runs `check.sh` had,
the one carrying the #75 guard is precisely the one kept.

MUT-A weak is reported rather than quietly dropped: the mutation **landed and was scored**
(100% → 86%, one "No Doc"), it just did not cross a 40% floor. A green there means "mutation
below threshold", not "gate asleep", and the two are easy to confuse. So the version that does
cross was measured too.

## Parity, read from `origin/main` and not a local checkout

(A CI rule was once wrong for days because it was read from a stale local `ci.yml`.)
`git show origin/main:.github/workflows/ci.yml`: exactly one `mix doctor`, at line 106, under
the `test` job's `MIX_ENV: test`; the `dialyzer` job (`MIX_ENV: dev`) runs only `dialyzer` and
`docs`. No dev-env doctor anywhere. The alias and the workflow now invoke the same command in
the same env over the same file set — parity by construction rather than a duplicate kept in
step by hand.

**Scope of that claim, checked rather than assumed:** it covers doctor ONLY. The alias's other
steps still run in `:dev` locally against `:test` on GH, but none derives its file set from
`elixirc_paths` — credo takes `included: ["lib/", "test/", "config/"]` from `.credo.exs`;
sobelow and format are path-based. Doctor was the one tool reading the beam, which is exactly
why #75 was doctor-specific. That divergence is pre-existing and untouched.

## Gate

`scripts/check.sh` → **exit 0** on this branch. Run deliberately *through* the modified
`ci.check` alias, since standalone `mix doctor` invocations do not exercise the line actually
changed. No bats pins `check.sh`'s content (it enters only the derived set of
`test/infra/shellcheck_gate_test.bats`, and only comments were touched there).

## Rejected

Raising the 40% floor so `BusyRetry` fits — the count is what is dishonest, not the threshold,
and moving a threshold to admit a failing subject is the same family as weakening an
assertion. Also rejected: a `.doctor.exs` excluding test-gated functions from the dev count,
which encodes the lie in configuration and leaves two doctor runs disagreeing about one module.


vjt:
Superseded by the union PR #1015, which carried this branch's commits and landed on main as f41952f3.

Gated as a union deliberately: #1013 and #1010 both edit `lib/grappa/networks/credentials.ex`, and #1014 changes the gate itself (`scripts/check.sh` + `mix.exs`), so three separate greens would never have covered the tree that actually merged. The union run: 640 tests, 640 passed, 25.1m.

Before opening it I checked, per PR, the count of lines this branch has that the union LACKS: 0. A cherry-picked union rewrites the SHAs, so GitHub cannot mark this MERGED by ancestry — closing by CONTENT, not because the work was dropped.

Issue #1018, opened 2026-08-07
cic: Alt+A / next-active cycle should skip muted windows — the #866 mute never reaches the #235 ordering


vjt:
## Ask

From vjt (#grappa, 2026-08-07 22:10):

> una finestra in "mute" dovrebbe essere esclusa dal giro che fa il bottone "alt+a"

A window the operator has muted should not be offered by the next-active-window
cycle. Today it is: the mute is invisible to that ordering, so a busy muted
channel keeps stealing the jump the operator uses to reach the windows they
actually read.

## Where it stands today

**The cycle.** `cicchetto/src/lib/activeWindows.ts` is the single ordering
(`#235`): DM/mention tier first, chronological within a tier, `messagesUnread`
as the activity gate (`#265` — presence churn excluded). Three doors dispatch
the same verb, so all three inherit whatever this issue changes:

- `cicchetto/src/lib/keybindings.ts:103` — Alt+A (physical `KeyA`)
- `cicchetto/src/lib/keybindings.ts:84` — Ctrl+N / Ctrl+P
- `cicchetto/src/Shell.tsx:292` — `nextUnread` → `jumpToNextActiveWindow()`, also
  behind the on-screen affordance button (BottomBar-right / sidebar-bottom-left)

**The mute.** `#866` already stores it: `NotificationPrefs.muted_targets`
(`cicchetto/src/lib/userSettings.ts:34`), keyed by the FOLDED conversation —
`canonicalChannel(channel)` for a channel, `canonicalChannel(peer)` for a DM,
never the row's `channel` field. Expiry is resolved by the READER
(`notificationPrefs()` in `cicchetto/src/lib/notificationPrefs.ts`, `withLiveMutes`),
so a consumer never needs a `now` argument. The only consumer today is
`pushTriggers.shouldNotify` (`cicchetto/src/lib/pushTriggers.ts:95`).

So both halves exist; nothing joins them.

## Proposed change

Filter muted windows out of the candidate set in `activeWindows.ts`, reusing the
existing key fold rather than re-deriving it — the DM case in particular must key
on the PEER, exactly as `pushTriggers.conversationKey` does, or every DM collapses
onto one mute entry.

Scope guard, so the implementer does not over-reach: **badges and counters stay
as they are.** `#866` Q4 was ruled explicitly — "push / beep / badge only;
`WindowCounts` mention counts and the sidebar severity dot are out of scope, a
muted target still counts". This issue extends the mute into the NAVIGATION
[...]

PR #1022, opened 2026-08-07
cic(#1018): the next-active cycle skips muted windows

vjt:
Issue #1018. A conversation the operator muted (#866) was invisible to the
#235 next-active ordering, so a busy muted channel kept stealing the Alt+A
jump.

## What changed

- **`activeWindows.orderUnreadWindows` takes the mute as a required input.**
  It is the ONE ordering behind all three doors (Alt+A, Ctrl+N/P, the
  affordance button), so all three inherit the skip without forking the
  ordering #235 unified. Required rather than optional on purpose: a future
  door that forgets the mute is a compile error, not a window that silently
  keeps stealing the jump.
- **`lib/conversationMute.ts` — one home for the mute key.** The DM case is
  the trap: the key is the FOLDED conversation (the channel for a channel,
  the PEER for a DM), and getting it wrong collapses every DM onto one entry.
  That derivation was private in `pushTriggers` and open-coded in the
  settings picker; this issue would have made three. Now `pushTriggers`,
  `activeWindows` and `SettingsDrawer` ask the same functions
  (`conversationMuteKey` / `windowMuteKey` / `isConversationMuted`).
- **No clock.** Expiry stays in the reader (#866 Q3): the memo passes
  `notificationPrefs().muted_targets`, already stripped of elapsed snoozes.

## Rulings taken (the issue's open questions)

1. **All three doors** — they are one verb; splitting them would fork #235.
2. **Every unread window muted → the cycle is a no-op**, falls out of the
   empty list.
3. **A mention inside a muted window is still skipped** — same as #866 Q2,
   the mute wins.

## Scope guard (#866 Q4)

Badges and mention counters are untouched; the muted window keeps its sidebar
unread badge (asserted in the e2e). The affordance's own count *does* drop the
muted window, because it reads the same ordered list — a button reading \"2\"
that cycles through 1 would be the lie.

## Gates

- \`scripts/bun.sh run test\` — **246 files / 4607 tests, exit 0.** Red first:
  the five new mute cases failed against the parent commit before the filter
  existed.
- \`scripts/bun.sh run check\` (biome + tsc) — exit 0, working tree unchanged
  before/after.
- **e2e NOT run** — \`issue1018-mute-skips-next-active.spec.ts\` is written but
  needs the integration stack, i.e. a lane I have not been allocated. It
  mutes through the SettingsDrawer picker (so the key the picker writes and
  the key the cycle reads are proven identical), then asserts Alt+A lands on
  the unmuted room, the affordance hides while the muted room is still
  unread, and the muted room's badge survives.

Two test-side edits ride along: `activeWindows.test.ts` passes the new input
everywhere, and `Shell.test.tsx`'s `channelKey` mock now keeps the REAL
`canonicalChannel` via `importActual` (the mute fold reaches it through
`activeWindows`; a hand-rolled stand-in would be the suite re-implementing
production's fold).

### DIFF
--- a/cicchetto/e2e/tests/issue1018-mute-skips-next-active.spec.ts
+++ b/cicchetto/e2e/tests/issue1018-mute-skips-next-active.spec.ts
@@ -1,0 +1,119 @@
+// GH #1018 — a MUTED conversation (#866) is not a stop on the next-active
+// cycle (#235: Alt+A, Ctrl+N/P, the on-screen affordance).
+//
+// Why an e2e and not just the unit test on `orderUnreadWindows`:
+//
+//   * the mute is created through the SettingsDrawer picker, so the folded
+//     key the picker WRITES and the key the cycle READS are proven to be the
+//     same string. Two unit suites on either side of that seam cannot see a
+//     fold mismatch between them — the same argument
+//     `push-866-conversation-mute.spec.ts` makes for the push port.
+//   * the outcome under test is "which window did the operator land on",
+//     which is only observable here.
+//
+// Scope guard under test too (#866 Q4): the muted channel keeps its sidebar
+// unread badge. The mute changes the NAVIGATION order, not the counters.
+//
+// Desktop only, deliberately: all three doors dispatch the SAME verb
+// (`jumpToNextActiveWindow`), and #235's own spec already gates the mobile
+// button against that verb. This spec's variable is the mute, not the door.
+//
+// Removing the feature turns this red at the first jump: the muted channel
+// heads the cycle (older activity, same tier) and Alt+A lands there.
+
+import {
+  loginAs,
+  openSettingsSection,
+  selectChannel,
+  sidebarMessageBadge,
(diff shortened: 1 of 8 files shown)


vjt:
Superseded by #1024, which landed on main as `68e8941e`.

This PR's commits were cherry-picked into that union and gated there **together** with the other half, rather than each against a base neither shared. Closing BY CONTENT: a cherry-pick rewrites the SHAs, so GitHub cannot detect the merge by ancestry and would leave this open with a head that still reads mergeable — an invitation to ship the same work twice.

Verified before the union was opened, as counts rather than "identical" (two empty sets also match): **lines this PR has that the union LACKS, restricted to this PR's own files = 0**. The union's `integration` was **643 passed / 0 failed / 27.7m**, and its head contained `dbdefa6b`, so that green attests the landed tree bit-for-bit.

Not deployed anywhere yet, and the issue stays open — it closes at the release, with a comment naming the tag.

Issue #1020, opened 2026-08-07
infra: cic build empties the SERVED dist in place — build beside, then swap


vjt:
## Ask

From vjt (2026-08-07 22:2x):

> *"fixa sta procedura di deploy, prima buildi e poi seappa, ma che cazzo"*

Every substrate builds the cic bundle **directly into the directory it is
currently serving**, with `--emptyOutDir`. So the served directory is wiped at
the *start* of the vite build and stays incomplete until it finishes: for that
whole window the running server serves an empty dist and **the SPA does not load
at all**.

## Where

- `infra/freebsd/jail_cic_build.sh:51` — `npm run build -- --outDir ../runtime/cicchetto-dist --emptyOutDir`. This is **production**: the BEAM self-serves that directory via `Plug.Static` since #485, resolved per request.
- `infra/linux/cic_build.sh:42` — same shape into `$OUT_DIR`.
- `scripts/deploy-cic.sh` / `scripts/deploy.sh` — Docker: builds into the bind-mounted `runtime/cicchetto-dist/` the container serves.
- `infra/packaging/build.sh:105` is **fine** — it writes to a staging dir nobody is serving. It is the shape the others should have.

Tell that the in-place wipe is already fighting the repo: three of these have to **restore `.gitkeep` after the build**, because `emptyOutDir` eats it.

## Why `--emptyOutDir` is there (do not just delete it)

`outDir` is outside vite's root (root is `cicchetto/`, output goes to
`../runtime/cicchetto-dist`, shared with Docker per `bd1d85cb`). Vite cleans
`outDir` by default but **refuses to empty an out-of-root directory** without
explicit consent, and that flag is the consent. The cleanup is wanted: vite emits
content-hashed assets, so without it every deploy leaves the previous bundle's
chunks behind forever.

**The flag is not the defect. The timing is.**

## Proposed change

Build into a sibling and swap:

1. `--outDir <served>.new --emptyOutDir` (still cleaned, but nobody is serving it)
2. on success, swap `<served>.new` into place with `mv`, then remove the old tree
3. on failure, leave the served directory untouched — today a failed build can
   leave it empty or half-written

The served path then goes complete-old → complete-new with no empty interval, and
because `Plug.Static` resolves per request the new tree is picked up immediately.
The window drops from a whole vite build to a couple of syscalls.

Do it for all three serving substrates, not just the jail — this is the same
substrate-parity problem #923 was filed for, and fixing one leaves the others.

## Honesty about what is measured

[...]

PR #1023, opened 2026-08-07
infra(#1020): build the cic bundle beside the served dist, then swap

vjt:
Fixes the deploy procedure for #1020 — *"prima buildi e poi seappa"*.

## The defect, as measured

Every serving substrate aimed vite's `outDir` at `runtime/cicchetto-dist`, the
directory `Grappa.Cic.Bundle.root/0` resolves **per request** (`Plug.Static` +
the SPA history-fallback; `root/0` stashes a path *string* in
`:persistent_term` and `File.read` runs on every call). Vite empties `outDir`
**before** it writes anything, so the served tree was blank for the whole build
— and a build that FAILED left it blank, because `set -eu` walked away from an
already-emptied directory.

`--emptyOutDir` is **not** the defect and is kept. `outDir` is out of vite's
root by design (root is `cicchetto/`, output goes to `runtime/cicchetto-dist`
so Docker and the jail share one server-side anchor, `bd1d85cb`), vite refuses
to clean an out-of-root dir without that consent, and the cleanup is *wanted* —
content-hashed chunks would otherwise accrete forever. The **timing** is the
defect.

### One correction to the brief I was given

`infra/freebsd/jail_cic_build.sh` — the **production** substrate — never had a
post-build `touch .gitkeep`. Only the linux and the three Docker launchers did
(`grep -rn gitkeep` over the parent commit: `infra/docker/deploy.sh:713`,
`infra/linux/cic_build.sh:49`, `scripts/deploy-cic.sh:54`,
`scripts/deploy.sh:147`; nothing under `infra/freebsd/`). So on the jail every
cic build silently deleted a **tracked** file and left `D
runtime/cicchetto-dist/.gitkeep` in the worktree — the same shape as the
DESIGN_NOTES 2335 incident where an unstaged `D` stalled a `git pull`. This
change closes that as a side effect (the bats case `jail: the new bundle lands`
is red on the parent for exactly that assertion), and the tree self-heals on
the first deploy that carries the fix.

## The change

`infra/lib/cic_dist.sh` — one POSIX-sh implementation, sourced (never executed,
same contract as `beam_wait.sh` / `deploy_common.sh`) by all three serving
substrates. `infra/packaging/build.sh` needs no change: it already stages, and
is the shape being generalised.

- **jail** (`infra/freebsd/jail_cic_build.sh`) and **native linux**
  (`infra/linux/cic_build.sh`): `--outDir <served>.next --emptyOutDir`, then
  promote.
- **docker**: the oneshot writes to whatever is bind-mounted at `/app/dist`, so
  the compose mount source became `${CIC_BUILD_OUT:-./runtime/cicchetto-dist}`.
  The three wrappers set it **on the build command only** (not exported at
  large, so the later `up`/`restart` compose calls see the normal mount) and
  promote afterwards. The **default stays the served dir** on purpose: a bare
  `compose --profile prod up` reaches the build through grappa's `depends_on`
  with nobody to promote, and there nothing is serving yet. A second compose
  service was the alternative and was rejected — two build definitions is
  exactly the drift `cic_version_export_test.bats` polices.

### Swap primitive and its failure window (asked for explicitly)

`mv` is `rename(2)`. There is no portable two-directory **exchange**
(`RENAME_EXCHANGE` is Linux-only and unreachable from `sh`), so the promote is
two renames:

```
mv <served> <served>.prev     # served path momentarily ABSENT
mv <staged> <served>
```

Between them the served path does not **exist** — `Plug.Static`'s `from:` misses
and the request falls through to the history-fallback / 404. It is **ENOENT,
not a half-written tree**, so a client gets the old bundle or the new one, never
a mix; and it is two syscalls against the whole vite build it replaces.

Die mid-swap and nothing is lost: `.prev` holds the complete previous bundle and
`.next` the complete new one, and the next run starts by clearing `.prev` and
re-staging `.next`. Only a crash inside the microsecond gap leaves the served
path missing; re-running the build fixes it. Both paths are siblings inside
`runtime/` so the rename cannot go cross-device and degrade to copy-then-delete.

`mv a b` where `b` is an existing DIRECTORY moves `a` INSIDE `b` — the served
path is therefore *renamed away*, not emptied, before the second `mv`, and that
failure has its own test case because it is silent (server keeps serving the
old bundle, every step reports success).

**A symlink flip would have been atomic and was rejected**: the served path is
TRACKED (`runtime/cicchetto-dist/.gitkeep`), and turning a tracked directory
into a symlink is a type change git reports as dirt — the very class of dirty
tree that stalls a deploy's `git pull --ff-only`.

### .gitkeep

Not relocated blindly: it stopped being a *repair* and became a *property of the
landing tree*. `cic_dist_promote` plants it in the staged tree **before** the
rename, so the tracked path is never absent for any interval, instead of being
restored shortly after it goes missing. The four post-build `touch` calls are
gone, and a bats case fails if one comes back.

### On failure the served tree is untouched — how that is proved

Two independent mechanisms, both tested:
1. the build writes into a directory nobody serves, so a non-zero vite exit
   under `set -eu` never reaches the promote at all;
2. the promote itself REFUSES a staged tree with no `index.html` (a build can
   exit 0 having written nothing reachable) and returns non-zero with the old
   bundle still in place.

`jail: a FAILED build leaves the previous bundle serving` and its linux twin
run the real scripts with `FAKE_BUILD_RC=1` and assert `<served>/index.html`
still reads `OLD`. Both are **red on the parent commit** (`cat: ... No such file
or directory` — the parent had already emptied it).

## Tests

`test/infra/cic_dist_swap_test.bats` — 17 cases. The fake builders **emulate
vite**: they empty `--outDir` first and, at that instant, record what the
SERVED directory holds. The assertion is on the *timing*, not on any string, so
it cannot be satisfied by a script that merely mentions the right path.

**Red on the parent commit** (production files restored to `HEAD~1`, test + lib
kept): 8 of 17 red, including both timing observations and both
failed-build cases —

```
not ok  9 jail: the served bundle is still serving while vite builds
not ok 10 jail: the new bundle lands, stale chunks and debris do not
not ok 11 jail: a FAILED build leaves the previous bundle serving
not ok 12 linux: the served bundle is still serving while vite builds
not ok 14 linux: a FAILED build leaves the previous bundle serving
not ok 15 docker: the cicchetto-build mount source is the CIC_BUILD_OUT seam
not ok 16 docker: every compose launcher aims the build at a staging dir and promotes it
not ok 17 docker: no launcher restores .gitkeep after the build any more
```

`linux: the new bundle lands` passes on the parent, correctly: building in
place *does* land the bundle and *does* clear stale chunks, and linux restored
its `.gitkeep`. It is a regression net, not a discriminator — the discriminators
are the five above.

**Mutations, both run and both read:**

| mutation | result |
|---|---|
| jail `--outDir "$staged"` → `"$served"` | `not ok 9, 10, 11`; linux 12 still `ok` (correctly scoped) |
| drop `touch "${_cic_staged}/.gitkeep"` from the promote | `not ok 4` (lib), `not ok 10` (jail), `not ok 13` (linux) |

Two harness families needed updating and the commit says why: the throwaway
clones now copy `infra/lib/cic_dist.sh` (they already copy `deploy_common.sh`,
for the same reason), and the `docker` stubs now model the oneshot *writing a
bundle* into `CIC_BUILD_OUT` — previously they answered `run --rm
cicchetto-build` with a bare `exit 0`, i.e. a build that produced nothing, which
the promote rightly refuses.

## Gates

- `scripts/bats.sh test/bin/ test/infra/ test/scripts/` → **354/354, exit 0**
- `scripts/shellcheck.sh` → **exit 0**, 71 derived scripts (the new lib is in
  the derived set: `--list` line 25)
- working tree porcelain empty before and after both gates

## HOT vs COLD, by content

Read out of `lib/grappa/deploy/preflight.ex` (not executed — that needs the
compile lane):

- **jail → HOT.** Changes are `infra/freebsd/jail_cic_build.sh` +
  `infra/lib/cic_dist.sh`: not `rc_d?`, not `systemd_unit?`, not `nginx?`, not
  `config/*.exs`, not `mix.exs`/`mix.lock`, not `application.ex`. And
  `docker_image?` is `filter_on(:docker, …)`, so `compose.yaml` is invisible to
  this substrate.
- **linux → HOT**, same reasoning.
- **docker → COLD**, via `docker_image?` (`String.starts_with?(path,
  "compose.")`). Worth stating the nuance: only the *oneshot's* mount changed,
  and `compose run --rm` re-reads the file every invocation, so nothing about
  the long-lived grappa container actually needs recreating. The `compose.`
  prefix class is deliberately conservative (H20 proved the enumeration failure
  mode twice) and I did **not** weaken it for this.

**The fix is self-applying on the jail**: `infra/freebsd/deploy.sh` runs `git
pull --ff-only` before `substrate_cic`, so the deploy that carries this change
is already the first one to build beside and swap.

## What I did NOT prove

- **The empty-dist window was never observed on a running jail.** Nobody may
  touch m42 or prod; the window is read out of `--emptyOutDir` + serve-at-
  request-time, exactly as the issue itself states. A structural reading says
  the path EXISTS; it says nothing about magnitude or frequency.
- **The Docker substrate was read, not exercised.** The build happens inside a
  container against a compose bind mount, so no bats can observe it. What is
  pinned is the compose mount source (byte-exact) and a scan that every launcher
  aims the build at `CIC_BUILD_OUT` and promotes. Nothing here proves a real
  container wrote where we think it did.
- ~~`docker compose config` was not run~~ — **since measured** (parse+interpolate
  only; no daemon, no container). `docker compose -f compose.yaml --profile prod
  config` returns **rc=0** both ways, and the mount interpolates to an absolute
  host path:

  ```
  # CIC_BUILD_OUT unset
  source: <repo>/runtime/cicchetto-dist
  target: /app/dist
  # CIC_BUILD_OUT=./runtime/cicchetto-dist.next
  source: <repo>/runtime/cicchetto-dist.next
  target: /app/dist
  ```

  What that proves is the interpolation and the resolved source path — nothing
  about a real container writing there. The Docker bullet above still stands.
- **The tests observe the window during the BUILD, not during the PROMOTE.**
  Reordering the two renames inside `cic_dist_promote` would not redden
  anything: the end state is identical and only the window width differs. That
  window is argued in the lib header, not measured.
- **The `.deb`/`.rpm` and e2e paths were left alone** — `infra/packaging/build.sh`
  already stages, and `cicchetto/e2e/compose.yaml` builds a separate
  `runtime/e2e/cicchetto-dist` that no production server serves.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -183,7 +183,17 @@ services:
     volumes:
       - ./cicchetto:/app
       - ./runtime/bun-cache:/cache
-      - ./runtime/cicchetto-dist:/app/dist
+      # #1020 — vite empties `dist/` before it writes, and `./runtime/
+      # cicchetto-dist` is the dir the BEAM serves PER REQUEST, so pointing
+      # this mount at it made every deploy serve an empty SPA for the whole
+      # build. The deploy wrappers (scripts/deploy.sh, scripts/deploy-cic.sh,
+      # infra/docker/deploy.sh) set CIC_BUILD_OUT to a staging sibling and
+      # rename it into place afterwards — one env var rather than a second
+      # service, so there is still ONE build definition.
+      # The DEFAULT stays the served dir on purpose: it is what a bare
+      # `compose --profile prod up` uses via grappa's depends_on, and there
+      # nothing is serving yet — no live server, no window to protect.
+      - ${CIC_BUILD_OUT:-./runtime/cicchetto-dist}:/app/dist
     tmpfs:
       # UID/GID must track the `user:` field above — bun writes to
       # HOME=/tmp, so a hardcoded uid=1000 tmpfs while the process runs
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -33326,3 +33326,63 @@ so the surfaced error is unforgeable proof the target survived cic's parser, the
 channel door and `Client.send_admin/2`. 423 and 447 stay unit-tested — the first
 needs a leaf with no `A:` line, the second a restricted-class user, and neither
 is reachable without reshaping the shared testnet for one spec.
+
+## 2026-08-07 — #1020: the cic build emptied the dist it was serving, so the build moved next door
+
+`--emptyOutDir` was never the defect. Vite refuses to clean an out-of-root
+`outDir` without explicit consent, and `outDir` is out of root by design (root
+is `cicchetto/`, the bundle lands in `runtime/cicchetto-dist` so Docker and the
+jail share one server-side anchor). The cleanup is *wanted*: vite emits
+content-hashed chunks, and without it every deploy accretes the previous
+bundle's assets forever and the dist stops being a truthful picture of what is
+served. **The timing was the defect.** Vite empties `outDir` BEFORE it writes,
+and `Grappa.Cic.Bundle.root/0` resolves that same directory PER REQUEST
+(`Plug.Static` + the SPA history-fallback), so every deploy blanked the SPA for
+the whole build — and a build that FAILED left it blanked, because `set -eu`
+walked away from an already-emptied tree.
+
+The fix is the shape `infra/packaging/build.sh` already had: build into a
+sibling nobody is serving, then rename it in. One implementation,
+`infra/lib/cic_dist.sh`, sourced by all three serving substrates — the jail's
+POSIX `/bin/sh` body, the native-Linux bash script, and the three compose
+launchers. `infra/packaging/build.sh` needs no change; it is the shape being
+generalised.
+
+**The swap has a window, and it is ENOENT rather than empty.** There is no
+portable two-directory exchange (`RENAME_EXCHANGE` is Linux-only and out of
+reach from `sh`), so the promote is two renames with the served path absent
+between them — two syscalls against the whole vite build it replaces. A client
+gets the old bundle or the new one, never a mix. Die mid-swap and both complete
+trees survive as `.prev` and `.next`; only a crash inside the microsecond gap
+leaves the served path missing, and re-running the build fixes it. A symlink
+flip WOULD have been atomic and was rejected for a concrete reason: the served
+path is TRACKED (`runtime/cicchetto-dist/.gitkeep`), and turning a tracked
+directory into a symlink is a type change git reports as dirt — the same class
+of dirty tree that once stalled a deploy's `git pull --ff-only` on a deleted
+`.gitkeep`.
+
+That `.gitkeep` moved from a repair to a property. Three scripts used to
+`touch` it back after the build because `emptyOutDir` ate it; the promote now
+plants it in the staged tree BEFORE the rename, so the tracked path is never
+absent at all rather than being restored shortly after it goes missing.
+
+**The Docker seam is a defaulted variable, on purpose.** The compose oneshot
+writes to whatever is bind-mounted at `/app/dist`, so the mount source became
+`${CIC_BUILD_OUT:-./runtime/cicchetto-dist}`. The deploy wrappers set it to the
+staging sibling and promote afterwards; the DEFAULT stays the served dir
+because a bare `compose --profile prod up` reaches the build through grappa's
+`depends_on` with nobody to promote — and there nothing is serving yet, so
+there is no window to protect. A second compose service would have been the
+alternative and was rejected: two build definitions is exactly the drift
+`cic_version_export_test.bats` exists to police.
+
+**What was proved and what was not.** The jail and native-Linux scripts are
+RUN in bats against a fake vite that empties its `outDir` first and records
+what the served directory holds at that instant — the assertion is on the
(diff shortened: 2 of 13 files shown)


vjt:
Superseded by #1024, which landed on main as `68e8941e`.

This PR's commits were cherry-picked into that union and gated there **together** with the other half, rather than each against a base neither shared. Closing BY CONTENT: a cherry-pick rewrites the SHAs, so GitHub cannot detect the merge by ancestry and would leave this open with a head that still reads mergeable — an invitation to ship the same work twice.

Verified before the union was opened, as counts rather than "identical" (two empty sets also match): **lines this PR has that the union LACKS, restricted to this PR's own files = 0**. The union's `integration` was **643 passed / 0 failed / 27.7m**, and its head contained `dbdefa6b`, so that green attests the landed tree bit-for-bit.

Not deployed anywhere yet, and the issue stays open — it closes at the release, with a comment naming the tag.

Issue #136, opened 2026-06-28
cicchetto: per-network operator-editable home-pane welcome/help text (split from #135)


vjt:
## What

Make the visitor home-pane **welcome text operator-editable, per network**,
instead of static-in-cic. Split out of #135 (which ships the home-pane rework
with a STATIC welcome text for now).

## Why

#135 hard-codes the welcome text in cicchetto, so every wording change is a code
edit + redeploy. Operators should be able to set a per-network welcome message
(and, soon, **help text** — see Notes) without a deploy.

## Scope

- Per-network welcome (and help) text stored + served via API; cic HomePane reads
  it instead of the static string from #135.
- Shares the same gap as #85 featured channels: operator-curated content with no
  admin UI today. Strongly consider ONE operator-content management surface
  (welcome + help + featured) rather than three bespoke ones — see #85 and #135.

## Notes

- Upcoming **help text** is the same shape of operator-editable per-network
  content; fold it in here rather than a third bespoke path.
- P2 — deferred behind the #135 static version. Requested by vjt.

PR #1033, opened 2026-08-08
Redesign the admin console onto a shared primitive layer

abonforti:
Ten admin tabs, four of which had CSS. One set of primitives, colours that
follow the active theme, and tables that fit a phone. No server changes, no
endpoint changes, no wire changes — zero files outside `cicchetto/`.

## The problem

Measured before touching anything:

- **Six of the ten tabs had no CSS at all.** Users, Vhosts, Credentials, Events,
  Session Log and Settings carried `admin-*` class names that matched no rule, so
  the browser rendered them with UA defaults.
- **Ten table class names, no shared base.** The
  `th/td { text-align:left; padding:.4rem .5rem; border-bottom }` block was
  copy-pasted three times; the other seven tables had nothing.
- **Semantic colour was hardcoded** — `#c00`, `#0a0`, `#c80`, eight literals
  outside the theme variables. That is why the console never followed a user's
  theme, and why it read badly under `mirc-light`.
- **Seven byte-identical copies** of the error row, seven of the refresh button,
  four different idioms for the same status badge.
- **No media queries for `.admin-*`**, and every control below the project's 44px
  tap floor.

### Colour follows the theme

An `--adm-*` layer derives every surface, line and semantic colour from the
existing theme variables via `color-mix`. Nothing was added to
`Grappa.Themes.TokenModel` — the server-side token set is closed and stays
closed, so a custom theme keeps working without knowing this exists. The eight
hardcoded hexes are gone.

### Nine shared primitives

| Primitive | Replaces |
|---|---|
| `AdminNav` | 220 hand-unrolled lines in `AdminPane` — ten tab buttons and ten panels |
| `AdminToolbar` | eight bespoke headers, seven copies of the refresh button |
| `AdminTable` | ten distinct table class names |
| `AdminStatus` | seven byte-identical error rows, plus loading and empty |
| `AdminBadge` | `LiveBadge` ×2, `CircuitBadge`, a bare class, and a raw ternary |
| `AdminCard` | the section container the dashboard shape needs |
| `AdminField` | label + control + message, for stacked forms |
| `AdminDetailPanel` | the four different inline-expand mechanics |
| `AdminFacts` | the columns a phone defers, per row |

### Tables that fit a phone

Nine columns is fine at 1440px and unreadable at 402px: panning sideways moves
the row's own name off screen, so you end up reading a number with no idea whose
it is. Sessions, Visitors, Credentials and Users keep *who*, *what state* and
*what can be done about it*; everything else defers into the row's detail panel.
Nothing is hidden on desktop.

Cards-per-row were the other candidate and are the wrong shape here. They read
better for one row and worse for a *list*, and a list is the job — you scan
these tables to find the odd one out, and cards put every value on its own line
where no two can be compared.

### Colour that means something

- **Session Log** — lifecycle stage, seven colours. Seven stages onto four
  severities forces collisions; two attempts proved it, first with `connected`,
  `registered` and `identified` all green, then with `connected` grey, which
  reads as "nothing happened".
- **Events** — the verb, five tones over 27 kinds. Severity does not survive 27
  values, because half of them are neither good nor bad. "What kind of change is
  this" does, and it is what an operator scans for.
- **Row actions** — reversibility. Disconnect is amber (parks a credential, the
  session comes back), Delete is outlined red (removes a row), Terminate is
  filled red because it kills a live pid, the only one you cannot undo by
  re-running the verb.

### 1 — the client restyle reverses a plan decision

The redesign plan said *dashboard for the admin console, irssi for the client,
deliberately different*. That is your line. The reviewer asked for the home
window to match, so `HomePane` now takes the same cards, tokens and button
metrics.

It is **isolated in its own commits** (the `cic(home)` ones, `1137c414` onward).
Drop them and everything before still stands — that is why they were kept apart
rather than folded into the pane work.

### 2 — how I read `ux-6-g-admin-mobile-h-scroll`

That spec requires pan-x to be *permitted* on every admin tab, and requires
exactly one tab — `DETERMINISTIC_WIDE_TAB`, Networks — to actually be wider than
the viewport so the permission is not trivially passing.

So Networks keeps every column and gets a frozen slug column instead; the other
four tables were free to narrow. If you meant that spec to pin *all* the wide
tables and not just the one it names, say so and I will put the columns back and
solve the width another way.

## Contracts held

- **No `data-testid` changed.** 31 e2e specs and 12 vitest files click them.
  Where a control moved — refresh out of the per-tab toolbars, Disconnect into
  the actions column — it carries its own id with it, because moving a button
  must not rename it.
- **`.admin-pane` and `.admin-tab-panel` keep their names, their
  `touch-action: pan-x pan-y` and their `max-height: var(--viewport-height)`.**
  Those are string literals in `Shell.tsx`, in the iOS selection carve-out and in
  four specs.
- **All ten `admin-tab-<key>` buttons stay one click away.** The Live /
  Configuration / Diagnostics grouping is visual: every tab still renders as one
  flat `role="tab"`.
- **Rebased on `main`,** including #982 and #985. #982 added a 28th admin-event
  kind and `tsc` refused the build until the tone map covered it — which is why
  that map is keyed on `WireAdminEvent["kind"]` rather than written as a loose
  lookup. #985 turned `.shell-chrome` into a zero-height box with a floating ☰
  over the pane's top-right corner, which is where the console now puts its close
  × — so suppressing that row on the admin window stops an overlap, where before
  it only saved a row.

## Two behaviour changes

- **The console opens on Sessions, not Visitors.** The old default predates the
  grouped nav; with the tabs in a deliberate order, opening on the second entry
  of the first group is arbitrary. `AdminPane.test.tsx` updated with it.
- **"Force Reap" is "Sweep visitors".** The old label named the internal process,
  not what the button does. Three specs updated.

## Not in scope, but adjacent

This does not touch #136 (operator-editable per-network welcome/help text): the
`HomePane` diff is seven `class=` attributes and none of the copy constants move.

Worth noting for when #136 lands, though. It asks for **one** operator-content
surface rather than three bespoke ones, and this PR builds the primitives such a
surface would be made of. The featured-channels editor — the other per-network
operator-curated content — is already reshaped into an `.adm-subsection` block
inside the Networks detail panel, so adding "welcome text" and "help text" beside
it is additive rather than an eleventh bespoke tab.

## How to review

1. `a75d6502` — the token layer and the primitives. Everything else is applying
   these.
2. `b63a0f9e` — `AdminPane` and the nav. The one place the one-click-per-tab
   constraint is enforced.
3. Any single tab migration to see the shape (`970f764c`, Sessions).
4. `a326d6ea` and `686e4d9a` — the mobile column strategy, and why the nested
   tables use a container query rather than a viewport one.
6. `f3c73cf2` — the scroll-owner fix and the three spec corrections it came
   with. Reading this one explains the `ux-6-g` question above better than the
   question does.
5. The `cic(home)` commits last, since they are the droppable ones.

## Gates

| Gate | Result |
|---|---|
| `bun run check` | pass — 517 files, 0 errors. The 59 warnings and 3 infos are present on the base commit. |
| `bun run test` | pass — 246 files, 4605 tests |
| `scripts/check.sh` | pass — 8 doctests, 56 properties, 5522 tests, 0 failures; bats green |
| `scripts/integration.sh` | pass — 642 passed, 0 failed, 23.4m (chromium + webkit-iphone-15) |

### Five e2e failures on the first full run, all fixed

Worth stating rather than hiding in the diff, because one of them matters
more than the other four.

Three were mechanical: an assertion on text inside a state badge that is now a
dot, a `/confirm delete/i` regex the scripted rename missed, and
`.admin-visitors-table` — a class `issue589-ios-admin-select` reaches for inside
a `page.evaluate`, which is why an audit that greps locators and testids never
saw it. That last one was fixed by putting the class back on the element rather
than rewriting the spec: when the code can satisfy the existing test, that is the
direction to go.

The other two are structural and are the reason this suite earns its runtime.
`.adm-table-wrap` took `overflow-x: auto` in the primitive layer, so the table
scrolled inside its own wrapper and the panel stopped overflowing. The gesture
still worked — nothing looked broken — but `ux-6-g`'s regression guard was left
measuring an element that can no longer overflow. It would have stayed green
while watching nothing, which is worse than a red test.

Below 900px the panel is the scroll owner again, which is what that spec locks
and why: it exists because of a real iPhone report ("horiz content, scrollbar,
but content doesn't move"). Desktop keeps the wrapper as the scroller, where the
card header staying put while columns pan is a genuine improvement and no spec
says otherwise.

Reviewed on a real device at 402×874 and 1180×820 across roughly a dozen rounds;
several commits here exist because that review found the first attempt wrong.


vjt:
## Review — orchestrator pass over `f3c73cf2`

Read the full diff (47 files, +4928/−1966), the 43 branch commits, the primitive
layer, the `--adm-*` block, every touched spec, and the `ux-6-g` spec on `main`.
**Not merged, and not merging:** CI is still pending, and two of the questions
below are vjt's to answer, not mine. What follows is findings first, then the two
open decisions, then what this review did *not* establish.

First, the part that is simply good, because it should be on the record and not
buried under the findings: the baseline is **measured**, not asserted — six of ten
tabs with no CSS at all, eight hardcoded semantic hexes, seven byte-identical
copies of the error row. The `--adm-*` layer derives from `--bg`/`--fg`/`--accent`
through `color-mix` and leaves `Grappa.Themes.TokenModel` alone, so a custom theme
keeps working without knowing this exists — that is the right seam. The specs were
updated to follow the product without being weakened, and one of them got
*stricter* (`/delete/i` → `/^delete$/i`). `.adm-btn` takes `min-height:
var(--tap-min)` rather than a fresh `44px` literal, which is how the tap floor
should be spelled. And `MatrixRain` is disciplined where decorative code usually is
not: canvas over DOM, ~15fps, `pointer-events: none`, a `ResizeObserver` because
`window.resize` misses a keyboard slide-in, and reduced-motion turning it **off**
rather than down.

### 1. The armed confirm buttons no longer name what they will do

Every admin call site now passes `confirmLabel="Confirm"` — 13 of them, across
Credentials, Networks, Sessions, Users, Vhosts and Visitors — with no `aria-label`
or `title` carrying the verb. Meanwhile every *non-admin* call site of the same
component keeps a descriptive armed label: `"really delete?"` (ArchiveModal),
`` `confirm removing ${passkey.name}` `` (PasskeySettings), `"apply — this
reconnects"` and `"save — this reconnects"` (SettingsDrawer), `"reconnect now
(drops + rejoins)"` (VhostSettingsPage). So the shared component now behaves one
way in admin and another way everywhere else.

`InlineConfirmButton`'s own moduledoc states the rule this crosses:

> Per `feedback_css_block_button_wraps_inline_prefix`: textContent IS the
> load-bearing UX signal.

Concrete case, from your own spec `m9b-admin-sessions-actions`: a Sessions row
carries **Disconnect** and **Terminate** side by side, and the mutex means arming
one disarms the other. Armed, the button now reads `Confirm` — with the verb gone,
the irreversible action and the recoverable one are distinguished only by which
box you last touched. That is the moment the two-tap gate exists for.

This is also the change with the widest blast radius in the PR: roughly ten spec
edits across seven files exist only to follow it. Keeping the verb (`Confirm
terminate`) — or at minimum an `aria-label` that does — would shrink the diff and
restore the signal. Your call to defend or change; I am flagging it, not vetoing.

### 2. `color-mix` has no fallback, and the whole console now depends on it

79 uses of `color-mix` in `default.css`; the five `@supports` blocks in that file
are all about `dvh`, none guards this. Where a `--adm-*` token cannot be computed
it becomes invalid at computed-value time, so surfaces, lines and every semantic
tone drop out together — not a degraded console, a broken one. `color-mix` landed
in Safari 16.2. Given cic is a PWA whose reason to exist is IRC on a phone, the
question is what iOS floor the project actually holds to, and whether a
`@supports (color-mix(in oklab, red, blue))` block with flat fallbacks is owed.
I have **not** measured what fraction of users that is, and I am not claiming it is
large or small — only that today the failure mode is total rather than graceful.

### 3. `prefers-reduced-motion` is read once, at mount

`MatrixRain` reads `matchMedia(...).matches` in `onMount` and returns early. There
is no `change` listener, so a user who turns the setting on while the Debug tab is
open keeps the animation until they navigate away. The moduledoc argues this
setting deserves *off, not down* — that argument applies to the live change too.
One listener, one `cancelAnimationFrame`.

### 4. `HomePane` now consumes the admin primitive layer — which makes the
### "drop the home commits" option less clean than it looks

Independently of the plan question below: `HomePane.tsx` now puts `adm-btn` and
`adm-btn--danger` on six client-facing buttons, so the *client* surface depends on
the *admin* console's primitives. That is a layering fact worth a decision on its
own merits.

It also means a block revert needs care, and I checked rather than assumed:

- `e3a6d64e` rewrote the home button CSS *around* `.adm-btn` as the base in the
  same sweep — `.home-pane-available-connect` has **0** rules on `main` and gains
  2 here, `.home-pane-network-disconnect` goes **2 → 1**. Reverting the `.tsx`
  alone would leave those buttons leaning on a base class they no longer carry.
- `ba39f2ca` is a **mixed** commit despite reading as a home one: it carries the
  home desktop grid *and* the creation of `MatrixRain` plus the `AdminCard`
  change, and `18cf93d1` later refactors exactly that code. Dropping it in block
  would take the admin Debug tab's decorative layer with it and conflict with the
  later commit.

Neither is an argument for keeping the home work — it is an argument that if it
goes, it goes as a deliberate re-slice, not a `git revert` of a commit range.

### 5. Neutral note

The default landing tab moves from Visitors to Sessions (`53e9b132`). User-visible,
probably deliberate, called out only so it is not discovered later as a surprise.

---

## Two decisions I am deliberately leaving OPEN — vjt's, not mine and not yours

**(a) The `cic(home)` commits from `1137c414` on.** They bring the home window onto
the console's dashboard shape, which runs against the standing plan decision
*"dashboard for admin, irssi for the client"*. Whether that decision still holds is
vjt's to say. If it does, see finding 4 for why the removal is a re-slice rather
than a revert.

**(b) How `ux-6-g-admin-mobile-h-scroll` should be read.** The spec on `main` pins
two different things at two different strengths: `touch-action: pan-x` on
`.admin-pane` **and** `pan-x` + `overflow-x: auto|scroll` on the panel for all
three of `visitors`/`sessions`/`networks`; and `scrollWidth > clientWidth` only on
`DETERMINISTIC_WIDE_TAB = "networks"`, because the other two are empty in the
baseline seed and would false-fail. Your `f3c73cf2` message reads the failure
correctly — `.adm-table-wrap` took `overflow-x: auto`, the scroller moved one level
in, the panel stopped overflowing, and the guard was left measuring an element that
can no longer overflow, which is worse than a red. The open question is which
contract that spec was written to hold:

- the **outcome** — "an operator can pan wide admin content on a phone" — in which
  case the scroller is free to move inward and the spec should follow it; or
- the **ancestor chain** — the pane and panel own the gesture and the overflow — in
  which case your below-900px restore is right and the design must keep the panel
  as the scroller on mobile.

The redesign's own premise ("tables that fit a phone") pulls at the positive twin
from a third direction: a table that fits no longer overflows at all. I am not
resolving that, and I would rather it be answered than patched around.

---

## What this review did NOT establish

- **Nothing visual.** I read the diff, not the rendered console. No screenshot, no
  device, no theme sweep — so I am not claiming `mirc-light` reads well now, only
  that the hardcoded hexes that broke it are gone.
- **I did not run the suite.** The CI on this head is the evidence, and it is still
  pending; the `cicchetto` and `shottino` jobs are green as of this comment.
- **I did not verify the `--adm-*` values against `shots/mockup-B.html`.** The
  commit claims byte-for-byte parity with the approved mockup; that mockup is not
  in the diff and I took the claim as stated rather than checking it.
- **Nothing about performance.** The `ResizeObserver` + rAF loop are argued in the
  code and unmeasured by me.

vjt:
Checked `78ee7f8b`. Both hold, and I looked at the parts that could have made them
cosmetic rather than real:

- **reduced-motion.** `raf` is declared and assigned *before* the listener is
  registered, so `cancelAnimationFrame(raf)` in the handler can never fire against
  an unset id; and since `draw` only re-schedules when it runs, cancelling the
  pending frame ends the chain rather than skipping one. The cleanup removes the
  listener. Not-resuming on the way back is the right asymmetry and the comment
  says why, which is better than the behaviour being discovered later.
- **`color-mix`.** `@supports not (color: color-mix(in oklab, red, blue))` is the
  correct shape of the test, and the flat values are the `irssi-dark` resolutions
  with an honest reason attached — a browser that cannot compute the mix cannot
  follow a custom theme either, so pretending otherwise would be theatre.

And a correction to my own framing, which your comment is right about: this file
already carried ~30 `color-mix` uses before the admin work. My finding should have
read "the console's structural colour now runs through it, turning a partial
degradation into a total one" — which is your sentence, not mine. Yours is the
accurate one.

Still open, unchanged: the armed-confirm labels (finding 1) and the
`HomePane`/`adm-btn` layering (finding 4), plus the two decisions that are vjt's —
the `cic(home)` commits and how `ux-6-g` should be read. Nothing here needs another
push from you; those want an answer, not a fix.

vjt:
`eb95daea` — finding 1 resolved, and resolved better than "revert it": the verb is
back on every armed admin button (`Confirm terminate`, `Confirm unbind`, `Confirm
sweep`, …) with only the trailing `?` dropped. That is a defensible copy change on
its own merits rather than a loss of signal, and the spec churn it drags with it
falls to a handful of one-character regex edits.

One leftover from the reverted sweep, cosmetic but confusing to the next reader:
`m9b-admin-sessions-actions.spec.ts` still has the test **title**
`"...transitions Disconnect → Confirm → fires"` while its assertion now pins
`/^Confirm disconnect$/`. The title describes a transition the test no longer
makes.

abonforti:
Thanks — three of the five are fixed, one is a decision I'd rather answer with
evidence than with a rebase, and one was already in the body.

## Fixed

**1 — the armed confirm buttons.** You're right, and my original change was the
wrong fix for the right complaint. The report I was acting on was that
`Confirm delete?` was *truncating* in a cramped actions column, and I solved a
width problem by deleting a word. The verb is back on all 13 call sites
(`eb95daea`), derived from `idleLabel` so the two cannot drift, and the Visitors
toggle mirrors its own `live_state` branch.

Worth noting the constraint dissolved on its own in the meantime: the secondary
columns moved into the detail panels, so the actions column now has room it did
not have when the truncation was reported. `Confirm delete` is also shorter than
the `Confirm delete?` that truncated — the word was never the problem.

**2 — `color-mix` fallback.** Added, `@supports not (color: color-mix(...))` with
flat values for every computed `--adm-*` token (`78ee7f8b`).

One correction to the framing, not to the conclusion: `default.css` already
carried **30 unguarded uses of `color-mix` on `main`** before this branch. So the
pattern is not new — what *is* new is that the console's structural colour now
runs through it, which turned a partial degradation into a total one. Your
conclusion holds either way; I'd just rather the record say I deepened an
existing exposure than that I introduced it.

**3 — `prefers-reduced-motion` read once.** Correct. It is watched live now, same
commit. Turning it back *off* deliberately does not restart the loop: resuming an
animation someone just asked to stop is the wrong side to err on, and reopening
the panel does it explicitly.

## Not fixed by rewriting history — answered by testing the removal instead

**4 — the `cic(home)` commits are not droppable as a range.** You are right, and
it is worse than you found: `ba39f2ca` does not touch `HomePane.tsx` at all
despite reading as a home commit, and `e3a6d64e` mixes the home buttons with the
Debug tab's amber phosphor. The cause is mine and worth naming — I committed
*review rounds* rather than *logical changes*, so each commit carries whatever a
single round of feedback happened to contain.

I considered an interactive rebase to re-slice them and decided against it, for
two reasons I'd rather state than have inferred:

- The mixed commits sit mid-branch and later commits refactor exactly that code
  (`18cf93d1` reworks the `MatrixRain` that `ba39f2ca` creates). A split cascades
  conflicts through everything after it, and I would end up with prettier history
  and *less* confidence — every gate would need re-running to earn it back, and
  your per-commit read would be invalidated.
- The home work **depends on this PR** (`adm-btn`, the `--adm-*` tokens), so it
  cannot become an independent PR either. Only a stacked one.

So instead of promising the removal is clean, I did it and measured it. On a
scratch branch off this head:

```
git checkout origin/main -- cicchetto/src/HomePane.tsx
# restore every `.home-pane*` rule to its main version
```

- `HomePane.tsx` → **byte-identical to `main`** (the whole home change there is 7
  `class=` attributes)
- `default.css` → 6 home rules restored, 12 branch-only removed, 11 dropped rules
  re-added: **0 `.home-pane*` rules differing from `main`**
- `bun run check` clean, and `HomePane.test.tsx` / `home.test.ts` /
  `focusVisible` / `sharedButtonRules` / `AdminPane` all green — **77 tests**

Net: `1 file changed, 114 insertions(+), 182 deletions(-)`, in `default.css` only.

So if decision (a) goes against the home work, the removal is one file, mechanical,
and verified green — not a range revert, and not a re-slice either. I've kept the
recipe; say the word and it lands as one commit.

## Still open, and still not mine

**(b) how `ux-6-g` should be read.** You framed the tension better than my PR body
did — a table that fits no longer overflows, which pulls at the positive twin from
a third direction. I have no standing to resolve which contract that spec was
written to hold, and I'd rather it be answered than patched around. My below-900px
restore is the conservative reading (ancestor chain), chosen because it keeps the
guard measuring something that can actually fail.

**5 — the default tab move** was called out in the PR body under "Two behaviour
changes"; flagging it again is fair, it is easy to miss in a diff this size.

Issue #1051, opened 2026-08-08
The floating ☰ covers the ✕ of the WHOIS / WHOWAS / LUSERS cards on non-channel windows


vjt:
The floating rail opener (☰) paints over the close (✕) of the top-pinned lookup cards — WHOIS / WHOWAS / LUSERS — so on a non-channel window the card cannot be dismissed with the mouse/finger.

Reported by vjt on a `$server` window; measured on `origin/main` (`70366a48`).

## Measurement

- The three lookup cards mount inside `.scrollback-overlay` (`ScrollbackPane.tsx:3528`, `:3534`, `:3542`) — the #133 top-pinned absolute layer, `pointer-events: none` on the container, `auto` on each card.
- `.scrollback-overlay` is `z-index: 5` (`themes/default.css:2553`).
- `.shell-chrome` is `z-index: 41` (`themes/default.css:2459`), zero-height, `justify-content: flex-end` — the lone ☰ overflows it and floats over the pane's top-right corner (#985).
- Each card's ✕ is `margin-left: auto` inside the card header (`.whois-card-close` `:9326`, `.lusers-card-close` `:9637`, same shape for whowas) — i.e. the card's own top-right corner.
- Both land in the same corner, and 41 > 5, so the ☰ wins the hit test.

## Radius is wider than the report

The float renders on **every non-channel window**, not only `$server`: the gate is `Shell.tsx:920` — `selectedChannel()?.kind !== "channel" && !isAdminPaneVisible()`. So a query window has the same collision. On a channel window the ☰ lives in the in-flow `TopicBar` band instead, so nothing overlaps.

This is the same root cause as #1050 (the ✕ of the directory pane), but the remedy there does **not** transfer: #1050 suppresses the row because `/list` does not need the rail. A server or query window does — bucket L says the settings cog must be reachable from every window kind, and on mobile this ☰ is the only door.

## Decision (vjt)

Keep it simple: the card wins. If a lookup card is open, the intent is to close it — reaching the ☰ underneath it is not a goal, because closing the card uncovers the ☰ anyway.

So: raise `.scrollback-overlay` above `.shell-chrome`.

## The non-obvious part

[...]

Issue #1050, opened 2026-08-08
cicchetto: in the /list window the floating ☰ covers the pane close ✕


vjt:
On mobile, in the channel directory window (`/list`), the floating ☰ rail opener sits on top of the pane's own close ✕. The two controls occupy the same corner, and the ☰ wins — it is painted at `z-index: 41` over the pane, so the tap that should leave the directory opens the rail instead.

## Measured on `origin/main` (`70366a48`)

Two independent controls, same corner:

- `.shell-chrome` is `height: 0` + `justify-content: flex-end` + `z-index: 41` (`cicchetto/src/themes/default.css:2459-2467`). Since #985 the row supplies no band: the lone ☰ floats over the pane's top-**right** corner.
- `.directory-close` (the ✕, #125) is the **last** child of `.directory-pane-header` (`cicchetto/src/DirectoryPane.tsx:372-380`), so it lands in that same top-right corner. The header is in normal flow; the ☰ is floated above it.

The overlap is recent, not original. #1039 retargeted the float's margin from its own `0.25rem / 0.5rem` to the topic bar's numbers, now the shared `--pane-chrome-inset-block: 0.5rem` / `--pane-chrome-inset-inline: 1rem` tokens (`default.css:224-225`, applied at `:2485`). That moved the glyph `0.25rem` down and `0.5rem` inward — off the very corner and straight onto the ✕. #1039 was right to make the two window kinds agree; the directory pane is collateral, because it is the one non-channel kind that already puts a control where the float now lands.

This is the **second** time the same #985 consequence has been paid. The admin window hit it first, and `cicchetto/src/ShellChrome.tsx` says so in as many words: the exclusivity between `ShellChrome` and the inline opener "got MORE load-bearing after #985 … the band was made zero-height and the lone ☰ floated over the pane's top-RIGHT corner — which is exactly where the admin pane now puts its close ×". The directory pane is the same shape and was never given the same treatment.

## Decision

Do **not** re-home the opener here: **the rail is not wanted on the directory window** (owner's call, 2026-08-08). The admin precedent moves the button into the pane header; this issue drops it for this kind instead.

Concretely, the gate at `cicchetto/src/Shell.tsx:920`:

```jsx
<Show when={selectedChannel()?.kind !== "channel" && !isAdminPaneVisible()}>
```

[...]

Issue #1039, opened 2026-08-08
cicchetto: the mobile hamburger jumps between window kinds — TopicBar and ShellChrome place it with different offsets


vjt:
On mobile, the ☰ does not sit in the same place when you switch between a channel window and
a window without a topic bar (query, server): on the latter it lands slightly higher and
slightly further right. Switching windows makes the button visibly jump, and it is the one
control that should be nailed down — it is the single door into the rail.

## Why it moves — measured on `origin/main`

There are two different buttons in two different containers. `Shell.tsx:911` states it:
*"EVERY window (channel: TopicBar ☰; non-channel: ShellChrome ☰ above …)"*, and the mobile
branch only mounts `TopicBar` under `<Show when={selKind() === "channel"}>` (`Shell.tsx:823`).

**Channel window** — `.topic-bar-hamburger` inside `.topic-bar`
(`cicchetto/src/themes/default.css:1795`):

```css
.topic-bar { display: flex; align-items: center; gap: 0.5rem; padding: 0.5rem 1rem; … }
```

Right offset: the bar's own `1rem` right padding. Vertical: `align-items: center` centres the
button in a bar whose height is driven by the tallest child — so it is at least `0.5rem` from
the pane top, and lower still whenever the topic wraps to its second line.

**Non-channel window** — `.shell-chrome-rail-opener` inside `.shell-chrome`
(`cicchetto/src/themes/default.css:2330` and `:2348`):

```css
.shell-chrome { position: relative; z-index: 41; height: 0; display: flex;
                align-items: flex-start; justify-content: flex-end; flex-shrink: 0; }
.shell-chrome .shell-chrome-rail-opener { background: var(--bg); margin: 0.25rem 0.5rem 0 0; }
```

Right offset: `0.5rem` of margin. Vertical: `0.25rem` of margin from the top of a zero-height
box, i.e. from the very top of the pane.

So the non-channel ☰ sits **0.5rem further right** and **at least 0.25rem higher** than the
channel one. That is the "slightly up and to the right" exactly.

Neither value is arbitrary in isolation — `#985` floated the lone opener over the pane after
the chrome band was dropped, and the CSS comment at `:2340` calls that margin *"breathing room,
NOT a notch allowance"* — but the two openers were never registered against each other.

## What "fixed" means

The glyph must land on the same pixel in both window kinds, so that switching windows never
moves it. Two things follow from the measurement above:

[...]

Issue #133, opened 2026-06-28
cicchetto: whois/lusers pane scrolls content instead of rendering as overlay


vjt:
## Problem
Opening the WHOIS or LUSERS pane in cicchetto causes the underlying content text to scroll. The pane is laid out inline, pushing/scrolling the message content rather than floating above it.

## Expected
- The WHOIS/LUSERS pane should render as an **overlay** on top of the content, not affect the scroll position of the underlying conversation.
- The **close (X) control should be larger** — currently too small to hit comfortably.

## Notes
Per chan request. Affects readability: returning from a whois/lusers view loses the reader's place in the channel buffer.

vjt:
Shipped to prod (HOT --cic bundle) — main 8cc0ced (fix) + c1d683e (DESIGN_NOTES). WHOIS/WHOWAS/LUSERS cards + peer-away banner now render in one absolute .scrollback-overlay layer floating above the scroll list, so opening one no longer shrinks the buffer or moves the reader's scroll position (root cause: all four were flex:1 siblings before .scrollback — fixed the class, not the named two). Close (×) enlarged to 44px (Apple HIG). pointer-events:none container + auto children = tap-through so a tall card can't cover the compose box. Gates green: 2198 cic tests, real-chromium e2e 3 card specs, tsc+vite build clean. Code-reviewed (subagent).

PR #1052, opened 2026-08-08
cic(#1050, #1051): one corner, two controls, two different remedies

vjt:
Two issues, one branch: **#1050** (the /list pane's ✕) and **#1051** (the
WHOIS / WHOWAS / LUSERS card ✕). Same root, adjacent code, and separating
them would only manufacture a conflict.

## The shared root

Since #985 the non-channel mobile chrome is a zero-height row whose lone
☰ overflows into the pane's top-**right** corner at `z-index: 41`. #1039
then retargeted that float's margin onto the shared
`--pane-chrome-inset-*` tokens so the channel ☰ and the float land on the
same rectangle — correct, and the reason both of these surfaced at once:
the glyph moved `0.25rem` down and `0.5rem` inward, off the very corner
and onto whatever the pane already put there. Two panes already put
something there.

## The remedies are NOT transferable

The gate decides whether the row exists; each window has to answer
whether it wants the rail at all.

| | #1050 `/list` | #1051 server / query |
|---|---|---|
| wants the rail? | no (owner's call) | yes — bucket L, on mobile this ☰ is the only door |
| remedy | suppress the **row** | raise the **card** |
| where | third exclusion in the `Shell.tsx` gate | `.scrollback-overlay` 5 → 42 |

**#1050** — whole row, not a `display: none` on the glyph: the row is
what carries the z-index, and its zero-height box would stay over the
header. Third exclusion beside `channel` (the TopicBar hosts the ☰ in
flow) and `admin` (which re-homed the same button into its own pane
header). Bucket L is knowingly relaxed for this one kind. The gate reads
`selKind()` — the memo the mobile `<Match>` already switches on — rather
than re-deriving the kind.

**#1051** — the cards live in `.scrollback-overlay` (#133) with each ✕ at
`margin-left: auto` in the card header, i.e. that same corner; 41 > 5, so
they could not be dismissed at all. vjt's ruling: the card wins, since
closing it uncovers the ☰ anyway.

## Two things the issues did not name

**The ceiling is 60, not 89.** Below the 89/90 drawer tier is true but
not tight — `.toast-stack` sits at **60** and must keep painting over a
card, or a toast raised while a WHOIS is up is invisible. 42 is the
value: one above the float.

**A second inversion.** `.scrollback-float-stack` (40 — scroll-to-bottom
+ next-active) carried a comment reading *"z-index above
`.scrollback-overlay` (cards)"*. Raising the overlay necessarily lifts it
over that stack too, so a full-height card now covers those buttons until
dismissed. Accepted as the same ruling applied consistently, and the
comment is corrected here rather than left to mislead. The alternative
cascades: lifting the float stack back over the overlay would also clear
`.shell-chrome` at 41 and invert #985's rule instead.

## The guard that could not fail

`shellChromeFloat.test.ts:46` asserts `40 < z(.shell-chrome) < 89` under
the heading *"the floated opener paints above the in-pane scrollback
floats"*, and its comment names **the pinned overlay itself** among those
floats. The assertion is on `.shell-chrome` alone, so raising the overlay
leaves it green while inverting the rule it documents. Both halves are
handled in this change: the comment is rewritten, and a sibling test
compares the numbers directly — `overlay > .shell-chrome`, `<
.toast-stack`, `> .scrollback-float-stack`. Red first, literal:

```
× the pinned card overlay paints ABOVE the floated opener, below the toasts
AssertionError: expected 5 to be greater than 41
  78|     expect(overlay).toBeGreaterThan(zOf(".shell-chrome"));
```

and for the #1050 gate, on the pre-fix tree:

```
× mobile /list window: the standalone .shell-chrome row is SUPPRESSED (#1050)
AssertionError: expected <header …(2)>…(1)</header> to be null
  1076|       expect(container.querySelector(".shell-chrome")).toBeNull();
```

(the `.directory-pane` barrier above it had already passed, so the branch
really had mounted).

## Test plan

- [x] `scripts/bun.sh run test` — **251 files / 4720 tests passed**.
- [x] `scripts/bun.sh run check` (biome + tsc) — **exit 0**.
- [x] `tsc --noEmit -p e2e` — 30 errors, all pre-existing, **none in the
      two new specs** (the e2e tree has no static gate; checked by hand).
- [ ] Integration CI — both new specs are `@webkit`, so they run on
      webkit-iphone-15 only. Not run locally.

**Spec shape follows the defect.** #1050 asserts the tap **closes** the
directory, not that the ✕ is visible — visibility is exactly what the bug
already satisfied, since the float painted over a perfectly visible
button. #1051 asserts `document.elementFromPoint` at the ✕'s own centre,
gated on a prior assertion that the two boxes still **intersect**:
without it, a layout change that merely moved them apart would keep the
spec green while quietly retiring what it guards.

Constraints honoured: `--pane-chrome-inset-*` and the float's margin are
untouched (#1039's outcome, asserted in
`issue1039-hamburger-corner.spec.ts`), and `.directory-close` is neither
moved nor restyled.

Refs #1050, #1051.

### DIFF
--- a/cicchetto/e2e/tests/issue1050-list-window-no-float.spec.ts
+++ b/cicchetto/e2e/tests/issue1050-list-window-no-float.spec.ts
@@ -1,0 +1,89 @@
+// @webkit — #1050. On the mobile /list window the floating ☰ sat on top of the
+// directory pane's own close ✕, and the ☰ won: the tap that should leave the
+// directory opened the rail instead.
+//
+// Two independent controls, one corner. `.directory-close` is the LAST child of
+// `.directory-pane-header` (#125), so it lands top-right in normal flow; since
+// #985 `.shell-chrome` is a zero-height row whose lone ☰ overflows into that
+// same corner at `z-index: 41`. #1039 is what brought them together — it
+// retargeted the float's margin onto the topic bar's `--pane-chrome-inset-*`
+// tokens, moving the glyph off the very corner and straight onto the ✕. The
+// admin window paid this first and was given an inline mount in its own header;
+// the owner's call for the directory is the opposite one, that this window does
+// not want the rail at all, so the whole row goes.
+//
+// WHY A NEW SPEC. Nothing existing covers it: all seven specs that assert
+// `shell-chrome-rail-opener` (issue1039-hamburger-corner,
+// issue71-inc2-mobile-rail-openers, issue985-mobile-floating-opener,
+// ux-4-z-cluster-journey, ux-5-a-hamburger-dedupe, ux-5-bm-mobile-hamburger,
+// ux-5-bt-narrow-chrome-compression) run on a channel, query, home or admin
+// window. The suppression therefore breaks no assertion — and without this one
+// it would come back unnoticed.
+//
+// WHY THE OUTCOME IS "THE ✕ CLOSES", not "the ✕ is visible. Visibility is
+// exactly what the bug already satisfied: the float painted OVER a perfectly
+// visible button. Only the click distinguishes the two.
+//
+// Mobile-only shape — `ShellChrome` has a single mount, inside Shell's
+// `isMobile()` branch — so this runs on webkit-iphone-15 alone via @webkit; the
+// chromium project grepInverts the tag.
+//
+// Parity per `feedback_e2e_user_class_parity_matrix`: a UI shape contract, no
+// subject-shaped branch. The registered seed suffices.
+
+import { loginAs, openRailMenu, selectChannel } from "../fixtures/cicchettoPage";
+import { AUTOJOIN_CHANNELS, getSeededVjt, NETWORK_NICK, NETWORK_SLUG } from "../fixtures/seedData";
+import { expect, test } from "../fixtures/test";
+
+const CHANNEL = AUTOJOIN_CHANNELS[0];
+
+test.setTimeout(90_000);
+
+test("@webkit #1050 — the /list window drops the floating ☰, and its ✕ actually closes the directory", async ({
+  page,
+}) => {
+  const vjt = getSeededVjt();
+  await loginAs(page, vjt);
+  await selectChannel(page, NETWORK_SLUG, CHANNEL, { ownNick: NETWORK_NICK });
+
+  // Reach the directory the way a phone does: the rail's rooms button. This
+  // also proves the rail is still reachable from the window we START on — the
+  // suppression is scoped to `list`, not a general retreat from bucket L.
+  await openRailMenu(page);
+  await page.getByTestId("mobile-panel-list").tap();
+
+  const pane = page.locator(".directory-pane");
+  await expect(pane).toBeVisible({ timeout: 15_000 });
+
+  // THE SUPPRESSION. Whole row, not a hidden glyph: a `display: none` on the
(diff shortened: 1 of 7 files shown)


vjt:
## The red was the spec racing an animation it opened itself. The remedy shape is unchanged.

**The #1050/#1051 split still holds** — nothing about either remedy moved. Only the #1050 spec changed.

### What the artifacts actually say

The row suppression **worked**. All three absence assertions passed:

```
expect(getByTestId("shell-chrome")).toHaveCount(0)              ✓
expect(getByTestId("shell-chrome-rail-opener")).toHaveCount(0)  ✓
expect(locator(".shell-chrome")).toHaveCount(0)                 ✓
```

So the floated ☰ was gone and something *else* won the corner. That something is the **members drawer, which this spec opens itself**.

Reaching `/list` the way a phone does goes through the rail, so `openRailMenu` → `openMembersDrawer`. `openListPanel` then closes it via the #291/#361 nav mutex (`setMembersOpen(false)`) — but on mobile that aside is:

```css
.shell-members { position: fixed; right: 0; z-index: 90;
                 transform: translateX(100%);
                 transition: transform 200ms ease-out; }
```

Closing is a **200 ms slide**, not an instant unmount. For the first tens of ms the drawer still covers the pane's top-right corner at `z-index: 90` — above everything in the pane — and the ✕ sits at x≈364 of a 393 px viewport, well inside an 18 rem drawer pinned right. The three `toHaveCount(0)` assertions in between are already satisfied when they run, so they cost almost nothing, and the probe landed inside the slide. `elementFromPoint` returned the drawer, correctly.

### Two corrections to the triage framing

1. **This is not webkit-specific.** The chromium project does `grepInvert: /@webkit/`, so chromium has *never run this spec*. "Chromium had no equivalent failure" is not a datapoint, and there is no engine difference to start from. The race is in the flow.
2. **The neighbouring `Target crashed` is not implicated** — and I am not claiming it is. 540 ran immediately after and passed, and the mechanism above is sufficient on its own and derived statically from the stylesheet, not from the crash.

### The fix

The assertion is **byte-identical**. What was missing was the pre-state, so it is established before the probe: the drawer is closed *and* its box no longer meets the ✕. Polled on the two rectangles rather than sleeping 200 ms, so the barrier states the condition that matters and cannot rot if the transition is retuned.

The drawer covering the corner mid-slide is the animation *working*. It is not the state #1050 is about, and it is not something to assert away.

### One thing I could not get from the artifacts, now fixed

The first red was unreadable, and this cost a round. `elementFromPoint` returns `null` for a point outside the viewport, so **"covered by an invisible layer" and "pushed off-screen" fail identically** through a bare boolean. The artifacts did not separate them either: the only screenshot is captured post-teardown and showed the page at a scale it never had at the assertion — I chased that for a while before the screencast frame nearest the failure showed a perfectly normal layout with the ✕ exactly where it belongs, which is what proved the covering layer was invisible rather than displaced.

The probe now reports the whole stack via `elementsFromPoint`, plus the rect, the point and the viewport. It costs nothing until something fails and makes the next red diagnose itself.

### Deliberately left alone

**#1051 does not share the race.** It reaches its window with `/q`, never opening the drawer before its probe, and opens it only in the final assertion — which is why it passed. I did not add the barrier there; a barrier for a race that cannot happen is noise.

Rebased onto `080fdc44`.

Issue #772, opened 2026-08-03
cicchetto: compose drafts do not survive a reload


vjt:
## What

An unsent compose draft is lost on every reload of the client. Verified against
`origin/main` at `6fadb5f9`.

`compose.ts` keeps per-channel state — including `draft` — in
`identityScopedStore` signals (`cicchetto/src/lib/compose.ts:228`), and neither
`compose.ts` nor `identityScopedStore.ts` touches `localStorage` or
`sessionStorage`. The state is pure in-memory, so it dies with the document.

## Why it matters now

This is **pre-existing** and applies to every reload path that already exists:
the #674 refresh banner's `Refresh` button, a manual browser reload, and the
#695 two-day stale-resume reload. Nothing here is a regression.

What changes is how easy it is to hit. #674 now applies a new bundle
automatically once the operator has been away past a 10-minute dwell, so
reloads start happening without anyone pressing anything. The dwell exists
precisely because a reload is destructive — but it bounds *when* the loss can
happen, it does not remove the loss.

Note the shape of the risk: the dwell means the operator was away for at least
ten minutes, so a draft they abandon mid-sentence and walk away from is exactly
the draft that gets discarded. #666 is the sibling case where losing or
duplicating draft content is itself the reported bug.

## Proposed shape

Persist the draft per `(identity, channel)` and restore it on boot. Scroll
position is the natural companion but is a separate question — the resume
scroll already has its own authority in `ScrollbackPane` (#535 / #608), so
folding drafts into that would cross two owners.

Open questions for whoever picks this up, not decided here:

- Storage tier. `localStorage` outlives the window but leaks a draft between
  tabs of the same identity; `sessionStorage` is per-window, which matches how
  #695 reasoned about the stale-resume stamp. The two answer different
  questions and the trade-off should be made explicitly, not inherited.
- Eviction. Drafts are unbounded in count and length; something has to decide
  when an abandoned draft for a channel the user never reopens goes away.
- Whether a restored draft should be visually marked as restored, or simply
  appear as if nothing happened.

If drafts become non-destructive, the #674 dwell mostly stops mattering and
auto-update could become the default — that is the more thorough fix the #674
description gestured at.

vjt:
**Approved by vjt (2026-08-06) with the shape fixed.**

> *"ok add and keep drafts in sessionstorage. and then we could do as well always auto reload but let's keep 674 in its current shape for now. if I am active I may not want a reload."*

[...]

PR #1054, opened 2026-08-08
cic(#954): a destroyed document is not a failed send

vjt:
Implements vjt's ruling on #954 (08:30Z): `handBack` must not fire on the one
rejection class where the server may already own the message. Narrow fix — no
delivery-confirmation protocol, no idempotency key, no correlation token. Part
(2) (the muted "sending" row) is untouched: it is undecided and nothing here
prepares for it.

## The defect

The #904 pump takes the text out of the composer at dispatch and owes it back
on every failing path. When the DOCUMENT is destroyed mid-flight — a reload,
the #674 auto-refresh applying itself, a closed tab — the POST aborts, but the
server may already have accepted and persisted it. Handing the text back
re-arms the composer, and #772 mirrors the draft into sessionStorage, so it
survives the reload: the operator lands on a page with the message in the
scrollback AND staged in the composer. One distracted Enter sends it twice.

## The discriminator is a lifecycle event, not the error

`fetch` rejects with the same `TypeError: Failed to fetch` for a dead Wi-Fi, a
DNS death, a CORS refusal, a vanished server AND a destroyed document. Matching
on that string would silently swallow real send failures — the exact opposite
of what the issue asks for. So the answer comes from a document-lifecycle event
this codebase already has a seam for: `pagehide` + `beforeunload`, the pair
`main.tsx` registers for the S3.3 away-hint. New module `lib/documentTeardown.ts`.

**A counter, not a boolean**, because a flag is wrong twice over:

* `pagehide` also fires on bfcache entry and the iOS PWA freeze, and those
  documents come back — a latch would condemn every send failure for the rest
  of their life;
* clearing that latch on `pageshow` does not fix it, because a frozen
  document's in-flight fetch rejects on the THAW, i.e. after `pageshow` already
  cleared the flag, so the one case this exists for would read false.

Comparing the epoch ACROSS a flight asks the only question the pump has. Both
directions are pinned by tests. `visibilitychange` is deliberately not a
trigger: a backgrounded tab keeps its fetches.

**The queued line is still handed back.** Only the line in the air can be owned
by the server; the one-deep #904 queue behind it never left the client, so
dropping it would lose a message outright rather than trade a duplicate for one.

## Where the trade stops being free — the limit I am NOT claiming away

The premise "the message very probably landed" holds and holds tight in the
reachable regime: the #954 harness (real Bandit listener, real TCP kill at a
controlled offset, N=20/row, three runs) measured 20/20 persisted at every
offset from +1 ms onward in both close modes, 95% ceiling on non-persistence
0.32%.

**It does not cover the tip.** An RST landing at +0 ms after the last body byte
was **0/20 — the message did not land**, and dropping the text there LOSES it.
Whether a destroyed document closes FIN or RST is UNMEASURED, and a killed tab
plausibly looks like RST. That is a real loss in a sub-millisecond window on an
otherwise idle path. It is recorded in the code comment where the drop happens
and in DESIGN_NOTES, and deliberately NOT engineered around: doing so is the
delivery-confirmation protocol the ruling excludes.

## What I refused to claim

* **That this is proven end to end in a browser.** It is proven at the unit
  seam: `pagehide` fired inside a real in-flight window, then the abort. No e2e
  spec drives an actual mid-POST document destruction, so "a real reload during
  a real send behaves this way" is inference from the seam, not a measurement.
* **That the drop is always safe** — see the RST +0 ms row above.
* **That the class is fully closed.** See below.

## The half I found and did NOT build

A paced multi-line drain (#666/#737) mirrors its OWN residue into the draft and
never reaches the pump's `handBack` (`keptBuffer` leaves `inHand` null), so an
aborted line inside a paste is **still re-armed**. Same hazard, second writer.

I wrote that fix and then reverted it, because it is not the same rule: the
drain's error mix includes explicit 429 refusals, which prove the line did NOT
land and must never be dropped, and the retry-cap path exits non-paced while
still holding a 429. Making the drop correct there means a second, different
predicate for one question — two patterns. Better as a decision of its own than
a guess made in passing. Flagged in the code comment, in DESIGN_NOTES, and here.

## Where I think the brief may be wrong

Nowhere on the ruling — it is right, and the measurement backs it. One note on
the brief's framing: it treats "look for existing infrastructure first" as
possibly finding a ready-made signal. There isn't one (`lifecycle.ts` is the
SESSION lifecycle, `documentVisibility.ts` deliberately excludes teardown), so
this adds a module rather than reusing one. That is the answer to question 1,
not a skipped step.

## Gate

`scripts/bun.sh run check` → exit 0 (biome clean of new findings, tsc ran and
passed — the 61 warnings are the pre-existing CSS specificity set).
`scripts/bun.sh run test` → exit 0.

Test count diff, measured both ways rather than computed: **4723 → 4733**
(+10), files **251 → 252**. The baseline run had my two test additions moved
aside with the production fix still in place, so it also shows the fix causes
**zero regressions** across the 4723 pre-existing tests.

**The red was read before the green.** With the guard neutralised to its
pre-fix behaviour (`const destroyedInFlight = false`), the #954 block reports
`3 failed | 2 passed`: the three positives fail (the in-flight line comes back,
it survives the reload, and it displaces the queued line), and the two
negatives PASS — an ordinary `Failed to fetch` with no teardown still hands the
text back, and a teardown that predates the dispatch does not condemn it. A
guard that cannot fire wide, proven by the two that stayed green.

### DIFF
--- a/cicchetto/src/__tests__/compose.test.ts
+++ b/cicchetto/src/__tests__/compose.test.ts
@@ -2192,6 +2192,142 @@ describe("#904 — one-deep send queue", () => {
   });
 });
 
+// #954 — the #904 pump owes the composer its text back on every failing path.
+// One "failure" is not one: a POST aborted because the DOCUMENT was destroyed
+// (reload, the #674 auto-refresh, a closed tab) may already have been accepted
+// and persisted by the server. Handing that text back re-arms the composer —
+// and #772 mirrors the draft into sessionStorage, so it survives the reload
+// and greets the operator with a line the channel is already showing. One
+// distracted Enter sends it twice.
+//
+// The abort itself is indistinguishable: `fetch` rejects with the same
+// `TypeError: Failed to fetch` for a dead Wi-Fi, a vanished server and a
+// destroyed document, so these tests drive the DOCUMENT-LIFECYCLE event
+// (`pagehide`), never an error string.
+describe("#954 — a send aborted by the document going away is not re-armed", () => {
+  const k = channelKey("freenode", "#a");
+
+  // Same controlled-ack shape the #904 block uses: the in-flight window is a
+  // real observable state, so the teardown can be placed INSIDE it.
+  const deferredSend = async () => {
+    const sb = await import("../lib/scrollback");
+    const acks: Array<{ body: string; resolve: () => void; reject: (e: unknown) => void }> = [];
+    vi.mocked(sb.sendMessage).mockImplementation(
+      (_slug: string, _target: string, body: string) =>
+        new Promise<void>((resolve, reject) => {
+          acks.push({ body, resolve, reject });
+        }),
+    );
+    return acks;
+  };
+
+  // What a destroyed document does to an in-flight POST: the page teardown
+  // fires, then the request rejects as a bare network failure.
+  const abortedByTeardown = (ack: { reject: (e: unknown) => void } | undefined): void => {
+    window.dispatchEvent(new Event("pagehide"));
+    ack?.reject(new TypeError("Failed to fetch"));
+  };
+
+  it("drops the in-flight line instead of putting it back in the composer", async () => {
+    localStorage.setItem("grappa-token", "tok");
+    const acks = await deferredSend();
+    const compose = await import("../lib/compose");
+
+    compose.setDraft(k, "probe954 already in the channel");
+    const done = compose.submit(k, "freenode", "#a");
+    await Promise.resolve();
+
+    abortedByTeardown(acks[0]);
+    expect(await done).toHaveProperty("error");
+
+    expect(compose.getDraft(k)).toBe("");
+  });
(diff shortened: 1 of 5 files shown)


vjt:
Superseded by #1058, which merged at `2c9a7ece`.

Its commits are in main, cherry-picked verbatim onto the main that actually received them and gated **together** with the other branch rather than each against a base neither had seen. The cherry-pick rewrote the SHAs, which is why GitHub cannot close this automatically and why it is being closed by content instead.

Verified before merging, per file: every code file in #1058 matches this PR's own insertion/deletion counts exactly, and `docs/DESIGN_NOTES.md` — the only file both branches touched — came out at 62 + 91 = 153, the exact sum. Nothing was lost or duplicated at the seam.

`integration` on the union: **658 tests, 658 passed.**

Issue #671, opened 2026-08-02
Auto-away never arms when the last visible socket goes stale before it dies


vjt:
## Summary

Auto-away stops firing when the last visible socket goes **stale before it dies**. Reported from the field: a user's device was left overnight and the peer never showed as away for hours.

The `#318` read-time staleness filter was scoped to push suppression only, but it also sits on the `before?` side of every auto-away transition — so it silently removes the very transition the FSM keys off.

## The chain

1. `WSPresence.emit_transition/4` only emits `:ws_all_hidden` on a genuine `true → false` flip of `any_visible?`.
2. `any_visible_in?/2` (`lib/grappa/ws_presence.ex:598-606`) counts a `:visible` pid **only while fresh**: `vis == :visible and fresh?(last, now, state.stale_ms)`.
3. `fresh?/3` (`ws_presence.ex:613-615`) is `now - last < stale_ms`, with `@default_stale_ms 60_000` (`ws_presence.ex:172`).
4. Staleness is **read-time and emits nothing**. A device that stops reporting — phone asleep, JS timers suspended, lid closed — ages out silently after 60s. No event, no transition.
5. When the socket actually dies, `handle_info({:DOWN, ...})` (`ws_presence.ex:568-586`) computes `before? = any_visible_in?(...)` — which is now **already `false`** because the pid is stale. `after?` is also `false`. `emit_transition` takes the `true -> :ok` branch: **no `:ws_all_hidden`**.
6. `Session.Server` therefore never arms the debounce (`lib/grappa/session/server.ex:2623-2635`), so `:auto_away_debounce_fire` never runs and no upstream `AWAY` is ever sent.

Net effect: the window in which auto-away can be armed is only as wide as `stale_ms`. The socket must die **within 60s of the last visibility report**, otherwise auto-away is lost permanently for that session.

## Same hole on the other two entry points

- `client_closing/2` (`ws_presence.ex:515-530`) computes `before?` the same way. A `pagehide` arriving after the pid already went stale produces no transition either.
- `set_visibility(user, pid, false)` — same, the pid is already discounted, so hiding it is a no-op transition.

## The docstring asserts the opposite

`ws_presence.ex:62-64`:

[...]

Issue #318, opened 2026-07-18
cicchetto/push: web push suppressed for iOS PWA backgrounded/closed but still :visible in WSPresence


vjt:
## Summary

Web Push is never delivered to an iOS cicchetto PWA user after they background/close the app, because their WebSocket session stays `:visible` in `Grappa.WSPresence`. Push dispatch treats a `:visible` session as actively present and suppresses the notification, so a user who has genuinely left the app receives nothing.

## Repro (iOS Safari PWA)

1. Install cicchetto to the home screen (standalone PWA).
2. Enable notifications: channel messages (all) restricted to a channel + mentions.
3. Background or close the PWA.
4. From another client, post in that channel (a plain line and/or a mention of the user).
5. Expected: a push notification. Actual: nothing arrives.

## Root cause (verified on prod)

Live-node + DB introspection confirmed the config is correct and the suppression is presence-driven:

- **`push_subscriptions`**: the user's subscription is registered (iPhone Safari UA) but `last_used_at` is `NULL` — no push was ever delivered to it. Other subscribers to the same channel had their `last_used_at` bumped at the exact moment of the test message, so dispatch *did* run and simply skipped this session.
- **`user_settings.notification_prefs`** persisted correctly: `channel_messages_all: true`, `channel_messages_only: ["<channel>"]`, `channel_mentions: true`.
- **`Grappa.WSPresence`** state for the session: exactly one socket, visibility `:visible`, `any_visible? = true` — even though the PWA had been closed/backgrounded.

So the session never transitioned to `:hidden` when the app was backgrounded/closed on iOS. `visibilitychange → hidden` is unreliable on the iOS PWA lifecycle (Safari does not reliably fire it on background/terminate). The socket stays `:visible`, presence believes the user is watching, and push is suppressed.

## Mitigation direction

Identify sessions that are not genuinely foreground/visible (backgrounded/closed PWAs whose `visibilitychange` never fired) and stop treating them as present for push-gating. Options, not mutually exclusive:

- **Client (cicchetto):** in addition to `visibilitychange`, listen to `pagehide` / `freeze` / `blur` and flip presence to hidden; and/or emit a client heartbeat the server can time out.
- **Server (grappa):** treat `:visible` as authoritative only while *fresh* — if a `:visible` socket has produced no client-confirmed activity/heartbeat within N seconds, downgrade it to `:hidden` for push-gating so a stale-visible socket cannot indefinitely suppress push.

## Acceptance

[...]

Issue #224, opened 2026-07-12
Auto-away permanently blocked by a stale :visible WS socket when the hidden/close signal is lost


vjt:
## Summary
On prod, a user showed as present/online indefinitely with no client foregrounded. Root cause: `WSPresence` held a socket pid marked `:visible` that had last reported visible ~2 days earlier and never reported `:hidden`. Auto-away only fires on the `any_visible?/1` `true → false` transition, so the stuck-visible socket blocked auto-away **forever** (and, keying off the same signal, push-suppression too).

## Why it got stuck
- The socket was still alive (TCP up) so its pid never `:DOWN`-ed.
- The client never sent `visibilitychange=hidden`, `pagehide`, or `client_closing` for it — consistent with the tab being killed by the browser/OS, the machine sleeping, or a half-open TCP where neither the unload event nor the socket teardown reached the server.

## Not a client-wiring bug
The client handlers exist and are correct (`cicchetto/src/main.tsx`: `pagehide` + `beforeunload` → `client_closing`; `visibilitychange` → `set_visibility`). The gap is **server-side**: nothing reconciles a socket stuck `:visible` with no recent liveness. A lost hidden/close signal is unrecoverable until pid `:DOWN`, which may never arrive for a half-open or idle-but-connected socket.

## Proposed directions (design open)
- Treat `:visible` as needing periodic re-affirmation (a visibility heartbeat), decaying a pid to `:hidden` if not refreshed within a window.
- Tie `WSPresence` liveness to the Phoenix channel heartbeat / a transport idle-timeout, so a non-heartbeating socket is dropped and `:DOWN`-cleared. Verify the websocket transport `:timeout` is actually configured.
- On `set_visibility(:visible)`, stamp a timestamp; a periodic sweep flips long-stale visibles to `:hidden`.

## Impact
User never goes away; auto-away and push-suppression both key off a stale signal.

## Field evidence
A live prod socket sat `:visible` for ~2 days (created ~2 days before observation, ~450k reductions), blocking auto-away across both of the user's upstream networks. Killing that one socket immediately let `any_visible?/1` go false, and both `Session.Server`s transitioned to `:away_auto` within the debounce window — confirming the socket was the sole blocker.

## Repro
Connect a desktop client that reports visible, then terminate it without firing any unload/visibility event (kill the browser process, or sleep the machine) while the TCP stays half-open. The socket lingers `:visible` and auto-away never fires.

vjt:
[...]

PR #1057, opened 2026-08-08
fix(#224): a stale :visible socket is demoted, not waited on

vjt:
## The defect

A prod socket sat `:visible` for ~2 days with nothing foregrounded. Auto-away
fires on the `any_visible?/1` `true → false` TRANSITION, and all three doors
that compute it (`set_visibility`, `client_closing`, pid `DOWN`) need a write
that never came. Read-time staleness (#318) already discounted the pid for PUSH
suppression, but it emits nothing when a pid merely ages out, so the FSM never
heard about it. Killing that one socket by hand let both `Session.Server`s reach
`:away_auto` within the debounce — it was the sole blocker.

## The question that decided the design, and its answer

The issue lists three directions, and which one is right turns on a fact nobody
had established: **was the stuck socket still heartbeating?** If not, a
transport idle-timeout would have reaped it and that is the cheap structural
fix. If yes, no timeout can ever help.

**It was heartbeating.** Not because of the "~450k reductions" the issue offers
— that number settles nothing, since reductions accrue from work done FOR a
socket too and every push fanned out to that user passed through it. The answer
comes from structure:

- Phoenix defaults the websocket `:timeout` to **60_000 ms**
  (`Phoenix.Transports.WebSocket.default_config/0`), and
  `Phoenix.Socket.Transport.load_config/2` merges the endpoint's `websocket:`
  list **over** that default. `endpoint.ex` sets no `:timeout`, so 60s is in
  force. "Not configured" is true of the source and false of the behaviour —
  worth flagging, since the triage comment on the issue reads the absence as
  meaning no timeout exists.
- Bandit honours it: `Bandit.WebSocket.Handler.handle_connection/2` returns
  `{:persistent, timeout}`, armed by ThousandIsland as a `:read_timeout` timer.
- That timer resets **only** via `handle_continuation/2`, reached from the
  `:tcp`/`:ssl` inbound clause. Bandit's generic `handle_info/2` — where every
  outbound PubSub push arrives — returns `{:noreply, …}` and never touches it.

So outbound traffic cannot hold a socket open, and one alive for two days was
receiving inbound frames at least every 60s: phx heartbeats, i.e. live JS
timers. **Direction 2 is not unbuilt, it is unavailable** — the mechanism
already exists, already runs, and correctly declined to fire. The connection was
never what died; the visibility report was.

(`WSPresence`'s own #318 "efficacy caveat" makes the same argument at 90 minutes.
I reached it from the transport source before reading that line, which is why
it is recorded as a derivation rather than a citation.)

## The fix

A periodic tick demotes each no-longer-`fresh?/3` `:visible` entry to
`{:hidden, nil}` and runs the **same** `emit_transition/4` door as every other
write. No new state: the stamp, the predicate, the window and the door all
existed. Three non-obvious points:

- **Demote, don't just emit.** Writing the entry is what makes a re-tick
  idempotent; the alternative is an "already emitted" flag beside it — the
  parallel bookkeeping that drifts. Measured, not asserted: the emit-only mutant
  re-fires `:ws_all_hidden` on every tick, forever.
- **The demotion IS the disarm.** A returning client reports `:visible` against
  a `:hidden` entry, which is a real `false → true` flip and cancels the
  debounce. Against a still-`:visible` entry that report emits nothing — so a
  sweep that only emitted would arm auto-away with nothing able to cancel it.
  Arm and disarm are one change.
- **#671's RAW/FRESH split survives.** Tempting to read the sweep as making the
  fresh view fully evented and collapse the two predicates. It does not: a pid
  that goes stale and dies before the next tick is removed by `DOWN` before the
  sweep sees it, and only the RAW `before?` there still produces its transition.
  Collapsing resurrects the #671 bug in a window one tick wide.

`:sweep_ms` defaults to `:stale_ms` (prod sets neither), so worst-case detection
is 120s against a 600s debounce. It is separately injectable for exactly one
reason, stated in the code: `WSPresence` is an app-wide singleton under
`max_cases: 1`, so a tick landing inside another test's `mark_stale_for_test/2`
window would demote a pid that test still expects reported-visible — a
cross-test failure with no local cause. `config/test.exs` parks it.

## Tests

`test/grappa/session/server_test.exs` — the user-visible outcome, the full chain
to a real `AWAY :auto-away` line on the wire: a device reports visible, then
just STOPS affirming it (no hidden report, no close, no DOWN), and auto-away
arms and fires. It asserts the debounce timer is `nil` BEFORE the sweep, so the
post-sweep assertion cannot pass on a pre-armed timer.

Six unit cases in `ws_presence_test.exs` cover demotion, idempotence, the
disarm, the fresh-sibling case, and the timer arming **and re-arming** (its own
short-interval instance; nothing sends `:sweep`, so only the periodic timer can
produce the events).

**The red was read, on two mutants, after committing the fix.**

- **M1, sweep is a no-op:** `35 tests, 5 failures`. The five are every new
  positive; the survivor is "a FRESH `:visible` socket survives the sweep
  untouched" — the negative that must stay green, proving the guard cannot fire
  wide. All 29 pre-existing cases stayed green, so the sweep regresses nothing
  and nothing pre-existing was already covering it. The e2e case fails with
  `left: nil` on the debounce timer.
- **M2, emit but don't write:** `35 tests, 4 failures` — a different set,
  killing exactly the cases that justify demotion over emission.

M2 caught a vacuous test of mine: the fresh-sibling case originally killed the
STALE socket and refuted `:ws_all_hidden`, which is a no-op transition in all
three worlds. It passed against both mutants and proved nothing. Rewritten to
kill the FRESH one, which only goes all-hidden if the stale sibling really was
written — that is the second commit.

## Gate

`scripts/check.sh` → **exit 0** (captured, not inferred): Credo `5318 mods/funs,
found no issues`, Sobelow, `Doctor validation has passed!`, Dialyzer
`Total errors: 0, Skipped: 0`, `8 doctests, 56 properties, 5550 tests, 0
failures`, bats. First run failed Credo on two unregistered Logger metadata
keys; fixed by using the existing `:user` and declaring `:silent_for_ms`.

No `scripts/integration.sh` run — the STACK lane is w2's and I did not take it.
I do not believe this needs one: the change is server-internal with no wire,
REST or cic surface, and the outcome is already proven to a real IRC `AWAY` line
against the in-process `Grappa.IRCServer` fake, in the COMPILE lane. If you
disagree, schedule it rather than take my word.

## What I refused to claim

- **That the field socket's client was foregrounded.** I established that it was
  sending inbound frames, which is what kills direction 2. WHY the visibility
  report stopped while phx heartbeats continued is not established — a lost
  push during a channel rejoin fits, so does a `visibilitychange` that never
  fired. The fix does not depend on knowing, which is the point of making it
  freshness-based rather than edge-based.
- **That this was reproduced against the real prod socket.** No worker touched
  m42. Everything here is repo-cold plus the local suite.
- **That the demotion is free.** A demoted entry is indistinguishable from a
  page that reported hidden, so `/admin/ws_presence` can no longer tell "swept"
  from "reported". A third visibility atom would preserve it at the cost of
  landing on every consumer of a closed type; instead the `:info` line carries
  the age. If the diagnostic matters more than I judged, that is the knob to
  turn.
- **That `2 × stale_ms` is a measured latency.** It is an upper bound from the
  construction, not an observation.

## Where I think the brief is wrong

Two places, both small.

1. It frames the "~450k reductions" warning as a reason the question might be
   unanswerable, and offers "design for both cases" as the legitimate fallback.
   The warning is right about the reductions and I did not use them — but the
   question turned out to be answerable from a different direction entirely
   (the timeout's reset path), which is stronger than either horn. The lesson I
   would draw is narrower than the brief's: reading code gives you structure,
   and here the STRUCTURE was sufficient because the survival of the socket was
   itself the measurement. Two days of survival under an enforced 60s idle
   timeout is a fact about magnitude that the code makes readable.
2. §4 says "a periodic sweep that flips a long-stale `:visible` to `:hidden` is
   cheap". True, but it undersells what makes it correct: the flip is not a
   tidy-up, it is the carrier of BOTH transitions. Had I built the "cheap"
   version as described — sweep, emit, leave the entry alone — auto-away would
   arm and never disarm, which is a worse bug than the one being fixed. The
   write is the design, not the housekeeping.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -559,7 +559,13 @@ config :logger, :console,
     :stop_ms,
     :spawn_ms,
     :welcome_ms,
-    :autojoin_ms
+    :autojoin_ms,
+    # #224 — how long each demoted socket had been silent when the
+    # WSPresence stale sweep flipped it to :hidden. The demotion makes the
+    # entry indistinguishable from a page that reported hidden, so the
+    # `/admin/ws_presence` snapshot can no longer show it; this line is
+    # where that age survives.
+    :silent_for_ms
   ]
 
 import_config "#{config_env()}.exs"
--- a/config/test.exs
+++ b/config/test.exs
@@ -282,3 +282,13 @@ config :grappa, Grappa.Vault,
       {Cloak.Ciphers.AES.GCM,
        tag: "AES.GCM.V1", key: Base.decode64!("Ot80AYbRqJG9htfEztMBqz6Eo9ALMWgu9Ze6w0CbbPg="), iv_length: 12}
   ]
+
+# #224 — park the WSPresence stale-demotion sweep for the whole run.
+# WSPresence is an application-wide singleton (see the `max_cases: 1`
+# note above), so its periodic tick is NOT scoped to the test that cares
+# about it: a tick landing inside a `mark_stale_for_test/2` window would
+# demote a pid the #671 cases still expect to be REPORTED-visible, i.e. a
+# rare cross-test failure with no local cause. Tests that exercise the
+# sweep drive it deterministically — either `send(pid, :sweep)` against
+# the singleton, or their own short-interval instance.
+config :grappa, Grappa.WSPresence, sweep_ms: 3_600_000
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -34049,3 +34049,94 @@ perfectly visible button. For #1051 it is `document.elementFromPoint` at
 the ✕'s own centre, preceded by an assertion that the two boxes still
 INTERSECT: without that precondition a layout change that merely moved
 them apart would keep the spec green while retiring what it guards.
+## 2026-08-08 — #224: the socket was alive, so no transport timeout could reach it
+
+A prod socket sat `:visible` for ~2 days with nothing foregrounded.
+Auto-away fires on the `any_visible?/1` `true → false` TRANSITION, and
+the three doors that compute it (`set_visibility`, `client_closing`, pid
+`DOWN`) all need a write that never came — the hidden/close signal was
+lost and the socket never died. Killing that one socket by hand let both
+of the user's `Session.Server`s reach `:away_auto` within the debounce,
+so it was the sole blocker.
+
+**The question that decided the design: was the stuck socket still
+heartbeating?** The issue offered "tie liveness to the transport
+idle-timeout" as a cheap structural fix, and "~450k reductions" as if it
+settled the question. It does not — reductions accrue from work done FOR
+a socket too, and every push fanned out to that user passed through it.
+Magnitude cannot be read off code.
+
+It is settleable from structure, though, and the answer is yes:
+
+- Phoenix's websocket `:timeout` defaults to **60_000 ms**
+  (`Phoenix.Transports.WebSocket.default_config/0`), and
+  `Phoenix.Socket.Transport.load_config/2` merges the endpoint's
+  `websocket:` keyword list OVER that default. `endpoint.ex` sets no
+  `:timeout`, so 60s is in force — "not configured" is true of the
+  source and false of the behaviour.
+- Bandit honours it: `Bandit.WebSocket.Handler.handle_connection/2`
+  returns `{:persistent, timeout}`, which ThousandIsland arms as a
+  `Process.send_after(self(), :read_timeout, …)`.
+- That timer is reset ONLY from `handle_continuation/2`, which is
+  reached from the `:tcp`/`:ssl` inbound-data clause. Bandit's generic
+  `handle_info/2` — the one every outbound PubSub push arrives through —
+  returns `{:noreply, …}` directly and never touches it.
+
+So outbound traffic cannot hold a socket open. A socket that survives two
+days received inbound frames at least every 60 seconds, which on this
+wire means phx heartbeats, which means the client's JS timers were
+running. The connection was never what died; the visibility report was.
+**Direction 2 is not merely unbuilt, it is unavailable** — the mechanism
+already exists, already runs, and correctly declined to fire.
+
+(The same argument, at 90 minutes rather than 2 days, is already written
+into `WSPresence`'s #318 "efficacy caveat". It was reached independently
+here from the transport source, which is why it is recorded as structure
+rather than as a quote.)
+
+**The fix.** A periodic tick demotes each no-longer-`fresh?/3`
+`:visible` entry to `{:hidden, nil}` and runs the SAME
+`emit_transition/4` door as every other write. Nothing new is stored:
+the stamp (`last_visible_at`), the predicate (`fresh?/3`), the window
+(`stale_ms`) and the door all predate it.
+
+Three things that are not obvious:
+
+- **Demote, don't just emit.** Writing the entry is what makes a re-tick
+  idempotent — the alternative is a flag beside the entry saying "already
+  emitted", i.e. exactly the parallel bookkeeping that drifts. A mutant
+  that emitted without writing re-fired `:ws_all_hidden` on every tick,
+  forever.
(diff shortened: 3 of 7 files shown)


vjt:
## Asked: can the #318 90-minute caveat and this derivation ever give different answers?

Short version: **not about the fact they share, but yes about what it licenses** — and chasing that turned up a residual case this PR does not close. Worth writing down.

### They cannot disagree about "was it heartbeating"

It is one inference at two magnitudes: survival ≫ one 60s idle period ⇒ inbound frames were arriving ⇒ the client's JS was running. 90 minutes is ~90 periods, two days is ~2880. The only regime where the inference gets weak is a socket observed alive for *roughly one* period — a fresh connection that has sent its handshake frames and is about to be reaped — and neither observation is anywhere near it. So no threshold exists at which one holds and the other does not.

### But they are about two different timers, and #318's caveat conflates them

Checked rather than assumed:

- **phoenix.js heartbeat** — `heartbeatIntervalMs || 3e4` (`phoenix.mjs:1124`), cic overrides nothing. Socket-level, runs unconditionally while the socket is open, knows nothing about visibility.
- **`visibilityHeartbeat.ts`** — also 30s, but `setVisible(false)` calls `stop()` and `clearInterval`s it. It runs **only while the page believes it is foreground**, and `reportVisibility` re-reads the live property each tick.

So socket survival proves the *phx* timer ran. It does not prove the *visibility* timer would have. A correctly-behaving backgrounded client keeps the socket alive and sends no visibility reports at all — that is not a malfunction, it is the design.

#318's caveat ("implying phx heartbeats (hence JS timers) kept running while backgrounded", used to doubt whether a backgrounded PWA really goes quiet) reads the one timer as evidence about the other. Its inference is sound; the conclusion it is used to support is stronger than the evidence carries. The field shape needs no client bug and no half-open TCP: a tab backgrounds, its `set_visibility(false)` never reaches the server (lost push during a rejoin, or `visibilitychange` never fired), the visibility heartbeat correctly stops, and phoenix.js keeps the socket up for days.

### The residual this exposed — the class is NOT fully closed

Both mechanisms — #318's read-time filter and this sweep — key on **a report that stopped arriving**. Neither touches a client that is *still reporting, wrongly*: if `document.visibilityState` never flips to hidden (the original #318 iOS root cause), `reportVisibility` re-reads it, still sees `visible`, and re-stamps every 30s. The entry never ages out, never gets demoted, and auto-away stays blocked exactly as before.

- The **#224 socket is not in that class** — its last report was ~2 days old, so it had stopped. This PR covers the reported defect.
- The **#318 socket may have been** — that one self-healed at ~90 min when the socket finally died. Today it would still block auto-away, because nothing here helps a fresh-but-lying report.

So the two windows can describe genuinely different animals, and that is the honest answer to the question: not a contradiction, a scope boundary. Closing the lying-report case means a signal that does not originate from the page's own belief about its visibility — which is a different design (and arguably a different issue), not a wider `stale_ms`.

I am **not** claiming to know which class the #318 socket was in. That is off-device behaviour I cannot observe from here, and no worker has touched prod. What I can say from the code is that the two classes exist, are distinguishable in principle (`age_ms` on `/admin/ws_presence` separates "silent" from "re-stamping"), and only one of them is fixed here.

vjt:
Superseded by #1058, which merged at `2c9a7ece`.

Its commits are in main, cherry-picked verbatim onto the main that actually received them and gated **together** with the other branch rather than each against a base neither had seen. The cherry-pick rewrote the SHAs, which is why GitHub cannot close this automatically and why it is being closed by content instead.

Verified before merging, per file: every code file in #1058 matches this PR's own insertion/deletion counts exactly, and `docs/DESIGN_NOTES.md` — the only file both branches touched — came out at 62 + 91 = 153, the exact sum. Nothing was lost or duplicated at the seam.

`integration` on the union: **658 tests, 658 passed.**

Issue #1070, opened 2026-08-08
sender_nick/1 drops the {:nick,…}/{:server,…} tag, so consumers cannot tell a server prefix from a user one


abonforti:
`Message.sender_nick/1` collapses a server prefix and a user prefix into the same bare
string, and that string is what gets persisted as `sender` and pushed to clients. The
distinction is present at parse time and dropped one function later, so no consumer can
recover it.

`lib/grappa/irc/message.ex:134-137`:

```elixir
def sender_nick({:nick, nil, _, _}), do: @anonymous_sender
def sender_nick({:nick, nick, _, _}), do: nick
def sender_nick({:server, server}), do: server
def sender_nick(nil), do: @anonymous_sender
```

The `{:nick, …}` / `{:server, …}` tag is exactly the information a client needs, and the
`user` / `host` in the nick tuple are discarded alongside it.

### Why it matters downstream

A client rendering scrollback back into IRC has to choose a prefix shape, and the shape is
the *only* thing on the wire that tells another client whether a NOTICE came from a user
or from a server — they route the two differently, and nick-derived tags and highlights
hang off the user form.

With `sender` alone, that choice cannot be made correctly. Concretely, on the `$server`
window: `persist_server_notice/2` (`lib/grappa/session/event_router.ex:3007`) files MOTD
and INFO numerics as

```elixir
build_persist(state, :notice, "$server", sender, body, %{})
```

`:notice` with a *server* sender — while a private NOTICE addressed to the user's own nick
lands on the same window, same kind, with a *user* sender. The two are indistinguishable
by the time a consumer sees the row.

This is not hypothetical: bicchierino (an IRC-facade bridge to grappa) got it wrong in
both directions within a day — first emitting a bare nick for user notices, then a
fabricated `nick!user@host` for server ones, because the only available signal was the
kind. It now guesses with "sender contains a dot ⇒ server", which happens to be exact
against bahamut (`isvalid()` in `include/common.h` excludes `.`, and `do_nick_name`
truncates at it) and is a guess anywhere else.

cicchetto does not hit this because it renders its own UI rather than reconstructing IRC
lines — the gap only shows up for consumers that have to put a prefix back on the wire.

### Suggested shape

Additive, no schema change: put the tag in the row `meta`, where `build_persist/6`
already threads a map through `put_sender_prefix/5`. Something like
`meta.sender_kind: "user" | "server"`, and — if it is welcome — `meta.sender_user` /
[...]

PR #1071, opened 2026-08-08
feat(#1070): carry the sender's wire shape in meta.sender_kind

abonforti:
Addresses #1070 (closed by hand after release, per repo convention).

### What it does

Adds `sender_meta/1` next to the existing `prefix_userhost/1` and threads it through the
three NOTICE persists — the channel arm, the non-channel arm (merged with `ctcp_meta/1`
so the CTCP classification is untouched), and `persist_server_notice/2`.
`sender_user` / `sender_host` come along on the nick branch, since they fall out of the
same tuple and are already allowlisted.

### Two things "additive" did not mean, both found by the suite

**`meta` is allowlisted.** `Meta.@known_keys` rejects an unregistered key, so the
changeset failed and the row never persisted. That surfaced as *a broadcast that never
arrived* — six tests failing with "no matching message after 2000ms" — rather than as a
meta mismatch, which took a baseline run on clean `main` to attribute correctly.
`sender_kind` is now registered in `@known_keys`, in both typespecs, and in the Logger
`:metadata` allowlist that `meta_test.exs` asserts against.

**Three tests asserted `meta == %{}` by exact equality.** Two meant something narrower
than they said:

- *"plain peer NOTICE gets no ctcp meta"* → now `refute Map.has_key?(meta, :ctcp)`, which
  is the claim in the title.
- *"MOTD does NOT double-persist via the routing path"* → the comment says the
  discriminator is `numeric` + `severity`, so it now asserts their absence. An empty map
  was standing in for that, and would have read as a routing regression here when routing
  is untouched.

The third is a server-prefixed **channel** NOTICE, and now expects the kind — a server
can notice a channel, which is precisely the case a consumer cannot otherwise resolve.

Worth flagging because it means the allowlist and those equality assertions are the real
cost of any future `meta` key, not the threading.

### New tests

- a server-prefixed NOTICE carries `sender_kind: "server"`, and no `sender_user` /
  `sender_host` is invented for it
- a nick-prefixed NOTICE carries `"user"` plus its real user@host
- a partial (+x-cloaked) prefix still reports the kind while `prefix_userhost/1` keeps
  refusing the half mask — the two facts are independent

### Verification

1599 tests, 0 failures across `test/grappa/session/`, `test/grappa/scrollback/` and
`test/grappa/irc/`. Compiles with `--warnings-as-errors`. `mix format` clean.

### Scope

The other ~19 `sender_nick/1` call sites are untouched. The NOTICE paths are where #1070
reports the ambiguity, and each remaining kind deserves its own look at whether its sender
can be a server at all — better as follow-ups than as one sweep. Happy to do them if you
would rather have it in one go.

### Where this came from

Building an isolated test stack for bicchierino, an IRC-facade bridge to grappa. It got
the prefix wrong in both directions within a day — bare nick for user notices, then
`nick!user@host` for server ones — because the kind was the only signal available. It now
guesses with "sender contains a dot", exact against bahamut and a guess elsewhere. With
this key it can stop guessing.

### Consequence worth stating

The `messages` table now holds two shapes of `meta` for the same `kind: :notice`,
discriminated only by which code path wrote the row: the five NOTICE persists carry
`sender_kind`, and the kinds deliberately left out of scope (privmsg, join, part, mode,
topic) do not. That is the cost of scoping this to where #1070 reports the ambiguity
rather than sweeping all 22 `sender_nick/1` call sites in one change.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -348,6 +348,7 @@ config :logger, :console,
     # re-prefix them. In the allowlist to satisfy the known_keys↔metadata
     # sync test even though no Logger call carries it today.
     :sender_prefix,
+    :sender_kind,
     # #591 — CTCP frame classification (Grappa.IRC.CTCP.verb_args/1) tagged onto
     # a :notice (peer's CTCP reply) or :privmsg (own /ctcp /ping self-echo) row.
     # In the allowlist to satisfy the known_keys↔metadata sync test even though
--- a/lib/grappa/scrollback/meta.ex
+++ b/lib/grappa/scrollback/meta.ex
@@ -164,6 +164,8 @@ defmodule Grappa.Scrollback.Meta do
             | :sender_user
             | :sender_host
             | :sender_prefix
+            | :sender_kind
+            | :sender_kind
             | :ctcp_verb
             | :ctcp_args
             | :ctcp_target
@@ -171,7 +173,7 @@ defmodule Grappa.Scrollback.Meta do
           ) => term()
         }
 
-  @known_keys ~w[target new_nick modes args numeric severity who who_target names names_target raw_verb raw_sender raw_params sender_user sender_host sender_prefix ctcp_verb ctcp_args ctcp_target nick_fallback]a
+  @known_keys ~w[target new_nick modes args numeric severity who who_target names names_target raw_verb raw_sender raw_params sender_user sender_host sender_prefix sender_kind ctcp_verb ctcp_args ctcp_target nick_fallback]a
 
   @doc """
   The atom-key allowlist. Exposed so the test suite can assert that
--- a/lib/grappa/scrollback/meta.ex
+++ b/lib/grappa/scrollback/meta.ex
@@ -165,7 +165,6 @@ defmodule Grappa.Scrollback.Meta do
             | :sender_host
             | :sender_prefix
             | :sender_kind
-            | :sender_kind
             | :ctcp_verb
             | :ctcp_args
             | :ctcp_target
@@ -199,6 +198,7 @@ defmodule Grappa.Scrollback.Meta do
           | :sender_user
           | :sender_host
           | :sender_prefix
+          | :sender_kind
           | :ctcp_verb
           | :ctcp_args
           | :ctcp_target
--- a/lib/grappa/session/event_router.ex
+++ b/lib/grappa/session/event_router.ex
@@ -428,7 +428,12 @@ defmodule Grappa.Session.EventRouter do
             else: Identifier.canonical_target(target, casemapping(state))
 
         notice_body = "CTCP VERSION query → grappa #{version}"
-        {state2, persist_eff} = build_persist(state, :notice, dm_channel, sender, notice_body, %{})
+        # sender_meta/1, not %{}: this row's sender is a peer nick off a real
+        # wire prefix, and leaving it bare would make an absent key mean
+        # either "prefix-less" or "a CTCP visibility row" — the ambiguity
+        # #1070 exists to remove, one layer down.
+        {state2, persist_eff} =
+          build_persist(state, :notice, dm_channel, sender, notice_body, sender_meta(msg))
 
         {:cont, state2, [{:reply, reply}, persist_eff]}
 
@@ -2785,8 +2790,16 @@ defmodule Grappa.Session.EventRouter do
         do: state.nick,
         else: Identifier.canonical_target(target, casemapping(state))
 
+    # Same as the VERSION row above: a real peer nick, so the key belongs.
     {state2, persist_eff} =
-      build_persist(state, :notice, dm_channel, sender, "CTCP PING query → answered", %{})
+      build_persist(
+        state,
+        :notice,
+        dm_channel,
+        sender,
+        "CTCP PING query → answered",
+        sender_meta(msg)
+      )
 
     {:cont, state2, [{:reply, reply}, persist_eff]}
   end
@@ -3082,6 +3095,9 @@ defmodule Grappa.Session.EventRouter do
   # Additive, same contract as `sender_prefix` above: a consumer that does
   # not know the key behaves exactly as before, and rows persisted earlier
   # simply lack it.
+  # The value set is closed — "user" | "server" — but Elixir typespecs do
+  # not admit string literals ("unexpected expression in typespec"), so it
+  # cannot be stated here. The two clauses below are what guarantee it.
   @spec sender_meta(Message.t()) ::
           %{optional(:sender_kind | :sender_user | :sender_host) => String.t()}
   defp sender_meta(%Message{prefix: {:server, _}}), do: %{sender_kind: "server"}
--- a/lib/grappa/session/event_router.ex
+++ b/lib/grappa/session/event_router.ex
@@ -458,7 +458,9 @@ defmodule Grappa.Session.EventRouter do
        when is_binary(channel) and is_binary(body) and
               byte_size(channel) > 0 and
               binary_part(channel, 0, 1) in ["#", "&", "!", "+"] do
-    {state, eff} = build_persist(state, :notice, channel, Message.sender_nick(msg), body, %{})
+    {state, eff} =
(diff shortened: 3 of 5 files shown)


vjt:
Reviewed. Good contribution — the shape matches what #1070 proposed, the additive-only rule is respected, and the three rewritten assertions move in the right direction (they were over-claiming with `meta == %{}`, and you say so out loud instead of quietly widening them). Three asks before merge, one of which is a real defect CI does not catch.

**1. `@spec known_keys/0` is now inconsistent with `@known_keys`.** In `lib/grappa/scrollback/meta.ex` the diff adds `| :sender_kind` to `@type t` **twice**, and the second one was plainly meant for the other typespec: `@spec known_keys/0` does **not** list it, while the `@known_keys` value it returns does. So the function is declared to return a list that cannot contain `:sender_kind`.

Worth stating why this survived a green build: Dialyzer's `overspecs`/`specdiffs` are off by default, so a return spec *narrower* than the success typing is accepted silently — `dialyzer` passed. No gate protects this file's specs from drifting, which is exactly why it should be fixed by hand rather than left. Please drop the duplicate in `@type t` and add the key to the `@spec known_keys/0` union.

**2. Five NOTICE persists, three carry the key.** Beyond the three you threaded, there are two more `build_persist(…, :notice, …)` sites — the CTCP VERSION query visibility row (`event_router.ex:431`) and the CTCP PING one (`:2789`). Both pass `%{}`, both derive their sender from a real wire prefix, and both land in the same DM/own-nick windows.

This matters more than a missing nice-to-have: the PR defines an absent key as "prefix-less, the line is the local connection's own". After this change, absence means *either* that *or* "one of the two CTCP visibility rows, whose sender is a peer nick" — the ambiguity #1070 exists to remove, reintroduced one layer down. Both sites already have `msg` in scope, so it is `sender_meta(msg)` instead of `%{}`.

The deliberate scoping away from privmsg/join/part/mode/topic is fine and well argued. One honest consequence worth a line in the description: the `messages` table now holds two shapes of `meta` for the same `kind: :notice`, discriminated only by which code path wrote the row.

**3. Tighten the local spec on the closed set.** `sender_kind` is a two-element closed set typed as `String.t()`. String values are the right call here — `"user" | "server"` is what #1070 proposed and `Meta` deliberately types values loosely — but nothing validates the *value*: `cast/1` and `dump/1` both gate on keys only, so `sender_kind: "banana"` would persist happily. The set is guaranteed by your two function clauses and by nothing else. `%{optional(:sender_kind) => "user" | "server", optional(:sender_user | :sender_host) => String.t()}` puts it where Dialyzer can see it, without touching `Meta`'s design.

**Housekeeping:** please drop `Closes #1070` from the body. We keep the keyword out of PRs because it once auto-closed an issue while production did not yet have the code; we close by hand after a release.

Verified rather than assumed, for the record: the allowlist is genuinely the only write door (`cast/1` + `dump/1` reject unknown keys, `load/1` is documented read-side leniency); `Scrollback.Wire` passes `meta` through verbatim so there is no second allowlist; and a client that does not know the key is unaffected — `cicchetto`'s `ScrollbackMetaT` is an open `Record<string, unknown>`, so no client type narrows on it. No `protocol_version` bump is needed.

abonforti:
Pushed `eff8fdd3`. Asks 1 and 2 done; ask 3 does not compile, details below.

**1 — the spec drift.** Confirmed exactly as described: `| :sender_kind` landed in
`@type t` twice and never reached `@spec known_keys/0`. Two `replace(…, 1)` calls in the
script I used to apply the change both matched inside `@type t`, and nothing downstream
noticed. Duplicate dropped, `@spec known_keys/0` union extended.

Worth recording that your diagnosis of *why* it survived is the useful half: a return
spec narrower than the success typing is accepted silently with `overspecs` off, so
`dialyzer` passing said nothing here. I would not have found this from a green build.

**2 — the two CTCP rows.** Done, and you are right that it is not a nice-to-have. With
those left bare, an absent key would have meant *either* "prefix-less, the line is the
local connection's own" *or* "one of these two visibility rows, whose sender is a peer
nick" — the exact ambiguity #1070 is about, reintroduced one layer down by the change
meant to remove it. Five of five `:notice` persists now carry the key.

I have added the consequence you asked for to the PR description: the `messages` table
now holds two `meta` shapes for `kind: :notice`, discriminated only by which path wrote
the row.

**3 — the tightened spec does not compile.** Elixir typespecs do not admit string
literals:

```
** (Kernel.TypespecError) lib/grappa/session/event_router.ex:3098:
   unexpected expression in typespec: "user"
```

`%{optional(:sender_kind) => "user" | "server", …}` is rejected outright, so this cannot
be expressed as written. Left as `String.t()` with a comment recording the closed set and
why it is not in the spec.

If having Dialyzer see the set matters more than the wire shape, `:user | :server` as
atoms *is* expressible — but that changes what `dump/1` writes and what every existing row
would have to be read against, which seems a poor trade for a two-value set already
guaranteed by two function clauses. Happy to do it if you would rather; say which.

**Housekeeping:** `Closes #1070` removed from the description.

vjt:
Fast turnaround, thanks. Two of three are settled; one is not, and I owe you a correction on a third.

**F2 — done, and done properly.** Both CTCP visibility rows now carry `sender_meta(msg)`, and the comments explain *why* rather than what. Good.

**F3 — you are right and my ask was wrong.** Elixir typespecs do not admit string literals, so `"user" | "server"` genuinely cannot be stated there; I asked for something the language does not allow. Your comment recording that the closed set is guaranteed by the two clauses is the correct resolution, and I would rather have that written down than a spec that lies. (If we ever want it enforced rather than constructed, that belongs in `Meta`'s `cast/1` as value validation, which is a change to `Meta`'s design and not this PR's job.)

**F1 — still outstanding.** The commit subject is *"align the known_keys spec, and cover the two CTCP rows"*, but the diff `288784c9..eff8fdd3` touches `lib/grappa/session/event_router.ex` only — 18 insertions, 2 deletions, one file. On the current head `lib/grappa/scrollback/meta.ex` still reads:

- lines 167–168: `| :sender_kind` twice in `@type t`
- `@spec known_keys/0` (line 185): ends `… | :ctcp_target | :nick_fallback` — no `:sender_kind`, while `@known_keys` on line 176 has it.

So `known_keys/0` is still declared to return a list that cannot contain the key it returns. Worth repeating why nothing catches this for you: Dialyzer's `overspecs`/`specdiffs` are off by default, so a return spec narrower than the success typing passes in silence — that is why the build is green and why this one has to be fixed by eye.

Once `meta.ex` carries the one-line spec change (and loses the duplicate), this is good to merge from my side.

abonforti:
`f6e29335`. You were right, and the miss was worse than the original defect: I reported F1
as done off the success message of the script that was supposed to make the edit, without
reading the file back. The change existed only in that message. The commit subject said
`meta.ex` and the diff did not touch it — and I did not look.

Verified on the pushed branch this time by reading it, not by trusting anything:

```
167:            | :sender_kind          # @type t, once
175:  @known_keys ~w[… sender_prefix sender_kind ctcp_verb …]a
201:          | :sender_kind            # @spec known_keys/0
```

Compiles with `--warnings-as-errors`, formatted.

Your point about *why* nothing catches it is the part I will carry forward. A green build
said nothing here twice over — once about the drift, once about the fix that was not
applied — because `overspecs` is off and a narrower return spec passes in silence. There
is no signal to lean on in this file, so it wants eyes, and I substituted a script's
stdout for eyes.

Thanks also for taking F3 back explicitly rather than letting it sit. Agreed that value
validation belongs in `Meta`'s `cast/1` if it is ever wanted enforced rather than
constructed, and that it is not this PR's job.

vjt:
F1 confirmed fixed on `f6e29335` — read off the file, not the commit subject: `:sender_kind` now appears exactly three times in `meta.ex`, once in `@type t` (duplicate gone), once in `@known_keys`, and once in `@spec known_keys/0` in the position matching `@known_keys`. Thanks also for dropping the closing keyword.

All three review points are settled — one of them by you correctly refusing mine, which is the right outcome. Approving from my side; I will merge once `integration` lands green on this head. The issue stays open until the release, per our convention.

Issue #1077, opened 2026-08-08
cic(sidebar): a muted conversation's badge should be dimmed, not full-strength


vjt:
## What

A muted conversation's unread badge should render **dimmed** — same colour, less bright — instead of at full strength like an unmuted one.

Requested by @vjt on IRC, verbatim: *"nei canali / conversazioni muted il badge deve essere anch'esso muted. non colore diverso ma meno bright"*. The constraint is explicit: **not a different hue**, a lower-intensity version of the badge that is already there. Whatever the badge means today it keeps meaning; it just stops shouting.

Minor / polish, but it is pre-1.0 chrome: the sidebar is the surface you look at all day.

## Why the current behaviour is deliberate, and why this revises it

The badge ignoring the mute is not an oversight — it was decided. `cicchetto/src/lib/activeWindows.ts` (around the `orderUnreadWindows` filter, `.filter((r) => r.unread > 0 && !r.muted)`) carries the note:

> the mute drops the window from the CYCLE, and it beats the tier: a mention inside a muted room is still not a stop (#866 Q2, "the mute always wins"). Everything else about the window is untouched — **its sidebar unread badge and the mention counters keep rendering exactly as before (#866 Q4)**.

So #866 Q4 answered "does the mute change the badge?" with *no*. This issue changes that answer to *yes, visually only*: the counts stay exact and the cycle behaviour is untouched — only the badge's brightness follows the mute. Whoever picks this up should update that comment rather than leave it contradicting the code.

## Where

Measured against `origin/main` @ `08f0fcb0`.

**The component** — `cicchetto/src/WindowBadges.tsx`. One component renders the whole triad for both surfaces; the surfaces differ only by the `BADGE_CLASSES` prefix (`sidebar` / `bottom-bar`) × (`messages` / `events` / `mention`). It currently takes no mute input at all. Fixing it here fixes sidebar and bottom bar at once — that is exactly what the file was extracted for.

**The precedent to copy** — the far-behind modifier in the same file: `export const FAR_BEHIND_CLASS = "far-behind"`, applied via `classList={{ [FAR_BEHIND_CLASS]: farBehind() }}` and qualified per badge class in the stylesheet. A muted modifier should be the same shape: one exported token, `classList`, qualified selectors. Do not interpolate class names.

[...]

Issue #1038, opened 2026-08-08
cicchetto: mute keys carry no network — muting #linux on one network silences it on every network


vjt:
Mute keys are network-blind. Muting `#linux` on Libera also silences `#linux` on Azzurra,
and the settings list shows only the channel name, so nothing on screen tells you which
network a mute belongs to — because it belongs to all of them.

vjt's ruling (2026-08-08, #grappa): **the mute key must always be prefixed by the network.**
This is a deliberate reversal of the current design, not a report of an accidental bug — the
network-blind key is documented as intentional in the code, and that decision is now withdrawn.

## Measured on `origin/main`

The key is the folded conversation and nothing else:

- `lib/grappa/push/triggers.ex:300` — `Map.has_key?(Map.get(prefs, :muted_targets, %{}), Identifier.canonical_target(key))`,
  where `key` is the folded channel for a channel row and the folded peer for a DM. No network
  component enters the lookup.
- `lib/grappa/user_settings.ex:114` — `@type muted_targets :: %{String.t() => %{String.t() => pos_integer() | nil}}`;
  the outer key is the folded target.
- `cicchetto/src/lib/conversationMute.ts:48` — `windowMuteKey(window)` returns
  `conversationMuteKey(window.channelName)`; the window's network slug is available at the call
  site and is discarded.
- `cicchetto/src/SettingsDrawer.tsx:537` states the intent outright: *"`muted_targets` is
  per-subject and carries no network: `#grappa` on two networks is ONE mute, and offering it
  twice would imply otherwise"* — and `muteCandidates()` dedupes by the folded key so the picker
  can only ever offer one row per channel name across all networks.
- `cicchetto/src/SettingsDrawer.tsx:1577` renders `muted.key` verbatim, which is why the muted
  list reads as a bare `#linux`.

So the behaviour is coherent with itself today; what changes is the requirement.

## The composite-key convention already exists

`cicchetto/src/lib/channelKey.ts` already defines the per-(network, channel) index used by
`scrollback.ts`, `selection.ts` and `subscribe.ts`: an opaque branded `ChannelKey` of shape
`` `${networkSlug} ${canonicalChannel(name)}` ``, with the space separator justified by RFC 2812
(space is not a valid chanstring byte) and `canonicalChannel` documented as a faithful mirror of
`Grappa.IRC.Identifier.canonical_target/1`. The mute store is the per-conversation store that
does not use it. Reusing that brand — rather than inventing a second composite shape — keeps the
[...]

PR #1085, opened 2026-08-08
cic(#1077): dim a muted conversation's badges

vjt:
Refs #1077.

## What

A muted conversation's badges now render dimmed — **same hue, lower
intensity**, per vjt's explicit constraint ("non colore diverso ma meno
bright"). Counts, cycle order and mute semantics are untouched.

`WindowBadges.tsx` gains `MUTED_CLASS`, following the `FAR_BEHIND_CLASS`
precedent exactly: one exported token, applied via `classList`, qualified
per badge class in the stylesheet. One component, both surfaces.

## Decisions worth reading

- **It revises #866 Q4.** That question — "does the mute change the
  badge?" — was answered *no*, deliberately. This changes it to *yes,
  visually only*. The `activeWindows.ts` comment stating the old answer
  is updated rather than left contradicting the code.
- **All three badges, mention included.** #866 Q2 already ruled a
  mention inside a muted room is not a stop; a full-strength `@4` beside
  two dimmed counters would contradict that. The issue flagged this as a
  possible shrink — say the word and the mention stays bright.
- **`opacity`, and that is why it composes.** `.far-behind` paints
  border/background/colour and never touches opacity, so muted + far
  behind keeps both treatments.
- **The events badge gets its own value (0.32, not 0.45).** It already
  sits at `opacity: 0.7`, and opacity does not compound across
  declarations on one element — reusing 0.45 would have made it
  *brighter* muted than unmuted. Written out rather than `calc`'d so the
  value you read is the value that renders.
- **Key derivation.** `decodeChannelKey` (the paired decoder), then
  `conversationMuteKey`, then the shared `isConversationMuted`. Not a
  hand-rolled `indexOf(" ")` and not a raw `===`: identifier matches
  fold, and the key shape has exactly one decoder by contract.

## Tests

Six new cases in `WindowBadges.test.tsx`, red before the change:
muted channel (all three badges) · muted DM keyed on the folded peer nick
· an unmuted neighbour left alone · **reactive unmute on the same DOM
node, no re-render** · composition with far-behind · the bottom-bar
variant.

The reactivity case is the load-bearing one, and I measured it rather
than assuming: replacing the signal read with a snapshot computed once at
setup kills **exactly that one assertion** and leaves the other twelve
green.

## The collateral, stated up front

`Sidebar.test.tsx` and `BottomBar.test.tsx` mocked `../lib/channelKey`
partially, without `canonicalChannel` — which now resolves through
`conversationMute`, so 102 tests in those two files went red. Both mocks
now spread `importOriginal` and override only the encoder they actually
mean to stub. Sidebar's also carried a byte-for-byte re-implementation of
`decodeChannelKey`; that is production logic copied into a test, and it
is gone.

I did **not** take the alternative of dropping the `conversationMuteKey`
fold to avoid touching the mocks. The name half of a `ChannelKey` is
already folded, so the fold is a no-op today and the issue says as much —
but "the string happens to be folded upstream" is the assumption
CLAUDE.md's identifier invariant exists to forbid, and it would break
silently if `channelKey()` ever stopped folding.

## What I am NOT claiming

- **The chosen opacities are unverified against real pixels.** 0.45 /
  0.32 are reasoned, not measured — I did not look at a rendered sidebar
  in either theme. If "meno bright" wants to be more or less than this,
  it is two numbers in `default.css`.
- **No e2e.** The stack is contended and I was told to ask first. The
  reactive-unmute acceptance is proven in vitest at the DOM level, not
  in a browser.
- **Custom themes are untested.** The rule is `opacity`, so it is
  theme-independent by construction, but I only reasoned that — I did
  not load a theme with a different badge palette.
- The issue's premise that #1038 may make the mute key network-scoped is
  not in play here: `origin/main` still has it network-agnostic, and this
  builds on that. If #1038 lands, this call site is one of the ones that
  needs the slug, and `decodeChannelKey` already hands it over.

## Gates

| gate | rc |
| --- | --- |
| `bun run check` | 0 |
| `bun run test` | 0 — 255 files, 4774 tests |
| `bun run build` | 0 |

### DIFF
--- a/cicchetto/src/WindowBadges.tsx
+++ b/cicchetto/src/WindowBadges.tsx
@@ -1,6 +1,8 @@
 import { type Component, Show } from "solid-js";
-import type { ChannelKey } from "./lib/channelKey";
+import { type ChannelKey, decodeChannelKey } from "./lib/channelKey";
+import { conversationMuteKey, isConversationMuted } from "./lib/conversationMute";
 import { mentionCounts } from "./lib/mentions";
+import { notificationPrefs } from "./lib/notificationPrefs";
 import { farBehindByChannel } from "./lib/scrollback";
 import { eventsUnread, messagesUnread } from "./lib/selection";
 
@@ -37,6 +39,17 @@ export type WindowBadgesVariant = keyof typeof BADGE_CLASSES;
 // the token itself is one string so the two surfaces cannot drift apart.
 export const FAR_BEHIND_CLASS = "far-behind";
 
+// #1077 — the modifier that says "this conversation is muted". Same shape as
+// FAR_BEHIND_CLASS for the same reason: one token, qualified per badge class
+// in the stylesheet, so the two surfaces cannot drift.
+//
+// It REVISES #866 Q4, which had answered "does the mute change the badge?"
+// with no. It still does not change what the badge MEANS or what it counts —
+// only how loudly it says it. Dimming, deliberately not a second colour: a
+// different hue would make the muted badge a different KIND of thing, and
+// the ask was for the same thing, quieter.
+export const MUTED_CLASS = "conversation-muted";
+
 export type BadgeKind = "messages" | "events";
 
 // What the badge SAYS, as opposed to what it shows. A bare "1832" inside a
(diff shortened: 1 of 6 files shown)


vjt:
Landed on main as `9566d9ce` (rebased onto `fb921a4f`, line counts identical: 6 files, 192+/16−). Closed by content rather than by GitHub: the branch was rebased locally and only `main` was pushed, so this PR's head is legitimately not an ancestor of main. Its `integration` was green after the failed job's rerun — the one red was `issue1050-list-window-no-float`, unrelated to this diff and now analysed on #1050.

PR #1090, opened 2026-08-08
docs(testing): name the biome FORMAT trap in the cic gate

vjt:
## What

One paragraph pair in `docs/TESTING.md`, in the cic-gate section next to the
other two `scripts/bun.sh` traps.

## Why

`scripts/bun.sh run check` is `biome check src && tsc --noEmit`. A pure
**format** violation therefore exits 1 — and the `&&` skips `tsc`, so the type
gate silently does not run.

The failure is cheap; the diagnosis is not. biome truncates its own output, so
the error in the file you just edited can sit inside `Diagnostics not shown: N`
while every file it *does* list is one you never touched. The summary reads
`Found 1 error` and nothing more. `grep "error TS"` returns empty, because it
is not a tsc error. `mix format` does not cover cic, so no earlier step catches
it. Everything on screen points at a pre-existing foreign red.

Paid for on 2026-08-08 while gating an unrelated cic change.

## The instruction

Format the files you touched **before** gating; scope `--write` to those files
only; pass `--max-diagnostics` to biome directly, never to `bun run check`
(the composite script forwards it to `tsc`, which rejects it).

## Test plan

- [x] Docs-only — no code path touched.

### DIFF
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -128,6 +128,25 @@ a raw `docker run oven/bun:1 …` to replicate it (that trips the docker
 guardrail above); the e2e stage already IS that clean build, through the
 wrapper.
 
+**`scripts/bun.sh run check` counts a biome FORMAT violation as an
+ERROR, and hides which file it came from.** `check` is `biome check src
+&& tsc --noEmit`: a line biome would reflow (a long `foo({ a, b, c })`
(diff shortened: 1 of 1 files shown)


vjt:
Landed on main as `d7075d07` (rebased onto `9566d9ce`, 1 file, 19+). Closed by content: branch rebased locally, only `main` pushed.

PR #1091, opened 2026-08-08
e2e(#1050): harden the corner barrier, and make it report itself

vjt:
Two hardening changes to the `#1050` spec's drawer barrier, plus the
instrumentation that would have made the last CI red readable. Both defects
were established by **reading the barrier and instrumenting it locally** — not
by a red this closes.

**This does NOT claim to fix the observed CI red.** The failing attempt of run
`31280208433` does not reproduce here: 20 local runs of the untouched spec on
the same containerised webkit, 0 red. The CI incidence rate is unknown — one
red on a denominator I cannot see. What follows stands on its own; whether it
also removes that red is untested, which is why there is no closing keyword.

## Commit 1 — a barrier must establish a state that STAYS established

**Defect A, the release condition was instantaneous.** "the drawer is not
under this point right now" is a property of a LIVE animation. Instrumented
locally it holds ~5–15ms before the slide ends — `getAnimations().length` still
`1`, drawer left at 377..389 against a final 393 — in 9 runs out of 10. The
assertion then samples at some later, unbounded instant: 4–15ms locally, 48ms
in the CI trace. A barrier that samples a moving target and hopes is a coin
flip by design, and naming the engine internal that decides the toss is not
required to call it one. The condition now also requires
`getAnimations().length === 0` — the transition FINISHED. That is a durable
pre-state; "momentarily off the point" never was one.

**Defect B, an empty hit stack was read as "clear".** `elementsFromPoint`
returns an empty list for a point outside the viewport — precisely the
conflation the probe below it was rewritten to eliminate, quietly reintroduced
one block earlier. A ✕ pushed off-screen would have released the barrier
instantly. An empty stack is now a state to keep waiting through; the probe's
own `inViewport` assertion stays the one that names it.

Neither relaxes an assertion, and neither can mask the `#1050` regression: the
float is `.shell-chrome`, a DIFFERENT element this barrier never consults, and
a drawer that genuinely failed to leave would still fail the hit-stack half.

## Commit 2 — self-reporting, on green runs too

The one red this spec produced cost a full reconstruction from the trace's
screencast — barrier release instant against painted drawer edge, fitted to the
200ms ease-out curve — to establish two facts the barrier could simply have
stated: that it polled **once**, and that it released 48ms before the probe with
the drawer still ~59px over the point. None of that is in any artifact; all of
it was in the barrier's own hands.

Barrier and probe now read the corner through one shared `inspectCorner`.
Sharing the oracle matters as much as printing it: the diagnosis turned on
comparing what the two saw at two instants, which is worthless if they are two
hand-rolled readings free to drift. Each reading carries the ✕'s rect, the hit
stack, the drawer's animation count, its computed transform and its gBCR left.

Reporting mirrors `issue988-rail-menu-first-row`: node-side `console.log` (the
`list` reporter prints stdout from passing tests) plus a `testInfo.attach` that
survives into the HTML report, both from a `finally` so a timed-out barrier
still hands over the samples that led there.

`docs/TESTING.md` gains the other half of that CI round: a trace declares its
screencast at the DEVICE resolution while writing the JPEGs at the CSS one, so
measuring a frame against the declared size overstates every offset by the
device pixel ratio — and flips the conclusion, as it did here once.

## Gates

`scripts/integration.sh --project webkit-iphone-15 --grep "#1050" --repeat-each 10`
→ **10 passed (47.4s)**, `rc=0`.

`tsc -p e2e/tsconfig.json --noEmit` → 26 errors, byte-identical to the
pre-change baseline, none in this spec. (`cicchetto/e2e/` has no static gate in
CI and `biome check` exits 0 without reading files under it, so that is the
only meaningful static signal available here.)

**The telemetry witnesses defect A on this host.** Across the 30 barrier
samples the 10 runs emitted, **9 samples in 9 distinct runs** have the drawer
already out of the hit stack while `nAnim === 1` and `drawerLeft` at 384–393
against a final 393. Those are exactly the instants where the OLD condition
would have released and the new one keeps waiting one more poll — the 9-in-10
figure, observed rather than predicted.

Defect B has **no witness in this run**: the stack was never empty here. It is
fixed on the strength of the read, not of a measurement.

## Not run

The full `scripts/integration.sh` suite. This touches one spec file and one doc,
and CI is being watched separately.

### DIFF
--- a/cicchetto/e2e/tests/issue1050-list-window-no-float.spec.ts
+++ b/cicchetto/e2e/tests/issue1050-list-window-no-float.spec.ts
@@ -116,6 +116,32 @@ test("@webkit #1050 — the /list window drops the floating ☰, and its ✕ act
   // This is NOT webkit-specific, despite only webkit having reported it: the
   // chromium project `grepInvert`s `@webkit`, so chromium has never run this
   // spec. The race is in the flow, not the engine.
+  //
+  // TWO THINGS THE HIT-STACK BARRIER STILL GOT WRONG, both fixed below and both
+  // established by reading, not by a red this closes.
+  //
+  // ONE — the release condition was INSTANTANEOUS where a barrier must be
+  // DURABLE. "the drawer is not under this point right now" is a property of a
+  // LIVE animation: instrumenting the barrier locally, it holds ~5–15ms before
+  // the slide ends (`getAnimations().length === 1`, drawer left at 377..389
+  // against a final 393) in 9 runs out of 10. The assertion then samples at
+  // some LATER, unbounded instant — 4–15ms locally, 48ms in the CI trace. A
+  // barrier that samples a moving target and hopes is a coin flip by design, no
+  // matter which engine internal decides the toss. So it now also requires the
+  // transition to be FINISHED: `getAnimations().length === 0`. That is the
+  // durable pre-state; "momentarily off the point" never was one.
+  //
+  // TWO — an EMPTY hit stack was read as "clear". `elementsFromPoint` returns
+  // an empty list for a point outside the viewport, which is precisely the
+  // conflation the probe below was rewritten to eliminate, quietly reintroduced
+  // one block earlier: a ✕ pushed off-screen would release the barrier
+  // instantly. An empty stack is now NOT free — it is a state to keep waiting
+  // through, and the probe's own `inViewport` assertion stays the one that
+  // names it.
+  //
+  // Neither relaxes anything, and neither can mask the #1050 regression: the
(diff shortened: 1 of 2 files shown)


vjt:
Landed as part of the union PR #1095, merged to main at `b160d9d4`. The cherry-pick rewrote the SHAs, so GitHub cannot mark this PR merged — closing by content: the commits from this branch are on main verbatim (the union's diffstat is the exact sum of #1091 and #1092: 630 insertions / 38 deletions over 5 files, no file overlap). Gated once as the union, which is the only gate that attested the two together.

Issue #1089, opened 2026-08-08
Switching into a window with unread flickers: the marker re-assert scrolls a second time without the hide


vjt:
Switching to a window that has unread messages flickers: the content paints once, then jumps. Reported by the maintainer.

**Discriminant already answered:** it happens on **every** switch — not only the first one after a reload. That rules out a cold-mount-only cause (`onMount`, first-focus-after-login) and points at the per-switch path.

## What the code does on a switch into an unread window

`cicchetto/src/ScrollbackPane.tsx`:

1. The channel-switch effect arms the marker latch and performs the establishing scroll **with the flicker hide on**:
   ```
   :2382  setMarkerActivationPending(true);
   :2383  applyActivation("marker-or-tail", true);   // withHide = true
   ```
   `withHide` drives `setActivating(true/false)` (`:2132`, `:2129`), which the pane renders as `visibility: hidden` (`:3651`) until the rAF×2 lands.

2. The rows commit then re-enters the applier. While the latch is still set and the divider is in the DOM, a `marker-activation` intent is produced:
   ```
   :2916  if (markerActivationPending() && listRef.querySelector('[data-testid="unread-marker"]'))
   :2917    intents.push({ kind: "marker-activation", ... });
   ```
   and the winning write is dispatched **without the hide**:
   ```
   :2868  scrollToActivation("marker-or-tail", false);
   ```

So a switch into an unread window performs *two* scroll writes: the first hidden, the second visible. A switch into a window with no unread never arms the second one — the latch has nothing to re-assert against, since the intent is gated on the divider being present.

## The part that has to be measured, not assumed

The `withHide` doc block states the re-assert is deliberately unhidden and claims it cannot paint:

> A RE-ASSERT (same key, driven by the length-effect when `markerActivationPending`) passes `false`: the rows-recreation reset happens in the SAME frame and the rAF×2 corrects it pre-paint, so the intermediate scrollTop=0 is never painted — no hide needed, and toggling `activating` on every rows change would itself flicker. (`:2101-2104`)

That invariant and the reported behaviour disagree. One of the two is wrong, and **which one has not been established** — I did not observe the flicker and did not instrument the frames. Do not "fix" this by flipping `:2868` to `true` on the strength of the reading alone: the same comment warns that toggling `activating` on every rows change is itself a flicker source, so a naive flip can trade one artifact for another.

## Suggested order of work

[...]

Issue #130, opened 2026-06-28
cicchetto: content flicker on channel/window change (scroll happens after paint)


vjt:
## Summary
Switching to a channel/window produces a visible flicker: the channel content paints first, then the scroll-to-bottom (or scroll-to-position) runs, so the buffer is briefly shown at the wrong scroll offset before snapping into place.

## Symptom
On window activation (channel change), the freshly-activated channel's content flickers — the user sees the content render, then jump as the scroll is applied.

## Root cause
The scroll adjustment that should occur **before** the window paints is instead running **after** the paint. Order is inverted:

- expected: set scroll position → paint
- actual: paint → set scroll position (visible correction)

## Fix direction
Apply the scroll synchronously before the browser paints the activated window — e.g. position the scroll in a layout-phase effect (`useLayoutEffect` / pre-paint) or set it before the window becomes visible, rather than in a post-paint effect. Goal: zero visible scroll correction on activation.

## Repro
1. Open cicchetto with at least one channel that has enough backlog to scroll.
2. Switch away to another window, then back.
3. Observe the content flicker/jump as the scroll snaps on activation.

vjt:
Fixed in b13c72f (merged to main, FF). cic-only.

Root cause: on window activation the scrollback container is the same DOM node across the swap, so its scrollTop is preserved from the leaving pane; the correcting scroll runs inside scrollToActivation's double-rAF (load-bearing — geometry is stale until Solid commits the new <For> rows AND layout settles), i.e. AFTER paint → visible jump.

Fix (SolidJS, no useLayoutEffect — that's React): an `activating` signal sets `visibility:hidden` on the container synchronously at activation (pre-paint) and reveals it only once the deferred scroll has settled — the wrong-scroll frame is never shown. visibility (not display) keeps scrollHeight readable; cold/empty windows skip the hide; reveal runs in every rAF exit path so it can't strand hidden. Covered by a new unit test that drives rAF manually and asserts the hidden→visible sequence.

Shipped to prod (irc.sniffo.org, hot cic bundle, hash C_5hl9wG); /healthz ok.

PR #1092, opened 2026-08-09
cic(#1089): re-assert the unread divider synchronously, not two frames late

vjt:
Refs #1089.

## What the measurement found

The issue nominated a suspect — the applier's marker RE-ASSERT runs
`scrollToActivation("marker-or-tail", false)`, without the #130 flicker hide —
and said, correctly, that the suspicion had not been measured. Measured, **the
hide is innocent**: the establishing write parks the pane on the divider and
reveals it correctly. The flicker is what lands *after* the reveal.

Sampling `visibility` plus the divider's on-screen offset every animation frame
through a switch, with the post-switch fetch at real-network latency, shows two
displacements:

1. **A `tail-only` activation writing through a live marker activation.** The
   ResizeObserver arm (`moved && followMode()`) fires — the container box does
   change between the $server pane and a channel pane, and the switch pre-arms
   `followMode` as an intent default. It tail-snapped ~18ms after the reveal and
   held the divider 365px off-screen ABOVE for ~400ms.
2. **The read-context page landing.** The eager join-ok `refreshScrollback`
   resumes from the READ CURSOR, so entering an unread window has only the unread
   rows loaded and `loadInitialScrollback`'s anchored `before` page ALWAYS arrives
   on focus. It prepends ~1049px ABOVE the viewport; `scrollTop` does not change,
   so the content moves under the reader, and the correction — deferred to an
   rAF×2 — concedes one COMPOSITED frame with the divider shoved off the bottom.

Timeline (viewport 300px, 400ms on the post-switch fetch):

```
main:  t=79 VIS divider 0px → t=91 VIS divider -368px (held ~400ms) → t=473 VIS +703px → t=512 0px
fixed: t=41 HID divider 0px → t=75 VIS 0px → t=480 VIS +23px → t=512 0px
```

## The change

One sync leg, in the idiom the same `switch` already uses for `overlay-freeze`
("re-assert SYNC — no transient frame for a reader to catch — then AGAIN across
rAF×2"). The applier runs post-commit, pre-paint, so scrolling the divider into
view synchronously lands in the same frame as the rows change that displaced the
pane; the rAF×2 still owns the settled read (`followMode` / `atBottomNow` from
the real distance) and corrects any pre-layout inaccuracy.

Displacement 1 needed no code of its own: with the pane parked from the commit
frame onward, the tail write stops firing. A precedence fix in `applyActivation`
written for it was measured **inert** — removing it left the spec green — so it
was dropped. The gap it addressed (`applyActivation` can only lose to
overlay-freeze, so it does not apply the documented precedence order) is real and
worth its own issue; it is simply not what produced this bug.

## The spec

`cicchetto/e2e/tests/issue1089-switch-into-unread-flicker.spec.ts` asserts the
visible outcome, per frame: **while the pane is VISIBLE, the unread divider never
leaves the viewport.** Hidden frames are excluded by construction — they are the
frames the operator cannot see — and the end state is pinned visible AND parked
on the divider, so the green cannot come from a pane that stayed hidden or never
rendered a divider.

It injects 400ms on the post-switch `/messages` route. Not a workaround: on the
local stack those fetches land inside the hide window and nothing is left to
displace the pane — the same spec ran green, green, RED on unmodified main
depending on which side of the reveal the fetch happened to land. The delay picks
the regime the report comes from (mobile, real link) deterministically instead of
leaving a coin-flip spec.

## Test plan

- [x] `integration.sh --project chromium --grep "#1089" --repeat-each 3` — `6 passed`, rc 0, on the rebased base
- [x] Control red: same spec, cure removed — `3 failed / 3 passed`, rc 1 (both displacements visible in the dump)
- [x] `bun run check` — rc 0
- [x] `bun run test` — 259 files / 4813 tests passed
- [x] 10 neighbouring scroll/marker specs — `1 failed / 19 passed`; the failure (`issue168 › operator paged UP → send snaps back to the bottom`) reproduces IDENTICALLY with the cure removed and is 3/3 green in isolation, i.e. a pre-existing order cascade in that batch, not this diff
- [ ] Full `scripts/integration.sh` suite — not run locally

### DIFF
--- a/cicchetto/e2e/tests/issue1089-switch-into-unread-flicker.spec.ts
+++ b/cicchetto/e2e/tests/issue1089-switch-into-unread-flicker.spec.ts
@@ -1,0 +1,412 @@
+// Issue #1089 — switching INTO a window that has unread must not flicker.
+//
+// ## The field report (vjt, prod)
+//
+// Entering a window with unread messages: the content paints once, then jumps.
+// Discriminating fact established in the issue: it happens on EVERY switch, not
+// only the first one after a reload — so it is not a cold-mount cause.
+//
+// ## What the measurement found (and what it acquitted)
+//
+// The issue nominated the marker RE-ASSERT running without the #130 flicker
+// hide (`scrollToActivation("marker-or-tail", false)`). Sampling every frame
+// through a switch says the hide is not the mechanism. Two displacements are,
+// and both happen AFTER the establishing write has already parked the pane on
+// the divider and revealed it:
+//
+//   1. a `tail-only` activation (the ResizeObserver arm: the container box does
+//      change between the $server pane and a channel pane, and the switch
+//      pre-arms `followMode` as an intent default) tail-snapping straight
+//      through the live marker activation — divider 365px off-screen ABOVE for
+//      ~400ms;
+//   2. the read-context page landing. The eager join-ok refresh only loads rows
+//      AFTER the read cursor, so entering an unread window ALWAYS fetches the
+//      ~50 rows of context below it; that page prepends ~1049px ABOVE the
+//      viewport, and the correction — deferred to an rAF×2 — concedes one
+//      COMPOSITED frame with the divider shoved off the bottom.
+//
+// Both are cured by giving the applier's marker-activation dispatch a SYNC leg
+// (see `dispatchScrollWrite`), which is the idiom the overlay-freeze case in
+// the same switch already uses. With the pane parked on the divider from the
+// commit frame onward, (1) stops firing as well.
+//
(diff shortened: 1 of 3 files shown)


vjt:
Landed as part of the union PR #1095, merged to main at `b160d9d4`. The cherry-pick rewrote the SHAs, so GitHub cannot mark this PR merged — closing by content: the commits from this branch are on main verbatim (the union's diffstat is the exact sum of #1091 and #1092: 630 insertions / 38 deletions over 5 files, no file overlap). Gated once as the union, which is the only gate that attested the two together.

Issue #143, opened 2026-06-29
cic: NamesModal occluded by mobile virtual keyboard — reposition + resize via visualViewport


vjt:
Follow-up bug on #140 (names modal, shipped 2026-06-29 main 7b5541d).

**Symptom (vjt, mobile):** the NamesModal renders UNDER the on-screen virtual
keyboard — occluded and mis-sized; unusable when the keyboard is up.

**Fix:** constrain the modal's max-height and re-anchor it within the VISIBLE
viewport using `window.visualViewport` (height / offsetTop), so it is never
hidden by the keyboard. Same mechanism #66 needs for the message-list — build a
reusable visualViewport helper, not a NamesModal one-off.

**Verification:** a real chromium e2e (or explicit visualViewport-resize assertion)
that the modal stays within the visible viewport with the keyboard shown — not a
hollow green spec. iOS keyboard/scroll physics differ from Playwright
(feedback_playwright_webkit_not_ios_scroll) → may need Mezmerize device dogfood
before final close.

cic-only. Precedes #78-redo / #142 in the pack (regression on a just-shipped feature).

vjt:
Also (vjt): too much vertical padding between nicks in the modal roster — tighten the per-row spacing in the names list so it reads denser (irssi-columnar). Same modal, fold into this fix.

vjt:
Also (vjt): the close × is too small — make it a proper 44px tap target (mirror #133's card close-× → 44px). Fold into this fix.

vjt:
Shipped to prod 2026-06-29 (cic-only, HOT --cic, no BEAM restart). Three mobile defects on the #140 names modal fixed + vjt device-confirmed:
1. Keyboard occlusion — backdrop now spans the visible region (top:0; height:var(--viewport-height)) instead of inset:0; reuses the UX-6-D --viewport-height infra (NO offsetTop — #297779 WebKit bug). main 242431d.
2. Denser roster rows (padding 0.25->0.1rem, min-height 32->28px, row-gap 0).
3. Close × -> 44px Apple-HIG tap target (#133 precedent).
4. iOS safe-area: full-height modal shrinks to the padding box between the safe areas, header clears the top controls. main 81ce8dd.
Bundles XUS0qRf_ then DwfXDbTu. TDD red->green e2e (names143-modal-mobile.spec.ts) asserts the CSS contract; iOS keyboard/safe-area divergence is device-only (Playwright != iOS) — vjt confirmed on-device. Announced #grappa.

PR #1128, opened 2026-08-09
#407: declare the modal chrome once, named for the shape it is

vjt:
## What

Twenty-four modal controls each carried a copy of one of three declaration blocks — eleven dialog-header buttons, ten keyboard-aware scrims, three older scrims. `#143` (44px tap target, safe-area insets) and `#987` (`align-items: center`, not `stretch`) are both visible in the history as fixes that landed on whichever modal their author had open, leaving the rest on the old numbers.

Net: **382 stylesheet lines out, 157 in.** No e2e selector moved.

## The pin lands first, and the extraction does not touch it

Deduping CSS is a promise not to move a pixel, and a promise is worth what it can be checked against. The first two commits add `modalChrome.test.ts` and pass **against the stylesheet with no base class in it**. It resolves each call site the way the cascade would — every rule reaching it, folded in source order, `:hover` on top — and compares it to maps transcribed from the sheet at `6e81170`.

The extraction commit changes **zero lines of that test file**. Same twenty-four expected maps, green on both sides of the change. That is the evidence, not a claim.

Two premises that would make the fold a fiction are asserted, not assumed:

- nothing reaches these classes except a bare single-class selector, so specificity is uniform and source order decides;
- the class list being resolved is the one the markup actually wears.

**Box shorthands expand to longhands**, which is not tidiness. Nine of the ten × rules spell their margin as one shorthand; `.links-modal-close` reaches the same four values through `margin: -0.6rem 0` plus a `margin-right` longhand in a second rule with the same selector. The two scrim geometries spell the same edges as `inset: 0` and as `left/right/top: 0`. Compared as written, identical boxes read as differences and a real difference hides under a spelling.

## Why the base is `.modal-chrome-button` and not `.modal-close`

The issue proposed `.modal-close`. Grepping made that look right; reading did not. `.links-modal-close` is the **second selector** of `.links-modal-zoom, .links-modal-close` — the links modal's zoom buttons wear the same block, two properties apart (`margin-right: 0`, because they are not the last control in the row, and `font-size: 1.4rem`).

A base honestly called `.modal-close` could not take the zoom, so the zoom would have kept a fourteen-declaration copy of the base standing right beside it: the defect surviving under a new name. `.modal-chrome-button` takes all eleven; the zoom keeps only its two deltas.

Same reasoning and same convention as **#740** (`.login-quiet-button`, worn beside the per-instance class, which stays a selector contract) — which is why not one existing test or e2e selector had to change.

## The scrim gets two named geometries and no default

`.modal-backdrop` carries the paint and the centring and deliberately **no geometry**. `.modal-backdrop-full` spans the layout viewport; `.modal-backdrop-viewport` spans only the visible region — the `#143` keyboard fix. Ten scrims have that fix and three do not, and with the iOS keyboard up that is a difference in behaviour, not in spelling.

Either default is a bad one: `-full` would hand the next modal the worse behaviour without anyone choosing it, and `-viewport` would move three modals' pixels under cover of a refactor. So the base alone is a zero-height box and every site names its geometry.

## Guards, and what killed them

Both guards were verified by mutation, each mutant run and reverted:

| mutant | result |
|---|---|
| `.share-modal-close` font-size 1.5rem → 1.4rem | 1 failed / 25 passed — `share-modal-close resolves to the one × shape` |
| `.banlist-modal-backdrop` z-index 1000 → 1001 | 1 failed / 25 passed — `banlist-modal-backdrop resolves to the one keyboard-aware scrim` |
| add `.some-shell .who-modal-close { }` | 1 failed / 25 passed — `reaches every call site through bare class selectors only` |
| copy the base block verbatim as `.newfangled-modal-close` | 1 failed / 28 passed — `leaves no rule outside the bases carrying a base's whole shape` |
| drop the geometry variant from one scrim's markup | 2 failed / 27 passed — the geometry guard **and** the pin (removing a variant is both) |

The clone guard is keyed on the **whole** resolved shape, so `.links-modal-zoom` — two properties apart — falls out of the rule by itself rather than needing an exclusion list.

## Gates

- `bun run check` — exit 0. 61 warnings / 5 infos, **the same counts as `origin/main`**; none added.
- `bun run test` — 266 files, 4941 tests, all passing, on the rebased tree.
- Working tree verified clean before *and* after both gates.

## What this does NOT prove

jsdom paints nothing: every assertion here is about what the cascade is *asked* to do. Existing e2e already hit-tests two of the three shapes in a real browser — `issue219` clicks `.names-modal-close` and waits for the modal to hide; `issue232` clicks `.mode-modal-backdrop` at `(6, 6)`, a corner click a mis-sized scrim would miss.

**The `-full` geometry had no browser witness, so this PR adds one** — `e2e/tests/issue407-modal-backdrop-full-bottom-edge.spec.ts`, a single untagged test (chromium only, so the e2e count moves by exactly +1).

It clicks the confirm modal's scrim at the **bottom** edge, not the corner `issue232` uses. That distinction is the whole point: a click at `(6, 6)` lands on anything anchored at `top: 0`, so it passes whether the scrim has a height or not and cannot tell `.modal-backdrop-full` from the bare base. `bottom: 0` is the single declaration the variant contributes, so the click goes where only that declaration can carry it. The box is measured first so a collapse reports as a geometry failure rather than as an out-of-bounds click position; the measurement is the diagnostic, the click is the oracle.

**It has not been run.** No e2e lane was available (`_build` is shared and held). It is expected to run in this PR's integration job. Its static gate did run and is green — and was proven to actually read the file: injecting a type error makes `bun run check` exit 1 naming `issue407-modal-backdrop-full-bottom-edge.spec.ts` at the injected line, restored clean afterwards. That is a typecheck, not a pass: **if the spec is wrong, CI is where it shows.**

### DIFF
--- a/cicchetto/e2e/tests/issue407-modal-backdrop-full-bottom-edge.spec.ts
+++ b/cicchetto/e2e/tests/issue407-modal-backdrop-full-bottom-edge.spec.ts
@@ -1,0 +1,88 @@
+// #407 — the one shape the extraction left without a witness.
+//
+// `.modal-backdrop` carries the scrim's paint and centring and DELIBERATELY no
+// geometry: on its own it is `position: fixed` with `left/right/top: 0` and no
+// bottom and no height — a zero-height box. Every scrim must also wear exactly
+// one geometry variant. `.modal-backdrop-viewport` adds the #143 keyboard-aware
+// height; `.modal-backdrop-full` adds the one declaration that makes the scrim
+// span the layout viewport: `bottom: 0`.
+//
+// `modalChrome.test.ts` pins all of that, but jsdom applies no stylesheet — it
+// proves what the cascade is ASKED to do, never what a browser paints. Two of
+// the three shapes already have a real-browser witness: issue219 clicks
+// `.names-modal-close` and waits for the modal to hide, and issue232 clicks
+// `.mode-modal-backdrop` at (6, 6). The `-full` geometry had none, and the two
+// modals that use it (confirm, delete-account) both dismiss on scrim click with
+// no spec clicking either.
+//
+// WHY THE BOTTOM AND NOT THE CORNER issue232 USES. A click at (6, 6) lands on
+// any box anchored at `top: 0`, so it passes whether or not the scrim has a
+// height at all — it cannot tell `.modal-backdrop-full` from the bare base.
+// `bottom: 0` is the single declaration the variant contributes, so the click
+// has to go where only that declaration can carry it: the bottom edge.
+//
+// RED / GREEN: drop `bottom: 0` from `.modal-backdrop-full`, or ship the scrim
+// wearing the base with no geometry variant, and the scrim collapses to zero
+// height — the box assertion names it, and the click falls through to the
+// window chrome behind while the modal stays open. Restore either and it
+// passes. The bottom-left corner is clear of occluders by construction: the
+// only rules at or above the scrim's z-index are `.error-banners` (z 1000) and
+// `.diag-float` (z 99999), and both are anchored to the TOP.
+//
+// The confirm modal is opened through the destructive close × (#195) and then
+// DISMISSED, which is the safe branch — it never PARTs. The spec asserts the
+// channel is still there afterwards, so a regression that turned dismissal into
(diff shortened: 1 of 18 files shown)


vjt:

PR #1129, opened 2026-08-09
test(#499): pin the circuit-open precondition instead of racing its cooldown

vjt:
Refs #499.

## The filed diagnosis did not survive measurement

#499 attributes an intermittent `login_test.exs` red to `NetworkCircuit`'s
singleton ETS table bleeding across tests serially (recycled sqlite rowids), and
proposes two test-infra routes: reset the ETS table between tests, or use
non-colliding ids. **Neither is implemented here, because neither addresses what
happens.**

Two facts falsify the premise:

1. `login_test.exs` has called `AdmissionStateHelpers.reset_network_circuit/0`
   from a per-test `setup` since **2026-05-14** (`8494a1de`), and an inline ETS
   delete loop before that. The mitigation the issue proposes as the fix had been
   in that file for two months before the issue was filed.
2. The instrumented failure shows **no foreign row**. Running the
   `network_circuit_open` case under host load with the ETS snapshot printed at
   the cast drain:

   ```
   snap_after_flush=[{1, 3, -576460741210, :open, -576460741150}]
   pre_check=:ok  login_elapsed_us=5749  result={:error, :no_server}
   ```

   One row, `count` exactly at threshold, written by that same test — and
   `check/1` on the very next line already returning `:ok`. Nothing bled in. The
   circuit the test had just opened had already closed itself.

## The actual mechanism

`NetworkCircuit.check/1` is the only clock-dependent reader: an `:open` row reads
back as `:ok` the instant `now >= cooled_at_ms`, with no writer involved — so no
reset can prevent it. `config/test.exs` sets `network_circuit_cooldown_ms: 50`
(±25% jitter → 38..62ms) to keep the cooldown-expiry cases cheap to sleep
through. That is shorter than the work a *consumer* test does between opening the
circuit and reading it back.

The same shape existed at ten sites across six files: establish "circuit is open"
through the production transition, then read it back through a clock-dependent
path (`check/1`, or `AdminWire`'s `retry_after_seconds`).

## The change

`AdmissionStateHelpers.open_circuit!/1` drives the real `record_failure/1`
transition, raises if it did not land, then pins `cooled_at_ms` a minute out, so
the precondition is a fact that cannot decay. The production module is untouched.

The split is principled, not an exclusion list: tests where open-ness is a
**precondition** use the helper; the four whose **subject** is the clock — the
`retry_after` upper bound, cooldown expiry, and the `:cooldown_expired` telemetry
pair — keep driving the real cooldown, because pinning would delete what they
test.

**Raising the cooldown was rejected**: same band-aid one size larger. The window
widens but stays a race, and it would slow the two expiry tests by exactly the
headroom bought.

## Evidence

- **Deterministic mutation.** Removing the pin from the helper reds exactly one
  assertion, every run: `assert {:error, :open, _} = check(net_id)` → `right: :ok`.
  23 tests, 1 failure. Restored → 23/0.
- **Incidence, measured on this worktree.** `login_test.exs` alone, unloaded:
  2 reds in the first 20 runs, then 0 in 25. Under 14-core host load, the
  `network_circuit_open` case fired **3 times in 30 runs** before, **0 in 20
  after**.
- **Full gate green.** `scripts/check.sh` rc=0 — 5599 tests / 0 failures,
  dialyzer 0 errors, format/credo/sobelow/doctor/bats clean; working tree clean
  before and after.

## What this does NOT fix

Under the same load `login_test.exs` reds on eleven other cases that never touch
the circuit (`433` nick negotiation, handshake waits, the mode-2 admission
guard). They share a substrate — the in-process `IRCServer` fake plus
`pool_size: 1` sqlite — not a cause.

One is worth naming because it is circuit-adjacent and could be mistaken for this
bug: the `#960 boundary` case fails on its **first** assertion, with
`Login.login` returning `{:error, :connect_timeout}` while the session process
logs `{:connect_failed, :econnrefused}` — the login surfaces a timeout for what
was actually a refusal, losing the diagnosis. That is a separate defect on the
login path; left for its own issue rather than widened into this one.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -35107,3 +35107,71 @@ second copy of the flow set in a module that has no business owning one —
 worse than the gap. The set is closed by construction upstream:
 `Admission.check_capacity/1` takes a `capacity_input` whose `flow` is typed
 `flow()`.
+
+## 2026-08-09 — #499: the circuit was not bleeding across tests, it was healing inside one
+
+**The filed diagnosis did not survive measurement.** Issue #499 attributed an
+intermittent `login_test.exs` red to `Grappa.Admission.NetworkCircuit`'s
+singleton ETS table bleeding across tests serially — sqlite recycles rowids, so
+a fresh network inherits an id a previous test already tripped the circuit for.
+It proposed two test-infra routes: reset the ETS table between tests, or hand
+the affected tests non-colliding ids. Neither was implemented, because neither
+addresses what actually happens.
+
+Two facts falsify the premise. First, `login_test.exs` has called
+`AdmissionStateHelpers.reset_network_circuit/0` from a per-test `setup` since
+2026-05-14 (an inline ETS-delete loop before that) — the very mitigation the
+issue proposes as the fix had been in that file for two months before the issue
+was filed. Second, the instrumented failure shows no foreign row at all. The
+`network_circuit_open` case, run under host load with the ETS snapshot printed
+at the cast drain, reported one row — `{1, 3, _, :open, _}`, count exactly at
+threshold, written by that same test — and `NetworkCircuit.check/1` on the very
+next line already returning `:ok`. Nothing bled in. The circuit the test had
+just opened had already closed itself.
+
+**`check/1` is the only clock-dependent reader, and the test-env cooldown is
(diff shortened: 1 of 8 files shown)


vjt:

Issue #949, opened 2026-08-06
cicchetto/iOS: the nick context menu carries #913's inset-blind clamp — placeAxis/.context-menu still cap against the full layout viewport


vjt:
Pointer issue, filed so #913 can be closed without losing the sibling it names.

## What #913 fixed, and what it left

#913 fixed the RAIL actions menu opening under the status bar. The fix has two halves, both on
`origin/main`:

- `cicchetto/src/themes/default.css` — `:root` defines `--safe-area-inset-top: env(safe-area-inset-top, 0px)`,
  and `.rail-actions-menu` caps at
  `max(0px, calc(var(--rail-menu-space-above, …) - var(--safe-area-inset-top, 0px)))`.
- `cicchetto/src/RailActions.tsx` publishes the JS-measured `--rail-menu-space-above` and
  re-measures on `resize` + `visualViewport.resize`.

`d129bfb0` states that the same origin bug lives in the context-menu math and "wants its own
issue". No such issue was filed. This is it.

## The residue, measured on `origin/main`

`cicchetto/src/lib/menuPosition.ts`:

```ts
export function placeAxis(click: number, size: number, viewport: number): number {
  if (size >= viewport) return 0;          // ← pins to layout-viewport y=0
  if (click + size <= viewport) return click;
  const flipped = click - size;
  return flipped >= 0 ? flipped : viewport - size;
}
```

`computeMenuPosition` feeds it `viewportHeight` and nothing else. `UserContextMenu.tsx:156-171`
supplies `window.visualViewport?.height ?? window.innerHeight` — keyboard-aware, and correctly so
per #487, but **inset-blind**. `index.html:7` sets `viewport-fit=cover`, so the layout viewport's
origin is the physical top of the display: `return 0` and `viewport - size` both address a
coordinate behind the status bar.

The CSS half has the same hole. `.context-menu` (`default.css:4529`) caps at
`max-height: var(--viewport-height, 100vh)` with **no** `--safe-area-inset-top` subtraction —
literally the pre-#913 spelling of the rule that #913 rewrote for `.rail-actions-menu`. The
`--safe-area-inset-top` variable #913 introduced is available; this rule does not read it.

Symptom class: on a notched iPhone, right-clicking / long-pressing a nick near the top of the
screen, or opening a context menu taller than the viewport, puts the first item under the status
bar and out of reach — #913's complaint at a different door.

## Not verified

- **Device behaviour. At all.** I have no iOS device and this repo forbids local browser
[...]

Issue #487, opened 2026-07-27
cicchetto: member-list context menu opens off-screen near the viewport bottom — clamp/flip its position


vjt:
Reported on #grappa by Sonic: right-clicking a nick in the member list near the bottom of the page opens the context menu partially below the fold — the lower items (kick / ban / whois / query, depending on where the click lands) cannot be reached or clicked.

## Root cause (read on `origin/main`, not the working tree)

The menu is positioned at the raw pointer coordinates, with no viewport clamping anywhere in the path:

- `cicchetto/src/MembersPane.tsx:130-132` — `onContextMenu` stores the event coordinates verbatim: `setMenuFor({ nick, x: e.clientX, y: e.clientY })`.
- `cicchetto/src/UserContextMenu.tsx:143` — those coordinates become the final style, unmodified:
  `style={{ position: "fixed", top: `${props.position.y}px`, left: `${props.position.x}px` }}`
- `cicchetto/src/themes/default.css:3949-3957` — `.context-menu` has no `max-height` and no `overflow`, so it cannot scroll itself out of trouble either.

The menu renders 8 items (op / deop / voice / devoice / kick / ban / whois / query), all of which stay in the DOM even when disabled, so its height is constant and non-trivial. Click at `clientY` closer to the bottom edge than that height and the tail of the menu is laid out past the viewport.

This is a known gap, not a regression: the header comment at `UserContextMenu.tsx:25-28` says so outright — *"No viewport-flip logic here … Overflow handling is expected to be addressed in browser smoke; jsdom doesn't give real viewport dimensions."* It was never followed up. `UserContextMenu.test.tsx:62` passes a fixed `{x: 100, y: 200}` and asserts nothing about placement.

## Requested behaviour

Sonic's own framing, and it is the right one: the opening position must be computed so the menu always lands inside the viewport.

Suggested shape, to be confirmed by whoever picks this up:

1. Measure the menu after mount (`getBoundingClientRect`) rather than guessing from the item count — item labels and the theme font size both move the numbers.
2. Flip rather than merely clamp on the Y axis: if `y + height > innerHeight`, open upwards from the click point (`top = y - height`). Flipping keeps the pointer on the menu's edge, which is what every native context menu does; clamping to `innerHeight - height` slides the menu away from the cursor.
[...]

PR #1131, opened 2026-08-09
cic(#949): clamp the context menu to the safe box, not the physical one

vjt:
Addresses #949. #913's residue at the second door its own commit named.

## What was wrong

`placeAxis` clamped over `[0, viewport)`. `index.html` has opted into
`viewport-fit=cover` since #205, so `0` is the **physical** top of the display
and `viewport` the physical bottom:

- the oversize pin returned `0` — behind the status bar, and the menu's own
  `overflow-y: auto` cannot recover a box that *starts* in the occluded strip;
- a flip could return `viewport - size` — tail under the home indicator;
- in landscape the same arithmetic on X addressed the notch and the rounded
  corners. `d129bfb0` said "both edges"; it is four.

`.context-menu`'s `max-height: var(--viewport-height, 100vh)` was literally the
pre-#913 spelling of the rule #913 rewrote for `.rail-actions-menu`.

## The fix

`placeAxis(click, size, start, end)` — the interval is the caller's. The caller
reads it off **`.context-menu-safe-area`**: a fixed, unpainted, `pointer-events:
none` box laid out at `inset: env(safe-area-inset-*)`, measured with
`getBoundingClientRect()`.

That indirection is the design, not plumbing. #913 established that JS must not
read an inset back out of a custom property: `getComputedStyle()
.getPropertyValue()` on an unregistered one can return the token stream, and the
`NaN` that follows is swallowed by any `|| 0` into a fix that looks applied and
does nothing. **A rect is a resolved length by construction** — the engine keeps
`env()`, JS reads geometry. One measurement yields all four insets, and it lands
in the same coordinate space as the press coordinates and the menu's own fixed
box, so nothing is converted.

The two bounds compose as a `min`. The ruler is laid out against the **layout**
viewport: it knows the notch, not the keyboard (iOS never shrinks the layout
viewport for it). `visualViewport.height` — #487's deliberate choice — knows the
keyboard, not the notch. Both share an origin while `visualViewport.offsetTop`
is 0, which holds for this non-scrolling, non-zoomable shell; the pre-#949 code
made the same assumption.

The height cap stays in **CSS**: the placement effect measures the menu's
*rendered* height, so a cap applied inline after that measurement would position
the box against a height it no longer has. Same division of labour as
`.rail-actions-menu`.

## What this does NOT prove

Where there is no inset the safe box is the full viewport and the placement is
bit-identical to #487's — which is why the existing e2e stays green, and why it
proves nothing here. Playwright resolves `env(safe-area-inset-*)` to 0 on every
engine in the suite and jsdom resolves neither `env()` nor `calc()`. The guards
are split three ways, none of which sees a real notch:

| guard | pins |
|---|---|
| `lib/menuPosition.test.ts` | the arithmetic, over an explicit interval |
| `__tests__/ContextMenu.test.tsx` | the wiring, against stubbed rects |
| `__tests__/contextMenuSafeArea.test.ts` | the stylesheet, at source level |

**The felt behaviour on a notched iPhone is confirmed by none of them.**

## No e2e, deliberately

A Playwright spec was considered and **declined**. Playwright resolves
`env(safe-area-inset-*)` to 0 on every engine in the suite, so at inset 0 this
change is bit-identical to the code it replaces: the spec would be green against
the broken build too. **A guard that cannot fail is not a guard** — it is a
green tick that buys false confidence, which is precisely how #588's assertion
stayed green for the whole life of #913.

The only way to make it bite is a differential against a stubbed inset, the
scaffolding #913 had to settle for. Here that would cost more than there: the
production path reads the inset from the ruler's **layout**, not from a
variable, so stubbing it without `!important`-overriding the very rule under
test would mean first re-routing the ruler through four `:root`
`var(--safe-area-inset-*)` indirections. That is production churn to make a test
exist, and it is not worth it.

**Device-verify owed:** the real check is vjt on a notched iPhone, portrait
**and** landscape — not `webkit-iphone-15` in CI. Until that happens this fix is
argued, not observed.

Two further instances of the same class were found and deliberately left alone —
see the issue thread.

## Test plan

- [x] `bun run check` — biome + `tsc` + e2e `tsc`, exit 0
- [x] `bun run test` — 267 files / 4933 tests, exit 0
- [x] Mutation: the y=0 pin restored, the ruler element deleted, the bottom
      inset dropped from the cap, the ruler's left edge blinded — each read red,
      each naming the intended assertion
- [ ] **Device dogfood on a notched iPhone (portrait and landscape)** — owed;
      no gate in this repo can stand in for it

### DIFF
--- a/cicchetto/src/ContextMenu.tsx
+++ b/cicchetto/src/ContextMenu.tsx
@@ -20,6 +20,21 @@ import { computeMenuPosition } from "./lib/menuPosition";
 // proven in the Playwright e2e (issue487-context-menu-viewport-clamp.spec.ts)
 // since jsdom gives no real viewport dimensions. Opacity-gated until measured
 // so the pre-measure frame never flashes off-screen.
+//
+// #949 — "inside the viewport" was the LAYOUT viewport, whose origin under
+// `viewport-fit=cover` is the physical top of the display. #913 fixed the same
+// arithmetic for the rail menu and named this door as carrying the residue.
+// The bounds now come from `.context-menu-safe-area`: a fixed, unpainted box
+// laid out at `inset: env(safe-area-inset-*)`, measured with
+// `getBoundingClientRect()`. That indirection is the point. #913 established
+// that JS must NOT read the inset back out of a custom property —
+// `getComputedStyle().getPropertyValue()` on an unregistered one can hand back
+// the token stream rather than a length, and the NaN that follows is swallowed
+// by any `|| 0` into a fix that looks applied and does nothing. A rect is a
+// resolved length by construction: the engine still owns `env()`, and JS reads
+// geometry, which is the one thing it can always trust. The box also yields
+// all four insets from one measurement, which is what the X axis (landscape
+// notch) and the bottom edge (home indicator) need.
 
 export type ContextMenuItem = {
   label: string;
@@ -50,6 +65,7 @@ const ContextMenu: Component<Props> = (props) => {
   };
 
   let menuRef: HTMLDivElement | undefined;
+  let safeAreaRef: HTMLDivElement | undefined;
   const [placement, setPlacement] = createSignal({
     top: props.position.y,
     left: props.position.x,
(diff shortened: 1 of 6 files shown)


vjt:

Issue #1127, opened 2026-08-09
Mobile: drop the outer bottom safe-area inset so the whole shell moves down and reclaims the wasted band


vjt:
## Ruling (revised — this supersedes the original hybrid proposal)

Move the **whole mobile shell down** so the wasted bottom band goes away and that vertical space is reclaimed for content. Decided by @vjt on IRC: *"in sostanza non voglio fare altro che spostare tutta la shell più giù, in modo da recuperare spazio e levare quel bottom sprecato"*.

Concretely: the bottom safe-area padding on the **outer** box — `.shell-mobile`'s `padding-bottom: env(safe-area-inset-bottom)` (`cicchetto/src/themes/default.css:6415`, inside `@media (max-width: 768px)`) — is what holds everything up. Drop it (or reduce it to a small residual) and the entire shell, `.bottom-bar` included, extends to the physical bottom edge.

## What is actually wrong (measured, not eyeballed)

The visible defect is a black band under the window-tab strip on mobile. It is **not** the bar being too tall.

Measured in Chrome with a `320x568x2,mobile,touch` emulated viewport on a live session:

- with `env()` = 0 (desktop Chrome always reports 0), `.bottom-bar` closes flush: `bottom: 568`, `innerHeight: 568`, no gap. **The defect does not reproduce without a real inset** — this needs a device, not a devtools guess.
- injecting the iOS home-indicator value (`padding-bottom: 34px` on `.shell-mobile`) reproduces it exactly: `.bottom-bar` now ends at `bottom: 534`, leaving a **34px gap**, and `document.elementFromPoint(160, innerHeight - 10)` returns `DIV.shell shell-mobile` whose computed `background-color` is **`rgba(0, 0, 0, 0)`**, while `.bottom-bar` is `rgb(49, 50, 68)`.

So the band is not wasted padding inside the bar — it is **nothing painting there**. The outer inset lifts the shell off the bottom edge and exposes the transparent area behind it. Removing that outer inset is exactly what the ruling asks for, and it reclaims the full inset height (~34px on a notched iPhone).

## Constraints

- **The tabs will now sit in the home-indicator strip. That is the accepted trade — do not silently re-add clearance to "fix" it.** iOS captures gestures there, so verify on a real device that tapping the bottom row of tabs still registers reliably; if a specific tab turns out to be unusable, report the measurement back rather than reverting the ruling on your own.
[...]

Issue #205, opened 2026-07-10
cicchetto: iPadOS PWA layout broken — UI clipped under status bar, settings unreachable


vjt:
## Report
Reported on IRC (#it-opers) — cicchetto running as an installed PWA on iPadOS.

## Symptoms
- Layout broken in **both orientations**, landscape worse than portrait.
- UI is **clipped vertically** — top of the interface renders **under the iOS status bar** (safe-area not respected).
- **Settings icon is not tappable** (neither touchscreen nor mouse pointer) — appears occluded / pushed out of the interactive region.

## Suspected cause
Classic iOS standalone-PWA viewport issue:
- `100vh` / fixed viewport height not accounting for the dynamic toolbar + status bar → content overflows and gets clipped.
- Missing `viewport-fit=cover` + `env(safe-area-inset-*)` padding on the top chrome, so the header (settings icon included) lands under the status bar and outside the hit region.

## Repro
- Device: iPad, iPadOS, cicchetto added to Home Screen (standalone PWA).
- Open in landscape (and portrait) → top bar clipped, settings icon dead.

## Next steps
- Verify `<meta name="viewport" ... viewport-fit=cover>` and safe-area insets on the top chrome.
- Confirm the settings control isn't sitting under an overlay / outside `100dvh` bounds.

Screenshot to be attached (requested on channel).

_Filed per channel request._

vjt:
## Visual confirmation (iPad landscape, standalone PWA)

Screenshot reviewed (kept private — not attached). It confirms the diagnosis:

- The app's **top chrome renders under the iOS status bar**: the top-left header/logo cluster collides with the OS clock, and the **settings gear (top-right) is pinned against the status-bar edge** — inside the system status-bar / gesture zone, which is why it doesn't take taps or pointer clicks.
- No `safe-area-inset-top` padding on the top bar → the whole header shifts up under the notch/status region.

**Fix direction confirmed:**
- Add `viewport-fit=cover` to the viewport meta.
- Apply `env(safe-area-inset-top/left/right/bottom)` padding to the top chrome (and bottom compose bar) so controls clear the status bar.
- Switch layout height from `100vh` → `100dvh` to stop the vertical clipping.

vjt:
Device datapoint (reporter): iPad Pro 11" M1 (3rd gen), iPadOS 26.5.2 — Safari/WebKit standalone PWA.

vjt:
[...]

PR #1132, opened 2026-08-09
Mobile: drop the outer bottom safe-area inset (#1127)

vjt:
Ruling from #1127: move the whole mobile shell down so the wasted bottom band goes away and that vertical space returns to content.

## What changed

`.shell-mobile` (inside `@media (max-width: 768px)`) now declares `padding-bottom: 0` instead of `env(safe-area-inset-bottom)`. The shell — `.bottom-bar` tab strip included — reaches the physical bottom edge.

**Zero, not a deleted declaration.** This is the part worth reviewing. Base `.shell` declares `padding-bottom: env(safe-area-inset-bottom)` at the *same* specificity (both single-class; an `@media` prelude adds none) and the mobile element carries **both** classes — `<div class="shell shell-mobile">`. `.shell-mobile`'s block wins only for as long as it declares the property. Delete the longhand and the inset cascades straight back in, with the diff reading like a removal. #205 already documented this trap from the other side, which is why that rule owns all four edges explicitly.

**The keyboard override is gone, its reason is not.** `.shell-mobile:has(textarea:focus, input:focus) { padding-bottom: 0 }` (UX-6 D1, removed at D9 on a speculative claim, restored at D11 from live device diag) collapses nothing against a zero base. It is deleted deliberately, and D11's finding — with the keyboard up iOS reports the inset relative to the *device*, not the shrunken visual viewport, so an inset here double-counts against `--viewport-height` — is now written on the surviving `padding-bottom: 0` declaration, where anyone re-adding an inset on this edge will read it as a precondition rather than as history.

**Bottom edge only.** Top / left / right insets are untouched and now pinned. **The tap target is untouched**: `.bottom-bar { min-height: 3rem }` is unchanged; the reclaimed space comes from the removed outer inset.

**Accepted trade, stated out loud:** the tabs now sit inside the home-indicator strip, where iOS also reads gestures. That is the ruling, not an oversight. If a specific tab turns out to be unusable on device, the answer is a measurement reported back — not unilaterally re-added clearance.

## The guard, and its red

`ipadSafeArea.test.ts` pinned `.shell` and never mentioned `.shell-mobile`, so this change would have landed green either way. It now collects every `padding-bottom` declared by a `.shell-mobile` rule and asserts the list is exactly `["0"]` — one assertion, because it must kill three distinct regressions. Proven with four deliberate mutants, each killing exactly one assertion (`src/__tests__/ipadSafeArea.test.ts`, 7 tests):

| mutant | result | killed |
|---|---|---|
| restore `env(safe-area-inset-bottom)` | `expected [ 'env(safe-area-inset-bottom)' ] to deeply equal [ '0' ]` | 1 failed / 6 passed, `:117` |
| delete the longhand (the cascade trap) | `expected [] to deeply equal [ '0' ]` | 1 failed / 6 passed, `:117` |
| append a second declaration after the `0` | `expected [ '0', 'env(safe-area-inset-bottom)' ] to deeply equal [ '0' ]` | 1 failed / 6 passed, `:117` |
| drop `padding-top: env(safe-area-inset-top)` | `expected '…' to match /padding-top:\s*env\(safe-area-inset-t…/` | 1 failed / 6 passed, top/left/right test |

New helper `nestedRuleBodies()` in `__tests__/helpers/themeCss.ts`: `ruleBody()` anchors at column 0 and cannot see rules inside an `@media` / `@supports` block. It returns *every* block for the selector on purpose — `.shell-mobile` has a second one under `@supports not (height: 100dvh)`, and an assertion about which value a property ends up with has to see them all.

## Three e2e specs asserted the old edge

- `ux-3-empty-toolbar-island.spec.ts` and `ux-z-cluster-journey.spec.ts` flip their expectation and keep guarding the edge at runtime, where CSSOM is a stronger oracle than source text. Both accept `/^0(px)?$/` (CSSOM may serialize the authored `0` as `0px`) and both would still fail on an *empty* string — i.e. on the cascade trap.
- `ux-6-d-keyboard-pattern.spec.ts` arm (d) loses its subject with the override and is deleted with its rationale. Worth noting: that arm was **already vacuous** under Playwright, which resolves every inset to 0 — it read `0px` whether or not the override applied.

## Gates

- `bun run check` (biome + `tsc --noEmit` + `tsc --noEmit -p e2e/tsconfig.json`) — **rc=0**. The 61 biome warnings / 5 infos are pre-existing descending-specificity diagnostics; the gate exits 0 with them.
- `bun run test` — **rc=0**, 265 files / 4914 tests passed.
- Working tree verified clean before and after the mutation runs.

## What I refuse to claim

- **That this looks right on a real iPhone.** Nobody has seen it. jsdom resolves no `env()`; Playwright's iPhone emulation resolves every inset to 0. **Device verification is still owed**, and so is the specific question the ruling flags: whether the bottom row of tabs still takes taps reliably inside the home-indicator strip.
- **That any before/after number here is mine.** Every measurement quoted — `.bottom-bar` ending 34px short of `innerHeight`, `elementFromPoint` returning `DIV.shell shell-mobile` at `rgba(0, 0, 0, 0)` against the bar's `rgb(49, 50, 68)` — is from the issue body, taken on a `320x568x2` emulated viewport **with the inset injected by hand** (`padding-bottom: 34px`), because desktop Chrome always reports `env()` as 0. I re-measured nothing.
- **That I could have measured it.** There is no faithful simulation available offline: `env(safe-area-inset-bottom)` cannot be given a value from CSS, and hand-injecting a `padding-bottom` would be overriding the very declaration under test — a circular check. The cascade claim in this PR is a reading of specificity and source order (`.shell` at `default.css:1158`, its bottom longhand at `:1195`; `.shell-mobile` at `:6364`), verified by mutant 2 at the source level, not by a rendered measurement.
- **That the e2e specs pass.** They were edited but **not run** — no lane, and these are `@webkit` specs needing the stack. Three specs are affected; CI is the first thing that will execute them.

### DIFF
--- a/cicchetto/e2e/tests/ux-3-empty-toolbar-island.spec.ts
+++ b/cicchetto/e2e/tests/ux-3-empty-toolbar-island.spec.ts
@@ -24,12 +24,19 @@
 //
 // Verification strategy. Walk document.styleSheets, find the
 // `.shell-mobile` rule (lives under the `@media (max-width: 768px)`
-// branch — CSSMediaRule), assert it declares both
-// `padding-top: env(safe-area-inset-top)` and
-// `padding-bottom: env(safe-area-inset-bottom)`. Webkit-iphone-15
+// branch — CSSMediaRule), assert it declares
+// `padding-top: env(safe-area-inset-top)`. Webkit-iphone-15
 // emulation matches viewport but doesn't inject real env() values —
 // computed paddingTop resolves to 0px regardless — so this is a
 // SOURCE-shape assertion, not a metric assertion.
+//
+// #1127 (2026-08-09) flipped the BOTTOM edge: the home-indicator inset
+// lifted the shell off the bottom and exposed the transparent area behind
+// it (a black band under the tab strip), so `.shell-mobile` now declares
+// `padding-bottom: 0` and the shell reaches the physical bottom. The edge
+// is still asserted here, with the opposite expectation — and the
+// declaration must still be PRESENT, since base `.shell` would otherwise
+// cascade the inset back in at equal specificity.
 //
 // Also assert `.topic-bar` rule NO LONGER contains `env(` /
 // `safe-area-inset-top` (the previous fix is reverted at the bar
@@ -99,7 +106,7 @@ test("@webkit UX-3 BIS — .shell.shell-mobile carries safe-area inset; bars do 
 
   // .shell-mobile rule lives inside @media (max-width: 768px) — must
   // recurse through CSSMediaRule.cssRules to find it. The rule MUST
-  // declare both top + bottom inset.
+  // declare the top inset, and MUST declare a flush bottom edge (#1127).
   const shell = await findRulePadding(page, ".shell-mobile");
   expect(shell).not.toBeNull();
   // The longhands MAY be authored or expanded; tolerate either.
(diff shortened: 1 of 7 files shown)


vjt:

Issue #1121, opened 2026-08-09
cicchetto: closing the media viewer strands the reader above the tail with no button and no unread badge


vjt:
## Symptom (reproduced on the testnet, 3 runs, reporter-driven)

Reader is caught up at the tail. They open the media viewer on an image and stay in it. Messages arrive while the overlay is up. On close:

* the new messages are **not visible**, and stay invisible indefinitely (the reporter waited a minute);
* **no floating scroll-to-bottom button appears** — there is no visible affordance saying anything is below;
* every message arriving *after* the close is also invisible;
* the moment the reader **sends their own message**, everything appears at once.

Reproduced deliberately three times on the testnet channel against the deployed client (image link, ~15-25 s wait, three lines pushed under the open overlay). The reporter's own words for the close-edge state: "quando la chiudo, continuo a non vedere un cazzo. non c'è nessun bottone di scroll".

## Mechanism (read on `origin/main`, not guessed)

The overlay freeze itself is intentional (#196 / #219-general): while `overlayCount() > 0` the pane's position is snapshotted and every scroll writer bails (`isOverlayFrozen()`, `ScrollbackPane.tsx:1260`; the length-effect tail-follow bails at `:2120`). That part is by design and works.

The damage is in the close-edge reconciliation, `applyOverlayRestore` (`ScrollbackPane.tsx:2985-3007`):

```ts
setFollowMode(
  listRef.scrollHeight - target - listRef.clientHeight <= SCROLL_BOTTOM_THRESHOLD_PX,
);
if (listRef.scrollTop === target) return;
```

`target` is the scrollTop captured on the **open** edge; `listRef.scrollHeight` is measured **now**, on the close edge. Two consequences, and it takes both to produce the field symptom:

1. **The follow intent is computed from mismatched epochs.** If content arrived under the overlay, `scrollHeight` grew by Δ, so the expression evaluates to `Δ + (original distance ≈ 0) > 50` (`SCROLL_BOTTOM_THRESHOLD_PX = 50`, `:151`) and `followMode` is set **false** — for a reader who was sitting exactly at the tail. Three chat lines are already well past 50 px. The code conflates *"the reader had scrolled up"* with *"the content grew while frozen"*. Tail-follow is now off for all subsequent messages too, which is why the pane stays dead after the close.

[...]

Issue #608, opened 2026-08-01
cicchetto: single scroll authority for ScrollbackPane — 12 writers, no arbitration


vjt:
## Problem

`cicchetto/src/ScrollbackPane.tsx` is 3113 lines with **17 `createEffect`s** and **12 distinct places that write the scroll position** (`listRef.scrollTop = …` or `element.scrollIntoView(…)`). There is no arbitration between them: whichever writer runs last in a given frame wins, and the ordering is not the same across engines (the field reports are all iOS/WebKit).

Current writers on `origin/main`:

| line | writer | intent it encodes |
|---|---|---|
| 1711 | overlay snapshot restore (rAF×2) | restore held offset |
| 1899 | `scrollToActivation` marker branch | jump to unread divider |
| 1921 / 1923 | `scrollToActivation` tail branch | tail |
| 2392 / 2398 | length-effect, overlay-frozen branch | re-assert held offset |
| 2446 / 2448 | length-effect, `atBottom()` branch (rAF×2) | tail-follow |
| 2543 | `loadMore` prepend | preserve position by height delta |
| 2742 / 2744 | `scrollToBottom()` | tail (own send #168/#580, gesture) |
| 2846 | mention jump | smooth-scroll to anchor |

They coordinate only through shared mutable state read at different times: `atBottom`, `markerActivationPending`, `markerCursorId`, `overlayScrollSnapshot` + `overlaySnapshotKey`, `lastScrollTop`, `lastInputEventAtMs`, `visibleTailSnapshot`. Several of these are read *inside* nested `requestAnimationFrame` callbacks, i.e. one or two frames after the decision that scheduled them, so a state change in between silently flips the outcome.

## Field symptom this is meant to explain

Residual of #580 (closed by 4ad9a8c9, confirmed deployed — prod jail HEAD `e3cbe718` contains it), reported 2026-08-01 on iOS PWA:

- send a message: the row **renders**, the pane does **not** scroll;
- the floating scroll-to-bottom button is **not** shown, i.e. `atBottom()` is `true` while the viewport is not at the bottom — the pane believes it is following;
- intermittent, and once it starts it keeps happening for a while;
- a force-close of the app clears it for a while ⇒ the deciding state accumulates over the session lifetime, it is not a per-send network race.

Two branches can produce exactly that pair (`atBottom() === true` + no scroll) without any of them being "wrong" on its own:

[...]

Issue #219, opened 2026-07-12
cicchetto: preserve scroll position when opening/closing a link overlay


vjt:
Reported via #it-opers.

**Observed:** opening a link scrolls the conversation view to the bottom.

**Expected:** opening a link should not move the scroll position; on closing the link/overlay, the conversation should resume exactly where the user had scrolled to (restore prior scroll offset).

**Notes:** client-side only, no wire change. Likely the overlay open/close is triggering a layout reflow or a scroll-to-bottom on the message list; capture scrollTop before opening and restore it on close (guard against auto-scroll-on-new-message stealing focus while the overlay is up).

vjt:
Shipped to prod (m42) in the 2026-07-14 03:00 CEST attended cold window: #216+#229 server cold-deployed (SHA d2fcaed) + client bundle hash=DNgcqTPq (--cic broadcast). Verified: jail healthz 200, public https://irc.sniffo.org 200, sessions recovered.

Issue #196, opened 2026-07-08
cicchetto: scroll position lost when opening image preview overlay (desktop)


vjt:
## Environment
- cicchetto PWA, desktop browser

## Steps to reproduce
1. Open a chat with image attachments
2. Click an image to open its preview overlay
3. Observe the message list behind the overlay

## Actual
Opening the preview overlay causes the underlying message list to scroll to the top (loses its scroll position). After closing the preview the user is left far from where they were, re-reading old messages as if they were new.

## Expected
Scroll position of the message list must be preserved while the preview overlay is open/closed (scroll anchoring). The overlay should not perturb the underlying scroll container.

## Notes
Likely the preview overlay mounts/reflows the message container (or toggles `overflow`/height on the scroll parent), dropping the scroll offset. Fix: capture and restore `scrollTop`, or isolate the overlay so it doesn't touch the list's scroll container.

Reported by @gmsrc. Filed on user's behalf.

vjt:
Fixed in 14daadc (desktop e2e hardened in aa37729 after the first version flooded/hung on CI), merged to main (aa37729). CI + integration green (desktop scroll-preserve spec passes, 1.7s). ScrollbackPane now snapshots the scroll container's scrollTop on the media-viewer open edge and re-asserts it across open/close, so the preview overlay no longer strands the reader. Awaiting the batched 03:00 CEST cold deploy to go live — issue stays open until the deploy verifies healthy.

vjt:
Shipped to prod in the batch-6 COLD deploy (2026-07-10 ~03:00 CEST, origin/main `3090c01`, `--full-restart`). Verified live: healthz OK, sessions reconnected, #89 verify_peer TLS to azzurra validated clean (1 credential :connected + 27 live sessions). Closing.

vjt:
Reopened per maintainer report — regression confirmed on desktop during dogfooding (2026-07-14): scroll position in the message list is still lost when opening the image preview overlay.

Note: #219 (closed today) fixed scroll preservation for *link* overlays, but the image-viewer open/close path still resets scrollTop.

Repro: scroll up in a channel with backlog (desktop), click an inline image to open the preview overlay, close it → the list jumps (back toward bottom / unread marker) instead of holding the pre-open scroll position.

vjt:
[...]

PR #1136, opened 2026-08-09
cic(#1121): reconcile the overlay close in the epoch each answer belongs to

vjt:
Ref #1121.

## What was broken

A reader sitting at the tail opens the media viewer, three lines land underneath, they close it — and the pane goes deaf. No new messages, no floating scroll-to-bottom button, no unread badge, and it stays that way until their own send. Reproduced three times on the testnet; reproduced here in jsdom.

The freeze under the overlay is **not** the bug (#196 / #219-general, deliberate, working). The damage is one function at the close edge, `applyOverlayRestore`, doing two things wrong with one expression:

```ts
setFollowMode(
  listRef.scrollHeight - target - listRef.clientHeight <= SCROLL_BOTTOM_THRESHOLD_PX,
);
if (listRef.scrollTop === target) return;
```

* **Mixed epochs.** `target` is the scrollTop captured on the OPEN edge; `scrollHeight` is read on the CLOSE edge. Growth under the overlay therefore reads as distance the reader put between themselves and the tail. A reader who never moved is reconciled to `followMode = false` — and stays there, which is why the pane ignores messages long after the overlay is gone.
* **Geometry never republished.** The freeze held the pane exactly at `target`, so the restore writes nothing, takes the early return, and emits no `scroll` event — and `onScroll` is the only thing that re-measures. `atBottomNow` goes on asserting a tail the pane has drifted away from. Both of its consumers repeat the lie: the floating button is `<Show when={!atBottomNow()}>`, and `readingAtTailKey` — published from the same predicate — is exactly what tells `selection.ts` to suppress this window's unread badge.

## The fix

The close edge asks two questions and they belong to two epochs.

* The follow **intent** ("was the reader following when the freeze began?") is an open-edge fact: capture the distance-to-tail alongside the px, reconcile against that.
* The **geometry** ("is the pane at the tail now?") is a close-edge fact, and the old expression answered it correctly all along — it was only *misaddressed*. Publish it to `atBottomNow`, measured against `target` (both exits leave the pane there), **before** the no-op early return, because the held-position case is the one that goes stale.

#608's reconciliation is preserved, not deleted: a mid-list snapshot carries a large distance in either epoch, so a reader who really had scrolled up is still not yanked to the bottom on close. That case gains something too — its `atBottomNow` was stale-true for the same reason, hiding the button from a reader who needed it.

## One site, two signals, three symptoms

The three reported symptoms are neither three defects nor one. They are **two independent signals**: the dead intent hides the messages; the stale geometry hides *both* affordances, because the button and the badge are one signal with two consumers by design.

Three mutants, two disjoint kill sets — measured, not predicted:

| mutant | kill set |
|---|---|
| revert the intent to the mixed-epoch read | *follow intent* only (1 failed, 2 passed) |
| remove the geometry publish | *button* + *badge* only (2 failed, 1 passed) |
| move the geometry publish after the early return | *button* + *badge* (same pair) |

The third is the honest result rather than a third axis: it proves the **placement** is load-bearing — the assertions exercise the no-op-write path, which is the field path — not that a third signal exists.

## Not the media viewer's bug

Twenty surfaces push the shared overlay refcount and all route through this one freeze and this one restore. The viewer is just the one a reader keeps open for twenty seconds while a channel talks underneath. The fix is at the shared site, so it is class-wide by construction. No other mixed-epoch distance exists in the pane: every other site reads all three terms off the live element in one instant.

## Deliberately not changed

* The freeze itself. A covered pane still holds its position; only the edge reconciliation is epoch-correct now.
* The reporter's proposal to stop freezing under fully-covering surfaces — a larger policy call, recorded in DESIGN_NOTES, not settled here.
* A restored follower parked above the tail is offered a button rather than snapped forward. Conservative direction; the opposite reading ("a follower's anchor is the tail, not a pixel") is defensible and also recorded.

## Gates

* `bun run test` — **4915/4915 passed**, 265 files, rc=0, clean tree before and after.
* `bun run check` (biome + `tsc` over `src/` and `e2e/`) — rc=0. 61 warnings are pre-existing and unchanged (959 → 960 files).
* The e2e spec `issue1121-overlay-close-tail-reader.spec.ts` is **typechecked but NEVER EXECUTED** — it has had no lane. It is modelled line-for-line on its `issue196-preview-scroll-live-arrival` twin, but a spec that has not run is a claim, not evidence. Treat it as unverified until it gets one.

## What the jsdom guard does not prove

It stubs every geometry it reads; jsdom never lays out, so `isSettled` can never hold there and the tail-follow write arrives on the fail-safe frame. `pushOverlay(null)` stands in for the real viewer, so the visualViewport shrink that #196 exists to survive is never exercised. The button assertion proves the element **mounts**, not that it is visible or hit-testable. The badge assertion proves the pane publishes `null`, not that `selection.ts` then renders a count. That gap is what the e2e is for.

### DIFF
--- a/cicchetto/e2e/tests/issue1121-overlay-close-tail-reader.spec.ts
+++ b/cicchetto/e2e/tests/issue1121-overlay-close-tail-reader.spec.ts
@@ -1,0 +1,148 @@
+// #1121 — closing a covering overlay over a reader who was AT the tail must not
+// strand them above it in silence.
+//
+// The mirror image of issue196-preview-scroll-live-arrival, which drives the
+// SCROLLED-UP reader: there the contract is "do not move me", here it is "do not
+// go deaf". Both run the same harness and the same gesture; only the reader's
+// starting position differs, and that difference is the whole bug.
+//
+// While the overlay is up the pane is frozen (#196 / #219-general) and lines
+// landing underneath do not move it — that part is deliberate and is what the
+// #196 spec pins. The damage was at the CLOSE edge, in `applyOverlayRestore`: it
+// reconciled the follow intent from `scrollHeight` measured on the close edge
+// against the scrollTop captured on the OPEN one, so the growth under the
+// overlay read as distance the reader had put between themselves and the tail.
+// A reader who had not moved lost tail-follow, and — because the restore writes
+// nothing when the pane is already where it was held — no scroll event fired,
+// `atBottomNow` stayed stale-true, and both rescue affordances were suppressed
+// with it: no floating scroll-to-bottom button, and `readingAtTailKey` kept
+// telling the badge derivation to hide this window's unread count. The reporter
+// waited a minute and only their own send released it.
+//
+// Two observables, one per broken signal:
+//   * the floating scroll-to-bottom button is VISIBLE right after the close
+//     (the geometry was republished for the close edge);
+//   * a line arriving AFTER the close is scrolled to (the follow intent
+//     survived the freeze).
+//
+// The button assertion runs FIRST and on its own: the follow-up line tail-snaps
+// the pane, which legitimately unmounts the button again.
+//
+// Desktop project only (untagged → chromium), like its #196 twin: desktop Chrome
+// reproduces desktop scroll physics (feedback_playwright_webkit_not_ios_scroll).
+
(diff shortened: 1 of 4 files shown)


vjt:

Issue #1124, opened 2026-08-09
cicchetto: dismiss the soft keyboard when the media viewer opens


vjt:
## Dismiss the soft keyboard when the media viewer opens

Requested by @vjt. On mobile, tapping an image link while the compose box has focus opens the media viewer **with the keyboard still up**.

That is not an accident: `cicchetto/src/lib/keepKeyboard.ts` installs a document-level capture listener that `preventDefault`s the `mousedown` focus shift whenever a focused input is followed by a tap on a non-input surface, precisely so the keyboard survives taps on buttons and rows. A tap on a media link is one of those taps, so nothing dismisses the keyboard today.

## Why it matters

`.media-viewer-modal` is capped at `max-height: calc(var(--viewport-height, 100dvh) - 2rem)` (`cicchetto/src/themes/default.css:11412-11426`), and `--viewport-height` tracks the mobile visual viewport. With the keyboard up, the viewport is roughly halved, so the image renders in half the space it could use.

## Implementation

`openMediaViewer` (`cicchetto/src/lib/mediaViewer.ts`) is the single entry point — both the scrollback's `renderRun` link handler and `MircText` route through it — so blurring the active element there covers every door, with no per-call-site wiring to forget (the same "one owner" argument `keepKeyboard.ts` makes for itself).

An explicit `blur()` is required: `keepKeyboard` only suppresses the implicit focus *shift*, it does not stand in the way of a deliberate blur.

## Accepted trade-off

On close, the keyboard does **not** come back by itself. The compose draft is preserved (compose state is per-window and untouched by the viewer), but a reader who was mid-sentence has to tap the field again to resume typing. @vjt accepted this explicitly when it was put to him.

Do not "fix" that by re-focusing the compose on viewer close: a programmatic focus that re-raises the keyboard on close would fight the same visual-viewport resize path that motivated the overlay freeze (#196 / #219-general, and see #1121).

vjt:
Shipped in **v0.16.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.16.0

Issue #125, opened 2026-06-28
cicchetto: /list channel directory — responsive layout, tap-to-open, close button, color rendering


vjt:
The channel directory / `/list` pane (shipped in #84) has layout + interaction problems: it scrolls horizontally on mobile, the columnar layout wastes space, joined rows have no action, there's no way to close the pane, and topic color codes aren't rendered.

## Problems

- **Horizontal scroll on mobile** — the columnar layout overflows the viewport width. Must be removed entirely.
- **Columnar layout wastes space**, especially on narrow screens.
- **Tapping an already-joined channel does nothing** (no-op today).
- **No close button** on the pane.
- **Color codes not rendered** — mIRC color/formatting (`\x03` etc.) in topics shows as raw control characters.

## Responsive layout

**Desktop — 3 columns:**
1. Channel name, with the **joined label *below*** the name
2. Topic
3. User count

**Mobile — 1 column, 2 rows per channel:**
- Row 1: channel name + user count, with the **joined label *beside*** the name
- Row 2: topic

The **joined label placement is responsive**: *beside* the name on mobile, *below* the name on desktop.

**Topic wrapping:** the topic **wraps fully — no truncation**. Rows grow to fit the whole topic. (Applies to the desktop middle column and the mobile second row.)

**Featured label (couples to #85):** a channel in its network's featured set shows a **"featured" label** on its row. No special positioning — featured rows keep the normal sort order.

## Sort order

Default sort is **user count, descending** (busiest channels first). Featured channels are **not** pinned to the top — they sort by user count like everything else, just with the featured label.

## Interaction

- Tapping/clicking an **already-joined** channel → **open that channel's window** (currently a no-op). Consistent with the #85 HomePane behaviour.
- Tapping an **unjoined** channel → keep the existing JOIN behaviour.
- **Close button** on the pane → **return to the previously active window** (restore the prior selection; don't just blank the pane).

## Must-haves

- **Remove horizontal scroll entirely** — the layout reflows within the viewport width at every size.
- **Color-code rendering works** — topic formatting (mIRC color/bold via `\x03` etc.) renders through the **same typed-formatting path as scrollback**, not as raw control characters. Per the one-parser invariant, cic does NOT parse IRC framing — this is the existing typed-formatting render path applied to the directory topic field (the server already supplies the topic string; cic renders its formatting).

## Component / wire

- Pane: `cicchetto/src/DirectoryPane.tsx`.
[...]

PR #1139, opened 2026-08-09
cic(#1124): dismiss the soft keyboard when the media viewer opens

vjt:
Refs #1124.

## What

`openMediaViewer` (`cicchetto/src/lib/mediaViewer.ts`) blurs the active element
before it flips the viewer state, so tapping an image link with the compose box
focused no longer opens the viewer under a raised keyboard.

## Why here

`.media-viewer-modal` is capped at `max-height: calc(var(--viewport-height) -
2rem)`, and `--viewport-height` tracks the **visual** viewport
(`lib/viewportHeight.ts`, the single writer, rewritten on every `vv.resize`).
With the keyboard up that is roughly half the screen, so the media renders in
half the height it could use.

Nothing dismissed the keyboard on its own, and that is deliberate:
`lib/keepKeyboard.ts` `preventDefault`s the `mousedown` focus-shift on a
`.scrollback-link` tap (media links carry that class) precisely so the keyboard
survives taps on controls. It suppresses the **implicit** focus shift; it does
not stand in the way of a **deliberate** `blur()`. Hence the explicit call.

The blur lives on the verb, not the call site, because the verb is the single
open door — see the correction below.

## One door — with a correction to the issue

The issue says *"both the scrollback's `renderRun` link handler and `MircText`
route through it"*. On `origin/main` that is **stale, in our favour**: there is
exactly **one** production caller, `MircText.tsx:244`. `ScrollbackPane`'s
`renderRun` stopped being a caller in #125, when `MircBody` was extracted into
`MircText` — `ScrollbackPane` has not mentioned `renderRun` since. Verified with
`git grep -n openMediaViewer HEAD -- cicchetto/src cicchetto/e2e` (only
`MircText`, plus comments). No third door exists: `mediaViewerState` is written
only by `openMediaViewer` / `closeMediaViewer` / the identity-rotation reset.

Two comments still named the dead `renderRun`; the first commit here fixes them.

**Deliberately not touched:** `playAudio` (`lib/audioPlayer.ts`) is a sibling
door for `kind: "audio"`, but it opens a *docked* mini-player, not a viewer
capped at the viewport height — it has no keyboard problem to solve, so it keeps
its behaviour.

## Accepted trade-off

On close the keyboard does **not** come back. The compose draft is per-window
state and is untouched, but a reader mid-sentence has to tap the field again.
@vjt accepted this explicitly. It is deliberately **not** "fixed" by re-focusing
the compose on close: a programmatic focus re-raises the keyboard through the
same visual-viewport resize path that motivated the overlay freeze (#196 /
#219-general, and #1121).

## Test, and what it cannot prove

`cicchetto/src/__tests__/mediaViewer.test.ts` — one test, on the store verb.
It asserts the pre-state (the textarea holds focus), calls the verb, and asserts
focus left it.

jsdom has **no soft keyboard and no visual viewport**, so the test pins the
CAUSE (focus leaves the field) and not the EFFECT (the viewport grows back and
the media gets the full height). **The effect is device-verify only — a real
iPhone is the only witness.** No e2e was added: Playwright/WebKit has no soft
keyboard either, so a spec there would assert the same cause with far more
machinery.

### Mutation evidence (disjoint kills, full 4913-test suite each)

| Mutant | Result |
|---|---|
| baseline | `266 files, 4913 passed`, rc 0 |
| **M1** — delete `focused.blur()` | `1 failed / 4912 passed`; the single failure is `mediaViewer.test.ts > openMediaViewer blurs the focused compose field…`. The new test is the **sole** witness of the fix. |
| **M2** — delete `setMediaViewerState({ href, kind })` | `20 failed / 4893 passed` (`MediaViewerModal` + `ScrollbackPane`), and `mediaViewer.test.ts` **passes**. The new test does not merely re-assert the pre-existing open. |

## Gates

- `scripts/bun.sh run test` — 266 files, 4913 passed, rc 0
- `scripts/bun.sh run check` — biome + `tsc --noEmit` (src and e2e), rc 0, re-run
  after deleting `*.tsbuildinfo` so the new file is not a warm-cache false pass
- No lane taken: cic-only change, no server compile, no e2e

### DIFF
--- a/cicchetto/src/MediaViewerModal.tsx
+++ b/cicchetto/src/MediaViewerModal.tsx
@@ -24,7 +24,7 @@ import { maybeEscapePwaClick } from "./lib/platform";
 // spec 2026-06-10; the spec bans lightbox-on-arrival, not click-to-view).
 //
 // Driven entirely by `mediaViewerState()` (lib/mediaViewer.ts);
-// ScrollbackPane's renderRun calls `openMediaViewer` for links that
+// `MircText`'s link handler calls `openMediaViewer` for links that
 // `classifyMediaLink` accepts. Mounted at Shell root in both branches
 // (PrivacyModal pattern).
 //
--- a/cicchetto/src/__tests__/mediaViewer.test.ts
+++ b/cicchetto/src/__tests__/mediaViewer.test.ts
@@ -1,0 +1,32 @@
+import { afterEach, describe, expect, it } from "vitest";
+import { closeMediaViewer, openMediaViewer } from "../lib/mediaViewer";
+
+// Media-viewer store (#1124). `openMediaViewer` is the single open door —
+// `MircText`'s media-link click handler is its only caller — so the
+// keyboard-dismissing blur is pinned on the VERB here, not at a call site.
+// Sibling shape: audioPlayer.test.ts.
+//
+// What jsdom cannot prove: there is no soft keyboard and no visualViewport
+// resize here, so this asserts the CAUSE (focus leaves the text field) and
+// not the EFFECT (the viewport grows back, the image gets the full height).
+// The effect is device-verify only.
+
+const IMAGE_URL = "https://grappa.example/uploads/abcdefghijklmnopqrstuvwxyz";
+
+afterEach(() => {
+  closeMediaViewer();
+  document.body.replaceChildren();
+});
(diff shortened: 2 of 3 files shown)


vjt:

Issue #1106, opened 2026-08-09
iOS: long-press Select… installs no visible selection while the compose keyboard is open


vjt:
On iOS, long-pressing a message row while the compose keyboard is open and choosing **Select…** produces no visible selection at all. With the keyboard closed the same gesture works.

Reported on IRC:

> stiamo ancora prevenendo il funzionamento della selezione a tastiera aperta. facendo long press e select a tastiera aperta la selezione non appare per niente

This is the keyboard-open half of the surface #79 and #1067 built: #79 duration-gates the `mousedown` `preventDefault` precisely so a long-press keeps the keyboard up "so the iOS selection survives", and #1067 replaced #366's select-all with the menu's `Select…`. On device the selection does not survive.

## What has been ruled out, by measurement

Replayed `lib/messageMenu.ts` `selectMessageText()` verbatim in Chromium 151 against a row plus a focused `<textarea>`, and captured the paint:

| case | activeElement | rangeCount | selection text | rects | painted |
|---|---|---|---|---|---|
| textarea blurred | `BODY` | 1 | full row | 8 | yes |
| textarea focused | `textarea` | 1 | full row | 8 | **yes** |

So the generic DOM mechanics are not the fault: `range.selectNodeContents(row)` + `addRange()` installs a live range that survives a focused text control, keeps its client rects, and is still painted, with focus left on the textarea. Whatever breaks is either WebKit-specific or lives in the entry path, not in `selectMessageText`'s three lines.

Chromium is not the reporting engine — this measurement narrows the search, it does not reproduce the bug. Neither jsdom nor Playwright webkit renders iOS selection UI, which `#1067` and `default.css` both already state about their own diagnoses.

## Leads to check on device, not yet measured

1. **WebKit paints one selection at a time.** With a focused editable, the visible selection may be the field's caret, and a document range elsewhere may install without ever being drawn. If so the fix has to choose: drop the keyboard before selecting (contradicting #79's whole premise) or find a WebKit-acceptable ordering.
[...]

PR #1141, opened 2026-08-09
#1106: pin the selection window open, and name the diagnostic that cannot decide

vjt:
Investigation of #1106 (iOS: `Select…` installs no visible selection
while the compose keyboard is open). **No fix — the one lead that could
be settled without a device turned out negative, and the negative is
worth pinning.**

## What is claimed

Lead 2 of the issue — "the `is-selecting` window may close immediately"
— is a question of ordering, not of rendering, so it is measurable here.
It is false, in the DOM contract:

- `selectionchange` is dispatched asynchronously, so the event
  `selectMessageText`'s own `addRange` provokes is delivered AFTER
  `ContextMenu` has run `props.onClose()` and unmounted the portal,
  straight into the listener registered a moment earlier. That is the
  real hazard shape, and it is worth a guard.
- With a focused `<textarea>` standing in for the open keyboard, three
  such events are delivered after the click, each carrying the full row
  text. The guard is reached, finds a live selection, returns early, and
  `is-selecting` survives.
- Nothing else in cic disturbs it: the only `removeAllRanges` in the
  tree is `selectMessageText`'s own, and no code refocuses the composer
  on menu close.

`MessageContextMenu.test.tsx` pins this. It uses jsdom's real Selection
rather than the stub `messageMenu.test.ts` uses — a stub cannot deliver
an asynchronous event, so the existing "arms the callout re-enable"
test asserts synchronously and would pass unchanged if the class were
stripped one tick later. Delivery is asserted explicitly so the guard
cannot go green for want of an event. Proven by deliberate red: a
teardown that clears the selection kills that one test and no other
(1 failed of 4965).

## What is refused

Nothing about what WebKit paints. There is no iPhone and no WebKit
engine here, so leads 1 and 3 are untouched — not weakened, not
narrowed, simply not addressed. jsdom is not an engine: it does not
paint, does not focus a button on click, and its Selection/text-control
interaction is already visibly divergent (focusing a textarea *after*
installing a range empties the document selection in jsdom, where the
issue's own Chromium table shows 8 rects still painted).

## One correction to the issue

The issue proposes capturing `keepKeyboard`'s
`kb: scrollback md held=<n>ms → HOLD keep-kbd` diag line on the failing
device, "which separates lead 1 from leads 2-3 before any code is
written". By `keepKeyboard`'s own documented device model it cannot:
that arm fires on `mousedown`, and the module states that on real iOS a
long-press synthesizes no `mousedown` at all — only taps do. The HOLD
line is predicted ABSENT on the very gesture under test, so its absence
is consistent with both "arm not reached" and "nothing was dispatched to
reach it". Seeing it PRESENT would be informative, but only by
falsifying that comment.

What would decide it needs the device and no new code: a remote Web
Inspector console, immediately after tapping `Select…` with the keyboard
up, reading `getSelection().rangeCount` / `.toString()`,
`document.documentElement.className`, and `getComputedStyle(row)`'s
`webkitUserSelect` / `webkitTouchCallout`. A live non-empty range plus
`is-selecting` present plus `text`/`default` resolved means WebKit
installed everything and painted nothing — lead 1 — and the fix belongs
in the ordering or the keyboard, not in the class or the CSS.

## Gates

- `bun run test` — 269 files, 4965 tests, all green
- `bun run check` — 0 errors (61 warnings, 5 infos: the pre-existing
  baseline shape)
- Deliberate red on the mutation described above, then reverted

### DIFF
--- a/cicchetto/src/__tests__/MessageContextMenu.test.tsx
+++ b/cicchetto/src/__tests__/MessageContextMenu.test.tsx
@@ -1,0 +1,105 @@
+// @vitest-environment jsdom
+import { fireEvent, render, screen } from "@solidjs/testing-library";
+import { beforeEach, describe, expect, it } from "vitest";
+import type { ScrollbackMessage } from "../lib/api";
+import { closeMessageMenu, openMessageMenu, SELECTING_CLASS } from "../lib/messageMenu";
+import MessageContextMenu from "../MessageContextMenu";
+
+// #1106 — the ORDERING half of "Select… installs no visible selection while the
+// compose keyboard is open".
+//
+// `selectMessageText` installs the range, puts `is-selecting` on <html>, and
+// only THEN registers the `selectionchange` listener that strips the class once
+// the selection is gone. `ContextMenu.handleItemClick` runs `item.action()` and
+// then `props.onClose()`, so the whole menu — portal, backdrop, buttons —
+// unmounts immediately after. `selectionchange` is dispatched ASYNCHRONOUSLY
+// (the spec queues a task), so the event the install itself provokes is
+// delivered AFTER that teardown, straight into the freshly-registered listener.
+// If the teardown emptied the selection, the class would be gone within a
+// macrotask and `html.is-ios.is-selecting .scrollback { -webkit-touch-callout:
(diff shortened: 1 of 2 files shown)


vjt:

Issue #1042, opened 2026-08-08
cicchetto: is the pane's top-right corner a contract for every window kind, or is the admin window legitimately different?


vjt:
Split out of #1039 rather than folded into it. #1039 fixed the two hosts it
was filed against — the channel `TopicBar` ☰ and the non-channel `ShellChrome`
float now read one inset token and land on the same rectangle. A **third**
host landed in #1033 (2026-08-07), after #1039 was written, and it places the
same control somewhere else entirely.

This issue is a question, not a defect report. It may well close as "working
as intended".

## What is there, measured on `origin/main` @ `d530a067`

Three containers host the one ☰ on mobile, mutually exclusive by construction:

| window kind | host | where the glyph lands |
|---|---|---|
| channel | `.topic-bar` (`themes/default.css:1879`) | top-**right**, `--pane-chrome-inset-*` from the pane corner |
| query / server / home / list / mentions | `.shell-chrome .shell-chrome-rail-opener` (`:2432`) | top-**right**, same token since #1039 |
| admin | `.admin-pane-header .shell-chrome-rail-opener` (`:4206`) | top-**LEFT**, `1rem` in from `.admin-pane`'s padding (`:4172`) |

`AdminPane.tsx:139` renders `RailOpenerButton` as the **first** child of a
`justify-content: space-between` header (☰, title, refresh, ×), and
`Shell.tsx:796` suppresses `.shell-chrome` for the admin kind
(`… && !isAdminPaneVisible()`) so the two mounts never coexist.

So entering the admin window does not move the ☰ by the 3.5px/7px #1039 was
about — it moves it across the pane.

## Why it is that way, and why #1039 did not touch it

Deliberate, and argued in `ShellChrome.tsx:73-90`: the admin window carries
its own header, so hosting the ☰ there spends one chrome band instead of two,
and — the part that matters here — #985 floated the opener over the pane's
top-**right** corner, *"which is exactly where the admin pane now puts its
close × and, on a phone, its refresh"*. The corner is occupied.

That is a real constraint, not an oversight, and reversing a placement one day
after it merged, without its author, is not a call to make inside a small
geometry fix. Hence this issue.

## The actual question

**Is "the ☰ is always in the pane's top-right corner" a contract that binds
every window kind?**

- If **yes**, the admin window has to give that corner up: × and refresh move
  (into the data card's own header, where #1033 already puts refresh on
  desktop via `refreshInCardHead`?) and the admin header loses its leading ☰.
[...]

Issue #358, opened 2026-07-20
cicchetto: day/night theme pairing — gallery theme should follow prefers-color-scheme like the base does


vjt:
Requested per a #grappa request (day/night auto theme switch).

Today the **base** theme already auto-switches with the OS: `cicchetto/src/lib/theme.ts` resolves `prefers-color-scheme` (`resolveAuto()`) between `mirc-light` and `irssi-dark`, and a `matchMedia("(prefers-color-scheme: dark)")` `change` listener (theme.ts:47-48) propagates OS light/dark flips live.

The **#75 theme gallery** (cog → themes) layers a single user-picked theme as inline CSS vars *over* that base, applied the same regardless of day or night. So a user who picks a gallery theme loses the automatic light/dark behavior the base has.

**Ask:** let the user pick a **day theme + a night theme** from the gallery and have cicchetto swap between them automatically, driven by the *same* `prefers-color-scheme` signal that already flips the base — no time-of-day scheduler and no geolocation (phones already auto-toggle dark mode on schedule/at sunset; we just follow the OS media query).

**Scope sketch:**
- Extend gallery selection state from one theme id to an optional `{ light, dark }` pair (single pick = same theme both modes → backward compatible).
- On the existing `matchMedia` change listener (theme.ts), re-apply the matching gallery layer for the resolved mode.
- Settings UI in the #75 gallery: a "use a different theme for light / dark" toggle exposing the second slot.

Baseline verified on `origin/main` (theme.ts). Related but distinct: #334 (new backgrounds), #293 (row/format layout in themes), #83 (AI-generated themes).

vjt:
Shipped in grappa 0.4.0 — cold-deployed to m42 prod 2026-07-25 (healthz 200, 33 sessions reconnected, migrations applied). Release: https://github.com/vjt/grappa-irc/releases/tag/v0.4.0

PR #1166, opened 2026-08-09
#1051: the wallpaper made the pane a stacking context, so 42 never met 41

vjt:
## What was wrong

`c9333004` raised `.scrollback-overlay` from 5 to 42, over `.shell-chrome`'s
41, and shipped in v0.15.0. The reporter still could not tap the ✕ of a WHOIS
card on a non-channel window.

The number was never the deciding factor. `:root.theme-has-bg
.scrollback-pane { isolation: isolate }` — the #75 wallpaper block, three
thousand lines away — makes the pane a **stacking context**, so the overlay's
42 resolves inside the pane and never meets `.shell-chrome` at all. What meets
it is the pane, at `z-index: auto`, and the ☰ therefore wins the corner at any
number the overlay could carry.

That class is present only for operators running a background image, which is
the whole shape of this incident: one build, fixed for half the userbase,
untouched for the other half, with every re-measurement that did not toggle
that one class reading green.

## The change

Drop the isolation. The block already carried the two rules that do the
ordering — `::before` at 0, `.scrollback` at 1 — and against the root they
resolve in the same relative order they did inside the isolated pane. The
wallpaper is unchanged; the overlay is free to outrank the chrome.

Rejected: keeping the isolation and moving the overlay or the ☰ so the two
share a context. DOM surgery in `Shell.tsx` for a one-declaration cause, and it
leaves the trap armed for the next in-pane layer that has to outrank the
chrome.

Not available: `::before { z-index: -1 }`. Measured, not assumed — see below.

## Why the guard had to change too

`src/__tests__/shellChromeFloat.test.ts` regex-extracts the two z-index values
from the stylesheet text and compares them. Both were correct throughout the
defect and the assertion passed the whole time. No text-level assertion can
observe a stacking context. That file stays (the numbers are still a contract)
but it is not the guard.

`e2e/tests/issue1051-card-close-above-float.spec.ts` now runs one shared body
twice, parametrised on whether a theme carrying a real built-in wallpaper is
active before boot. Each arm asserts its own premise (`html.theme-has-bg`
absent / present) so neither can decay into a copy of the other.

Measured on `webkit-iphone-15`:

| | no-wallpaper arm | wallpaper arm |
|---|---|---|
| before the CSS change | pass | **fail** — `elementFromPoint` at the ✕'s centre returns `.shell-chrome-rail-opener` |
| after | pass | pass |

## The half the fix put at risk

Removing the isolation is a fix only if the wallpaper still paints, and still
paints behind the conversation — which no hit test can see. A third test
measures it in pixels at opacity 1.0, toggling the wallpaper through the #358
day/night pair (`emulateMedia` flips light↔dark live: same document, same
scroll offset, one variable). Both themes copy the same built-in palette, so
only `background.builtin` differs, and a first phase runs the identical flip
with the control theme in *both* slots and requires byte-identical images —
measuring the flip's inertness instead of asserting it.

The first version of that oracle compared screenshots either side of a
`page.reload()` and was a **false green**: mutating the layer to `opacity: 0`
still passed it, because a reload changes the pane for reasons of its own. The
corrected oracle was then re-mutated, each mutation killing exactly one
assertion:

- `::before { opacity: 0 }` → only *"an opaque wallpaper must change what the scrollback area paints"* fails.
- `::before { z-index: 2 }` → only *"a message arriving must be visible over the wallpaper"* fails.
- `::before { z-index: -1 }` → the first claim fails; the wallpaper stops painting entirely. That mutation read **green** under the pre-correction oracle, which is why -1 is ruled out on measurement rather than on reasoning.

## What this does not establish

- `webkit-iphone-15` only — the float is mobile-only, and that the desktop shell never mounts it is an argument from `Shell.tsx`, not a measurement.
- The pixel claims cover one catalogue background at opacity 1.0 on one viewport. Nothing here speaks to blending at the 0.3 default or to the other thirteen entries.
- Nothing was verified against a production install.
- #1042 stays open and untouched: whether the pane's top-right corner is a contract for every window kind is still undecided.

### DIFF
--- a/cicchetto/e2e/fixtures/grappaApi.ts
+++ b/cicchetto/e2e/fixtures/grappaApi.ts
@@ -173,7 +173,16 @@ export async function loginVisitor(
 // author-nick e2e. The runner talks to grappa directly on port 4000
 // (bypassing nginx), so these mirror the cic `themesApi` verbs without the
 // browser. Only the fields the spec asserts on are typed.
-export type ThemeWire = { id: number; author: string; built_in: boolean };
+export type ThemeWire = {
+  id: number;
+  author: string;
+  built_in: boolean;
+  // The sanitized token map (`Grappa.Themes.TokenModel`). Left as an open
+  // record: specs that need a palette COPY one off a built-in rather than
+  // spell 27 colour keys, so nothing here is read field-by-field except
+  // `background`, which #1051 rewrites.
+  payload: Record<string, unknown>;
+};
 
 export async function listGalleryThemes(token: string): Promise<ThemeWire[]> {
   const res = await fetch(`${GRAPPA_BASE_URL}/themes`, {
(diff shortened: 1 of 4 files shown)


vjt:
Superseded by #1180, which replays this branch's commits onto current `main` together with the other held PR so the union is gated once rather than each half separately. A per-PR green attests `branch + base`, not `branch + whatever lands beside it`.

Left open until #1180 merges, so this stays the fallback if the union goes red. The cherry-pick gives the commits new SHAs, so GitHub cannot close this one by itself — it will be closed by hand, naming #1180.

vjt:
Landed via #1180, which replayed this branch's commits onto current `main` together with the other held PR and gated the union once. The cherry-pick gave the commits new SHAs, so GitHub could not close this by itself — closing by content, naming the union.

Merged as `4d9c245f`.

Issue #1163, opened 2026-08-09
admin: binding a credential does not start the session — only a restart does (unbind stops it live)


vjt:
## What happens

Binding a credential from the admin console writes the row and returns `201`, but the
session never dials out. The operator sees a credential that exists, is listed, reports
no live session, and does nothing — until the node is restarted.

Unbinding, by contrast, takes effect immediately: `DELETE /admin/credentials/:user_id/:network_id`
delegates to `Credentials.unbind_credential/2`, which detaches the credential **and stops
the live session** (the controller's own docstring says so).

So the two halves of the same admin control disagree: **unbind is live, bind is not.**

## Measured on `origin/main` (`3383abf6`)

* `GrappaWeb.Admin.CredentialsController.create/2` — the whole success path is
  `bind_with_conflict_classification/3` → `emit_credential_bound/4` → `LiveIntrospection.lookup_session/2`
  → render. `lookup_session` only *reads* whether a session happens to be live; nothing
  starts one.
* `Grappa.Networks.Credentials.bind_credential/3` is a plain
  `%Credential{} |> Credential.changeset(attrs) |> Repo.insert()` (plus preload). No
  `Session.start_session/3`, no supervisor call.
* The spawn producer is `Grappa.Bootstrap`, and it runs **once at boot** — a `Task` with
  `restart: :transient` in the supervision tree. Its own moduledoc states the contract:
  "Bootstrap re-reads the DB every boot, so the next deploy picks up new bindings without
  any config edit." Every boot. Not on write.
* `POST /admin/reload` is not a workaround: `Grappa.HotReload.migrate_and_reload/0` is
  code hot-reload (migrations + module reload), not a Bootstrap re-run.

## Why this is worth fixing rather than documenting

It lands hardest on exactly the person we least want to lose — a first-time self-hoster
following the only path available to them. Reported from the field tonight by an operator
running `ghcr.io/vjt/grappa:latest`: they created the account, hit
"no networks", and went to the admin console because on a **release image the CLI has no
`bind-network` verb at all** (see #1158 — the release ships no Mix, so every `boot|` verb
of the repo dispatcher is absent). The console is not one door among several for them.
It is the only door, and it silently half-works.

The failure is also indistinguishable from a real connection problem: no error, no state
change, no log line the operator would think to look for. The natural next move is to
debug IRC connectivity — a network that was never dialled.

## Suggested shape

Make `POST /admin/credentials` spawn the session on success, mirroring what unbind
[...]

PR #1171, opened 2026-08-10
fix(#1163): an admin credential bind dials the upstream

vjt:
Fixes the asymmetry in #1163: `DELETE /admin/credentials/:user_id/:network_id`
stops the live session immediately, while `POST /admin/credentials` wrote the
row, returned `201`, and never dialled. Its success path only *read* whether a
session happened to be live (`LiveIntrospection.lookup_session/2`).

## What was actually broken

Worse than "no session": `Credential.connection_state` defaults to
`:connected`, so a console bind wrote a row that **claimed** a session it did
not have. cic rendered CONNECTED — nothing for the operator to press — while
`POST /messages` answered *"That network or resource doesn't exist."* That is
the state the U-0 fix eliminated on the `PATCH /networks/:id` door
(`networks_controller.ex:395` records it verbatim), reproduced by construction
on every admin bind.

## The door

`Grappa.Operator.connect_credential/1`, the user twin of the Visitors-tab
`reconnect_session/2`, reaching the shared
`Grappa.SpawnOrchestrator.spawn/4` (admission -> `Backoff.reset` -> spawn) that
`Bootstrap`, `NetworksController`, `Admin.SessionsController`, `Operator` and
`Visitors` already reach. No second copy of the sequence, so bind-time and
boot-time cannot drift.

`GrappaWeb.NetworkSpawn.orchestrate/4` — the web-layer helper — was the wrong
door despite being the closer one: it builds the capacity input from the
*request's* IP, correct when a user connects their own network, wrong when an
admin acts for somebody else (the admin's IP would count against the target
user's per-IP cap). `source_ip: nil` + a new `:admin_credential_bind` flow,
exactly as `reconnect_session/2` does for the same on-behalf-of reason.

## Ordering and failure posture

* **U-0 ordering.** Bind `:parked`, spawn, and only a live `Session.Server`
  promotes the row to `:connected`.
* **A refused spawn is not a failed bind.** The row is the operator's input, so
  it stays — unlike accretion (`SessionController.spawn_or_rollback/4`), which
  unbinds on refusal because there the *user* asked to join a network. It stays
  `:parked` (recoverable via the owning user's `PATCH /networks/:id`), the
  response is still `201`, and it says why.
* **Wire, additive, snake_case, no `protocol_version` bump** — proposed shape,
  mirroring the `session_action` the PATCH response already carries:
  * `session_action`: `"spawned" | "not_spawned"`
  * `session_error`: `null`, or the refusal tag (`"resolve_failed"`,
    `"user_cap_exceeded"`, `"network_circuit_open"`, `"start_failed"`, ...)
* **The servers-bound invariant did not move**: the resolver is Bootstrap's own
  `SessionPlan.resolve/1`, so "no enabled server" fails at bind time the same
  way it fails at boot — not through a softer second copy of the rule.

## Tests — declared red before the fix, and observed red

| test | failure without the fix |
|---|---|
| bind dials out (fake upstream sees `NICK`, pid registered, row `:connected`) | `wait_for_line` -> `{:error, :timeout}` |
| network with no enabled server: `201`, row kept `:parked`, body says not-spawned | `session_action` absent (`nil`); row reads `"connected"` with `live_state: null` — the U-0 lie, visible in the failure output |
| circuit **closed**: both the boot door (`Bootstrap.run/0`) and the bind door bring a session up | bind half: no pid |
| circuit **open**: neither door brings a session up | `session_error` absent |

The last two are the "one code path" acceptance criterion. Asserting only that
each door spawns would leave the plausible wrong implementation — an inline
copy that skips admission — fully green; running both doors against the same
network under the same gate kills it.

## Gates

* **Correction to an earlier claim in this PR body.** It said `scripts/check.sh`
  was green *including* the wireTypes drift gate. That was wrong, and CI caught
  it. The drift gate is the LAST step of `check.sh`, and I read the final run as
  green from the ABSENCE of test-failure headers in a filtered log — absence of
  a failure I never proved the run had reached. Re-run directly, the gate was
  red. Now green, verified by the gate's own words: `wireTypes.ts is in sync.` /
  `wireSchema.ts is in sync.`
* Regenerating exposed more than a stale file — see the third commit: a named
  public type in a `*Wire` module IS a wire contract, and `session_action` is
  one key that must carry one union.
* What is green, and on which tree — stated precisely, because the claim above
  went wrong by being vague:
  * On the tree at commit 2 (`style(#1163)`): full `scripts/check.sh` stage 1 —
    format, credo --strict, sobelow, deps.audit, doctor, **dialyzer**, docs —
    plus the suite, 5647 tests / 0 failures.
  * On the tree at commit 3 (the regeneration): the drift gate
    (`in sync.` twice), the admin credentials test file (39/39), and cic
    `bun run check` (biome + tsc + e2e tsc) exit 0, 0 TS errors.
  * **Not re-run against commit 3: dialyzer, the full suite, bats.** Commit 3
    changes two `@spec`s and one `@type`, no runtime code, so I expect them to
    hold — but expecting is not measuring, and CI is paying for it rather than
    the host, which is busy with another worker's e2e.
  One earlier `check.sh` run of the same tree reported 3 failures that were not
  captured and did not reproduce across two subsequent runs — recorded rather
  than dropped; the shared test DB is contended across worktrees on this host.
* e2e not run locally (STACK lane not held). Blast radius looks nil by
  inspection: every e2e spec that binds a credential creates a **server-less**
  network, so the new spawn stops at `resolve_failed` and never dials the
  testnet. The only visible change there is the badge reading `parked` instead
  of a fictitious `connected`.

## Not covered

* No browser-level check of the admin console; the tests drive the HTTP door.
* cic renders no copy for the two new fields yet — it casts the response
  without validating, so they are inert client-side. The honest signal it
  *does* now show is `parked` instead of the fictitious `connected`. Wiring
  the toast is a follow-up, pending the wire shape being accepted.
* `PATCH /admin/credentials/...` against a live session (issue's point 3) is
  untouched — out of scope here.

### DIFF
--- a/cicchetto/e2e/tests/issue1163-admin-bind-dials.spec.ts
+++ b/cicchetto/e2e/tests/issue1163-admin-bind-dials.spec.ts
@@ -11,10 +11,31 @@
 // That spec binds against networks it creates with `POST /admin/networks` and
 // no server row. A server-less network fails plan resolution, so even the
 // FIXED code does not dial there — the scenario is blind to this bug by
-// construction. This spec therefore adds the one thing that was missing: an
-// enabled server, pointed at the same `bahamut-test:6667` testnet the seeder
-// uses. Everything else stays throwaway (own user, own network, own nick), so
-// nothing here touches the seeded vjt / m9b credentials other specs depend on.
+// construction. What this spec needs, and it has to get it somewhere, is a
+// network with an ENABLED SERVER.
+//
+// ## Why it borrows the seeded network instead of building its own
+//
+// The first version of this spec built a throwaway network and attached a
+// server to it. It passed — and 26 minutes later, in the same run, it failed
+// two assertions in `ux-6-g-admin-mobile-h-scroll.spec.ts`, which measures
+// that no admin table is wider than a phone panel: 393px against a 365px
+// panel, 28px of overflow, read identically by that spec's scrollLeft clamp.
+//
+// The throwaway network could not be deleted. `Networks.delete_network/1`
+// refuses a network that has scrollback, and this is the only spec whose
+// network gets a LIVE session — the moment it dials, the server's `$server`
+// notices are persisted against that network. The teardown's DELETE was
+// refused, the best-effort `catch` swallowed the refusal, and a row with a
+// 27-character slug sat in the Networks tab for the rest of the run
+// (`workers: 1`, so every later spec saw it). The width failure was this
+// spec's leak, arriving late and wearing another spec's name.
+//
+// So it binds a throwaway USER to the SEEDED `bahamut-test` network instead.
+// That network is already in the Networks tab and the baseline already fits
+// it, so this spec adds nothing to any measured table. The seeded vjt / m9b
+// CREDENTIALS other specs depend on are untouched: a second credential on the
+// same network is additive, and unbind removes it and stops its session.
 //
 // ## The oracle
 //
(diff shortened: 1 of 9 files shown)


vjt:
Superseded by #1180, which replays this branch's commits onto current `main` together with the other held PR so the union is gated once rather than each half separately. A per-PR green attests `branch + base`, not `branch + whatever lands beside it`.

Left open until #1180 merges, so this stays the fallback if the union goes red. The cherry-pick gives the commits new SHAs, so GitHub cannot close this one by itself — it will be closed by hand, naming #1180.

vjt:
Landed via #1180, which replayed this branch's commits onto current `main` together with the other held PR and gated the union once. The cherry-pick gave the commits new SHAs, so GitHub could not close this by itself — closing by content, naming the union.

Merged as `4d9c245f`.

Issue #1157, opened 2026-08-09
cic(admin): a phone should not pan the admin tables — rows become cards below 900px


vjt:
vjt on `#grappa` (2026-08-09 23:05), setting the 1.0 bar: *"dobbiamo avere un admin decente e
utilizzabile prima"*, and reporting that the admin console **still scrolls horizontally on a
phone** on production `0.15.0-7a907188`.

## This is not a leftover — it is the current design, measured

`#1074` ("no horizontal scroll on a phone, and open the row detail in place") shipped in
`v0.15.0` (`fb921a4f`, `git tag --contains` = `v0.15.0`) and did what its title says about the
**row detail**. It did not remove the pan, it **moved the scroller**:

`cicchetto/src/themes/default.css`, the `UX-6-G ownership` block (~L5470):

```css
@media (max-width: 899px) {
  .adm-table-wrap { overflow-x: visible; }
  .adm-card { overflow: visible; width: max-content; min-width: 100%; }
}
```

with the comment stating it outright: *"BELOW 900px the PANEL owns the horizontal scroll, not
the table wrapper… on a phone the wrapper stops scrolling, the card is allowed to grow to its
widest content, and the overflow lands back on the panel where the spec looks for it."* The
lock is guarded by `ux-6-g-admin-mobile-h-scroll`, which exists because of a real iPhone report
("horiz content, scrollbar, but content doesn't move").

So today, on a phone, a wide admin table **is meant to pan**. The report is not a regression
and not a cache artifact — it is a disagreement with the current design.

## What the fix has to be, then

Panning a table on a 393px viewport is the thing to remove, not to relocate. The tables carry
columns that a phone cannot show side by side, so the shape has to change on mobile:

- **Rows become cards below 900px** — one record per block, label/value stacked, no column
  alignment across records. That is the only layout that makes the pan unnecessary rather than
  hidden.
- The **row detail already opens in place** (#1074), so the interaction model for "see one
  record fully" exists; this issue is about the list around it.
- Column priority has to be decided per tab: which one or two fields identify a record at a
  glance (Sessions, Vhosts, Accounts, Networks all differ). That is a product call, not a
  refactor.

## Constraints for whoever takes it

1. **Do not just re-hide the scroller.** Moving `overflow-x` one level in or out again keeps
   the pan and invalidates the guard — the exact trap the CSS comment already documents.
   `ux-6-g-admin-mobile-h-scroll` must either keep passing honestly or be replaced by an
[...]

Issue #1074, opened 2026-08-08
cic(admin): no horizontal scroll on a phone, and open the row detail in place


vjt:
## What

Two complaints from vjt's admin brief (`#it-opers`, 2026-08-08), in scope for **1.0**:

1. *"l'expansion delle righe non deve portare top, dovrebbero aprirsi in line nella stessa
   posizione dove sono le righe"*
2. *"su mobile poi admin non dovrebbe proprio fare horiz scroll e usare l'expansion che è
   stata introdotta"*

They are the same knot, and the second is what causes the first.

## Measured (`origin/main` 08f0fcb0)

**The detail panel is outside the table, and rendered BEFORE it.** In every tab that has one
— Users, Sessions, Visitors, Networks, Credentials — `AdminDetailPanel` is a sibling of the
table card, placed above it. `AdminUsersTab.tsx`: panels at `291-353`, table card at
`356-437`. On open it calls `el.scrollIntoView({ block: "nearest" })`
(`admin/AdminDetailPanel.tsx:64-65`). Tap a row far down the list and the viewport travels up
to the panel — that is exactly the "porta in cima" being reported.

Both halves are deliberate and documented in that file's header comment:

> An expand row lives in a `<td colspan>` of a table that is deliberately WIDER than the
> viewport — the UX-6-G contract keeps admin tables wide and horizontally pannable rather than
> collapsing them to cards. Anything inside that cell inherits the TABLE's width, so a form in
> there is unreachable on a phone without swiping sideways [...]
>
> A previous attempt pinned the cell with `position: sticky; left: 0` and capped its children
> at `92vw`. That put a horizontally-scrolling nested table inside a sticky viewport-capped
> cell of another horizontally-scrolling table; the two scroll contexts fought, and panning
> the outer one clipped the nested table. Reverted.

And the scroll:

> The panel renders BEFORE the table for the same reason (so closing it leaves you at the top
> of the list rather than stranded past it).

So: the panel cannot go back inline **while the table is wider than the viewport**. Kill the
h-scroll and the blocker is gone — the two requests resolve each other. Don't attempt them
separately, and don't retry the sticky-cell approach; it has already been tried and reverted.

## The h-scroll is already half gone

`a669e8cf` introduced `.adm-col-detail { display: none }` (`default.css:5050`, restored at
`min-width: 900px`) — secondary columns are dropped on a phone and shown in the row's detail
[...]

Issue #388, opened 2026-07-24
Flavor-agnostic post-registration identity detection (unblocks #349 wizard beyond Azzurra)


vjt:
Split out from #349 (registration wizard) per chan request. Decision: ship the wizard on Azzurra now, track cross-network identity detection here.

## Problem (verified in code)
The registration wizard keys "registration complete" off the lowercase user mode `+r`. That signal is emitted **only** by bahamut/Azzurra. On other network flavors the wizard cannot detect completion:

- **Libera / solanum**: no `+r` registered umode. Identity is signalled via account — WHOIS numeric 330 (RPL_WHOISACCOUNT) and the `account-notify` capability (`ACCOUNT` messages).
- **OFTC**: uses uppercase `+R` for the equivalent registered state.

grappa currently negotiates only `sasl` + `labeled-response` and does **no** `account-notify` / `ACCOUNT` handling, so its sole identity signal is lowercase `+r`. Result: the wizard works on Azzurra only.

## Scope
Grow a flavor-agnostic identity / registration-state detection layer:
- [ ] Negotiate the `account-notify` capability; handle inbound `ACCOUNT` messages.
- [ ] Parse WHOIS-330 (RPL_WHOISACCOUNT) for self as a post-identify confirmation.
- [ ] Detect uppercase `+R` (OFTC) in addition to lowercase `+r` (bahamut/Azzurra).
- [ ] Wizard consumes a normalized "identified/registered" event instead of a raw mode.

Blocks the non-Azzurra rollout of the #349 wizard.

vjt:
## Live repro from a self-hoster, and a fresh measurement confirming the diagnosis

A self-hoster configured Libera.Chat on their own instance tonight and reported that
NickServ identification works, but the UI still shows `identified: no` in the sidebar.
That is this issue, observed in the wild — so the impact is wider than the #349 wizard:
the **session sidebar reports the wrong identity state** on any non-bahamut network.

### Measured on Libera.Chat just now

Queried from an already-identified connection on Libera (`tantalum.libera.chat`):

```
< 330 vjt-claude vjt-claude vjt-claude :is logged in as
< 221 vjt-claude +Ziw
```

So on solanum the account is exposed through **numeric 330**, while the user mode set of
an identified user is `+Ziw` — there is **no `+r`**. This confirms the analysis in the
issue body from the live network rather than from documentation.

### Why the sidebar inherits the same defect

[...]

PR #1172, opened 2026-08-10
cic(admin): rows become cards on a phone, and the pan comes out

vjt:
Fixes the #1157 report: the admin console still scrolls sideways on a phone
on production `0.15.0`.

It was not a regression. #1074 shipped the `UX-6-G ownership` CSS block,
whose own comment said "below 900px the PANEL owns the horizontal scroll,
not the table wrapper" — a phone was *meant* to pan a wide admin table.
vjt reversed that premise after dogfooding. Below 900px a row is now a
card; above 900px nothing moves.

The pan is removed, not relocated. That distinction is the issue's first
constraint, because relocating is what #1074 did: it keeps the gesture and
invalidates whatever guard was watching the old element.

## What is in here

* The unified Sessions view merges the deleted Visitors tab, row-backed so
  parked / failed / expired-but-unreaped subjects stay visible.
* Vhost addresses that can only answer 409 are gone from the create picker,
  and the #543 derivation block is dropped server-side.
* `last_seen_at` on the two row-backed admin wires (additive, snake_case, no
  `protocol_version` bump).
* Cards below 900px for every admin table, `data-label` on every `<td>`,
  `adm-cell-title` on the cell that heads a record.
* The identity column's CSS, which the merge commit emitted classes for and
  never wrote.

## Three things the browser found that review would not have

**The old guard was green while six tabs panned.**
`ux-6-g-admin-mobile-h-scroll` looped three tabs but gated its width
assertion on `networks` — the one tab vjt did not ask to change — and
conceded in a comment that the other two were empty and would pass
trivially. Reproducing the #1074 arrangement under the new oracle shows
**six** tabs overflowing, including `events`, `session_log` and `debug`,
which contain no `<td>` at all.

**The identity column was shipped without CSS.** Six classes emitted by the
merge commit had no rule in the project's only stylesheet, so three of the
four dictated properties of column 1 did not exist while a code comment
claimed one of them outright. jsdom has no layout, so the unit suite stayed
green throughout. Measured red at 7.64px, then 16px, then green.

**A row-backed list broke a count.** `m9b`'s row-count canary reads 4 in a
scoped run and 6 in a full one: rows now outlive their pids, so the total
became a function of suite execution order. It asserts per-subject presence
instead. My first diagnosis of this was wrong (I said stray visitors; they
are users) and the retraction is in the commit.

## Guards

The pan guard now asserts that at 393px no scroll container inside the admin
pane has `scrollWidth > clientWidth`, on all nine tabs, seeded with real
rows — a claim no relocation of the scroller satisfies. Demonstrated by
watching the offender move from `.adm-table-wrap` to `.admin-tab-panel` when
the old block is restored. `.adm-nav`, the nine-chip tab strip, is exempt by
name: it is a navigation affordance you swipe to choose a tab, not a record
you pan to read.

Every layout claim is pinned by a mutation that kills exactly one assertion.

## Not done, on purpose

The dictation's point 4 asks for disconnect/terminate to collapse into a
single button with a dropdown on mobile. With the card layout its reason is
gone — the buttons fit in a block and the pan guard is green without it —
so it is not built. Flagged here because that is a product call, not mine.

## Verification

`bun run check` clean, vitest 270 files / 4996 tests green, and the full
`scripts/integration.sh` suite run three times. The only reds left are
disjoint across runs (`p0b-peer-away` once, `issue168-scroll-authority`
once, never the same twice), both on desktop chromium where the
`max-width: 899px` block is inert and no added selector applies.


vjt:
Heads-up before this can merge: `main` moved to `4d9c245f` (the #1051 + #1163 union), and rebasing this branch onto it conflicts in `lib/grappa_web/controllers/admin/credentials_controller.ex` — #1157's truthful last-seen against #1163's bind-dials rewrite. Seven files overlap in total (`admin_wire.ex`, the controller and its test, `wireSchema.ts`, `wireTypes.ts`, `default.css`, `DESIGN_NOTES.md`); the controller is the only one git could not auto-merge.

That resolution is semantic, so it belongs to this branch's author rather than to me. Queued for w2 right after #388's PR is open.

Issue #1224, opened 2026-08-11
cic(admin): ended sessions belong in their own sub-view with post-mortem fields, not inline in the live list with three em-dash columns


vjt:
vjt, same iPhone dogfood pass on staging `b0d5342d` (`#grappa`, 2026-08-11 15:59), one screenshot
of the Sessions tab further down the list:

> direi che poi i visitor deleted son utili ma vanno su sotto pagina con campi diversi perché son
> sempre offline

Sibling of #1223 (same pass, different defect) and the direct follow-up to #1158 item 4 / #1221,
which is what put these rows on screen in the first place.

## What the rows look like today

`session_log`-origin rows — the ones for a subject that no longer exists — sit inline in the same
list as live sessions, carrying the `deleted` badge. On a phone card that reads:

```
▸ [user]    ccping   ● deleted
  azzurra
  LAST SEEN   —
  CHANNELS    —
  ACTIONS     —
```

Three of the four dictated columns are an em-dash, on every such row, structurally and forever:
`last seen`, `channels` and `actions` are all live-process facts, and by construction there is no
process and no DB row. The drill-down then spends its space saying the same thing in prose —
`CONNECTION: no DB row`, `LIVE: BEAM has no pid`, `UPSTREAM: —`.

So the list is paying full column width for fields that can never be populated, and the operator
scans past rows that look broken rather than finished.

## What these rows actually have

`SessionLogWireT` (`cicchetto/src/lib/wireTypes.ts:1322`) is the whole record behind them:

| field | use |
|---|---|
| `nick`, `subject_kind`, `network_slug` | identity — already rendered |
| `event` | what ended it (`:disconnected`, …) |
| `at` | when it ended |
| `reason`, `clean` | how it ended — `:shutdown (clean)` is already in the panel |
| `duration_ms` | how long it ran |
| `old_nick`, `delay_ms`, `attempt` | reconnect/rename history |

Not one of them is `last seen` / `channels` / `actions`. The record is a *post-mortem*, and the
table it is being rendered into is a *live console*.

## What vjt is asking for

A separate sub-page (or sub-view under Sessions) for ended sessions, with **its own columns** drawn
from the list above — ended-at, reason/clean, duration, network — instead of the four live ones.
They are always offline, so nothing there needs a liveness column or an actions cell.

Points that need deciding before anyone writes code, and they are product calls:

1. **Where it lives** — a sixth tab, or a link/toggle inside Sessions ("show ended"). The tab strip
[...]

Issue #1223, opened 2026-08-11
cic(admin): on a phone the detail panel repeats what the card already shows, the row name is the only tap target, and the facts grid throws away half the width


vjt:
vjt, dogfooding staging on an iPhone at `b0d5342d` (`#grappa`, 2026-08-11 15:41):

> admin stiamo buttando un sacco di spazio orizzontale e abbiamo tap target solo sul testo
> quando invece dovrebbe esser tutta l'area, parlo ad es dei nomi visitor

> poi abbiamo quest'idiozia di mostrare colonne già mostrate

Three defects on a phone, one of which is the cause of another. Measured on `origin/main`
= `b0d5342d` = the staging build he was looking at. Not a duplicate of #1157: that one is
about the LIST panning (rows → cards); this is about what the row card and its detail panel
do once the pan is gone.

## 1. The detail panel repeats fields the card already shows — because nothing is dropped

`cicchetto/src/themes/default.css:5289`

```css
.adm-col-detail { display: none; }

@media (min-width: 900px) {
  .adm-col-detail { display: table-cell; }
}
```

The whole mobile design rests on that `display: none`: secondary columns are meant to leave
the table on a phone and come back inside the panel. `admin/AdminFacts.tsx` says so in its
header comment — *"Nothing is lost, only deferred: every dropped column reappears here"*.

They never leave. Same file, `:5684`:

```css
@media (max-width: 899px) {
  .adm-table, .adm-table tbody, .adm-table tr, .adm-table td { display: block; ... }  /* :5688 */
  .adm-table td { display: flex; ... }                                                 /* :5711 */
}
```

`.adm-table td` is `(0,1,1)`; `.adm-col-detail` is `(0,1,0)`. **The stacking rule wins**, so
below 900px every `td.adm-col-detail` renders as a labelled line of the card. `thead` is
hidden outright, so the `th.adm-col-detail` partner does disappear — which is why this reads
as "designed" rather than broken.

Consequence, exactly as reported: on Users the card shows `LIVE 2` and `INSERTED 2026-07-27
12:22`, and the panel underneath — subtitled *"the columns the table drops on a phone"* —
shows `LIVE SESSIONS 2` and `INSERTED 2026-07-27 12:22`. The same two values, twice, one
under the other, with a subtitle asserting the opposite. On Sessions the panel likewise
repeats `NETWORK`, which the identity cell already prints under the nick.

Two ways out, and it is a product call, not a refactor:

[...]

Issue #1158, opened 2026-08-09
cic(admin): creating a user must be one flow in the Users tab — no separate Credentials pane


vjt:
vjt on `#grappa`, 2026-08-09 23:07, setting the 1.0 bar for the admin console:

> per 1.0 però voglio assicurarmi che creare utenti sia flawless e adesso invece mi pare che
> sia ancora necessario fare i giri di crud tra user e credential. user e credential panes non
> dovrebbero esistere invho, ma solo il tab users.

Most of the admin console's remaining work is explicitly **post-1.0** by the same ruling. This
one is not: user creation is the flow that has to be flawless before 1.0.

## Measured on `origin/main` — creating a usable account is two tabs and two writes

- **Users tab** (`cicchetto/src/AdminUsersTab.tsx`): `POST /admin/users` with `name`,
  `password`, `is_admin`. That is the grappa account and nothing else — it cannot connect
  anywhere.
- **Credentials tab** (`cicchetto/src/AdminCredentialsTab.tsx`): `POST /admin/credentials`, a
  bind form over *user + network + nick + auth_method* plus optional password / sasl_user /
  realname / autojoin. The tab's own header comment lists `GET /admin/users` and
  `GET /admin/networks` as **sidecar fetches at mount** — it exists only to join data that
  lives elsewhere.
- Server side both are real, separate resources: `router.ex:174-178`
  (`GET|POST /credentials`, `PATCH|DELETE /credentials/:user_id/:network_id`) against
  `lib/grappa/networks/credential.ex`, which carries `nick`, `ident`, `realname`,
  `sasl_user`, `auth_method`, `autojoin_channels` and the encrypted secrets.

So the operator's mental model is right: today you create a row in one tab, then go find it
again in another tab to make it mean something.

## What 1.0 asks for

**One flow, in the Users tab.** Creating a user offers its network binding in the same form;
the separate Credentials tab goes away as an operator-facing surface.

## Decisions this needs before code

1. **A user can have several credentials** (one per network) — that is the whole point of the
   composite key `:user_id/:network_id`. So the Users tab needs a shape for "this user, on
   these networks", not a single embedded credential. Row detail is the obvious home (#1074
   already opens it in place), but the layout is a product call, and it lands on top of the
   mobile rework in #1157.
2. **The destructive semantics must survive the move.** `PATCH /admin/credentials/...` kills
   the live session when password or auth_method change and answers with
[...]

Issue #1073, opened 2026-08-08
cic(admin): reuse the channel top bar for admin — live key stats, trailing ☰ with the expanded menu, no × and no refresh


vjt:
## What

The admin console's header band gets restructured. Brief from vjt on `#it-opers`
(2026-08-08), in scope for **1.0** — *"si si pre 1.0 voglio un admin al bacio per la release"*.

**The band should be the SAME bar the channel windows use, with different content in it** —
vjt: *"la top bar admin dev'esser possibilmente la stessa barra che abbiamo per i canali, solo
con dentro roba diversa"*. Not a lookalike rebuilt on the `--adm-*` layer: the same component.

Contents, one row across the top of the admin pane:

- **left** — live key stats: number of sessions, number of visitors, hostname, loadavg, and
  the **server version**
- **right** — the ☰, opening the rail with the menu **already expanded, as on home**
- **gone** — the `×`. vjt: *"la x sparisce"*
- **moved** — the refresh button leaves the header for the rail actions. vjt: *"il refresh
  può tranquillamente stare tra le actions nel rail, non serve così prominente"*

**It has to stay short.** vjt: *"si, non invadente"*; Hypnotize: *"deve essere piccola, che già
la tastiera occupa una madonna"*. Vertical space on a phone with the keyboard up is the
binding constraint — five stats do not get five rows.

## Reusing the channel bar buys the ☰ side for free (measured, `origin/main` 08f0fcb0)

The admin ☰ is **not** a second hamburger. `AdminPane.tsx:139` mounts the very same
`RailOpenerButton` (`ShellChrome.tsx:92`) that non-channel windows use. The genuinely
different one is `.topic-bar-hamburger` (`TopicBar.tsx:368-370`, `default.css:1960`) — the
channel bar's own, and it is that bar's **last flex child**, i.e. already on the right.

The side differs today because the two hosts are different boxes:

- `.shell-chrome` (`default.css:2510`) is `height: 0` + `justify-content: flex-end`, z-index 41
  — it **floats** the glyph into the top-right corner over the pane.
- `.admin-pane-header` (`default.css:4260`) is a real band, `display: flex` +
  `justify-content: space-between`, and the opener is its **first child** — so it lands left.

Moving admin onto the `.topic-bar` shell therefore resolves the placement by construction
rather than by a new CSS override. What has to be preserved from the current arrangement is
the reason it exists at all, documented at `AdminPane.tsx:61`: hosting the opener inline is
[...]

PR #1233, opened 2026-08-11
cic(#1223 p1, #1224): make the phone column drop real, and move ended sessions to their own sub-page

vjt:
Four commits: #1224 had to be built with #1223 item 1 already applied rather than inheriting the duplication (vjt, on the issue), and the last two are audit results rather than features.

## 1. `cic(#1223)` — the phone column drop, and BOTH duplications

vjt's ruling (2026-08-11): *"1223 punto 1: direi drop no?"* — the secondary columns really leave the card on a phone; the detail panel stays. Dropping the panel was the rejected arm.

**One symptom, two defects.**

- **CSS specificity** (Users, the user page). `.adm-col-detail { display: none }` is (0,1,0) and the stacking block that turns rows into cards re-declares `.adm-table td` (0,1,1), so every "dropped" column came back as a labelled line of the card while the panel printed it again under a subtitle promising the opposite. `thead` is hidden outright, which is why it read as designed. Fixed with a `td`-qualified twin (0,2,1) **inside** that block — a rule outside a media query cannot outrank one inside it.
- **JSX duplication** (Sessions). Sessions drops no column at any width; `detailFacts` carried a `network` fact while `.admin-session-network` printed the slug under the nick, **on desktop too**. Raising the selector would not have touched it. The fact goes.

**The 769–899 band.** The console's regime changes at 900px (nav rail, form grid, table stacking, `.adm-col-detail`); `isMobile()` is the shell's 768px, and three admin gates were reading it. Fixing only the CSS would have made that band strictly worse — columns gone AND no disclosure to reach the panel they went into. So the console gets its own `isAdminNarrow()` (899px), read by `AdminRowName` and the Networks column split. `refreshSlot` keeps `isMobile()` deliberately: where the refresh button lives is a question about the shell's rail, not the table.

**Where each half is tested.** The drop and the band are real-browser claims (jsdom cannot see paint — #1073), asserted with the report's own oracle: no value on the card AND in the panel, both sides asserted non-empty first so an empty intersection cannot pass vacuously. The Sessions duplication is plain JSX and is pinned in vitest, with a mutation check (re-adding the fact fails exactly that one assertion).

## 2. `cic(#1224)` — ended sessions on their own sub-page

Three of the four dictated columns were an em-dash on every log-only row, structurally: `last seen`, `channels` and `actions` are live-process facts, and those rows have neither a process nor a DB row.

vjt's three rulings, applied: a **sub-page of Sessions** (not a sixth tab — the strip already overflows on a phone); **no `deleted` badge** (the list that stays is strictly live); the **ring caveat is not a blocker but travels with the page** — a page named after a population implies it holds all of it in a way an inline badge did not, so it is the sub-page card's subtitle.

- The merge is untouched: one row set, one composite key, the log still enriching a live row's drill-down. `liveSessions`/`endedSessions` partition it — a partition, not two filters, so no row class can fall off both screens. `EndedSessionRow` narrows `last_event` to non-null, so the page needs no guard for a case its constructor makes impossible.
- The Sessions card now renders whenever the data loaded, with its empty state inside it: "no live sessions, some ended ones" is a real state after the split, and gating the card on the row count would have taken the door away exactly when it is wanted.
- Page columns: who / last event / at / lasted / how. `lasted` and `how` render an em-dash when the log's last word was not a disconnect, rather than inventing an ending.

## 3. `e2e(#1224)` — the assert the move dropped

Removing a product fact edits the specs that asserted it, and this branch edited two it does not own. That is exactly where a test weakens without anyone noticing, so here is the line-by-line accounting of what changed in each, and whether the assertion **moved** or was **removed**.

### `issue1158-item4-deleted-session-row.spec.ts` — three assertions touched

| Was | Now | Verdict |
| --- | --- | --- |
| `expect(admin-session-gone-${key}).toBeVisible()` — the `deleted` badge marks the row so the operator can tell "gone" from "down" | `expect(admin-session-gone-${key}).toHaveCount(0)` in the live list, **plus** the row asserted present on the ended-sessions page | **Moved, and inverted into a witness.** The distinction is now carried by the surface the row is on, which is a stronger separation than a badge on a shared list. The inverted assert additionally proves #1224 really removed the badge instead of leaving it somewhere unseen — a claim the old spec could not make. |
| `expect(detail).toContainText("last event")` — the drill-in panel carries a last-event fact at all | `expect(admin-ended-session-event-${key}).not.toBeEmpty()` | **Moved and strengthened.** `toContainText` on a panel matched the label; `not.toBeEmpty()` on the cell asserts the value. Still not `logged.event` verbatim: terminating the session can append a newer lifecycle row between the barrier and the render. |
| `expect(detail).toContainText("session log only")` — provenance: this row exists only because the log outlived the deleted subject | *(nothing, in the first pass)* | **Removed.** The fact itself is gone from the product by design — the sub-page states it as a premise rather than repeating it per row. But the premise was asserted nowhere on a populated page: the empty-state wording is pinned in the unit twin, and an empty screen cannot witness a listed row's provenance. |

Commit 3 restores the third one: `expect(admin-ended-sessions-card).toContainText("From the lifecycle log")`, on the populated page. It is the same guarantee in a new place — the surface naming the log as the record — and it now sits where the row it describes is visible.

### `m8-admin-visitors-delete.spec.ts` — no assertion touched

+5/−3, all of it comment. The spec deliberately does not count rows (the identity-wide property lives on the verbs), and #1224 gave that choice a second reason: the count now depends on a screen this spec never opens. Nothing was weakened here because nothing executable changed.

## 4. `e2e(#1224)` — the pan guard enumerates surfaces, not tabs

Auditing the whole test tree for what #1224 moved (636 files: 389 e2e specs, 247 unit) turned up no other red spec — but it did turn up a guard that had gone stale in the same way the assert above did.

`ux-6-g-admin-mobile-h-scroll` states that "the pan is a property of the pane, not of the three tabs that happened to have tables when this spec was written", and then loops the tab list. A tab is not the only thing the pane shows: `AdminUsersTab` has opened `AdminUserPage` from inside the Users panel since #1158 (which is where the Credentials tab's job went), and #1224 adds the ended-sessions page behind the Sessions card header. Both are full-pane surfaces with tables at 393px that a loop over `ADMIN_TABS` structurally cannot reach. **The user page has been outside this guard since #1158** — #1224 would have been the second surface to be born uncovered.

- The width oracle now walks a **surface** list: every tab, plus each sub-page with the steps that open it.
- The **tab half maintains itself** — a count assertion against the pane's rendered `role="tab"` buttons fails when a tab is added without a line in the list.
- The **sub-page half cannot**, and the comment says so instead of implying coverage it lacks: a door has no marker. `admin-sessions-ended-open` is a static testid on a card header; `admin-user-networks-<id>` is a per-row button sharing its actions cell with Delete. Nothing can be crawled safely or matched honestly. One marker attribute per door would make the list derive itself; that is named as the fix, not hidden.
- Both sub-pages get an anti-hollow-green barrier on their **table**, not just their root — an empty-state card cannot overflow. It earned itself immediately: the first run used `.first()` for the user row, which lands on `admin-vjt`, a user with no IRC bind and so no networks table. The barrier failed instead of measuring an empty page.

## Gates

- `bun run check` (biome + tsc + e2e tsc): **green**, exit 0.
- `bun run test`: **green**, 280 files / 5243 tests.
- **Full `scripts/integration.sh`: green — 710 passed, 0 failed, 26.9m, exit 0.** Not a scoped `--grep`: the whole suite, on the STACK lane, against the committed tree.
- Both guards proven by mutation, each killing exactly its own assertion and leaving the siblings green:
  - *Does the new surface get measured, or just visited?* Mutant: give the ended-sessions sub-page a pannable scroller (`overflow-x: auto` + a 900px table). The oracle failed and **named it** — `{cls: "adm-table-wrap", scrollW: 900, clientW: 331, surface: "subpage:ended-sessions"}`. The other three tests in the file stayed green.
  - *Does the #1223 oracle fail without the fix?* Mutant: delete the `td`-qualified twin (`.adm-table td.adm-col-detail`). Exactly the two drop assertions went red, with the reported defect printed literally — `values printed twice — card [...,"0","2026-08-11 19:42",...], panel ["0","2026-08-11 19:42"]` — plus the 769–899 band test. The three panel-layout tests, which that rule does not govern, stayed green.

Docs: two `DESIGN_NOTES.md` entries, one per feature commit.

### DIFF
--- a/cicchetto/e2e/fixtures/cicchettoPage.ts
+++ b/cicchetto/e2e/fixtures/cicchettoPage.ts
@@ -1025,6 +1025,27 @@ export async function openAdminSessionsTab(page: Page): Promise<Locator> {
   return table;
 }
 
+// #1224 — the ended-sessions sub-page, reached the only way an operator can:
+// through the Sessions tab's card header. Deliberately NOT routed through
+// `openAdminSessionsTab`: that one waits on `admin-sessions-table`, which does
+// not render when the live list is empty, and an empty live list is exactly
+// when a spec is most likely to be looking for a session that is over. The
+// door itself is the barrier — it renders as soon as the tab's data loads.
+// Idempotent about how it got there: a spec that already opened the Sessions
+// tab must not be sent back through the rail launcher, which is a TOGGLE and
+// would close the pane it is standing in.
+export async function openAdminEndedSessions(page: Page): Promise<Locator> {
+  const door = page.getByTestId("admin-sessions-ended-open");
+  if ((await door.count()) === 0) {
+    await openAdminConsole(page);
+    await page.getByTestId("admin-tab-sessions").click();
+  }
+  await door.click({ timeout: 15_000 });
+  const subpage = page.getByTestId("admin-ended-sessions-page");
+  await expect(subpage).toBeVisible({ timeout: 10_000 });
+  return subpage;
+}
+
 // A row's testid suffix is the composite `<kind>:<subject_id>:<network_id>` —
 // the same string the `/admin/sessions/:id/*` verbs parse. A spec knows the
 // subject id (a minted visitor id, a seeded user id) but not the network's
--- a/cicchetto/e2e/tests/issue1158-item4-deleted-session-row.spec.ts
+++ b/cicchetto/e2e/tests/issue1158-item4-deleted-session-row.spec.ts
@@ -107,11 +107,19 @@ test("#1158 item 4 a deleted visitor's session survives on the ended-sessions pa
     // claim is that the page names the event it is showing.
     await expect(page.getByTestId(`admin-ended-session-event-${key}`)).not.toBeEmpty();
 
+    // PROVENANCE. Before #1224 the drill-in carried `record: session log
+    // only — the subject was deleted, the event outlived it`, and the
+    // assert on it is the one claim the move did not relocate by itself:
+    // the sub-page states it as its premise instead of per-row, but a
+    // premise nobody asserts is not covered. So it is asserted here, on a
+    // POPULATED page — the empty-state wording is pinned in the unit
+    // twin, and an empty screen cannot witness this.
+    const card = page.getByTestId("admin-ended-sessions-card");
+    await expect(card).toContainText("From the lifecycle log");
+
     // The caveat survived the move — a page named after a population
     // implies it holds all of it, and a bounded ring does not.
-    await expect(page.getByTestId("admin-ended-sessions-card")).toContainText(
-      "not evidence the session never ran",
-    );
+    await expect(card).toContainText("not evidence the session never ran");
 
     // No credential to park, no pid to stop: every verb would resolve to
     // a subject that is gone, so none is offered on either screen.
--- a/cicchetto/e2e/tests/issue1158-item4-deleted-session-row.spec.ts
+++ b/cicchetto/e2e/tests/issue1158-item4-deleted-session-row.spec.ts
@@ -10,11 +10,16 @@
 // `<kind>:<id>:<network>` composite the admin session rows are, which is
 // what makes the unified Sessions view able to list them.
 //
-// What this spec pins is the SURFACE contract, in the one direction the
-// product actually promises:
+// #1224 moved the SURFACE, not the property. Those rows used to sit inline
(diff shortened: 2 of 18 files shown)


vjt:

Issue #1029, opened 2026-08-07
ops: nginx access log has no $request_time or $request_id — request-level diagnosis is blind


vjt:
## What is missing

The m42 host vhost (`/usr/local/etc/nginx/sites/irc.openssl.it`, `server_name irc.sindro.me irc.sniffo.org`) writes its access log with `log_format main`:

```
$remote_addr - $remote_user [$time_local] "$request" $status $body_bytes_sent
"$http_referer" "$http_user_agent" "$http_x_forwarded_for"
```

Status, path and user-agent — **no duration, no correlation id, second-resolution timestamps**.

The in-repo substrates are worse: `infra/freebsd/nginx.conf`, `infra/linux/nginx.conf` and `cicchetto/e2e/nginx-test.conf` define **no `log_format` and no `access_log` at all**, so they inherit nginx's `combined` default. Same blindness, three more places.

## What it blocks — a measured cost, not a hypothetical one

The #394 re-measurement counted 20 client aborts over 45,135 requests at the edge and **could not attribute a single one of them**, because every available discriminator needs the same missing field:

| question | field it needs |
|---|---|
| is an abort our own 8s `bootFetch` ceiling, or a client teardown? | `$request_time` — an 8.0s cluster vs sub-second scatter |
| is an abort the 60s edge ceiling wearing a 499? (a 499 preempts a 504) | `$request_time` |
| is `GET /me` genuinely an outlier, or simply the biggest target? | `$request_time` — aborts per second-in-flight, not per request |
| did the server actually take long, or was it fast and the client left? | `$request_id`, to join nginx to the BEAM log |
| is one suspend event being counted as N aborts? | `$msec` — ms resolution |
| is a 502 a connection-level race or an application failure? | `$request_time` plus the BEAM-log join (see #1030) |

The BEAM already carries `request_id` in its Logger metadata (`config/config.exs:295`) and `Plug.RequestId` sets the `x-request-id` response header. But the BEAM log has no user-agent and nginx has no request id, so **the two logs cannot be joined today**. That join is one field.

## Retention is the other half of the same blindness

Retained logs reach back only to 2026-08-06 20:00 (current + 5 rotated, roughly 1.5 days). The 2026-07-25 sample #394 was filed on is gone, and could not be re-checked at all — its provenance is now permanently unknown. Whatever fields we add, we will be equally blind three weeks from now unless retention covers the gap between a report being filed and somebody looking at it.

## Proposal

[...]

Issue #394, opened 2026-07-25
cicchetto: iOS clients abort requests 277x more than macOS — GET /me worst hit


vjt:
## Symptom

iOS clients abort requests at a rate that macOS does not: over the last 20k requests, **277 client aborts from iOS against 1 from macOS**. The most-aborted path is **`GET /me` (134 aborts)**.

Surfaced while investigating #393 (DB/perf). **This is a separate client-side problem** — it is not the A2 fold-predicate issue and does not go away with it. The #393 index work has already cut the server-side cost of the counts that `GET /me` builds, so an abort rate this asymmetric points at the client (or its request lifecycle on iOS), not at the query.

## Why it matters

`GET /me` is the cold-load path — it builds the unread counts for every window. An abort there means the client tore down a request the server was still serving: wasted DB work, and from the user's side a home screen that may render with missing or stale counts.

## Worth checking

- iOS Safari / PWA suspend-resume behaviour tearing down in-flight fetches (backgrounding the tab, lock screen).
- Any client-side timeout or `AbortController` in the `/me` path firing earlier than the server can answer under load.
- Whether the aborts cluster in time with the slow-request windows recorded in #393, or are spread evenly (the two tell different stories).

## Evidence

- last 20k requests: 277 client aborts on iOS vs 1 on macOS
- most-aborted path: `GET /me`, 134 aborts

vjt:
Adding the client-side retry angle, which is the part not yet covered here, plus two measurements that constrain the diagnosis.

## Retry amplification (the missing half)

The abort count alone reads as "iOS gives up". The request volume says it also *comes back*: over the same 20k-request sample from the jail's nginx `access.log`, split by platform:

| platform | requests | client aborts (499) | 500s |
|---|---|---|---|
| iOS | 10,924 | 277 | 158 |
| macOS | 3,484 | 1 | 29 |
| Android | 2,595 | 14 | — |

iOS issues roughly **3x the requests of macOS** from a user population that is not 3x larger. That is consistent with abort-then-retry rather than abort-and-stop, and it makes the loop self-feeding: `GET /me` is slow for a subject with many windows → the client aborts on its own timeout → it retries → the extra concurrent `/me` work makes the next one slower. Desktop clients simply wait and get their answer, which is why the same server-side latency is invisible there. Android PWA on Firefox was reported working fine, matching its low abort count.

[...]

PR #1237, opened 2026-08-11
ops(#1029): give the in-repo nginx substrates the production log format

vjt:
The in-repo half of #1029. The host half (m42's `/usr/local/etc/nginx/sites/irc.openssl.it`) is
out of this repo and **already live** — vjt applied and reloaded it on 2026-08-08.

## What was blind

`infra/linux/nginx.conf` and `cicchetto/e2e/nginx-test.conf` defined **no `log_format` and no
`access_log` at all**, so they inherited nginx's `combined` default: status, path, user-agent, and
nothing else. The #394 re-measurement counted 20 client aborts over 45,135 requests and could not
attribute a single one, because every discriminator it needed is a field `combined` does not carry —
was this our own 8s `bootFetch` ceiling or a client teardown, did the server take long or did the
client leave, is one suspend event being counted as N.

## What this does

One new `infra/snippets/log-format.conf`, included by both substrates:

```nginx
log_format main '$remote_addr - $remote_user [$time_local] "$request" '
                '$status $body_bytes_sent "$http_referer" '
                '"$http_user_agent" "$http_x_forwarded_for" '
                'rt=$request_time urt=$upstream_response_time us=$upstream_status '
                'rid=$sent_http_x_request_id msec=$msec';

access_log /var/log/nginx/access.log main;
```

- **Defined once, not pasted twice.** Two nearly-identical copies stop being comparable the first
  time somebody edits one, and comparability is the entire point. It is a *second* snippet rather
  than a block inside `locations-api.conf` because `log_format` is http-context only and that one is
  included inside `server { }`.
- **Byte-identical to the live host vhost, with the tail APPENDED.** Those logs are shipped by
  syslog-ng into a collector that parses `combined` *positionally*; a field inserted anywhere before
  the user-agent silently mis-assigns every one after it.
- `rid=$sent_http_x_request_id` is the join key to the BEAM log — Phoenix already sets the header
  (`Plug.RequestId`) and already logs `request_id`; nginx just never wrote it down.
- `install_nginx.sh` now installs the new snippet. The e2e stack already bind-mounts the whole
  `infra/snippets` dir, so it needs no wiring.

### Correction to the issue text

`infra/freebsd/nginx.conf` does not exist — the jail has no nginx (its host vhost is the one out of
repo). The in-repo substrates with an `http { }` block are **two**, not three.

## Verification

| gate | result |
|---|---|
| `scripts/bats.sh` (full set) | exit 0 — 445 ok, 0 not ok |
| new `test/infra/nginx_log_format_test.bats` | 7 cases, all green |
| mutation run, 7 mutants | 7 killed, **one assertion each** |
| real `nginx -t`, `infra/linux/nginx.conf` | `test is successful` |
| real `nginx -t`, `cicchetto/e2e/nginx-test.conf` | `test is successful` |
| deliberately corrupted snippet | `[emerg] … test failed` — the check can fail |

Run against the digest-pinned `nginx:alpine`, with the installer's two placeholders rendered exactly
as `install_nginx.sh` renders them. A real nginx was then started and probed, to confirm the line
actually comes out — and comes out on stdout, as the snippet's comment claims:

```
192.168.65.1 - - [11/Aug/2026:19:53:43 +0000] "GET /healthz HTTP/1.1" 502 150 "-"
"grappa-1029-probe/1.0" "-" rt=0.000 urt=0.000 us=502 rid=- msec=1786478023.235
```

`rid=-` because no BEAM was behind it (hence the 502); vjt's live m42 line shows it populated.

The guard is derived from the tree rather than hardcoded — "every conf with an `http` block includes
the one snippet", "the installer installs every snippet the conf includes" — so a new substrate, or
a new snippet, is in scope the day it appears.

## Still manual, and still open

- **The m42 host vhost half is NOT in this PR** and stays a manual host-side apply. It is already
  done and live; nothing here touches it, and nothing here connects to m42.
- **Retention is untouched** — the other half of the blindness the issue raises, and a host-side
  decision. Worth noting: this change adds **71 bytes per request** (measured on the live m42 line),
  so a *size*-based rotation now retains proportionally less time than it did yesterday.
- `infra/nginx-tls-frontend.example.conf` is deliberately left alone: it is a server-context fragment
  dropped into an operator's own nginx, so it cannot define a format or reference one this repo owns.

## Deploy

**nginx config only — applies with a `reload`. No BEAM restart, no migration, no code.** Nothing here
is on the hot/cold axis: it does not touch `config/`, `mix.exs`, `VERSION`, or any `.ex`.

### DIFF
--- a/cicchetto/e2e/nginx-test.conf
+++ b/cicchetto/e2e/nginx-test.conf
@@ -24,6 +24,12 @@ http {
     keepalive_timeout 65;
     server_tokens   off;
 
+    # Access-log shape — the same snippet the native-host substrate includes
+    # (#1029), so a diagnosis rehearsed here reads like the production one.
+    # http context: `log_format` is not valid inside `server { }`. Resolved
+    # from the ../../infra/snippets bind mount, like locations-api.conf.
+    include /etc/nginx/snippets/log-format.conf;
+
     upstream grappa_upstream {
         server grappa:4000;
         keepalive 32;
--- a/infra/linux/install_nginx.sh
+++ b/infra/linux/install_nginx.sh
@@ -2,7 +2,8 @@
 # Install the dumb reverse-proxy nginx config + shared snippet on a native
 # Linux host, enable/reload the service. The BEAM self-serves the SPA and
 # owns every header (#485), so there is no cicchetto-dist symlink and no
-# security-headers snippet — only the substrate-agnostic proxy snippet.
+# security-headers snippet — only the two substrate-agnostic snippets: the
+# proxy locations and the shared access-log format (#1029).
 #
 # Idempotent — re-run after `git pull` (or to change
 # LISTEN_ADDR/TRUSTED_UPSTREAM_CIDR) to refresh the config.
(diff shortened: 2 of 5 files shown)


vjt:

Issue #1235, opened 2026-08-11
cic(reply): cap the quoted body at 42 chars with an ellipsis, and put a space before the trailing <<


vjt:
Change request from vjt on `#grappa`, verbatim intent:

> sul reply limitiamo a 42 i caratteri di cui facciamo reply, se sforano mettiamo un ellipsis `...`, e poi mettiamo sempre uno spazio prima del `<<` finale

Two changes to the reply-quote wrapper. Both are in `cicchetto/src/lib/replyQuote.ts`; measured against `origin/main` `5bb1d168`.

## 1. Cap the quoted body at 42 characters, ellipsis when it overflows

Today there is no cap at all: `quotableBody` returns the whole body and `replyQuote` wraps it as-is, so replying to a long line drags the entire line into the compose box.

**The cap belongs in the WRAPPER, not in `quotableBody`.** That helper is shared with `addQuoteCommand` (#1107) — truncating there would silently clip `!addquote`, which archives the line and must stay whole. `replyQuote.ts:21` is the seam that owns the reply shape.

Open decisions, all cheap but they change the output:
- Does the 42 count the ellipsis or sit before it? (`42 chars + "..."` vs `39 + "..."`)
- Literal `...` (three dots, as written in the request) or `…` U+2026? The request says `...`.
- Cut at 42 flat, or back off to the last word boundary so the quote does not end mid-word?
- Does the cap apply to the body only, or to the whole `<nick> body` head-included? Read as body-only — the nick is not what overflows.

## 2. Always a space before the trailing `<<`

`REPLY_QUOTE_TAIL` is `"<< "` (`replyQuote.ts:14`) and the return is `` `${head} ${body}${REPLY_QUOTE_TAIL}` `` — so today the output is `<vjt> ciao mondo<< `, the tail flush against the last word. Requested shape: `<vjt> ciao mondo << `.

**Measured, and it is the good news:** the de-nesting regex that strips a previous hop is `^(?:<nick>|\* nick) [\s\S]*<<(?: |$)` (`quotableBody.ts:35`). The greedy `[\s\S]*` absorbs a space just as happily as a word character, so **the nesting cut keeps working with the new tail, and it also keeps working on lines persisted with the old one**. No migration, no compatibility branch. Verify it, do not take my word.

## Blast radius

[...]

Issue #1123, opened 2026-08-09
cicchetto: replying to a reply nests the whole quote chain instead of truncating at the previous <<


vjt:
## Replying to a reply nests the quote instead of truncating it

`replyQuote` (`cicchetto/src/lib/replyQuote.ts:23-31`) builds the quote from the whole message body:

```ts
const body = mircPlainText(msg.body ?? "").trim();
if (body === "") return null;
return `<${msg.sender}> ${body}${REPLY_QUOTE_TAIL}`;   // REPLY_QUOTE_TAIL = "<< "
```

So when the message being replied to is *itself* a reply, its body already contains a quote plus the `<< ` tail, and the new quote wraps the old one:

| step | compose content |
|---|---|
| A replies to B | `<B> original<< answer` |
| C replies to A's line | `<A> <B> original<< answer<< ` |
| D replies to C's line | `<C> <A> <B> original<< answer<< reply<< ` |

Each hop carries the entire history forward, so the line the reader is actually answering ends up buried mid-string, and on a narrow client the visible part of the message is all quote and no content.

## Expected

Quote only the words of the person being replied to: everything up to and including the **last** `<< ` in their body is a previous quote and should be dropped before building the new one. Step C above should yield `<A> answer<< `.

## Implementation note

Do not strip on a bare `<<` substring search: `<<` occurs in ordinary text (`shift << 2`, C++ stream operators, `cat <<EOF`), and eating a real message on that would be worse than the nesting.

Strip only when the body has the shape the client itself produces — a leading `<nick> ` followed by text and the `<< ` tail — i.e. anchor the match at the start of the body with a nick-shaped first token, and cut at the last such tail. The nick shape is worth deriving from what the wire actually allows rather than inventing a new charset here.

This is quote-building only; the sent message and the wire format are unaffected.

vjt:
Shipped in **v0.16.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.16.0

Issue #1107, opened 2026-08-09
Message menu: !addquote item that pre-fills the compose box


vjt:
Add an **!addquote** item to the message long-press menu. Picking it puts `!addquote ` followed by the message text into the compose box, leaving the operator to review and hit send — cic does not send anything by itself, and does not need to know what `!addquote` means. The command is interpreted by whatever bot is in the channel; this is purely a compose-filling convenience.

Requested on IRC:

> addquote dev'esser semplicemente una voce di menu !addquote che appunto mette nel composebox "!addquote" e poi il messaggio e ovviamente sempre caret alla fine visibile

## Where it goes

`MessageContextMenu.tsx` already builds the item list — Copy, Reply, Select… — from `messageMenu()`'s target. This is a fourth entry alongside them, and it reuses the machinery Reply already has:

- `lib/replyQuote.ts` `replyQuote(msg)` is the precedent for turning a `ScrollbackMessage` into compose text: it gates on `isContentKind` (a JOIN/PART is not something to quote), rejects an empty sender or an empty body, and runs the body through `mircPlainText` so mIRC control bytes do not round-trip into a message the operator never formatted. An `!addquote` payload wants the same treatment, minus the `<nick> …<< ` wrapper.
- `lib/composeAppend.ts` `appendToCompose(networkSlug, channelName, text)` is the insertion verb.
- Same `enabled:` posture as Reply — visible but disabled on a row with nothing to quote, so the menu's shape does not jump between rows.

## Open question

Whether the payload is the bare body or carries the nick (`!addquote <vjt> ...`). Bots differ; state the choice in the implementation.

## Depends on

The caret must land at the end **and be visible**, which is the defect tracked separately: the compose textarea is `rows=1` and `appendToCompose` never scrolls it, so any insertion that wraps past the first line hides the caret. `!addquote ` plus a message body will overflow essentially every time, so this item is unusable-feeling until that one lands. Fix that first.

vjt:
Shipped in **v1.0.0** — closing.

10 dedicated commits landed for this issue between `v0.16.0` and `v1.0.0`:

- `1e81e0b2` test(#1107): guard the !addquote menu item before building it
- `a2a93165` cic(#1107): lift the quotable-body rules out of the reply wrapper
- `284b2378` cic(#1107): add the !addquote item to the message menu
- `97e35521` test(#1107): kill the two guards that were passing vacuously
[...]

Issue #1105, opened 2026-08-09
Reply quote leaves the caret below the compose fold on drafts that wrap


vjt:
Replying to a message that does not fit on one line leaves the caret below the visible area of the compose box: the quote is inserted, the caret is placed at the end, but the textarea is never scrolled, so you type blind.

Reported on IRC after the #1067 reply gesture shipped:

> bug, con reply a messaggi lunghi non facciamo lo scroll del compose box, quindi il cursore risulta invisibile

## Reproduced

Measured in Chromium 151 on a 393px-wide viewport, replaying the exact `appendToCompose` sequence (set value → `focus({preventScroll:true})` → `queueMicrotask(setSelectionRange(end,end))`) against a `rows=1` textarea carrying the `.compose-box textarea` rules from `cicchetto/src/themes/default.css` (`flex:1`, `padding:0.5rem`, `line-height:1.4`, `resize:none`):

| quote | length | scrollHeight | clientHeight | scrollTop after append | caret in view |
|---|---|---|---|---|---|
| `<vjt> asd<< ` | 12 | 36 | 36 | 0 | yes |
| `<vjt> ok adesso si puo fare swipe per reply<< ` | 46 | 55 | 36 | 0 | **no** |
| a 110-char quote | 110 | 75 | 36 | 0 | **no** |

So the threshold is not "very long": **any quote that wraps past the first line** already hides the caret. Setting `scrollTop = scrollHeight` in the same place brings it back into view in every case (19 and 39 respectively for the two failing rows).

Note this repro measures the mechanism in a faithful standalone textarea, not the running app; the code path below is the app's.

## Cause

`cicchetto/src/lib/composeAppend.ts` — `appendToCompose` (L17) writes the draft, focuses, and places the caret, but never touches `scrollTop`:

```ts
ta.focus({ preventScroll: true });
queueMicrotask(() => {
  ta.setSelectionRange(next.length, next.length);
});
```

`setSelectionRange` does not scroll the element. The compose textarea is `rows={1}` (`ComposeBox.tsx` L585) with `resize: none`, so an overflowing draft is an internal scroll container left pinned at `scrollTop === 0`, with the end-caret below the fold.

This is the same defect #173 already fixed for history recall, and the fix is one line already present in the file — `scrollRecallCaretIntoView` in `ComposeBox.tsx` (L191-198):

```ts
queueMicrotask(() => {
  const end = el.value.length;
  el.setSelectionRange(end, end);
[...]

PR #1238, opened 2026-08-11
cic(#1235): cap the reply quote at 42 chars and space the trailing <<

vjt:
vjt, verbatim: *"sul reply limitiamo a 42 i caratteri di cui facciamo reply, se
sforano mettiamo un ellipsis `...`, e poi mettiamo sempre uno spazio prima del
`<<` finale"*.

Two changes, both in the reply-quote WRAPPER (`cicchetto/src/lib/replyQuote.ts`).

## Where the cap lives, and the control group

The cap is in `replyQuote`, never in `quotableBody`. That helper is shared with
`addQuoteCommand` (#1107), which archives the line and must keep it whole, so a
cap there would have silently clipped `!addquote`.

**`addQuote.test.ts` is the control group and it does not move — zero lines
changed in that file.** If one of its seven `<<`-shaped expectations had shifted,
the cap would have landed in the wrong module.

## The four open readings, and who decided them

The issue left four cheap-but-visible choices open. They were **decided by the
orchestrator and since ratified by vjt**, not by the implementation, and they are
parked in one constant plus one function so a change of mind stays a one-line
edit:

- **42 counts BODY characters and the ellipsis is ADDED past them** — the quoted
  run is 45, not 42 with three spent on dots.
- **The marker is the literal `...`**, three ASCII dots as the request spells it,
  not `…` U+2026.
- **The cut is flat at 42**, with no backing off to a word boundary.
- **The cap applies to the body only** — the `<nick> ` head is outside the count.

One further call was made at the keyboard, and it is an implementation
correctness matter rather than a product one: **the cut counts CODE POINTS, not
UTF-16 units.** A `slice(0, 42)` can land between the halves of a surrogate pair
and put a lone surrogate in the compose box, and from there onto the wire.

## The de-nesting claim, tested rather than cited

The issue asserts that #1123's de-nesting regex (`[\s\S]*<<(?: |$)`) keeps
working with the new tail and on lines persisted with the old one. That is now
four assertions, not a citation: old flush tail, new spaced tail, a round trip
through the wrapper's own output, and a round trip through a CAPPED quote (where
the tail follows an ellipsis rather than a word).

## Mutation results — which test dies, and to what

Five mutants, each applied to the real file, each verified to have landed before
the run (`git diff` non-empty), each reverted after. Baseline is 40 passed.

| # | Mutant | Killed by |
|---|--------|-----------|
| 1 | cap removed (`${body}` back in the template) | **5** tests, all in the #1235 cap block |
| 2 | leading space removed from the tail | **28** tests |
| 3 | ellipsis eats into the 42 (`39 + "..."`) | **4** tests |
| 4 | `...` replaced with `…` U+2026 | **4** tests, incl. the hardcoded-dots one |
| 5 | `[...body]` replaced with `body.split("")` | **1** test — the astral one, exactly |

Mutant 4 is why two ellipsis tests exist: the one written against
`REPLY_QUOTE_ELLIPSIS` is circular and survives it; the one that hardcodes `...`
is what kills it.

## What was measured, and what is still only reasoned

**The e2e ran, on the STACK lane, and it is green.** Union of every spec that
exercises the reply quote (`issue1105|issue1067|issue1156`), fresh stack with no
`KEEP_STACK` because cic `src` changed and the served bundle had to be rebuilt:
**12 passed**.

**#1105's rebuilt fixture measures — proven by displacement, not by passing.**
Its regression needs the compose box to overflow by at least 40px, and the cap
shortens the longest possible quote to ~61 characters. Dropping the fixture to a
single reply and rerunning turns the spec RED on exactly the overflow assertion:

```
expect(g.scrollHeight).toBeGreaterThan(g.clientHeight + minOverflowPx)
Expected: > 82      Received: 73
```

With `clientHeight` at 42, one reply buys **31px of overflow against the 40px
bar** — the earlier PR text called that "nowhere near", reasoned; it is now a
number. Three replies pass. So the three are not a guess: at one it breaks, and
the spec is not green for the wrong reason.

**Still not claimed:** that re-swiping the same row cannot race its own snap-back
transition. Each append is awaited before the next swipe fires, and the run was
green, but that is ONE green run — it is evidence, not a flake verdict.

**No `DESIGN_NOTES.md` entry.** The durable rationale sits at the seam it
governs — the four rulings on the constant, the wrapper-not-helper argument on
`replyQuote`, the cap-vs-overflow consequence in #1105's own header. An entry
would restate them a second time. Happy to add one if the call goes the other
way.

### DIFF
--- a/cicchetto/e2e/tests/issue1067-swipe-reply-message-menu.spec.ts
+++ b/cicchetto/e2e/tests/issue1067-swipe-reply-message-menu.spec.ts
@@ -173,7 +173,7 @@ test("issue1067 — a left→right swipe on a message quotes it into the compose
   expect(result.transformAfter).toBe("");
 
   // THE outcome: the quote, exactly as the issue spells it, caret at the end.
-  const quote = `<${specNick()}> ${body}<< `;
+  const quote = `<${specNick()}> ${body} << `;
   await expect(composeTextarea(page)).toHaveValue(quote);
   const caret = await composeTextarea(page).evaluate(
     (el) => (el as HTMLTextAreaElement).selectionStart,
@@ -293,7 +293,7 @@ test("issue1067 — the menu's Reply item fills the compose with the same quote 
   await menuItem(page, "Reply").click();
 
   await expect(menu(page)).toHaveCount(0);
-  await expect(composeTextarea(page)).toHaveValue(`<${specNick()}> ${body}<< `);
+  await expect(composeTextarea(page)).toHaveValue(`<${specNick()}> ${body} << `);
 });
 
 // The clipboard is stubbed at the browser boundary rather than driven through
@@ -408,7 +408,7 @@ test("issue1156 — a join row refuses the swipe while the privmsg above it stil
     ]);
     expect(onSpeech.prevented).toBe(true);
     expect(onSpeech.transformDuring).toMatch(/^translateX\(\d/);
-    await expect(composeTextarea(page)).toHaveValue(`<${specNick()}> ${body}<< `);
+    await expect(composeTextarea(page)).toHaveValue(`<${specNick()}> ${body} << `);
   } finally {
     await peer.disconnect("1156 witness done");
   }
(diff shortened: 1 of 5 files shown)


vjt:

Issue #1228, opened 2026-08-11
cic(settings): on a phone the notifications sub-page overflows the drawer's right edge — inputs, select and both 'remove' buttons are cut by the screen


vjt:
On a phone, the settings drawer's **notifications** sub-page spills past the right edge of the screen. From vjt's iPhone dogfood of staging (`b0d5342d`), #grappa: «qui abbiamo degli overflow sulla destra as well da fixare».

## What is cut off (read off the screenshot, right edge of the viewport)

- `enable browser notifications` — the master toggle label runs off the edge.
- the `only in channels:` text input (showing `#sbiffo, #grappa`) — right border is off-screen.
- the `only from nicks:` text input (placeholder `alice, bob`) — same.
- the `mute:` `pick a conversation…` select — same.
- the **devices** rows: both `remove` buttons are sliced vertically by the screen edge.

The left column (scrollback) is behind the drawer and is expected; the defect is that the drawer's own content is wider than the drawer.

## Not yet measured — do this first

The cause is **not** established. The drawer itself is bounded (`default.css:3971` — `width: 22rem; max-width: min(28rem, 100vw)`), so something inside is refusing to shrink to it. Reproduce at ~390 px CSS width with the inspector and find which box is the widest before touching anything.

Candidates worth checking in that repro, none of them confirmed:

- `.settings-drawer fieldset` (`default.css:4245`) sets border/padding/margin but no `min-inline-size`. A `<fieldset>` defaults to `min-inline-size: min-content` and, unlike other boxes, will not shrink below it.
- `.settings-drawer label` (`:4257`) is `display: flex` with no `min-width: 0` on the items — flex items keep `min-width: auto` and refuse to shrink below their content.
- `.settings-drawer label.prefs-list` (`:6517`) is the column/stretch variant used by the two text inputs; the inputs themselves get only padding rules (`:6528`), no width, so they fall back to the UA's `size`-derived intrinsic width.
- the devices rows (`.devices-list li`, `:6569`) are flex; `.device-ua-name` has `min-width: 0` (#964) but the `remove` button sits outside it.

## Wanted

Nothing in the drawer renders past its own right border at phone width, and the tap targets (`remove`) stay whole.

Related but separate: #1227 (the redundant `upload duration:` label pushing its select off-screen in the *general* sub-page) — same family of symptom, different sub-page.

vjt:
Shipped in **v1.0.0** — closing.

[...]

Issue #1044, opened 2026-08-08
One secret slot, several roles: a credential cannot hold a server PASS and a NickServ password at the same time


vjt:
A credential has exactly one secret slot (`password_encrypted`) and `auth_method`
decides which single role that slot is spent on. That models a network where the
roles are alternatives. On a network where they are not — a server that requires a
`PASS` to let you in *and* services that require a NickServ IDENTIFY — the model has
nowhere to put the second secret, and the user has to choose between connecting and
identifying.

**This is not an oper-side edge case.** It applies to every private or
password-gated server: a `PASS` on connect is the ordinary way such a network keeps
the public out, and it is unrelated to, and independent of, whatever account the user
then identifies to with services. Any user of any password-gated network hits this,
not just staff. The affected population is "everyone on a private server who also has
a registered nick", which on a client meant to connect to arbitrary networks is a
plain-vanilla configuration rather than a corner.

Reported by Mezmerize on IRC, from a live self-hosted setup.

## Measured today (against `origin/main`)

- `@auth_methods` is `[:auto, :sasl, :server_pass, :nickserv_identify, :none]`
  (`lib/grappa/networks/credential.ex:67`) — an exclusive choice. `password_encrypted`
  is the `PASS` wire token on `:server_pass`/`:auto`, the SASL PLAIN payload on
  `:sasl`/`:auto`, and the NickServ secret on `:nickserv_identify`. One column, three
  meanings, never two at once.
- The perform expander resolves a **fixed** alternation of three tokens —
  `~r/\$(nickserv_pass|oper_pass|nick\b)/` (`lib/grappa/session/perform_list.ex`) —
  each backed by its own dedicated column (`oper_pass_encrypted`,
  `credential.ex:201`). Adding a secret means touching schema, regex and API. It does
  not scale, and #124 relieved the pressure by removing one rather than generalising.
- `pending_password` is staged **only** for `:nickserv_identify`
  (`lib/grappa/session/server.ex:1325`), so on a `:server_pass` row `$nickserv_pass`
  binds to `nil` and the expander emits an empty string — a bare `IDENTIFY` reaches
  the wire.
- `autojoin_channels` is `{:array, :string}` in plaintext (`credential.ex:225`), so a
  channel key cannot live there without storing a secret in the clear.
  `auth_command_template` (`credential.ex:224`) is likewise a plain `:string`.
- `Network.services_flavor` already enumerates `:azzurra | :atheme | :oftc | :unknown`
[...]

Issue #349, opened 2026-07-19
cicchetto: registration wizard — per-network 'register nick' button + guided modal (Azzurra/Libera/OFTC)


vjt:
## Summary

A per-network **"register nick"** button, shown only while the user is **not `+r`** (unregistered) on that network, opening a guided modal that registers the nick with the network's NickServ. The button hides once the user becomes `+r`.

## Placement

The button lives in **Home, post-connect** — it appears only *after* the session has connected, and only while the user is **not registered / not `+r`** on that network. It is **not** on the login page. A pre-connect login-page placement was considered and rejected as excessive: *"joini prima registri poi"* — join first, register after. Registration is a deliberate post-connect action taken from Home, decoupled from the connect/login flow. Once `+r`, the button disappears.

## Flow (modal wizard — each step has a "next")

1. **Intro** — short copy on *why* registering matters. → next
2. **Email** — collect the email. → next
3. **Password** — collect the password. → next
4. **REGISTER** — generate the **network-dependent** `/msg NickServ REGISTER …`, send it, then **wait for and display NickServ's response** (async — no deterministic terminator).
5. **Confirmation code** — tell the user to grab the code emailed to them and paste it. → next
6. **Confirm / auth** — generate the **network-dependent** NickServ confirm/auth command, send it, **wait for NickServ and show the outcome**.

Once the user is `+r`, the "register" button disappears.

Every step gives **visual feedback** and surfaces NickServ's actual replies — it is not a silent connect-and-forward.

## Network-dependent command templates (must define, per network)

REGISTER + confirm/verify syntax differs per services implementation — support **Azzurra, Libera, OFTC**:
- **Azzurra** (Services 2.2.3): `REGISTER <password> <email>` → emailed code → verify (confirm the exact verb against `nickserv.c` — `CONFIRM`/`VERIFY`/`AUTH`).
- **Libera** (Atheme): `REGISTER <password> <email>` → `VERIFY REGISTER <account> <code>`.
- **OFTC**: confirm its REGISTER + activation flow.

Needs a per-network table of `{register, verify}` templates + the services nick. Where it lives (cic config vs grappa `Network` metadata) is a design call — grappa already owns per-network identity.

## Shared infrastructure (with #124, #347, #129)

- **`+r` detection** — the self-`MODE +r` observation (`event_router.ex:545-557`, per #129/#124) drives the button show/hide and confirms step 6 succeeded.
[...]

PR #1248, opened 2026-08-12
cic(#1157): the five admin residuals — collapse, autojoin, enabled, fieldset, auth

vjt:
## What this is

The tail of vjt's admin dictation on #1157, plus the two calls #1224 left
open. **Most of #1157 had already shipped** — rows become cards below
900px and the pan is gone (`e45a6733`), Visitors merged into one
row-backed Sessions view (`ffc8c044` / `e296e697`), all four Vhosts
requirements, #1224's ended sub-page (`9deecc60`). This is what was left,
one commit per item.

| item | commit |
|---|---|
| column 4 collapses into one button + dropdown on a phone | `91162728` + e2e `a8d7bb3a` |
| `autojoin` removed from the admin console | `2defe025` |
| no bind flow — a section per configured network with an `enabled` tick | `84f3cba4` |
| the settings form is a fieldset, one field per row | `d5ba4d46` |
| the auth selector offers `sasl` / `server_pass` / `none` | `387f79bf` |
| DESIGN_NOTES | `2d5745e4` |

## Two judgement calls worth reviewing

**1. The server's `auth_method` closed set is deliberately NOT narrowed.**
The selector offers three; the enum keeps five. Both survivors have an
owner that is not this console — `nickserv_identify` is written by the
SERVER (`registration_changeset/2`, when the #349 wizard sees `+r`), and
`auto` is the Bahamut/Azzurra PASS-handoff path documented in `AuthFSM`.
Narrowing the enum is a migration over live rows and belongs to **#1044**,
which #1157's own text defers this to ("settled there and rendered here,
not decided twice").

That makes the interesting case the one the narrowing CREATES: a
credential holding a method the console does not offer. A `<select>`
whose value matches no option renders as though the first option had
been chosen, and the next save carries that out — silently converting a
wizard-set `nickserv_identify` into `sasl`. So the stored value is added
to the options, labelled `(set elsewhere)`, and a save that did not touch
the field sends no `auth_method` at all. Both asserted.

**2. "One button with a dropdown" is implemented as "one control in the
cell".** A visitor row carries exactly one verb (`rowActions`), and
putting a list of one behind a tap costs an interaction to say nothing.
Only a multi-verb row collapses — which is the disconnect/terminate pair
vjt named. Picking from the menu ARMS the verb rather than running it, so
the destructive confirmation is not traded for a menu tap; the collapsed
cell grows a Cancel because the button that opened the menu is gone while
a verb is armed.

## Evidence

- `bun run check` — rc=0 (biome + tsc + e2e tsc)
- unit — 5294 passed, 281 files
- **`scripts/integration.sh` — 724 passed, 27.1m, rc=0** (full suite, not a scoped grep)
- `scripts/check.sh` — rc=0
- **17 unit mutants killed**, one assertion each where the assertions are
  independent
- **3 e2e mutants killed, with disjoint attribution**: removing the
  collapse kills the two @webkit tests and leaves the desktop one green;
  dropping the breakpoint from the gate kills the desktop test and leaves
  the two @webkit ones green; making the tick write immediately kills the
  unbind-arms test.

`ux-6-g-admin-mobile-h-scroll` now exercises the new `<fieldset>` at
393px, which is the oracle for its `min-inline-size: 0` — the #1228 trap,
one storey up.

## Test plan

- [x] full integration suite green
- [x] both sides of the mobile/desktop branch proven on a real iPhone 15
      viewport and on desktop chromium
- [x] no horizontal overflow on any admin surface at 393px (`.adm-nav`
      exempt, per vjt)
- [ ] operator dogfood of the per-user page on a real phone

### DIFF
--- a/cicchetto/e2e/tests/admin-user-networks.spec.ts
+++ b/cicchetto/e2e/tests/admin-user-networks.spec.ts
@@ -67,17 +67,21 @@ async function openUserPage(page: import("@playwright/test").Page, userId: strin
   await expect(page.getByTestId("admin-user-page")).toBeVisible({ timeout: 10_000 });
 }
 
-/** Fill and submit the `+` form for one network. */
-async function addNetwork(
+/**
+ * Give the user one network, the #1157 way: tick the section, fill the
+ * nick the bind needs, Save. There is no `+` and no network picker —
+ * every configured network already has a section, and the tick is the
+ * whole statement about access.
+ */
+async function enableNetwork(
   page: import("@playwright/test").Page,
   networkId: number,
   nick: string,
 ): Promise<void> {
-  await page.getByTestId("admin-user-network-add").click();
-  await expect(page.getByTestId("admin-user-network-add-form")).toBeVisible();
-  await page.getByTestId("admin-user-network-add-network").selectOption(String(networkId));
-  await page.getByTestId("admin-user-network-add-nick").fill(nick);
-  await page.getByTestId("admin-user-network-add-submit").click();
+  await page.getByTestId(`admin-user-network-enabled-${networkId}`).check();
+  await expect(page.getByTestId(`admin-user-network-form-${networkId}`)).toBeVisible();
+  await page.getByTestId(`admin-user-network-nick-${networkId}`).fill(nick);
+  await page.getByTestId(`admin-user-network-save-${networkId}`).click();
 }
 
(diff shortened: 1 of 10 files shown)


vjt:
Superseded by #1254, which cherry-picks these seven commits onto current main together with #1253's three and gates the union once. The commits are unchanged in content; only their SHAs differ, which is why GitHub cannot close this by itself.

Issue #1250, opened 2026-08-12
cic: the paste flood guard is bypassed by the GBoard clipboard chip — an IME insertFromPaste fires no paste event, so the 5-message cap never runs


vjt:
The paste flood guard (#80 / #816) is bypassed on Android when the text is inserted from the GBoard clipboard chip instead of the long-press **Paste** menu: the block lands in the compose box whole, no "Paste as N messages?" dialog, no "Upload as .txt" door, and submitting it sends the full burst the cap exists to prevent.

## Confirmed by the reporter

A self-hoster on `0.16.0-9632e34d` (measured by CTCP VERSION, and `git merge-base --is-ancestor 3d5695e2 9632e34d` confirms #816 is in that build — this is not a stale client). Their own A/B: **GBoard clipboard chip → no dialog, inserts directly; long-press Paste → dialog appears as designed.**

## Why

`routeClipboardPaste` (`cicchetto/src/lib/pasteRoute.ts`) is reached from exactly two places — the textarea's `onPaste` (`ComposeBox.tsx`) and the boot-time document listener (`lib/globalPaste.ts`, #352). Both are `paste` ClipboardEvent listeners.

An IME insertion is not a `paste` event. GBoard's clipboard chip commits text through the input method, which surfaces as `beforeinput` / `input` with `inputType: "insertFromPaste"` and fires **no** `paste` event at all. There is nothing to intercept, so `classifyPaste` never runs.

```
$ git grep -rn "beforeinput\|beforeInput\|insertFromPaste" origin/main -- cicchetto/src
(no output)
```

The guard is therefore not "weak on Android" — on that path it does not exist. Note the same reasoning covers any other IME-mediated insertion, and probably the desktop X11/Wayland middle-click primary-selection paste, which likewise does not always emit a `paste` event.

## Consequence

`PASTE_HARD_MESSAGE_LIMIT = 5` is the flood ceiling (`lib/pasteFlood.ts`), and per vjt's 2026-08-06 ruling the over-limit arm is meant to be a hard close — the block never enters the composer, upload-as-file is the only door. Via the chip the block enters the composer regardless of size, so an operator can send an arbitrary burst without ever being asked. The cap is bypassable by picking a different paste gesture, which makes it advisory rather than enforced.

Note the burst is not sent silently on arrival — it still takes a submit — but the operator is never shown the count, and the `.txt` alternative is never offered.

## Fix sketch

[...]

Issue #352, opened 2026-07-19
cicchetto: global paste (Ctrl/Cmd+V) should focus the compose bar and paste, even when unfocused


vjt:
## Summary

Pressing **Ctrl+V / Cmd+V while the compose input is NOT focused** currently does nothing — paste only works with the textarea already focused. It should **focus the compose bar and paste the content** (text or image), so the user can paste without first clicking into the input.

## Current state (source-anchored)

- Paste is handled **only on the compose textarea**: `onPaste` (`ComposeBox.tsx:353`) reads `e.clipboardData` items (text + image branches), wired at `ComposeBox.tsx:513` on the `<textarea>`. It fires only when the textarea already holds focus.
- The compose textarea can be focused programmatically: `ta.focus()` (`ComposeBox.tsx:348`).

So a paste with focus anywhere else (scrollback, body, a just-closed menu) is dropped on the floor.

## Proposed change

Install a **document-level `paste` listener** (same boot-time global pattern as `lib/keepKeyboard.ts`): when a `paste` `ClipboardEvent` fires and the active element is **not** the compose textarea (nor any other text-entry field), focus the compose textarea and route the clipboard payload through the **existing** paste path:
- text → insert at caret (reuse the `#80` confirmed-insert logic already in ComposeBox),
- image/file → route to `handleFiles` / the existing image-paste branch (upload).

Don't duplicate the text/image branching — reuse `onPaste`'s logic.

## Feasibility (vjt's open question — "non so se sia possibile stare in listen su un tale evento")

It **is** possible: browsers dispatch a `paste` `ClipboardEvent` to `document` even when focus is on a non-editable element (Chrome/Firefox expose `clipboardData` there). **Caveat — Safari/iOS is stricter:** clipboard read may be denied outside a user-gesture-bound *editable* target, so a programmatic focus-then-replay might not see the payload. Degrade gracefully: focus the compose input anyway so the user's paste (or next keystroke) lands in it, even if the first event's data can't be read. Cross-check the iOS notes in `keepKeyboard.ts` / `docs/DESIGN_NOTES.md`.

## Boundary

- **Don't hijack paste** when a different editable field owns focus (search, settings input, another modal) — only act when focus is on a non-input surface (scrollback / body) or nothing editable.
- Respect open modals with their own paste semantics (media viewer, theme editor).
- Client-side (cicchetto) → no cold deploy.

vjt:
[...]

PR #1253, opened 2026-08-12
cic(#1250): guard the paste flood cap on the IME path, where it did not exist

vjt:
Refs #1250.

GBoard's clipboard chip fires **no `paste` event** — it commits through the
input method, which surfaces as `beforeinput` with an `insertFromPaste`
inputType. Both doors into `routeClipboardPaste` are `paste` listeners, so
`classifyPaste` never ran: the guard was not weak on that path, it was
**absent**, and `PASTE_HARD_MESSAGE_LIMIT` was advisory rather than enforced.
Reporter's A/B: chip → no dialog, long-press Paste → dialog. Verified on
`origin/main`: `git grep beforeinput -- cicchetto/src cicchetto/e2e` returns
nothing.

`routePastedInput` is the third door and converges on the same decision —
`routeGuardedText` now owns the drain check and the three-way classify, so the
textarea paste, the #352 document listener and the IME insertion cannot drift.
`insertFromPasteAsQuotation` rides the same arm.

## Not asking twice, and why there is no bookkeeping

A native paste is reported twice: as `paste`, and as the `beforeinput` for the
insertion it causes. The first implementation was a module-level gesture claim
released on a macrotask. **It was reverted**: it is order-independent, but it
also makes two pastes *in one task* indistinguishable from one — a state no
browser can produce and every existing paste test does. Two test files had
already been taught to yield between cases before that became obvious, and a
guard that disarms itself for the next author is worse than the assumption it
removes.

What ships is structural: the UA fires `paste`, and only an **uncancelled**
paste performs the insertion that fires `beforeinput`. Both guarded arms
cancel, so the second report never exists. The pass-through arm lets both fire
and is idempotent.

## The ordering was measured, not assumed

The issue says some engines report `beforeinput` first, which would break that
argument. A synthetic `ClipboardEvent` cannot settle it — an untrusted event
runs no default action, so no `beforeinput` ever follows and the measurement
would be vacuous. The new spec drives a **real** clipboard gesture and pins the
observed sequence:

- **chromium**: `["paste"]`
- **webkit (iPhone 15)**: `["paste"]`

A guarded paste is reported exactly once, as `paste`, on both engines we ship.
The issue's claim is not reproduced on either. If a future engine differs, that
pin reddens instead of leaving a double dialog for an operator to find.

## Test plan

- [x] `bun run test` — 5290 passing, incl. 17 in `pasteRoute.test.ts`
- [x] `bun run check` — biome + tsc + tsc `-p e2e/tsconfig.json`
- [x] `scripts/integration.sh --grep "#1250"` — 4/4 across both projects
- [x] **Mutation**: unwiring the IME listener from ComposeBox kills exactly the
      two IME tests and leaves the two ordering tests green. (The first mutant
      broke `tsc`, so the suite never ran and its red meant nothing — a mutant
      has to compile before its red is a measurement.)

## Stated limits

The IME tests dispatch a synthetic `beforeinput[insertFromPaste]`: Playwright
cannot drive GBoard and no CDP verb commits one. What is proven is that cic
answers the event GBoard emits — that GBoard emits it rests on the reporter's
device A/B. The issue's closing question about the `text === ""` iOS arm is
left open, not folded in.

### DIFF
--- a/cicchetto/e2e/tests/issue1250-ime-paste-flood-guard.spec.ts
+++ b/cicchetto/e2e/tests/issue1250-ime-paste-flood-guard.spec.ts
@@ -1,0 +1,210 @@
+// #1250 — the paste flood guard on the IME path.
+//
+// GBoard's clipboard chip does not fire a `paste` event. It commits the text
+// through the input method, which surfaces as `beforeinput` with an
+// `insertFromPaste` inputType, and both pre-#1250 doors into the router were
+// `paste` listeners. The guard was therefore not weak on that path, it was
+// ABSENT: `classifyPaste` never ran, and an operator could send a burst of any
+// size by choosing the chip over the long-press Paste menu — never shown the
+// count, never offered the .txt door that the 2026-08-06 ruling made the only
+// way out above the cap. Reporter's A/B: chip → no dialog, long-press → dialog.
+//
+// What each test here is for:
+//
+//   1-2. The IME door itself. A synthetic `beforeinput[insertFromPaste]` is the
+//        closest a browser harness gets to the chip — Playwright cannot drive
+//        GBoard, and no CDP verb commits an insertFromPaste. This is a stated
+//        LIMIT of the simulation, not a claim about GBoard: what is proven is
+//        that cic answers the event GBoard emits. The oracle is the visible
+//        outcome the reporter did NOT get — a dialog, and a composer that
+//        stays empty.
+//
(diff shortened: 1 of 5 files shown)


vjt:
Superseded by #1254, which cherry-picks these three commits onto current main together with #1248's seven and gates the union once. The commits are unchanged in content; only their SHAs differ, which is why GitHub cannot close this by itself.

Issue #1256, opened 2026-08-12
bug(uploads): text/plain is served without a charset, so UTF-8 pastes render as mojibake


abonforti:
`GET /uploads/:slug` serves `text/plain` with no `charset` parameter, so a browser has no declared encoding and falls back to its locale default (windows-1252 in Western locales). Any uploaded text containing non-ASCII renders as mojibake — `è` shows as `Ã¨`, `—` as `â€”` — even though the stored bytes are valid UTF-8.

Hit on a paste generated from IRC text: the file on disk decodes cleanly as UTF-8, and the mojibake in the browser matches a UTF-8 → windows-1252 misread exactly, character for character.

## Response

```
content-type: text/plain
content-disposition: inline; filename="paste.txt"; filename*=UTF-8''paste.txt
x-content-type-options: nosniff
```

Note the mismatch: `content-disposition` labels the *filename* as UTF-8, while the body itself is served unlabelled.

## Where it comes from

`lib/grappa_web/controllers/uploads_controller.ex:185`:

```elixir
|> put_resp_header("content-type", row.mime)
```

`row.mime` is the value stored at upload time, taken from the closed allowlist whose text entry is bare `"text/plain"` (line 85). There is no code path that attaches a charset.

The charset is not lost in transit either — it cannot get in. `validate_mime` (line 277) matches the client-declared content type by exact equality:

```elixir
case Map.fetch(@mime_categories, ct) do
  {:ok, category} -> {:ok, ct, category}
  :error -> sniff_audio(ct, filename)
end
```

An upload declaring `text/plain; charset=utf-8` misses `@mime_categories`, falls through to `sniff_audio`, is not a generic sniffable MIME, and gets 415. So a client that labels its encoding correctly is rejected, and one that doesn't gets stored unlabelled.

## Repro

1. `printf 'perch\xc3\xa9 \xc3\xa8 cos\xc3\xac\n' > t.txt` (UTF-8, no BOM)
2. Upload it via the normal paste/upload path.
3. Open the returned `/uploads/<slug>` URL in a browser with a Western locale.
4. The accented characters render as `Ã©` / `Ã¨` / `Ã¬`. `curl` the same URL and the bytes are correct UTF-8 — the file is fine, only the declared type is wrong.

## Scope

[...]

PR #1259, opened 2026-08-12
uploads: declare the charset of uploaded text instead of guessing it (#1256)

vjt:
## What

`GET /uploads/:slug` served `text/plain` with no charset, so a browser had no declared encoding and fell back to its locale default (windows-1252 in a Western locale). A UTF-8 paste rendered as mojibake — `è` as `Ã¨`, `—` as `â€”` — while the bytes on disk were valid the whole time.

The charset could not get in either. `validate_mime/1` matched the whole client-declared content type against the closed MIME allowlist, which made that allowlist accidentally a closed *parameter* allowlist: `text/plain; charset=utf-8` missed, fell into the audio extension rescue, and earned a 415. The client labelling its encoding honestly was the one being refused.

Per vjt's ruling on the issue, the fix labels only where the bytes are ours, server first.

## How

- **`Grappa.Uploads.ContentType`** — `parse/1` splits `type/subtype` from the parameter run and reduces the declared charset to an ATOM from a closed set; `header/2` rebuilds the response header by re-spelling that atom. Reached through the context (`Uploads.parse_content_type/1`, `Uploads.content_type_header/2`), following the `MimeExt` precedent.
- **`uploads.charset`** — new nullable column, no backfill, no default. NULL is not missing data: it is the record that those clients declared nothing. Mapped through an `Ecto.Enum` fed from `ContentType`, so a charset the header builder cannot spell canonically cannot be persisted. The migration is additive-nullable and classifies HOT, so it joins the preflight's pinned HOT set.
- **`UploadsController`** — matches the type, persists the charset beside the MIME, and rebuilds the `content-type` on `show/2`. `nosniff` unchanged.
- **cic** — `baseMime` makes the upload gates parameter-tolerant (an exact-string `categoryOf` would have dropped a correctly-labelled File client-side, before it could even reach the 415), and the paste-as-`.txt` File declares `charset=utf-8` — a fact for that path, since the File constructor encodes a USVString part as UTF-8 by specification.

## Decisions worth reviewing

- **The rebuild is the security half.** Accepting parameters turns the declared type into client-controlled text one hop from a response header, and `nosniff` only pins the browser to whatever we declare. Storing the raw parameter run would trade mojibake for content-type confusion and undercut the protection at `show/2`. Nothing the client wrote reaches the header.
- **An unrecognised charset is dropped, not rejected.** `charset=iso-8859-1` still lands a 201 and is served unlabelled — the pre-fix behaviour, deliberately preserved. Refusing it would regress the mirror-image case (a genuinely Latin-1 `.txt`, correct today because nothing is declared) into a 415 to protect it from a display bug.
- **A charset is read only off a `text/*` type**, derived from the type rather than kept as a second allowlist beside `@mime_categories`.

## Test plan

- [x] Red measured before the fix: 5 of the 6 new controller assertions failed with `unsupported_media_type`; the sixth — the mirror-image guard — passed, as it had to. On cic, `categoryOf("text/plain; charset=utf-8")` answered `null` and the paste File's type read bare `text/plain`.
- [x] `scripts/check.sh` green: 8 doctests, 59 properties, 5961 tests, 0 failures; Credo 5713 mods/funs no issues; Dialyzer `Total errors: 0`; Sobelow no medium-or-above; bats green.
- [x] `bun run test` 281 files / 5315 tests, `bun run check` clean.
- [ ] e2e `issue1256-upload-charset.spec.ts` — pastes an accented block, uploads it, navigates to the served URL and asserts the RENDERED text (with `Ã` as a negative control, a byte that can only appear under a windows-1252 misread). Not yet run: the stack lane was held elsewhere. To be measured before merge.

## Notes

- Not for the 1.0.0 cut, per vjt on the issue.
- The wire contract is untouched; the admin uploads listing is unchanged.

### DIFF
--- a/cicchetto/e2e/tests/issue1256-upload-charset.spec.ts
+++ b/cicchetto/e2e/tests/issue1256-upload-charset.spec.ts
@@ -1,0 +1,125 @@
+// #1256 — an uploaded UTF-8 paste must RENDER as the text that was
+// pasted, not as mojibake.
+//
+// `GET /uploads/:slug` used to serve `text/plain` with no charset, so a
+// browser had no declared encoding and fell back to its locale default
+// (windows-1252 in a Western locale): `è` painted as `Ã¨`, `—` as
+// `â€”`. The bytes on disk were always fine — the DECLARATION was
+// missing, and it could not get in either, because the upload
+// allowlist matched the client-declared content type by whole-string
+// equality. A client labelling its encoding honestly
+// (`text/plain; charset=utf-8`) was the one earning a 415.
+//
+// The fix is two-sided and this spec is the only gate that sees both
+// halves at once: cic labels the File it builds from a JS string
+// (pasteRoute.ts — UTF-8 by File-constructor spec, a fact and not a
+// guess), and the server splits the parameter off the type before
+// matching its allowlist, persists the charset in a closed set, and
+// rebuilds the response header from it (Grappa.Uploads.ContentType).
+//
+// The oracle is deliberately the RENDERED text, read out of a real
+// navigation to the upload URL — the response header is asserted too,
+// but a header assertion alone would not prove the browser stopped
+// guessing. Server-side unit coverage (uploads_controller_test.exs)
+// pins the header matrix; vitest pins the File's label; only a real
+// browser can show the accents surviving the trip.
+
+import type { Page } from "@playwright/test";
+import {
+  composeTextarea,
+  confirmModal,
+  confirmModalYes,
+  loginAs,
+  selectChannel,
+} from "../fixtures/cicchettoPage";
+import { AUTOJOIN_CHANNELS, NETWORK_SLUG } from "../fixtures/seedData";
+import { expect, specNick, specUser, test } from "../fixtures/test";
+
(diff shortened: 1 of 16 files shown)


vjt:
### e2e measured

`scripts/integration.sh --project chromium --grep "#1256"` — **1 passed** (16.2s).

And the oracle was checked against a mutant, because a rendered-text assertion that has never been seen to fail is not evidence: if Chromium auto-detected UTF-8 on its own, the spec would pass without the fix.

**Mutant:** `show/2` reverted to `put_resp_header("content-type", row.mime)` — the upload still succeeds and is still stored with `charset: :utf8`, only the header stops carrying it. The header assertion was removed for that run alone, so the *rendered* text was the only thing under test and the kill is attributable to it.

**Result: 1 failed**, on exactly that assertion, with the windows-1252 misread character for character:

```
Received string: "9e7c7608 riga 0 â€” perchÃ© Ã¨ cosÃ¬, piÃ¹ o meno: Ã Ã¨Ã¬Ã²Ã¹
```

So the browser does *not* recover the encoding by itself, the declaration is what fixes it, and the spec discriminates. Both files restored; tree clean at `abd65a4f`.

Issue #1249, opened 2026-08-12
channel modes: CHANMODES type-A list modes (b/e/I) are written into the channel-mode set — a single +b makes the header read +blrnt


vjt:
Setting a ban makes the channel-mode indicator grow a `b`: a channel that is really `+lrnt` renders as `+blrnt` after a single `MODE #chan +b <mask>`. Reported by a self-hoster; the mode string shown in the channel header is the server-side `channel_modes` cache, so this is not a cicchetto rendering choice.

## Where

`lib/grappa/session/event_router.ex` — `walk_channel_modes/5`.

The walk tests `ISupport.user_prefix/2` first (membership modes: skipped, they belong to the roster), and treats **everything else** as a channel-level flag:

```elixir
:error ->
  takes_param = ISupport.takes_param?(isupport, mode, direction)
  {arg, remaining} = if takes_param, do: pop_arg(args), else: {nil, args}
  entry = toggle_channel_mode(entry, mode, arg, direction)
```

`ISupport.takes_param?/3` correctly says "type A takes a param on both signs", so the mask is consumed and the arg list stays aligned — but the letter is then written into `entry.modes` and the mask into `entry.params[mode]`, exactly like `+k`/`+l`.

CHANMODES type A (`b`, `e`, `I` — `Grappa.Session.ISupport` `@default_chanmodes` `a:`) is a **list**, not a channel flag. RFC 2811 §4.3 / the ISUPPORT `CHANMODES=A,B,C,D` contract: type A letters address a per-channel list, they are never part of the channel's mode state, and an ircd never reports them in 324 RPL_CHANNELMODEIS. The classifier already knows the class — the walk just does not consult it for anything but arity.

## Consequences

1. **Phantom flag.** One `+b` and the header shows `b` forever, alongside real flags. Same for `+e` (ban exception) and `+I` (invite exception).
2. **One param slot for a whole list.** `entry.params["b"]` holds the *last* mask set; every earlier ban's mask is overwritten. The cache claims a single-valued mode where the ircd has an N-entry list.
3. **Unban of a different mask clears the flag.** `toggle_channel_mode(entry, "b", _, :remove)` does `List.delete(entry.modes, "b")` — removing ban #2 of 3 drops the `b` from the display while two bans are still in force.
[...]

PR #1261, opened 2026-08-13
Add UnifiedPush as a second Web Push provider for push_subscriptions

LucentW:
## Summary

Adds [UnifiedPush](https://unifiedpush.org/) support to `push_subscriptions`, so a native client that can't piggyback on a browser's Push API (e.g. an Android app) can still register for push notifications without an always-on WebSocket.

The key finding driving the shape of this: **a UnifiedPush registration is wire-compatible with the existing W3C Web Push one.** The client generates its own P-256 keypair + auth secret client-side (the same way `PushManager.subscribe()` does internally for Web Push), so `Grappa.Push.Sender`'s existing VAPID-signed `aes128gcm` encrypted send path works against a UnifiedPush endpoint completely unchanged. **`Push.Sender` needed zero code changes.**

What differs is only *how a device registered* — so that's the only thing that got a new field:

- `push_subscriptions` gains a `provider` column (`"webpush"` | `"unifiedpush"`, default `"webpush"`). Plain `ADD COLUMN`, classified `:hot` by `Grappa.Deploy.Preflight` (pinned into the allowlist test).
- `POST /push/subscriptions` accepts an optional `"provider"` tag on the same wire body it already took. Omitted → the schema default preserves every existing caller's behavior byte-for-byte.
- `GET /push/subscriptions` surfaces `provider` in each device summary, so a "see + revoke my devices" UX (cic's B3 settings drawer, and the equivalent I'm building client-side in Resentin) can show what kind of device each row is. `DELETE` revokes either identically — no branch needed there either.

An earlier draft of this went the other way — plaintext, unencrypted UnifiedPush POSTs, no keys — on the theory that UnifiedPush's own spec doesn't *require* encryption. Reversed after review: this codebase already has a working, audited encrypted pipeline, an Android client can produce a real keypair as easily as a browser can, and a second delivery mechanism would have been real new surface (second HTTP client path, second failure taxonomy) for a privacy property the existing path already gives for free. Full writeup in the `docs/DESIGN_NOTES.md` entry.

## Why draft

This touches the `push_subscriptions` schema on a server people are running in production — opening as a draft so you (or your review agents) have room to look it over before it's mergeable, per your CLAUDE.md contribution posture. Happy to adjust anything about the shape.

## Client side

I'm implementing the Resentin (native Android) side of this on a matching branch — registration flow, device list with the same provider-aware "see + revoke my devices" UX, and incoming-push handling. Wanted the server contract settled first since the client work depends on this exact wire shape.

## Test plan

- [x] `mix test --warnings-as-errors` — 5946 tests, 0 failures (full suite, not just the touched area)
- [x] `mix format --check-formatted`
- [x] `mix credo --strict` — 0 issues
- [x] `mix sobelow --config --exit Medium` — 0 Medium+ findings
- [x] `mix doctor` — 0 failed modules, coverage unchanged
- [x] `mix grappa.gen_wire_types --check` — in sync (this PR doesn't touch any Wire typespec, but ran it to be sure)
- [x] New coverage: `Subscription.changeset/2` provider defaulting/validation, controller create/list with `provider`, `Grappa.Deploy.Preflight` hot-set pin for the new migration
- [ ] Not yet run: `mix hex.audit` / `mix deps.audit` (no dependency changes in this PR, so skipped — let me know if you'd like it run anyway)


vjt:
## Review — UnifiedPush provider (head `9ec413da`)

Thanks for this — the design instinct here is the right one,
and the reasoning is written down where the next reader will
find it. Below is a full review.

Every claim is labelled **MEASURED** (I ran it, literal output
included) or **DERIVED** (source inspection only). Nothing has
been promoted from one to the other to make it sound stronger.

**Verdict: mergeable with changes.** Four items — one
house-rule violation, one measured boundary defect, one
measured surviving mutant, one migration-hygiene fix.

---

### The central thesis holds

DERIVED (source inspection at `9ec413da`). Delivery really is
provider-agnostic. `provider` appears in
`lib/grappa/push/sender.ex` on lines 11, 13, 27 and 234 only —
all four are documentation. The full path:

* `Push.create/2` → `do_insert/2` → `Subscription.changeset/2`
  → `Repo.insert/1`. No branch.
* `touch_last_used/1` uses a bare `Repo.update/1` with no
  changeset, so it never touches the column.
* Dead-endpoint cleanup keys on `sub.endpoint` alone.
* `delete/2` keys on id + subject. No branch.

The claim is true. The +22 lines in `sender.ex` being entirely
moduledoc is the right call: the reason a field does *not*
branch is exactly what a future reader needs written down.

Also checked rather than assumed:

* **`:hot` classification — correct.** I read the classifier
  rather than the prose. `hot_alter_op?/1`'s 3-arg clause
  requires a literal column, a literal type, and
  `old_inserts_still_valid?/1`, which returns
  `Keyword.has_key?(opts, :default)` for `{:ok, false}`.
  `add :provider, :text, null: false, default: "webpush"`
  satisfies all three. The allowlist pin in `preflight_test.exs`
  is present.
* **Additive-only wire contract respected.** Nothing repurposed
  or removed, `provider` is snake_case, no `protocol_version`
  bump owed.
* **The client is safe.** `cicchetto/src/lib/push.ts:174-183`
  decodes the index with a bare `as` cast, and cic ships no
  runtime validator (no zod/valibot/ajv/io-ts in
  `package.json`), so the extra key is ignored.
  `PushSubscribeRequest` (`push.ts:108-118`) has no `provider`
  key, so there is no collision. And
  `GrappaWeb.PushSubscriptionJSON` is not a `*Wire` module, so
  `mix grappa.gen_wire_types --check` does not cover it — no
  codegen drift, no client change owed.
* **`:text` vs `:string` is not a finding.** I checked before
  raising it: `:text` is used 14× against `:string`'s 57×
  in-tree, and both are TEXT affinity in SQLite.
* The DESIGN_NOTES entry is appended at EOF dated 2026-08-13,
  directly after the other 2026-08-13 entry. Chronology fine.
* Docs updated in the same commit, as the house rule requires.

**Baseline, MEASURED:** both touched test files at `9ec413da`,
unmodified — `54 tests, 0 failures`.

---

### 1. `provider` should be `Ecto.Enum`, not `:string` + `validate_inclusion`

`lib/grappa/push/subscription.ex:71,73,97,173`

**DERIVED** — this is a consistency judgment, not a runtime
fact. The census behind it is measured.

CLAUDE.md: "**Atoms or `@type t :: literal | literal` — never
untyped strings** for closed sets." Line 73 is
`@type provider :: String.t()`, which is the shape that rule
names, and it is re-exported into the published summary type at
`push_subscription_json.ex:42`.

MEASURED census of closed-set columns in `lib/`:

| field | module |
|---|---|
| `kind` | `Scrollback.Message` |
| `auth_method` | `Networks.Credential` |
| `connection_state` | `Networks.Credential` |
| `services_flavor` | `Networks.Network` |
| `event` | `SessionLog.Event` |
| `subject_kind` | `SessionLog.Event` |
| `passkey_mode` | `Accounts.User` |

All seven are `Ecto.Enum`. `validate_inclusion` occurs **zero**
times in `lib/`, `test/` and `priv/`, so this would introduce
the codebase's first instance — against "Total consistency or
nothing. Half-migrated creates two patterns."
`docs/project-story.md:163-167` records the same call being
made for `Message.kind`.

The reason I would not treat this as a nit is that the change
looks close to free: `Ecto.Enum` casts the inbound wire string,
Jason re-encodes the atom back to the same wire string, and an
out-of-set value still yields the `"is invalid"` token — so the
422 test and the JSON contract are both untouched. Only the two
struct-level assertions become `:webpush` / `:unifiedpush`. The
default would be spelled `default: :webpush` in the schema and
`"webpush"` in the migration, matching `connection_state` and
`passkey_mode`.

In fairness, there is one real counter-argument: `Ecto.Enum`
raises when *loading* a value it does not know, so a row written
by a newer node breaks an older node's read. That hazard is
genuine — but all seven existing fields already accept it, so it
does not single this case out. If you disagree, that is the
argument to make, and it would belong in DESIGN_NOTES as a
deliberate exception rather than passing silently.

---

### 2. MEASURED — `"provider": ""` returns 201 and silently becomes `"webpush"`

`lib/grappa_web/controllers/push_subscription_controller.ex:127`

The guard is `when is_binary(provider)`. The sibling helper ten
lines above, `maybe_put_supersedes/2` at `:117`, guards
`when is_binary(sup) and sup != ""`. This one omits the second
half.

Mechanism: `Ecto.Changeset.cast/3`'s default `empty_values` is
`[""]` (`deps/ecto/lib/ecto/changeset.ex:372`), so `""` is
stripped from params before validation, `validate_inclusion/3`
never fires, and the schema default lands.

This was measured, not reasoned. Two temporary tests posted a
body with `"provider": ""`. Literal output:

```
MEASURED-A status=201 stored_providers=["webpush"]
```

and, asserting what the code comment claims:

```
1) test ... MEASUREMENT B — the code comment's claim:
   empty provider 422s
   ** (RuntimeError) expected response with status 422,
      got: 201, with body:
   "{\"id\":\"fc589b0e-0c45-4841-81f2-297d22c7edf7\",
     \"created_at\":\"2026-08-13T07:27:57.814481Z\"}"
```

So a row is really created, with a real id, at 201.

The comment at `:122-126` states: "`validate_inclusion/3`
rejects anything outside the closed set, so an invalid string
surfaces as a 422, not silently coerced." That is not true for
`""`, and an inaccurate comment is worse than no comment,
because the next reader will rely on it.

Failing scenario: an Android client whose provider setting is
uninitialised POSTs `"provider": ""`, receives a 201, and its
row shows up in the device list as a browser tab. The blast
radius is small (display-only field, safe default), but it is a
silent coercion at a boundary, which CLAUDE.md calls out
explicitly.

Fix: mirror the sibling — `and provider != ""`.

Related, and **not** separately measured: the same guard also
swallows a non-string. `"provider": 123` falls through to the
default and 201s, whereas a bad *string* 422s. Worth deciding
deliberately; the empty string is the case I would definitely
close.

---

### 3. MEASURED — surviving mutant: the GET wire never observes `"unifiedpush"`

`test/grappa_web/controllers/push_subscription_controller_test.exs:305`

Line 305 is the only assertion made against the GET response
body, and it asserts `"webpush"`. The two tests that do observe
`"unifiedpush"` (`:178`) and the default (`:165`) read
`stored.provider` out of the database via
`Push.list_for_subject/1`, not out of the JSON.

Four-step measurement:

1. Baseline, both files unmodified: `54 tests, 0 failures`.
2. Mutant injected — `push_subscription_json.ex:67` changed
   from `provider: sub.provider` to `provider: "webpush"`.
   Re-run: `Compiling 1 file (.ex)` (confirming the mutant was
   actually built rather than served from a stale build), then
   **`54 tests, 0 failures`** — identical to baseline. The
   mutant survives the entire added suite.
3. Added the test below with the mutant still in place:
   `24 tests, 1 failure` — exactly one, the new one:
   `left: "webpush" / right: "unifiedpush"`.
4. Mutant reverted, test kept: `24 tests, 0 failures`.

So the mutant is real, the current suite cannot see it, and the
proposed test kills it and nothing else.

The behaviour the feature exists for — a device list that can
tell a browser tab from a UnifiedPush registration — is the one
behaviour not covered. Suggested test, verified by the run
above:

```elixir
test "index surfaces provider: unifiedpush on the wire",
     %{conn: conn} do
  {user, session} = user_and_session()

  {:ok, _} =
    Push.create({:user, user.id}, %{
      endpoint: "https://ntfy.example/up/list-up",
      p256dh_key: "k3",
      auth_key: "a3",
      user_agent: "ua-up",
      provider: "unifiedpush"
    })

  conn = conn |> put_bearer(session.id)
              |> get("/push/subscriptions")
  assert %{"subscriptions" => [a]} = json_response(conn, 200)
  assert a["provider"] == "unifiedpush"
end
```

To be fair to the rest of the suite: the other tests are real.
They assert outcomes rather than call sequences, and each kills
a distinct mutant (schema default, `@optional` membership,
`validate_inclusion`, the 422 shape at the HTTP boundary). The
one at `push_test.exs:216` is closer to a mirror — nothing in
this PR could break it — but it pins a real invariant cheaply,
so I would leave it in.

---

### 4. Hand-typed migration stamp

`priv/repo/migrations/20260813000000_add_provider_to_push_subscriptions.exs`

CLAUDE.md:526 is marked 🔴: "GENERATE migrations with
`mix ecto.gen.migration <name>` — NEVER hand-write the
timestamp", naming #1044/#1038, which collided on
`20260810120000`. Git fuses two such files with no conflict
marker, and on a database that has already applied the version,
neither migration ever runs and the deploy reports success.

Worth saying explicitly, since the existing tree looks like
cover and is not: MEASURED — 21 of the last 25 migrations on
`origin/main` carry hand-typed exact-hour stamps (40 of 86
overall). The rule text names that legacy as the cause of the
incident, so the rule is newer than the pattern.

Severity is latent, not live. MEASURED: this is the only open
PR touching `priv/repo/migrations/`, there are no duplicate
versions on main, and `20260813000000` sorts correctly after
main's latest `20260812220753`. Nothing is broken today —
regenerating the stamp is cheap insurance.

---

### 5. Nit — "DELETE revokes both" is true but unmeasured

The claim appears in four prose locations (Sender moduledoc,
Subscription moduledoc, controller moduledoc, DESIGN_NOTES) and
in zero tests. It *is* structurally true — `delete/2` keys on id
+ subject with no branch — so there is little that could
regress, which is why this is a nit rather than a finding. But
four assertions and no measurement is a ratio worth noticing.

---

### What I measured, and what I did not

**MEASURED by running code:** the baseline of both touched test
files (54/0); finding 2's literal 201 and `["webpush"]`, plus
the 422 assertion failing with `got: 201` and a real row id;
finding 3's full four-step mutant chain, with
`Compiling 1 file (.ex)` confirming the rebuild.

**MEASURED by inspection:** the `provider` path through `lib/`;
the preflight classifier predicates; the migration stamp census;
the duplicate-stamp check across open PRs; the `Ecto.Enum`
census; Ecto's `@empty_values` default; the client's push
module, `package.json` and the `gen_wire_types` glob; every
`provider` assertion in both test files.

**NOT measured, stated plainly:**

* **The full suite.** I ran only the two touched test files — no
  `check.sh`, no Dialyzer, no Credo, no Sobelow, no integration.
  **I make no claim that this branch is green overall.**
* Finding 1 is a consistency judgment; its census is measured,
  the recommendation itself is not something one can run.
* The non-string `"provider": 123` variant of finding 2 was not
  separately exercised.
* CI status was not part of this review.

---

### One claim I am not able to confirm

The DESIGN_NOTES entry states as settled fact that "only a
Web-Push-vendor's OWN push service actually verifies the VAPID
JWT, so an unverified-but-present header on a UnifiedPush POST
is simply inert, not wrong."

That is a claim about third-party distributor behaviour (ntfy,
NextPush, Sunup, and others), not about this codebase, and
nothing in this PR tests it against a live distributor. It is
plausible — UnifiedPush 3.x adopted RFC 8030/8291 Web Push — but
plausible is not measured, and the decision log is where this
project records things it has actually established. I would
suggest either one real end-to-end delivery against a live
distributor, or softening the sentence to say it is reasoned
rather than observed.

This does **not** block the merge: if the claim turns out to be
wrong, the blast radius is confined to `unifiedpush` rows and
surfaces through the existing failure taxonomy. No webpush
behaviour changes either way.

Relatedly, I did not check whether any distributor caps the
encrypted payload below what `Push.Sender` emits — unexamined
rather than cleared.

LucentW:
Thanks for the review — this is an unusually thorough one, and every item lands. Addressed all four plus the nit, on top of the rebase onto current `main`:

1. **`provider` → `Ecto.Enum`** (`cdb57bde`): matches the other seven closed-set fields in `lib/`, `validate_inclusion/3` removed (no longer the codebase's only instance). Wire shape is unchanged — Jason encodes the atom back to the same string.
2. **`"provider": ""` boundary defect** (`11a8667d`): confirmed the mechanism — `Ecto.Changeset.cast/3`'s `empty_values` strips `""` before validation regardless of the controller-level guard, so it 201s and defaults, same as an omitted field. That's the intended fallback, not a bug to reject; fixed the inaccurate comment instead and added the `provider != ""` guard explicitly (mirroring `maybe_put_supersedes/2`) plus a regression test pinning the behavior.
3. **Surviving mutant** (`11a8667d`): added the suggested test — GET now has a test asserting `provider` through the actual JSON view, not just via `Push.list_for_subject/1`.
4. **Hand-typed migration stamp** (`11a8667d`): regenerated via `mix ecto.gen.migration`, `20260813000000` → `20260813074128`; `preflight_test.exs`'s `@expected_hot` updated to match.
5. **Nit — DELETE cross-provider claim untested** (`11a8667d`): added a DELETE test for a `unifiedpush` row.
6. **Unmeasured VAPID-relay claim** (`91d6288b`): reworded both `DESIGN_NOTES.md` and `Push.Sender`'s moduledoc to say plainly this is reasoned (RFC 8030/8291), not observed against a live distributor — with the blast radius spelled out if it's wrong.

Re-ran the full local gate suite (compile --warnings-as-errors, format, credo --strict, sobelow --exit Medium, doctor, gen_wire_types --check) plus the full test suite after all of the above: **6005 tests, 59 properties, 8 doctests, 0 failures**.

vjt:
### Rebased onto `main` (maintainer action) — and one repo-specific trap you should know about

`#1249` landed on `main` a few minutes ago and touched `docs/DESIGN_NOTES.md`, which put this
branch into a conflicting state. Since the fixes here were already green at `91d6288b`, I rebased
the branch onto current `main` rather than ask you for another round-trip.

**What changed:** nothing in your work. Same six commits, same authorship, same content — only
replayed onto the new base. New head is `3a80c5c2`.

**The one thing worth flagging,** because it is a recurring failure mode in this repository and it
would have shipped silently:

The rebase reported `Successfully rebased`, exit 0, **zero conflicts, no dropped commits** — and it
still ate three lines. The `\n---\n\n` separator that opens a `DESIGN_NOTES.md` entry got resolved
away as "already present", welding this entry onto the tail of the one `#1249` had just added. It
shows up as **additions down by exactly 3, deletions zero**, which reads like a clean rebase if you
only check that nothing was deleted.

It was caught by diffing a `git diff --numstat` of the branch's contribution pinned *before* the
rebase against the same numstat after (64 → 61 on that file), then restored inside the entry's own
docs commit — never as a separator-only commit, since a standalone three-line commit gets eaten the
same way on the next rebase.

If you rebase this branch again, pinning that numstat first is the cheap defence.

**CI:** because this is a fork PR, the force-push invalidated the previous approval, so the workflow
runs are sitting at `action_required` until a maintainer re-approves them. That is on our side, not
yours.

Issue #1262, opened 2026-08-13
denoise: fold status-prefix MODE churn into the presence filter, without hiding structural channel modes


vjt:
## The ask

Channel MODE rows are noise on busy channels and should be foldable by the denoise
filter, the same way join/part/quit/nick_change already are (#222 render filter,
#458 server history filter).

## The standing ruling this collides with — read it before writing code

`lib/grappa/scrollback/message.ex` pins the suppressed set to a **deliberately
narrow** four:

```elixir
# The broad presence/control kinds (:mode, :topic, :kick,
# :server_event) carry operator-relevant signal and MUST stay visible —
# suppressing them would be a bug.
@suppressed_presence_kinds [:join, :part, :quit, :nick_change]
```

So the naive change — appending `:mode` to that list and to cic's
`SUPPRESSED_PRESENCE_KINDS` (`cicchetto/src/lib/presenceFilter.ts`) — is
explicitly ruled out by #458, and it would hide `+b` / `+k` / `+m` / `+i` / `+l`
transitions from the operator who is denoising a big channel. That is not what is
being asked for and it must not be what ships.

## What is actually noisy

The churn is **status-prefix modes on members**: services and bots re-opping and
re-voicing on every join, mass `+o` after a netsplit, `+v` sprays on a large
channel. Those rows say nothing an operator reads. The **structural** channel
modes (type A/B/C list and parameter modes — bans, exceptions, invex, key, limit —
and the type-D flags `+m`, `+i`, `+t`, `+n`) are exactly the operator-relevant
signal the #458 comment is protecting.

So the feature is a **split inside `:mode`**, not a fifth entry in the kind set.

## Where the data already is

`lib/grappa/session/event_router.ex` persists both mode branches with structured
metadata, not a rendered string:

- channel MODE → `build_persist(state, :mode, target, sender, nil, %{modes: modes, args: args})`
- own-nick MODE → same shape, on the synthetic `$server` window (#154(b))

So the letters needed to classify a row are already stored. No schema migration is
needed to *decide*; the question is where the decision is made.

## The constraint that shapes the design

The two filters are not symmetric:

- cic filters at **render** time (`ScrollbackPane`'s `rows()` memo) and can inspect
  the whole row, metadata included, for free.
- the server filters at **query** time (#458) so that `limit` counts visible rows —
  it omits **by kind** in SQL. Reaching into the metadata JSON per row to decide is
  the part that gets expensive and untestable.

[...]

Issue #239, opened 2026-07-14
cic: hidden control messages leave the unread counter stuck — mark trailing hidden control msgs read on window display


vjt:
## Bug (regression from #222)

With the #222 presence filter active (join/part/quit/nick hidden on large channels), the **unread counter can't be reset**. A channel accumulates trailing control messages; the server-owned read cursor still counts them as unread, but cic never renders them, so the user opens the window, sees nothing new, and the badge **stays stuck > 0** with no way to clear it.

## Expected

When a window is displayed/focused **and control messages are being hidden for that channel**, mark-as-read all the **trailing hidden control messages** so the counter resets. If the trailing unread includes a *visible* message, the counter must still reflect the visible unread (only skip the hidden trailing tail).

## Root cause

The read-cursor advance cadence is driven by *visible* messages settling in view. Hidden control messages at the tail never settle visibly → `last_read_message_id` never advances past them → the unread count (computed against stored message ids, including the hidden ones) stays non-zero.

## Fix direction

- On window display, when the channel's presence filter is hiding control messages, advance the read cursor to skip the **trailing** run of hidden control messages (up to the latest message id if the whole tail is hidden, otherwise up to the last message before the first *visible* unread).
- Read state stays **server-owned** (cursor = `last_read_message_id`); cic originates the read-position signal via the existing read-marker path — it does NOT compute/own the count locally. One code path, same as normal read-marker sends.

## Interactions to respect

- **#233 monotonic clamp:** the cursor is now advance-only (`id > current`). This fix only *advances* → compatible. Don't regress the clamp.
- **Explicit mark-as-unread** escape hatch: don't auto-advance over a window the user explicitly marked unread.
- **cic in-pane divider freeze contract:** auto-advancing past hidden trailing control msgs must not disturb the visible "new messages" divider for real unread.
- Unread count should be computed over **visible** (non-hidden) messages — hidden control msgs must not inflate the badge.

## Scope

- **P0**, cicchetto. Likely pure-client (read-marker send + count over visible rows) — rides `--cic` HOT. Confirm no server change needed; if the count is server-computed, extend the existing cursor path, don't fork.
- REAL e2e: large channel with trailing hidden join/part → open window → badge clears to 0 (assert the visible outcome, not a hollow spec).

_Filed on vjt's request (2026-07-14), appended to the end of the current pack (after #235)._

vjt:
[...]

Issue #154, opened 2026-06-29
cicchetto: ops-verb errors swallowed (fire-and-forget) + inbound MODE/umode not rendered


vjt:
Two client-side display gaps surfaced on #grappa (Mez, visitor socket): /umode and /mode produce no visible output at all.

1. Fire-and-forget pushes. `pushChannelOp/Kick/Umode/Mode` (cicchetto/src/lib/socket.ts:276,305,335,346) call `_userChannel.push(...)` with no `.receive` chain, unlike `oper` (381-383) and `raw` (398-400). The server `{:error, ...}` reply (visitor_not_allowed pre-#153, or invalid_*) is discarded -> silent no-op. compose.ts also sets `result = { ok: true }` synchronously regardless of the server reply (compose.ts:535-536,543-544). Wire `.receive("ok"/"error")` and surface errors like oper/raw do.

2. Inbound MODE/umode not rendered. Mez identified to NickServ and the server applied umode +a (RPL_UMODEIS 221 = +aiorS on query) but cicchetto showed nothing — no notification, no scrollback row. Incoming MODE changes on own nick / channel and the 221 numeric need a render path.

Verify: a rejected verb shows an error row; an applied own-nick umode shows a scrollback/notice row; a server MODE on a joined channel renders.

vjt:
**code complete — PENDING NIGHT --cic BATCH (rides #153/#155 window)**

Merged to `main` as `2b0a4ab` (FF off `4e7a98e`). **NOT deployed** — build-deferred to tonight's batch. Left OPEN for post-deploy close.

⚠️ **Scope flipped to plan B (COLD + --cic), vjt-approved.** #154 turned out to be TWO bugs, and the render fix needed a **server** change — so #154 is no longer `--cic`-only.

### What the guardrail proved (server-vs-cic, static trace before any edit)
- **221 RPL_UMODEIS** — `numeric_router` `:scan` → `{:server, nil}` → `:notice` on `$server`. **Already reaches cic AND already renders.** No change.
- **Channel MODE** (`:op MODE #chan +o nick`) — EventRouter channel branch → `:mode` row + broadcast. **Already reaches cic AND already renders** (`ScrollbackPane` `case "mode"`). No change.
- **Own-nick MODE echo** (`:you MODE you +a`) — EventRouter's user-MODE-on-self branch **DELIBERATELY dropped it** ("user-modes aren't channel events"). This was the ONLY render gap → **needed a server change** → flipped #154 to COLD.
[...]

PR #1265, opened 2026-08-13
#1262 — fold channel MODE into the denoise filter, as the withdrawn #458 rule now allows

vjt:
## What this is

Issue #1262 as **ruled**, not as filed. The body designs a write-time split
inside `:mode` (a `:status_mode` kind, or a boolean column) so status-prefix
churn folds while structural modes stay. vjt's comment of 2026-08-13
**withdraws the #458 rule** that made the split necessary, so `:mode` enters
the suppressed set as a plain fifth kind: no sub-classification, no column,
no migration. The write-time split is demoted to a possible follow-up.

- `Grappa.Scrollback.Message` — `:mode` appended to `@suppressed_presence_kinds`
  and its `@spec`.
- `cicchetto/src/lib/presenceFilter.ts` — `"mode"` appended to
  `SUPPRESSED_PRESENCE_KINDS`, same order.

## The accepted cost, stated where it will be found

While a channel is denoised, structural channel-mode transitions (`+b` `+k`
`+l` `+m` `+i`) are folded along with the status-prefix churn the issue aimed
at. A ban set during a denoised session is not in that session's transcript.
That is the trade vjt took knowingly; it is written into `DESIGN_NOTES` so
nobody rediscovers it as a bug report.

Own-nick mode rows (#154(b) — `/umode +i`, the services-pushed `+r`/`+a` at
IDENTIFY) are untouched: they land on the synthetic `$server` window, which
has no member count, and `Grappa.PresenceFilter` resolves an unknowable count
to SHOW.

## Two behaviour changes no line of the issue asked for

1. `count_after_split/6` and `ReadCursor`'s `hidden_channels` both derive from
   the same SSOT list, so `:mode` also stops counting toward the `events`
   badge on a hidden window. Correct direction — the pane no longer renders
   those rows, and a badge the operator cannot clear by reading is the
   #239/#505 defect — but it is a behaviour change, named here rather than
   left to be discovered.
2. The denoise button's `title` / `aria-label` now say `mode` too. It is
   rendered copy, so it was measured before being touched: the specs select
   that button by `data-testid="presence-toggle"` and never by its label.
   Leaving it would have made the UI understate what the toggle takes away.

## Gating

- **`Message.suppressed_presence_kinds/0` ↔ `SUPPRESSED_PRESENCE_KINDS` is now
  GATED**, which it was not before. `test/grappa/presence_filter_test.exs`
  parses the literal out of the `.ts` and asserts equality *and order*. The
  gate was hardened until each of two mutants produces a distinct, readable
  message (empty array → "Parsed ZERO kinds out of …"; renamed constant →
  "Could not locate …"). To be precise about what that bought: it bought a
  LEGIBLE failure, not a silent green averted — with `cic_kinds` empty the
  equality failed anyway, since the server list is a non-empty literal.
- `$server` resolving to SHOW is pinned **beside a live oversized channel on
  the same network**. Without that neighbour the assertion would pass merely
  because no session existed, which is a different clause.
- Red measured before green: with `lib/` reverted to the base, `Compiling 6
  files (.ex)` and 4 failures (the parity gate, the two on
  `suppressed_presence_kinds/0`, the `fetch hide_presence` one). Green after
  restore.
- Two `window_counts_test` tests used a `:mode` row as the witness of the
  control that SURVIVES hiding. Witness moved to `:topic` (already the in-tree
  idiom — the third test in the same file says so), the `:mode` rows kept and
  the expected counts moved, so those tests now also prove `:mode` stops
  counting.

## Verification

- `scripts/check.sh` — **rc=0**. `8 doctests, 59 properties, 5998 tests, 0
  failures`; bats suite all `ok`. Sobelow: the gate passes; not "zero
  findings" — the two pre-existing `Traversal.FileModule` Low Confidence hits
  on `lib/grappa/cic/bundle.ex` are still there.
- `bun run check` — rc=0, 1013 files, 60 warnings / 5 infos (the pre-existing
  baseline; the file count rose by one because the new e2e spec is genuinely
  covered by the e2e tsconfig).
- `bun run test` — rc=0, 281 files / 5331 tests passed.

🔴 **What the local `check.sh` does NOT cover.** It ran to green *before* the
last commit. That commit (`docs(#1262): correct the prose that ENUMERATES the
suppressed set`) touches `lib/grappa/scrollback.ex` and
`lib/grappa/user_settings.ex` — **comments and a typedoc only, no branch, no
predicate, no kind set** — plus four cic comments and the button copy. The cic
side was re-gated after the fact (both numbers above are post-commit). The
Elixir delta's verdict is **this PR's CI**, not a second local run. Local is
the sample; CI is the verdict.

**The e2e spec has never been executed.** `cicchetto/e2e/tests/issue1262-mode-denoise.spec.ts`
passes only the static gate, which says nothing about whether its selectors
find anything. It needs a stack, and it will be run together with the three
neighbours that consume the same filter — `issue222-presence-filter`,
`issue458-presence-page-yield`, `issue239-hidden-msg-unread` — chosen by who
consumes the shared path, not by filename prefix.

## Conflict watch

`docs/DESIGN_NOTES.md` is contended with **two** open PRs, #1260 and #1261.
The branch's contribution to that file is pinned at 81 additions / 0 deletions
and will be re-diffed after any rebase, boundary shape read on the file — the
separator-eater tell is additions down by 3, not deletions.

### DIFF
--- a/cicchetto/e2e/tests/issue1262-mode-denoise.spec.ts
+++ b/cicchetto/e2e/tests/issue1262-mode-denoise.spec.ts
@@ -103,14 +103,61 @@ test("#1262 — a channel MODE row folds when denoised (server-side too), the $s
     await expect(modeRows).toHaveCount(0, { timeout: 10_000 });
     await expect(contentRow).toHaveCount(1); // a filter, not a blanket drop
 
-    // --- THE SERVER HALF: reload → the page arrives without the mode row --
-    // RED before #1262's server change: `:mode` was not in
-    // `suppressed_presence_kinds/0`, so the cold-load page still carried the
-    // row and cic (with only its own side moved) would render it.
+    // --- THE SERVER HALF: the cold-load PAGE arrives without the mode row -
+    //
+    // Asserted on the RESPONSE BODY, not on the pane. Measured, because the
+    // first shape of this arm did the latter and was VACUOUS: with only
+    // `message.ex` reverted and cic's own filter intact, the render filter
+    // hides the row the server still sends, so the pane shows 0 either way
+    // and the spec passed with the server defect fully present. A server-only
+    // mutant survived it; that is what promoted this from a DOM count to a
+    // wire assertion. The pane count below stays as the operator-visible
+    // half, but it is the JSON that speaks for the server.
+    // The path must be ANCHORED, not `includes`-matched: the cold load also
+    // issues `/messages/count`, whose URL contains `/messages` and whose body
+    // is `{count: N}` — an object with no `.some`. Measured, not foreseen: the
+    // loose predicate matched it first and the arm died on a TypeError.
+    // EVERY such response is collected, not just the first one awaited. The
+    // first shape awaited a single response and read an array with no privmsg
+    // in it — the cold load issues more than one GET for a channel, and which
+    // one carries the rows is not something this spec should be pinning. The
+    // claim is about what the server served for this channel, so the subject
+    // is the UNION of what it served.
+    const servedPages: { kind: string }[][] = [];
+    page.on("response", (r) => {
+      if (
+        r.request().method() !== "GET" ||
+        r.status() !== 200 ||
+        !new URL(r.url()).pathname.endsWith(`/channels/${encodeURIComponent(channel)}/messages`)
+      ) {
+        return;
+      }
+      // `GET .../messages` answers a BARE ARRAY (api.ts's `fetchMessages`),
+      // not an envelope. Anchored on the path because `/messages/count`
+      // answers `{count: N}` and an `includes` match ate it first.
+      void r
+        .json()
+        .then((body) => {
+          if (Array.isArray(body)) servedPages.push(body as { kind: string }[]);
+        })
+        .catch(() => {});
+    });
+
     await page.reload();
     await selectChannel(page, NETWORK_SLUG, channel, { awaitWsReady: false });
+
+    // DOM barrier FIRST: the content row being back is the durable signal
+    // that the cold load has actually landed. Reading the collected pages
+    // before it would race the fetches still in flight.
     await expect(contentRow).toHaveCount(1, { timeout: 15_000 });
     await expect(modeRows).toHaveCount(0);
+
+    const served = servedPages.flat();
+    // Non-vacuity, twice over: a page set that was never populated, or one
+    // that carried nothing, would satisfy "no mode row" for the wrong reason.
+    expect(servedPages.length).toBeGreaterThan(0);
+    expect(served.some((m) => m.kind === "privmsg")).toBe(true);
+    expect(served.filter((m) => m.kind === "mode")).toEqual([]);
 
     // --- $server is NOT denoised by the channel's pin ---------------------
     // The pin is per-channel (`"<slug> <channel>"`), and `$server` has no
--- a/cicchetto/e2e/tests/issue1262-mode-denoise.spec.ts
+++ b/cicchetto/e2e/tests/issue1262-mode-denoise.spec.ts
@@ -1,0 +1,127 @@
+// #1262 — channel MODE rows fold into the denoise filter, own-nick mode rows
+// on `$server` do not.
+//
+// `:mode` was carved OUT of the suppressed set by #458 on the rule "the broad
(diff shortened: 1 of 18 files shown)


vjt:
Rebased onto `3b5fbf20` (#1249 in). Two things worth recording, because one of them nearly shipped as a silent loss.

**The rebase ate the `DESIGN_NOTES.md` entry separator.** `rc=0`, 3/3 commits, zero conflicts, nothing on stderr — and three lines gone: the blank / `---` / blank that opens the entry. It is the known failure: a separator is three lines identical to three already upstream, so the rebase computes a matching patch-id and drops them as already-applied.

Caught by the two-sided check, not by reading the diff. The branch's contribution to that file was pinned to a file **before** the rebase at `81 additions / 0 deletions`; afterwards it read `78 / 0`. **The tell is the additions being down by three, not the deletions** — no line was deleted, three were never re-applied. The boundary shape on the file confirmed it independently: the previous entry's last line was immediately followed by this entry's heading.

Repaired by folding the three lines back **into the commit that introduces the entry** (`84788820`), via `--fixup` + `--autosquash`. Never as a commit of its own: a separator-only commit is exactly the shape the next rebase eats again. Post-repair, both sides re-checked — numstat back to `81 / 0` against the new base, boundary shape back to text / blank / `---` / blank / heading, and the 81 additions split `66 + 15` across the two commits that legitimately touch the file, with no third.

**The previous green CI attested a pre-#1249 base**, so it does not carry over; this push re-gates from scratch.

Still outstanding, unchanged: the e2e spec has passed in CI on a clean runner, but **the mutant has not been run** — the oracle has not been seen to fail, so it is not yet an oracle. That run is queued behind the e2e lane and will remove `:mode` from the suppressed set to confirm the spec dies.

vjt:
## Mutant run — and it found a hole in my own spec

Four runs on the e2e lane. Reporting the uncomfortable one first.

**Baseline, four specs, rebased branch:** all green — `issue1262` plus the three
neighbours that consume the same filter (`issue222-presence-filter`,
`issue458-presence-page-yield`, `issue239-hidden-msg-unread`), 4 passed in 26.3s.
The neighbours had never run on this branch before.

**Mutant 1, both sides** (`:mode` out of `@suppressed_presence_kinds` AND out of
`SUPPRESSED_PRESENCE_KINDS`): **killed**, on the right assertion —
`[data-kind="mode"]` expected 0, received 1, after the login/join/PUT barriers
had all passed. Not a barrier death.

**Mutant 2, server only** — and this is the finding. `message.ex` reverted, cic's
filter left intact: **the spec PASSED.** The render filter hides the row the
server is still sending, so the pane counts 0 in both worlds. The spec's
headline claim — that the row folds *server-side too* — had no live oracle, and
the green CI reported earlier was, for that half, true by construction. Mutant 1
had not exposed it: it killed on the FIRST assertion, the render one, and
Playwright stops there, so the server arm was never reached.

I had also written a comment in that arm asserting the opposite would happen
("cic, with only its own side moved, would render it"). That was an inference I
never measured, and it is wrong. Retracted, and replaced by what was measured.

**The fix moves the observation, it does not soften the claim.** The arm now
reads the cold-load `GET .../messages` response BODY and asserts no
`kind: "mode"` row is in it. Re-measured, in order:

- unmutated → **green**;
- server-only mutant → **killed**, on exactly that assertion, reporting the
  `+m` row the server served (`Received + 15`, the mode row with
  `meta.modes: "+m"`).

Two more vacuities surfaced while writing it, both of the "read nothing, pass"
family, and both caught by a red rather than by review: the first URL predicate
used `includes` and matched `/messages/count`, whose body is an object with no
`.some`; and awaiting a single response caught a page carrying no content row,
because a cold load issues more than one GET per channel. The arm now collects
every matching response, asserts the shape instead of trusting the cast, waits
on a DOM barrier before reading, and requires the content witness in the union.

The general rule is in DESIGN_NOTES: **when two layers suppress the same thing,
a test observing the OUTER layer cannot see the inner one fail.** Mutate per
layer; observe at the layer being claimed.

**Not claimed:** the three neighbours were run green *before* this spec edit.
The edit touches only `issue1262-mode-denoise.spec.ts`, so I do not expect them
to care — but I have not re-run them since, and I am not asserting that I have.
`bun run check` is rc=0 with 0 `error TS` after the edit.

Issue #1263, opened 2026-08-13
Per-worker build caches so the COMPILE lane can stop serialising workers


vjt:
## Problem

Two workers on the same host cannot compile or run Elixir gates at the same
time. The orchestrator serialises them with an exclusive `COMPILE` lane, so a
worker that is otherwise ready sits idle waiting for the other one's
`check.sh` / `mix` run to finish. That serialisation is real work lost, every
day, on every pair of concurrent tasks.

The lane is not arbitrary: it exists because the caches are shared **by
design**. `scripts/_lib.sh:36-40` resolves `REPO_ROOT` through
`--git-common-dir`, so every worktree binds the main repo's `_build`, `deps`
and `priv/plts`. Two concurrent `mix compile` runs against one `_build` are
exactly the contamination class already documented (a red naming a file the
branch never touched).

## Proposal

Make the cache root **opt-in per worker** instead of always-shared, so the
lane can be dropped when isolation is requested:

1. `scripts/_lib.sh`: honour an env knob (e.g. `GRAPPA_CACHE_ID`) that, when
   set, points `_build`/`deps`/`priv/plts` at a per-id directory under the
   main repo instead of the shared one. Unset = today's behaviour, byte for
   byte.
2. Test database: already solved upstream — `config/test.exs:4` builds the
   SQLite path from `MIX_TEST_PARTITION`, so `MIX_TEST_PARTITION=w1|w2` gives
   each worker its own `runtime/grappa_test*.db`. Nothing to change, just
   set it.
3. Docker/e2e (`STACK`) is a separate resource: the compose project name and
   host ports would have to be per-id too. Out of scope here unless someone
   wants both lanes gone at once — this issue is `COMPILE` only.

## Cost, stated honestly

Per-worker caches mean a full first compile per worker and a **separate
dialyzer PLT per worker** (the expensive one), plus the disk for two copies
of `_build`/`deps`. That is the price of parallelism; it is paid once per
worker, not per run.

## Acceptance

- With `GRAPPA_CACHE_ID` set differently in two worktrees, two concurrent
  `check.sh` runs both go green and neither sees the other's artefacts.
- With the knob unset, `_build`/`deps`/`priv/plts` resolve exactly as today.
- The orchestrator can then stop granting `COMPILE` exclusively.

vjt:
Shipped in **v1.1.0** — https://github.com/vjt/grappa-irc/releases/tag/v1.1.0

[...]

PR #1266, opened 2026-08-13
#1263 — opt-in per-worker build caches so the COMPILE lane can go

vjt:
## The shape

`GRAPPA_CACHE_ID=<id>` binds `_build`, `deps` and `priv/plts` from
`.caches/<id>/` instead of the shared main-repo tree. Unset, `CACHE_VOLUMES`
stays empty and expands to nothing — today's docker invocation, byte for byte,
which was the non-negotiable part of the ask.

The fix is at the same layer as the cause. The lane exists because `_lib.sh`
resolves `REPO_ROOT` through `--git-common-dir`, so every worktree binds one
`_build`; the lane is a mutex protecting a directory. The worktree source
overrides already establish "bind something else over `/app/x`" as how this
codebase redirects a container path, so the knob reuses that verb rather than
introducing machinery.

## Acceptance: PROVEN, with the oracle stated

Two `check.sh` runs, **two separate worktrees**, ids `w2a` and `w2b`, launched
two seconds apart, **both cold**:

| | run A (`w2a`) | run B (`w2b`) |
|---|---|---|
| exit | `rc=0` | `rc=0` |
| mix | `8 doctests, 59 properties, 5995 tests, 0 failures` | same |
| bats | 0 `not ok` | 0 `not ok` |

They overlapped for their entire run and finished five seconds apart.

Green is only half the claim. The other half — that neither run reached the
shared cache — is asserted by **displacement, not inspection**: the shared
`_build`/`deps`/`priv/plts` were timestamped before launch, and afterwards
`find -newermt` reported **0 files modified**, with the file count unchanged
at **8261**. Asserting what the runs *did* write would have passed even if
they had also written somewhere they should not; this cannot.

🔴 **No wall-clock number is offered, deliberately.** The only window in which
two cold gates ran had the full e2e suite running beside them on the same
host. The isolation claim is structural and survives that; any timing reading
would not. **In particular the dialyzer PLT BUILD TIME — the figure that would
justify a follow-up sharing just the PLT — remains unmeasured.** I am not
estimating it.

## What the issue got wrong, measured

Point 2 says the test DB is already solved — `config/test.exs` builds the
sqlite path from `MIX_TEST_PARTITION`, "nothing to change, just set it."

Half true, and the missing half is the one that matters: the *config* side
reads the variable, but **nothing on the shell side ever set it, and
`docker compose run` does not inherit host environment**. The evidence is a
grep with five hits and no sixth: `MIX_TEST_PARTITION` appears in
`config/test.exs` (twice — the sqlite path and the uploads dir), in
`OPERATIONS.md` and `DESIGN_NOTES.md` as prose, and **nowhere in any `.sh` or
in `compose.yaml`**. Setting it on the host was a silent no-op, and two
isolated `_build` trees would still have fought over one
`runtime/grappa_test.db`. So the knob derives the partition from the id and
forwards it with `-e`. An explicitly set partition still wins.

## Two decisions worth defending

**The knob forces a oneshot.** `in_container_or_oneshot` would otherwise exec
into the live container on a main checkout, and a running container cannot be
remounted — the caller would receive the shared cache while believing they
asked for isolation. Silent loss of the isolation is the one failure this must
not have. (This case sits outside the "a worktree always goes oneshot"
reasoning: it only bites from a *main* checkout with the stack up.)

**Nothing is seeded from the shared cache.** Cloning the warm `_build` and PLT
would make adoption nearly free (APFS `cp -c` is instant), but those artefacts
were compiled from another worktree's source, which is the exact contamination
class this issue exists to end. A fresh id is cold, deliberately.

The id is validated before any side effect — it becomes a directory name and a
`docker -v` argument, so a slash, a `..`, a leading dot or a shell
metacharacter is refused rather than sanitised.

## Cost, measured — and a corrected estimate

**113M per id**: `_build` 70M (dev + test), `deps` 32M, `priv/plts` 11M.

An earlier revision of this branch said ~209M. That number was read off the
*shared* tree, which also carries `prod` and PLTs left over from older OTP
versions — neither of which a fresh id reproduces. The estimate was wrong by
almost half and is corrected in the same docs it was written into.

## Gating

- Two full cold `check.sh` runs, both green (above). They also cover the
  `CLAUDE.md` edit: that commit predates the launch and both trees were clean.
- `scripts/bats.sh` full set on this branch: **rc=0, 458 `ok`, 0 `not ok`**,
  shellcheck gate included.
- Ten new cases in `test/scripts/cache_id_isolation_test.bats`. Red read before
  green: 9 of 10 failed, and the one that passed was the knob-unset invariance
  case, which had to pass before the change.
- The forced oneshot is pinned by a **pair**: one case asserts no exec happens,
  the other is the reference box proving the fixture can reach the exec branch
  at all. It was needed — the fixture initially passed for the wrong reason,
  because on macOS `\$TMPDIR` resolves through a symlink and the fake main
  checkout was silently classified as a worktree.
- The final commit is docs-only (`OPERATIONS.md`, `DESIGN_NOTES.md`) and
  postdates those runs. No test reads either file (checked).

## Scope held

The docker stack — compose project name, host ports — is untouched, so
`integration.sh` and e2e remain single-occupancy and the `STACK` lane survives.
This kills one lane, not both. Concurrent `deps.get` under two ids is also
**not** proven: the hex/rebar caches (`.hex`, `.mix`) stay shared, so the two
were run sequentially during setup.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
 # ── Elixir / Erlang
 /_build/
+# #1263 — per-worker build caches (GRAPPA_CACHE_ID): one
+# _build/deps/priv/plts triplet per id, opt-in, never shared.
+/.caches/
 /cover/
 /coverage/
 /deps/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -778,6 +778,13 @@ is due. Don't just look at todo.md.
   a forced rebuild (`scripts/mix.sh --env=dev compile --force`) before you
   believe it, and say in the report which of the two you established. The
   COMPILE lane serialises access; it does NOT make the artefacts yours.
+  **Opt out with `GRAPPA_CACHE_ID=<id>` (#1263):** the three caches then come
+  from `.caches/<id>/` and `MIX_TEST_PARTITION` follows the id, so two ids can
+  run `mix` concurrently. Unset changes nothing. A fresh id is COLD — full
+  compile plus its own dialyzer PLT, nothing seeded (seeding would import the
+  contamination described above). Does NOT isolate the docker stack: compose
+  project and ports are still shared, so `integration.sh` / e2e stay
+  single-occupancy. See `docs/OPERATIONS.md`.
 - **NEVER install hex packages on the host.** Add them to `mix.exs`,
   rebuild the image (`scripts/mix.sh deps.get`).
 - **Don't touch the IRC parser without re-running parser tests.**
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -40725,10 +40725,28 @@ operator may be partitioning for an unrelated reason.
   `cp -c` clone is instant — but those artefacts were compiled from another
   worktree's source, which is precisely the contamination class this issue
   exists to end. A fresh id starts empty and pays a full compile plus its own
-  PLT. Cost measured on this host: ~209M per id (`_build` 148M, `deps` 33M,
-  `priv/plts` 28M), disk that is free at 885G available. The PLT *build time*
-  is the number that would justify revisiting this, and it has not been
-  measured yet.
+  PLT. **The disk cost was over-estimated before it was measured**: reading
+  the shared tree gave ~209M per id, but a real cold `check.sh` produces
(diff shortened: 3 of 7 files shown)


vjt:

Issue #1264, opened 2026-08-13
The !addquote context-menu entry must carry the sender nick in the payload


vjt:
## What changes

The `!addquote` entry in the message context menu must put the **sender's
nick** in the payload, not just the message text.

Today (`cicchetto/src/lib/addQuote.ts:26-29`) the command is the bare body:

```
!addquote ciao mondo
```

Wanted:

```
!addquote <vjt> ciao mondo
```

## This reverses the #1107 ruling — deliberately

The bare-body shape is not an accident, it is a ruling with its reasoning
written down in `cicchetto/src/lib/addQuote.ts:10-16` and pinned by an
assertion in `cicchetto/src/__tests__/addQuote.test.ts:51`, with the rationale
restated at `:12-16` of the test. The argument then was that a bot storing its
input verbatim would bake `<nick>` into every stored quote.

vjt has now ruled the other way: the attribution belongs in the payload. So
the comment block and the pinned test are part of the change — flip them, do
not work around them, and leave the new reason in place of the old one so the
next reader is not misled.

## Scope

- `addQuoteCommand` prefixes `<sender>` before the body.
- ACTION rows: the current code deliberately strips the `* nick` form
  (`addQuote.ts:23-25`). Decide one shape and pin it — an ACTION with
  attribution in a different position from a PRIVMSG is exactly the
  unpredictability the old comment warned about.
- Rows with nothing to quote keep refusing (`quotableBody` gate, unchanged).
- Update the tests that assert the old payload
  (`__tests__/addQuote.test.ts`, `__tests__/MessageContextMenu.test.tsx:158`).

## Acceptance

- Clicking `!addquote` on a PRIVMSG from `vjt` fills the compose box with
  `!addquote <vjt> ` plus the body, caret at the end and revealed.
- The old bare-body assertions are gone, not merely relaxed, and the file's
  comment states the current rule.

vjt:
Ruling from vjt (2026-08-13, on IRC): the bare-body shape was a mistake. The
payload carries the nick, and the open question about ACTION rows is closed —
an ACTION keeps its rendered form.

So:

- PRIVMSG / NOTICE → `!addquote <vjt> ciao mondo`
- ACTION → `!addquote * vjt pees over the fence`

That is the same text the scrollback shows for each kind, which is the point:
what gets quoted is what the operator read. `addQuote.ts:23-25` currently
strips the `* nick` form on purpose — that stripping goes, and the comment
explaining it goes with it.

[...]

Issue #1126, opened 2026-08-09
Reply to an /me quotes the raw CTCP envelope and re-emits \x01 on the wire


vjt:
## What happens

Replying to an `/me` line quotes the raw CTCP wire form instead of the action the user actually sees.

Observed on `#it-opers` today, replying to `* Fairy si dà alla fuga`. What went out on the wire (captured from an independent IRC client's raw log, `cat -v`):

```
PRIVMSG #it-opers :<Fairy> ^AACTION si dM-CM- alla fuga^A<< avevo letti "alla figa"
```

Two distinct defects in one line:

1. **Cosmetic** — the quote reads `<Fairy> ACTION si dà alla fuga<< `, i.e. the CTCP verb leaks into the visible text. The reader of that message never saw the string `ACTION`; they saw `* Fairy si dà alla fuga`.
2. **Protocol** — the two `\x01` delimiters (`^A` above) are re-emitted verbatim inside an ordinary PRIVMSG. We are putting CTCP framing bytes in the middle of a non-CTCP message. Strict clients may treat the payload as a malformed/embedded CTCP; at minimum it is a byte we generated and never intended to send.

## Why (read on `origin/main`, not from memory)

- `cicchetto/src/lib/replyQuote.ts:23-31` builds the quote from `mircPlainText(msg.body ?? "")`. `mircPlainText` deliberately leaves `\x01` alone — see `cicchetto/src/__tests__/mircFormat.test.ts:32-34`, which asserts `parseMircFormat("\x01ACTION waves\x01")` yields a single run with the bytes intact. So the envelope survives into the compose box, and from there onto the wire.
- The render layer *does* unwrap it: `stripCtcpAction` in `cicchetto/src/ScrollbackPane.tsx:460-467` (`CTCP_ACTION_PREFIX = "\x01ACTION "`), covered by `cicchetto/src/__tests__/ScrollbackPane.test.tsx:398-422`. But that helper is a module-local `const` in `ScrollbackPane.tsx` — it is not exported, so `replyQuote` has no access to it and the two layers disagree about what an action's text is.
- `isContentKind` (`cicchetto/src/lib/api.ts:660`) includes `action`, which is correct — an `/me` *is* conversation and *should* be quotable. The bug is the shape of the quote, not the fact that it quotes.

## What the fix has to decide

[...]

Issue #1078, opened 2026-08-08
e2e: a far-behind re-seeding spec poisons one downstream victim, and which victim depends on run position


vjt:
Measured 2026-08-08 on the dev host: five full `scripts/integration.sh` runs (660 tests, 1 worker,
chromium, deterministic file order), one dedicated worktree per side, nothing else on the box.

## The claim

A "far-behind" e2e spec — one that calls `resetSubject` and re-seeds ~260 rows into `#bofh` with a
planted mid-list read cursor — leaves state behind that **reddens exactly one victim downstream of
itself**, and *which* victim it reddens depends on **where the spec sorts in the run order**.

## The measurements

| run | tree | re-seeding spec position | reds |
|---|---|---|---|
| A1 | `40c5576b` | #380 | `p0b-peer-away:30` @ #443, `scroll-on-window-switch:410` @ #477 |
| B  | `49804a8e` | #91  | `issue168-scroll-authority:169` @ #114, `issue625-single-send-scroll-jump:227` @ #328 |
| A2 | `40c5576b` | #380 | `p0b-peer-away:30` @ #443 |
| C  | `49804a8e`, **spec renamed only** | #380 | `p0b-peer-away:30` @ #443 |
| D  | `49804a8e`, **spec deleted** | absent | *(neither victim — see below)* |

Anti-correlated 5/5 across every observation:

* spec @ **#91** → `issue625:227` RED, `p0b` green
* spec @ **#380** → `p0b:30` RED, `issue625:227` green

## Why this is displacement, not correlation

**Run C carries the suspect production diff in full, byte-identical to run B's tree, and moves ONE
variable: the spec's filename**, so it sorts back where the previous far-behind spec used to sit
(#380 instead of #91). No assert weakened, no timeout bumped, no product code touched.
**The red moved with it** — run C reproduced the baseline profile exactly, with the suspect code present.

**Run D removes the spec entirely.** Prediction registered *before* the run: `issue625:227` GREEN and
`p0b-peer-away:30` GREEN. Both held (659 tests; `issue625` green @ #327, `p0b` green @ #442).
Present-and-upstream → red; moved downstream → red relocates; removed → both clear. Three ways.

## Consequences worth writing down

* **An iso-rerun does NOT clear a red of this class.** Alone in a fresh stack there is no poisoner
  upstream, so 3/3 green in isolation says nothing. That is exactly what happened when one of these
  reds was iso-rerun ×3.
[...]

PR #1269, opened 2026-08-13
#1264 — !addquote carries the sender, reversing the #1107 payload ruling

vjt:
## The shape, as ruled

* PRIVMSG / NOTICE → `!addquote <vjt> ciao mondo`
* ACTION → `!addquote * vjt pees over the fence`

vjt's comment on the issue closes the open question about ACTION rows: an action
keeps its rendered form. The heads therefore **differ between kinds**, and under
the new rule that is the predictable shape rather than an exception to it —
#1107 read a difference between kinds as unpredictability, but each of these is
the line its own row showed. #1126 had already ruled identically for Reply, from
the other side: rendering `* vjt waves` as `<vjt> waves` puts a sentence in
someone's mouth.

## The reversal went through the pin, not around it

#1107 did not just implement the bare body, it argued for it in a comment block
and pinned it with `carries no nick — the payload is the bare body`. Both are
part of the change:

- the assertion is **replaced by its inverse**, not deleted — a rule with no pin
  is a rule the next reshaping takes silently, which is the part #1107 got right;
- the comment blocks in `addQuote.ts` and at the head of its test are **rewritten
  in place**, so the next reader finds the current rule where the withdrawn one
  used to be, rather than a gap.

The new reasoning, stated so it can be argued with later: the old tiebreak was
"a bot that wants attribution can read the nick off the channel". That holds at
the moment the command arrives and not one second afterwards — and a quote is a
thing you recall months later, when that context is gone.

## One refactor, and why it is not scope creep

`attributionHead` moves into `lib/quotableBody.ts` and `replyQuote` loses its
inline copy. That module exists **because** #1107 argued that duplicating quote
rules would let the two doors drift on the next ruling — and the head has just
become one of those rules, identical on both sides. Left duplicated, the next
kind gets taught to one door only. `replyQuote`'s output is unchanged, and its
own suite still asserts it.

One of the new tests asserts the two doors agree by comparing against
`replyQuote`'s head rather than against a literal, so a future divergence fails
instead of quietly passing on both sides.

## Gating

- Red read before green: **12 assertions failed** across `addQuote.test.ts` and
  `MessageContextMenu.test.tsx`, all of them payload-shape. The 12 that passed
  were the refusal gates (`quotableBody`), which this change does not touch —
  the failure set matching the blast radius is itself the check that the change
  is where I think it is.
- `bun run test`: **rc=0, 281 files / 5332 tests passed**.
- `bun run check`: **rc=0**, 1013 files, 60 warnings / 5 infos — the unchanged
  pre-existing baseline, 0 `error TS`.
- No lane used: this is cic-only.

## The e2e — extended, NOT run

`cicchetto/e2e/tests/issue1107-addquote-menu-item.spec.ts` had three assertions
of the old bare-body payload. A spec asserting the withdrawn world is rewritten
onto the decided behaviour, not left to fail and not softened: all three now
expect `!addquote <${specNick()}> ${body}`, and the spec's header comment says
why.

🔴 **I have not executed it.** It passes the static gate only, and a static gate
says nothing about whether the assertion holds in a browser. No new e2e was
added — this is a string-shape change on an existing feature whose outcome is
the compose box's contents, which the component tests already observe; a new
spec would have claimed a stack for a string.

One risk I am naming rather than hiding: the expected nick is `specNick()`. That
is sound here because #1078 gives every spec its own subject, derived from a
sha1 of the title path, and a collision fails loudly at provision rather than
silently renaming — but the assertion does now depend on the live nick matching
the provisioned one, which it did not before.

### DIFF
--- a/cicchetto/e2e/tests/issue1107-addquote-menu-item.spec.ts
+++ b/cicchetto/e2e/tests/issue1107-addquote-menu-item.spec.ts
@@ -1,6 +1,6 @@
-// #1107 — the message menu's `!addquote` item puts `!addquote ` plus the
-// message text into the compose box, with the caret at the end and visible,
-// and sends NOTHING. The command is for whatever bot is in the channel; cic
+// #1107 — the message menu's `!addquote` item puts `!addquote `, the sender's
+// `<nick>` head (#1264) and the message text into the compose box, with the
+// caret at the end and visible, and sends NOTHING. The command is for whatever bot is in the channel; cic
 // only fills the box.
 //
 // Harness + limits:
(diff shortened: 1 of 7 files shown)


vjt:

Issue #1277, opened 2026-08-13
cic(reply): raise the quoted-body cap from 42 to 100 characters


vjt:
vjt on #grappa: «aumentiamo i char del reply da 42 a 100 va». The #1235 cap is too tight in practice — raise `REPLY_QUOTE_BODY_LIMIT` from 42 to 100.

Only the number changes. Every reading ruled on in #1235 stays exactly as it is:

- 100 counts BODY code points and the ellipsis is **added past** them (a capped quote runs 103, not 100 with three spent on dots);
- the marker stays the literal `...`, not `…` U+2026;
- the cut stays **flat** at the limit, with no backing off to a word boundary;
- the head (`<nick> ` / `* nick `) stays outside the count.

## Sites (measured on `origin/main`, `6d466227`)

1. `cicchetto/src/lib/replyQuote.ts:29` — `export const REPLY_QUOTE_BODY_LIMIT = 42;`. The one-line edit the #1235 ruling was written to make cheap. The comment block above it at `:17-28` quotes the old request verbatim («limitiamo a 42 i caratteri») and needs rewording so the file does not argue for a number it no longer implements — keep the four rulings, restate them against 100 and note that 42 was the original ask.
2. `cicchetto/e2e/tests/issue1105-reply-quote-caret-visible.spec.ts:64` — `const QUOTED_BODY_LIMIT = 42;`, a **deliberate mirror**: the e2e package does not import from `src`, so the house convention is a mirrored constant plus a lockstep comment (`:61-63`). It must move in the same commit or the spec's own arithmetic goes wrong. The prose at `:53-56` also states the capped length ("`<nick> ` + 45 + ` << `, about two [lines]") — restate for 103.
3. `cicchetto/src/__tests__/replyQuote.test.ts` — the asserts already use the constant (`:224`, `:241-242`), so they follow by themselves; the **prose** does not. Test name at `:238` ("keeps a full 42 characters…") and comments at `:218`, `:236-237`, `:255`, `:277` all name 42. A test that says 42 while asserting 100 is worse than no comment.

## The one thing to verify rather than assume

[...]

Issue #246, opened 2026-07-14
Message splitting silently drops bytes at each split boundary (source prefix not reserved in 512 budget)


vjt:
## Summary

Outbound message splitting silently **drops bytes at every split boundary** when a `PRIVMSG`/`NOTICE` is long enough to require splitting. The loss is silent: no error is surfaced, and because the cut lands mid-token the client display looks plausible, so the corruption is easy to miss.

## Repro

Send a single `PRIVMSG` to a channel whose UTF-8 payload is 600 bytes:

```
"ABCDEFGH IJKLMNOP QRSTUVWX YZ " repeated 20 times   # 30 bytes × 20 = 600
```

grappa splits it into two lines on the wire. Captured the raw traffic on a receiving client and reconstructed:

| part | delivered length | starts at payload offset |
|------|------------------|--------------------------|
| 1    | 468 bytes        | 0                        |
| 2    | 106 bytes        | **493** (should be 468)  |

- `part1` is a clean prefix: `payload[0:468]` exactly.
- `part2` resumes at payload offset **493**, not 468 → payload bytes `[468, 493)` (25 bytes: `QRSTUVWX YZ ABCDEFGH IJKL`) are **silently eaten**.
- The final trailing byte is also dropped.
- Total: **26 of 600 bytes lost**, reconstructed length 574.

## Root cause

`468 = 512 − 44`, where 44 is the framing overhead of the **relayed** line:

```
:nick!user@host PRIVMSG #channel :<text>\r\n
└──── 44 bytes of overhead ────┘         (25 source prefix + 17 command/target + 2 CRLF)
```

The splitter budgets the 512-byte limit against the message text (or against only the client→server framing `PRIVMSG #channel :…`), **without reserving space for the source prefix `:nick!user@host ` that the server prepends when relaying to other clients.**

Consequence:
- The client→server line is ≤ 512, so it passes validation and is accepted.
- The **relayed** line (with the source prefix added) exceeds 512, so the server truncates its tail at the wire limit — here to 468 bytes of text.
- The next chunk resumes at the splitter's *intended* (pre-truncation) offset, so a hole of `intended_offset − wire_limit` bytes opens at each boundary.

This is why it "looks fine on screen": no error is raised, and the mid-token cut reads as continuous.

## Fix

When computing the split budget, reserve the **worst-case relayed framing**, not just the client-side command:

```
budget = 512
       − len(":" + nick + "!" + user + "@" + host + " ")   # source prefix, worst case
[...]

PR #1278, opened 2026-08-13
cic(#1277): raise the reply quote cap from 42 to 100

vjt:
Raises `REPLY_QUOTE_BODY_LIMIT` from 42 to 100. Refs #1277.

Only the number moves. Every reading ruled on in #1235 stands: 100 counts BODY
code points and the ellipsis is added past them (a capped quote runs 103), the
marker stays the literal `...`, the cut stays flat at the limit with no backing
off to a word boundary, and the head (`<nick> ` / `* nick `) stays outside the
count. Prose was rewritten wherever it argued for 42, in the source comment and
in both test files.

## Sites

- `src/lib/replyQuote.ts` — the constant plus its comment block.
- `e2e/tests/issue1105-reply-quote-caret-visible.spec.ts` — the deliberate
  mirrored constant (the e2e package does not import from `src`) and its
  lockstep prose, same commit.
- `src/__tests__/replyQuote.test.ts` — prose, plus two tests that build their
  bodies against the number rather than the constant.

## Measured, not assumed

- **The two hardcoded tests really do break.** Kept at their 42-sized bodies
  they go red against a 100 cap (2 failed / 38 passed) rather than passing
  vacuously; re-sized against 100 and run with a deliberate 99 cap, they are the
  exactly two tests that fail. The issue's site list expected only prose here.
- **The e2e fixture is left alone.** `FILLER` clears the new cap by ~37
  characters where it cleared the old one by ~95, so the run answered instead of
  the arithmetic: three stacked quotes measure `scrollHeight` 269 against
  `clientHeight` 42 at the 390px viewport — 227px of overflow against the spec's
  `MIN_OVERFLOW_PX` guard of 40, where the 42-char cap produced 109px. The guard
  does not fire, so nothing was lengthened; lengthening would also have pushed
  the posted body toward the #246 split budget for no gain.
- **Compose room at 390px.** One quote now occupies three wrapped lines of the
  `rows=1` box (`scrollHeight` 112) where it occupied two (73). The caret stays
  in view — that is what the spec asserts — but the quote is one line taller
  than before per reply.

## Test plan

- [x] `bun run test` — 282 files / 5352 tests green
- [x] `bun run check` — biome + tsc over `src` and `e2e`, RC 0
- [x] `scripts/integration.sh --grep issue1105` — passes on the committed state
- [ ] full integration suite in CI

### DIFF
--- a/cicchetto/e2e/tests/issue1105-reply-quote-caret-visible.spec.ts
+++ b/cicchetto/e2e/tests/issue1105-reply-quote-caret-visible.spec.ts
@@ -45,14 +45,16 @@ const CHANNEL = AUTOJOIN_CHANNELS[0];
 // already wraps and already hides the caret. This body is far past it, so the
 // ROW wraps and the quote it produces is a capped one, while staying well
 // inside one PRIVMSG so the server-side split budget (#246) never turns it
-// into two scrollback rows.
+// into two scrollback rows. #1277 raised the cap to 100 and left the filler
+// alone: the posted body runs 161 code points with the timestamp, still 61
+// past the cap, and the overflow below was measured rather than assumed.
 const FILLER =
   "che va a capo parecchie volte perche' il textarea e' rows=1 e non cresce, " +
   "quindi il caret finisce sotto la piega e non si vede piu' nulla";
 
-// #1235 capped the quoted body at 42 characters plus a literal `...`, so a
-// single reply can no longer BE six wrapped lines: the longest quote the
-// gesture can now produce is `<nick> ` + 45 + ` << `, about two. The overflow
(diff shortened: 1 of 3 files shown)


vjt:
Ruling from the project owner on #grappa, recorded here so the tradeoff is not re-litigated: **100 stands.**

The number was asked for before the cost was known, so it was put back to him with the measurement from this PR — at a 390px viewport one quote now occupies three wrapped lines of the `rows=1` compose box (`scrollHeight` 112) where it occupied two (73), caret still in view. Answer: fine as is, no intermediate value.

Nothing to change in the diff — this only removes the open question above it.

vjt:

Landed as: `275950d1`.

Issue #1192, opened 2026-08-10
cic: CTCP submenu in the nick context menu


vjt:
Requested on #grappa: a CTCP entry in the nick context menu — starting from VERSION, then widened by the reporter to "all the verbs, maybe in a CTCP submenu".

## What already exists (measured on `origin/main` `8c8b7c79`)

* The verb works when typed: outbound `/ctcp` self-echoes and renders as `→ CTCP VERB [args] to <target>` in `cicchetto/src/ScrollbackPane.tsx:693-718`, off typed `meta.ctcp_verb` / `meta.ctcp_args` (server-side SSOT `Grappa.IRC.CTCP.verb_args/1`).
* The nick context menu is `cicchetto/src/UserContextMenu.tsx` — it stops at WHOIS (`:76`) and Query (`:82`).

## What is missing

Only the menu affordance. Same shape as the WHOWAS request: shipped machinery, no entry point.

## Design note

A submenu keeps the nick menu from growing by six items. Useful verbs are few: VERSION, TIME, PING (latency), CLIENTINFO, USERINFO, SOURCE.

Two properties worth writing down before anyone implements this:

* CTCP replies come back as **asynchronous NOTICEs**, outside the WHOIS numeric burst. A "power WHOIS" that folds client info into the WHOIS card would have to correlate them after the fact, with no guarantee they ever arrive.
* Clients that suppress CTCP (bots, hardened setups) simply never answer, so any field built on this must degrade to empty rather than spin.

Neither VERSION nor CLIENTINFO is verifiable: both are strings the remote client chooses for itself. VERSION is a free-form name/version; CLIENTINFO is the list of CTCP verbs that client answers — capability discovery, not identity.

## Scope note

Filed under the feature freeze as a record, deliberately **not** queued.

vjt:
**Ground measured on `305e2df8` before writing any code — plus one finding that decides whether this feature is worth shipping in the shape the issue describes.**

- `UserContextMenu.tsx` is only the item list (8 entries). The chrome — portal, backdrop, Escape, the measured #487 clamp and the #949 safe-area — lives in `ContextMenu.tsx`, shared with `MessageContextMenu` (#1067). `ContextMenuItem` is `{label, enabled, action}`: **flat, with no nesting**.
- Both mount sites (`MembersPane.tsx:205`, `ScrollbackPane.tsx:3891`) pass `channelName` = the **source** window, which is exactly the key a CTCP wants.
[...]

Issue #719, opened 2026-08-03
cicchetto: non-PING CTCP replies land in the server window, not the window the query was sent from


vjt:
## Problem

Send `/ctcp <nick> VERSION` (or TIME, or anything that is not PING) from a query window: the request self-echo renders in that window correctly (#640), but the **reply** renders in the **server window** instead. The two halves of one round trip end up in two different places.

`/ping` does not have the problem — its reply comes back to the window it was typed in.

Reported on IRC.

## Why (traced on `origin/main` @ `42ee2013`)

Two independent mechanisms, and only one of them covers the general case:

1. **Server side, by design.** `route_non_channel_notice/3` in `lib/grappa/session/event_router.ex:2546` routes *any* CTCP-framed NOTICE to `$server`:

   ```elixir
   if CTCP.framed?(body) do
     {"$server", body}
   ```

   That is #591 doing what it was asked to do: before it, a CTCP reply took the regular-nick branch, persisted under the peer, and `maybe_open_query_window/2` minted a tab — pinging somebody left a window open with a row of control characters in it. The comment says as much, and adds that the body stays verbatim so *a client can recognise its own outstanding query and report the round trip*.

2. **Client side, PING only.** That client-side half exists exactly once: `cicchetto/src/lib/pingCorrelation.ts` registers `(networkId, folded nick, token)` → source window at send time (`compose.ts` `case "ping"`), and `subscribe.ts` resolves the reply back to the source window and synthesizes the RTT line there.

`compose.ts` `case "ctcp"` registers **nothing**. So for every verb other than PING there is no correlation entry, nothing claims the reply, and it renders under its normal `$server` routing — which is the observed behaviour, not a regression.

## Wanted

Extend the correlation table to any CTCP verb, rather than moving the routing decision back into the server:

- register `(networkId, folded target nick, verb)` → source window on `/ctcp <target> <VERB> [args]`, same shape and TTL sweep as `registerPing`
- on a `:notice` carrying `meta.ctcp_verb`, resolve against that table first; a claim renders the reply in the source window, no claim leaves the row exactly where it is today
- keep PING's token-keyed entry as-is — it needs the token for the RTT; the generic entry keys on the verb because non-PING replies echo no token

Keeping the fix client-side preserves what #591 bought (no auto-minted query window for a peer who merely answered a probe), and matches irssi, where the reply is shown in the window the question was asked from.

## Notes

- Non-goal: routing CTCP replies to a query window server-side. That reopens #591.
[...]

Issue #641, opened 2026-08-02
cicchetto: uncorrelated CTCP notice renders raw delimiters in the window (notice arm has no ctcp_verb branch)


vjt:
An inbound CTCP reply that matches no pending `/ping` renders its raw body, `\x01` delimiters included, straight into the window. The "one IRC parser, on the server / cic never shows `\x01`" invariant holds for the privmsg arm only.

### Measured (staging `grappa.bad.ass`, `0.9.0-654f158f`, browser)

`/ping NickServ` from `#ccping-test`. The `$server` window shows, literally (`^A` below stands for U+0001, which is present verbatim in the DOM):

```
09:39:08 -NickServ- ^APING^A
```

The service does answer, and the framing is correct on the wire — captured independently:

```
> PRIVMSG NickServ :^APING 1785644999^A
< NickServ NOTICE :^APING^A
```

NickServ does not echo the token back, so `maybeConsumePingReply` finds nothing to correlate and returns `false` (correct, and tracked in #637). The fall-through is the defect: the unmatched reply is appended and rendered raw.

### Why

`ScrollbackPane.tsx` reads `msg.meta.ctcp_verb` only in `case "privmsg"`, where a typed verb produces the `-> CTCP VERB [args] to <target>` line. `case "notice"` has no such branch, so a server-classified CTCP notice falls through to the generic notice rendering and prints its body verbatim, delimiters and all. The server has already classified it — `meta.ctcp_verb: "PING"` is present on the row; nothing needs to parse the delimiters to fix this.

### Expected

A CTCP-classified notice renders as a human line built from the typed meta (`<- CTCP PING reply from NickServ`, or whatever wording the display layer settles on), and U+0001 never reaches the screen. Independent of whether the reply correlates to a pending `/ping`.

### Notes

Same shape as #14, for the notice path.

Distinct from #637: #637 is "no RTT line because the token is missing"; this is "the uncorrelated reply is shown raw". Fixing #637's correlation would hide this particular instance while leaving every other unmatched CTCP notice (a stray `VERSION` reply, a `TIME` reply from a client that answers late) rendering raw.

vjt:
Shipped in **v0.11.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.11.0

The tag carries `.deb`, `.rpm` and `pkg.tar.zst` assets, so self-hosters have something to pull.

Note this is a **release, not a deploy**: the hosted service is unchanged and still runs the previous build until the code-review queue closes.

PR #1279, opened 2026-08-13
cic(#719): correlate every CTCP verb, not just PING

vjt:
## What

`/ctcp <nick> VERSION` (or TIME, or anything that is not PING) sent from a
query window echoed its **question** in that window (#640) and rendered its
**answer** in `$server`. `/ping` did not, because PING was the only verb
anything registered a correlation entry for.

The correlation table now covers every verb. The reply lands in the window the
query was sent from; an uncorrelated reply keeps falling back to `$server`.

## The issue's premise had moved — two corrections

**1. The fix does not belong where the issue says.** #719 was traced at
`42ee2013` and states that `compose.ts` `case "ctcp"` "registers nothing".
That is no longer the shape of the code: #1192 moved `/ctcp`, `/ping` *and*
the nick context menu's CTCP submenu onto one seam, `sendCtcpQuery`
(`lib/ctcpQuery.ts`), which owns the #600 register-before-send ordering — and
registers PING only. Registering in `compose.ts` as written would have missed
the menu caller entirely, so the branch went in the seam.

**2. shottino does not do what the issue says it does.** The Notes cite
`event_router.ex:2745` — "shottino renders it as a card in the window the
question was asked from" — and use it to argue the clients would agree. That
sentence is wrong, measured:

- `card()` (`frontends/shottino/shottino.c:7429`) files a row under the window
  that has **FOCUS when the answer arrives**, falling back to that network's
  `$server` when the focused window belongs to another network.
- It does this for **every verb uniformly** (`render_ctcp_reply`, `:3645`
  claimed-PING, `:3656` uncorrelated-PING, `:3660`/`:3661` any other verb).
- Its `(network, nick, stamp)` table holds **no source window**: `ping_claim`
  (`:3455`) returns a bool, and all it decides is RTT-vs-plain-line and
  backfill suppression.
- The card block's own comment (`:7410`) is accurate where the event_router one
  is not: *"the window that had focus when the answer arrived"*.

So the two clients now follow different rules. They coincide in the common case
(ask, stay put) and diverge when the operator switches windows while a reply is
in flight: cic files it where the question was asked, shottino where the
operator is looking. Recorded in DESIGN_NOTES rather than papered over. **The
`event_router.ex:2745` sentence is left wrong on purpose** — a server-tree
comment fix is not this PR's business, and it needs a compile lane.

## Shape

- `pingCorrelation.ts` — a second map keyed `(network, ASCII-folded nick,
  upper-cased VERB)`. Two maps, not one three-part key: a PING token is opaque
  operator text and could *be* a verb, so `/ping frank VERSION` and a VERSION
  reply from frank would otherwise fight over one entry (pinned in both
  directions). Same identity-scoped store, same #637 TTL sweep — now run over
  both maps from either register, so a session that only ever runs `/ctcp`
  still bounds the ping side. No RTT: there is no token to time against.
- `ctcpQuery.ts` — PING keeps its token entry, every other verb takes a verb
  entry, both before the send is awaited (#600).
- `subscribe.ts` — `maybeConsumePingReply` becomes `maybeConsumeCtcpReply`. A
  claimed non-PING reply is **re-keyed, not rewritten**: it keeps its body and
  its typed ctcp meta so `ScrollbackPane`'s notice arm draws
  `← CTCP VERSION reply from <peer>` in the source window. The PING arm strips
  meta because it *synthesizes* an RTT body; stripping here would drop the row
  into the generic body render and leak raw `\x01` (#641).

Design constraints from the issue, all held: the routing decision stays
client-side (`route_non_channel_notice/3` untouched, #591 intact); an
uncorrelated reply still falls back to `$server`; PING's token entry is
unchanged; same shape and TTL sweep as `registerPing`.

## One deliberate consequence beyond the report

The verb now folds on the **resolve** side too. The send side already folded
(`sendCtcpQuery` compares `verb.toUpperCase()`), so a case-sensitive compare on
the way back was a standing asymmetry: a peer answering `ping` in lower case
missed a token-table entry it had every right to. One fold, both ends —
covered by its own test.

## Test plan

- [x] Load-bearing red first: `#719 — a non-PING CTCP reply claiming a pending
  query renders in the SOURCE window` failed with the source window `undefined`
  before the fix, while its sibling (`...claiming NOTHING stays in $server`)
  passed pre-fix — the fallback is preserved behaviour, not new behaviour.
- [x] Three mutants, each killing exactly one test and no others: drop the verb
  fold → only the lower-cased-PING test dies; strip meta on the claimed branch
  → only the source-window test dies; claim every reply regardless → only the
  `$server`-fallback test dies. Source restored green after each.
- [x] `scripts/bun.sh run check` — RC=0 (biome + `tsc --noEmit` + the e2e
  tsconfig; 60 warnings, all pre-existing, none on a touched file).
- [x] `scripts/bun.sh run test` — RC=0, 282 files / 5366 tests. The full suite
  caught a regression the scoped run could not: `compose.test.ts` replaces the
  whole `pingCorrelation` module, so its mock had to learn the new register or
  the seam threw and took the send with it.
- [x] `scripts/design-notes-gate.sh` — RC=0, "1 new entry heading(s), separator
  and marker present" (run against committed state; it reports "nothing to
  check" on an uncommitted tree).
- [ ] No e2e written. The visible outcome is which window a row renders in,
  which the subscribe-level tests assert against the real store; an e2e would
  need a peer that answers CTCP VERSION, and bahamut does not.

## Not verified

- No compile or stack lane was used, so nothing server-side was exercised. The
  shottino findings are from reading its source, not from running it.

### DIFF
--- a/cicchetto/src/ScrollbackPane.tsx
+++ b/cicchetto/src/ScrollbackPane.tsx
@@ -775,9 +775,16 @@ const renderBody = (msg: ScrollbackMessage, handlers: NickHandlers): JSX.Element
       // body is the raw \x01VERB [args]\x01 frame, classified ONCE on the server
       // (SSOT Grappa.IRC.CTCP.verb_args/1) into typed meta.ctcp_verb /
       // meta.ctcp_args. A CORRELATED /ping reply is consumed upstream
-      // (subscribe.ts maybeConsumePingReply) and never reaches here; an
-      // UNCORRELATED CTCP reply (a stray VERSION/TIME, or a token-less PING that
-      // matched no pending /ping) does. Before this arm it fell through to the
+      // (subscribe.ts maybeConsumeCtcpReply) and never reaches here, because that
+      // gate SYNTHESIZES an RTT line and strips the meta. Everything else does
+      // reach here: an UNCORRELATED reply (a stray VERSION/TIME, or a token-less
+      // PING that matched no pending /ping) in its `$server` routing, and — since
+      // #719 — a CORRELATED non-PING reply too, re-keyed to the window the query
+      // was sent from with its meta deliberately INTACT so this arm can draw it.
+      // That is why the gate must not strip meta on the #719 branch: doing so
+      // would drop the row into the generic body render below and leak raw \x01,
+      // which is the very thing this arm exists to prevent.
+      // Before this arm it fell through to the
       // generic body render below and leaked the raw \x01 into the DOM. Render a
       // human INBOUND line from the TYPED meta instead — cic NEVER parses \x01
       // (the "one IRC parser, on the server" invariant). This is the notice twin
--- a/cicchetto/src/__tests__/ScrollbackPane.test.tsx
+++ b/cicchetto/src/__tests__/ScrollbackPane.test.tsx
@@ -660,8 +660,9 @@ describe("ScrollbackPane", () => {
 
   it("renders an uncorrelated inbound CTCP notice from typed meta, never raw \\x01 — #641", () => {
     // #641 — an inbound CTCP reply (a NOTICE carrying \x01VERB [args]\x01) that
-    // matches no pending /ping is NOT consumed by subscribe.ts's
-    // maybeConsumePingReply (that swallows only CORRELATED PING replies). It
+    // matches no pending query is NOT consumed by subscribe.ts's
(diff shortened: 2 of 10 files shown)


vjt:

Landed as: `6a0b6a1b + ea7cdf7d`.

Issue #630, opened 2026-08-01
Inbound flood protection: throttle misbehaving clients, 429 then sever the web session


vjt:
## Why

A misbehaving (or hostile) client can flood us **inbound** and we have almost nothing in the way.

Today the only inbound limiter in the codebase is #340's per-`(subject, network)` send bucket, and it guards
exactly one door: `POST /messages` (`messages_controller.ex:197`). **`lib/grappa_web/channels/` contains zero
rate-limit code** — every `handle_in` verb on `GrappaChannel` (`visibility`, `away`, `recover`, `watchlist`,
`client_closing`, …) is unmetered. That is the cheaper and faster door, so the guarded one is not the one an
abusive client will use.

Goal, in vjt's words: *"protect ourselves from misbehaving clients"* — a user must not be able to flood us, and
a **sustained** flood must cost them their socket, not cost everyone else the service.

## The ladder (vjt's call, 2026-08-02)

1. **Throttle** — a per-subject budget across **WS verbs AND all REST writes**, one shared budget so a flooder
   cannot dodge it by switching surface. #340's per-`(subject, network)` send bucket stays **on top** as the
   finer-grained send limit; this is the coarse outer gate, not a replacement.
2. **429 first** — an over-budget request is refused, not queued: HTTP `429` on REST; on WS, where there is no
   status code, an **error frame carrying the same 429 semantics** plus a retry hint.
3. **Sever on sustained abuse** — a client that keeps flooding past the 429s gets its **socket closed with a
   specific close code AND its auth session invalidated**, so it must re-authenticate.
   🔴 **The IRC session (`Session.Server`) is NOT touched** — deliberately. A client-side bug must not cost the
   user their IRC presence or look like a netsplit to everyone in their channels. Web session dies, bouncer lives.

## Design constraints (read `CLAUDE.md` first — these are not negotiable)

- **Reuse the verbs, not the nouns.** `Grappa.RateLimit.TokenBucket` (ETS, per-key, `take/4`) and the
  `FailureWindow` escalation shape already exist. The escalation counter is a *window*, not a new parallel state
  machine — do not duplicate state that can be derived.
- **One feature, one code path, every door** — the REST plug and the WS `handle_in` guard must call the SAME
  context function. Two implementations that drift is the failure mode here.
- **Additive wire contract** (`docs/CLIENT_PROTOCOL.md`, #447): a new error frame kind and a new close code are
[...]

Issue #480, opened 2026-07-27
Send throttle should mirror the upstream oper allowance, not a fixed per-user one


vjt:
## Summary

The inbound message-send throttle (`config :grappa, :send_throttle`, capacity 5 / 0.5 tokens per second) is tuned to the flood allowance bahamut applies to an **ordinary** client. For a session whose upstream connection is oper'd, that allowance is dramatically higher — so grappa's bucket stops being a shock absorber and becomes the binding constraint, 429'ing a client the IRC server would have happily served.

Raised on #it-opers by Sonic (per channel request), who hit it as an admin+ircop.

## Why the current numbers are right for a normal user and wrong for an oper

The throttle exists to 429 a flooding client *before* the upstream k-lines it — that is stated in `config/config.exs`, and the numbers were read off bahamut's flood allowance. On Azzurra's bahamut the oper path is different in three places:

- `include/config.h:510` — `#define NO_OPER_FLOOD` is **on** in this build.
- `src/parse.c:225-236` — with `NO_OPER_FLOOD`, an oper's `since` penalty is incremented only on a weighted fraction of messages (`(cptr->receiveM % 10) ? 1 : 0`) instead of the ordinary `2 + i / 120`; and with umode `+F` (`NoMsgThrottle`, `OFLAG_UMODEF`) it is not incremented at all.
- `src/s_bsd.c:1791-1795` — the `CLIENT_FLOOD` RecvQ check is skipped entirely for opers.

So for an oper (and especially an oper with `+F`) there is effectively no upstream ceiling to protect, while grappa still enforces burst 5 + 1 line / 2s.

## The axis this should key on

Not a grappa-side identity tier ("trusted user", admin flag, account role). The throttle is not an authorization gate — it is a mirror of the *upstream* allowance, so it should read *upstream* state: is this session's IRC connection oper'd, and does it carry `+F`?

That state is already tracked. `Grappa.Session.Server` keeps a per-session `umodes` set (#229, `server.ex:638-645`), populated at connect by an explicit umode query (`Grappa.IRC.Client.send_umode_query/2`, `client.ex:753-767`, because ircds do not volunteer the user's own umodes) and kept current from `:mode`-on-`$server` events. `handle_call(:get_umodes, …)` (`server.ex:1889`) already exposes it.

## Proposed behaviour

In `GrappaWeb.MessagesController.take_send_token/2`, pick the bucket parameters from the session's umodes:

[...]

PR #1280, opened 2026-08-13
send: mirror the upstream oper flood allowance in the send throttle

vjt:
Refs #480.

The inbound send throttle (#340) 429s a flooding client before the ircd k-lines
it, and its numbers were read off the allowance bahamut applies to an
**ordinary** client. On a connection the ircd meters on its oper path the same
bucket stops mirroring anything and becomes the binding constraint, refusing a
line the upstream would have carried.

## The axis

Upstream state, not a grappa-side identity tier — the throttle is not an
authorization gate. `Grappa.Session.FloodAllowance` (pure, the
`IdentityState` shape) folds the per-session umode set #229 already tracks
plus the network's services flavour into `:exempt | :oper | :ordinary`;
`Session.get_flood_allowance/2` serves it; `MessagesController` picks bucket
parameters from the class and never sees a mode letter.

- `:exempt` — skip the bucket. The outer #630 request budget still meters the
  door, so this is not an unmetered path.
- `:oper` — the same bucket and key with a wider pair (`oper_capacity` /
  `oper_refill_per_sec`), and its own `retry-after`.
- `:ordinary` — today's numbers, unchanged.

Cost: one extra serialized call per POST, to the session that is about to
handle the send anyway. Mirroring umodes into ETS for a lock-free read would
duplicate state that already exists.

## The two decisions the issue asked for

**Where the numbers live.** Still `Application.compile_env`, like the existing
pair. Moving the whole `send_throttle` block to runtime config is a second
change with its own risk surface and the oper axis does not depend on it.

**Which letter means oper.** `+o`, read flavour-independently — RFC 1459 fixes
it. `supported_umodes` cannot answer this: 004 param 3 is a signless
concatenation of letters parsed with no meaning attached, so it says `F`
EXISTS on this ircd, never that `F` means NoMsgThrottle. The per-flavour table
is `IdentityState.registered_umode/1`'s precedent.

## Two calls worth reviewing

**An unclassified network gets no exemption.** bahamut's `F` (`OFLAG_UMODEF`)
is the one letter read at source; solanum assigns no `F` in core and nothing
was verified for hybrid. This is the OPPOSITE default from #388, deliberately:
there is no prior behaviour to preserve here, and the directions are not
symmetric — a withheld exemption costs an oper headroom the `:oper` tier
mostly gives back, a wrongly granted one switches the throttle off for a
connection the upstream still meters. An operator who wants it sets
`services_flavor: :azzurra` via the admin surface #349 ships. The exemption
also requires `+o`: free where the letter is an oper flag, a bound on the
damage where it is not.

**The oper numbers are chosen, not measured.** 10x the ordinary pair. The
issue's source citations establish that the oper path is qualitatively
different (penalty on a fraction of messages, RecvQ check skipped) — a PATH,
not a magnitude. Nothing was measured against a live bahamut; the pair stays a
config knob and the DESIGN_NOTES entry says so.

Out of scope, as the issue states: the `Grappa.Admission` per-IP clone cap is
untouched.

## Test plan

- [x] `scripts/check.sh` RC 0 on a clean tree — 8 doctests, 59 properties,
      6072 tests, 0 failures; dialyzer 0; credo strict 0; bats green
- [x] a failing-then-passing test per branch: exempt, oper, ordinary, plus the
      withheld-exemption case and the `:no_session` fallback
- [x] five mutants, each killed by the intended tests and only those:
      oper capacity 6→3 (3 controller tests), oper refill 0.25→0.5 (only the
      retry-after test), `:exempt` taking the oper bucket (only the exempt
      test), unverified flavours defaulting to `F` (2 unit + 1 controller +
      1 session), the exemption no longer requiring `+o` (1 unit)
- [x] `scripts/design-notes-gate.sh` RC 0
- [ ] CI

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -140,9 +140,23 @@ config :grappa, :busy_retry,
 #   * capacity — burst allowance (5, per bahamut's flood burst).
 #   * refill_per_sec — sustained rate: 1 token every 2s = 0.5/s (bahamut's
 #     ~2s-per-line drip once the burst is spent).
+#
+# GH #480 — the oper pair. On a connection the ircd meters on its OPER
+# path the ordinary numbers stop mirroring anything and start being the
+# binding constraint, refusing a line bahamut would have carried. These
+# are NOT read off the source: `parse.c` / `s_bsd.c` describe a PATH (the
+# penalty accrues on a fraction of messages; the RecvQ check is skipped),
+# never a magnitude. 10x the ordinary pair is a CHOSEN headroom — wide
+# enough that grappa stops being the ceiling, still a bucket rather than
+# an open door, and still a config knob for the operator who measures
+# their own network. `Grappa.Session.FloodAllowance` decides which
+# session gets them; a session the upstream throttles not at all skips
+# the bucket entirely.
 config :grappa, :send_throttle,
   capacity: 5,
-  refill_per_sec: 0.5
+  refill_per_sec: 0.5,
+  oper_capacity: 50,
+  oper_refill_per_sec: 5.0
 
 # GH #630 — coarse per-subject INBOUND request budget spanning EVERY WS
 # `handle_in` verb AND every REST write (POST/PUT/PATCH/DELETE). This is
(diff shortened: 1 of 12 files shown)


vjt:

Landed as: `65f0b970 + d76145ed`.

Issue #93, opened 2026-06-26
Phase 5: multi-server failover logic


vjt:
The Phase 2 schema includes `network_servers` (irssi shape: priority + enabled) but only the first server is used. Phase 5: try server 0 -> on connect fail try server 1 -> ... -> exponential backoff -> reset on success.

Phase 5 hardening cluster.

vjt:
Folded con #271 per ordine vjt: implementare insieme come failover a **due livelli** — server = outer loop (#93), leaf = inner loop (#271). Design + macchina a stati nel commento su #271.

Portata a `status:queued`, P1 (ride sul P0 #271). Lo schema `network_servers` è già in Phase 2; qui resta la logica try-server-N + backoff esponenziale + reset-on-success, come inner-loop la rotazione-leaf di #271.

vjt:
**Stale-premise audit** (backlog sweep against `origin/main` `8ed4f2e0`, 2026-08-06). Evidence only — nothing closed, relabelled or edited here; triage is vjt's. Full sweep: #258.

**Verdict: PARZIALE — the backoff half is already shipped.**

Already there: `Grappa.Session.Backoff` (ETS, survives a `:transient` restart) plus the #100 `@connection_stable_ms` reset gate (`server.ex:215-225`). Exponential backoff and reset-on-success are done.

Still true and still the whole point: `lib/grappa/networks/servers.ex:89 pick_server!/1` takes the **head** of `filter(&enabled) |> sort_by({priority, id})` — one server, no rotation. `git grep failover|next_server` over `lib` → zero hits. The module's own doc still says Phase 5 fail-over "is the natural evolution from here".

Remaining: server-index advance on connect failure (0 → 1 → …) and re-resolution across the list.

_Repo-cold evidence only: `git show origin/main:<path>` + symbol presence. No test, stack, browser or CI run backs this._

vjt:
Shipped in **v1.2.0** — https://github.com/vjt/grappa-irc/releases/tag/v1.2.0 (main `c0d3a252`). Self-hosters: that is the tag to pull.

PR #1282, opened 2026-08-13
#93: walk the server list on the backoff ladder

vjt:
## What

`network_servers` has carried irssi's `(host, port, tls, priority, enabled)`
shape since the Phase 2 schema, but only ever the head of that list was
dialled. A network whose preferred endpoint went down parked every session on
it until an operator intervened, however many working rows sat behind it.

This is the outer loop of the two-level fail-over machine agreed on #271 —
server as outer, resolved leaf as inner. The inner half shipped with #271
(`Grappa.IRC.Client` resolves the RR set itself, shuffles, and rotates across
leaves).

## The finding: the outer loop needed no loop

The iteration was already running, once per failure, spread across process
lifetimes:

`do_connect/5` fails → the Client stops → the linked `Session.Server` exits
abnormally → `terminate/2` calls `Backoff.record_failure/2` **synchronously**
(its docstring says why: so the count is visible to the next `init/1`) → the
`:transient` supervisor restarts the child → `init/1` calls the plan's
`refresh_plan` closure.

Every element of `for server in servers_by_priority(enabled)` was in place
except the part where the body picks a different server: the closure
re-resolved the same head forever. So the endpoint is now
`pick_server!(network, attempt)` with `attempt` read from the counter that
already exists. The ring position is **derived**, not tracked — no per-server
row of state, nothing to keep in sync, nothing to collect when a server row is
deleted under a live session.

No change to `lib/grappa/irc/client.ex`. It was not needed.

## Two doors, deliberately different

`resolve/1` — Bootstrap, the `/connect` verb via `SpawnOrchestrator`,
`Operator` — stays pinned to ring position 0. Not a convenience default: those
doors mean "start this session now" and clear the ladder before spawning, and
reading the counter there would additionally **race**, because
`Backoff.reset/2` is a `cast` and a resolve immediately behind it can still
observe the pre-reset count. The respawn door runs at the one moment the
counter is authoritative, and it is the only one that walks the ring.

`pick_server!/2` takes no default argument (CLAUDE.md): a caller that does not
say which position it wants is a caller that has not decided which door it is.

## Consequence worth stating

The exponential ladder stays keyed on `(subject, network)`, so it is shared
across the ring rather than one ladder per server row — with three dead
endpoints the exponent grows 1, 2, 3, … across them instead of restarting at
the base for each. That is the intended reading of "derive instead of
duplicate", and it preserves the property the ladder exists for: the rate at
which we approach an upstream is identical to today's whatever the server
count. A single-server network is byte-for-byte unchanged, every attempt
folding back onto position 0.

## Evidence

Red first: the plan-level tests were committed before the implementation and
failed with real assertion diffs (`left: "primary"`, `right: "secondary"`),
RC=2, 5 failures.

Green: `scripts/check.sh` RC=0 — credo `5761 mods/funs, found no issues`,
sobelow `No vulnerabilities found`, doctor passed, `8 doctests, 59 properties,
6062 tests, 0 failures`, dialyzer `Total errors: 0 … done (passed
successfully)`, bats 471 tests with zero `not ok`.

Mutation: 5 mutants injected, 5 killed, attributed so that policy and
plumbing are proven separately (one mutant covering both sides would be
vacuous).

| mutant | injection | killed |
|---|---|---|
| m1 | `Enum.at(ring, 0)` — `attempt` ignored | 8 (3 pure ring units, 4 plan, 1 visitor) |
| m2 | `rem(attempt + 1, …)` | 10, including the "spawn door stays on the primary" guard |
| m3 | `enabled` filter dropped | 7, of which 4 pre-existing (`no_server`, `add_network/3` refusal) |
| m4 | user respawn door → `resolve_attempt(cred, 0)` | **only** the 4 user-plan tests |
| m5 | visitor respawn door → `…, 0` | **exactly 1**, the visitor test |

m4/m5 are the load-bearing ones: each side's wiring is covered independently
of the policy in `pick_server!/2`, and the visitor side is not passing in the
user side's wake. m2 is what keeps the primary-pinning guard from being
decorative — no other mutant touched it.

## Not measured, not claimed

No wall-clock magnitude for fail-over. `@connect_failure_sleep_ms` (30s) and
the ladder's 5s base are source reads, not observations; no stack was stood up
and no end-to-end fail-over was timed.

The tests also stop at the plan boundary: they assert that the respawn door
resolves the next endpoint, not that a real refused connect reaches that door.
That chain is pre-existing behaviour and this change does not touch it — but
it is not re-proven here either.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -41082,3 +41082,77 @@ an existing one, resist adding a parallel delivery path just because the
 *registration* handshake differs — a discriminator field for display, not a
 branch in the sender, keeps one audited path instead of two half-audited
 ones.
+<!-- entry #93 -->
+
+---
+
+## 2026-08-13 — #93: the outer fail-over loop was already running, it just never advanced
+
+`network_servers` has carried irssi's `(host, port, tls, priority, enabled)`
+shape since the Phase 2 schema, and for just as long only ever the head of
+that list was dialled. A network whose preferred endpoint went down parked
+every session on that endpoint until an operator intervened, however many
+working rows sat behind it. #271 folded this together with its own problem
+into one two-level machine — server as the outer loop, resolved leaf as the
+inner one — and shipped the inner half: `Grappa.IRC.Client` resolves the full
+RR set itself, shuffles it, and rotates across leaves before giving up.
+
+The 2026-08-06 stale-premise audit on the issue called the remainder PARTIAL,
+on the grounds that the backoff half was already shipped (`Session.Backoff`
+plus #100's `@connection_stable_ms` reset gate) and only the server-index
+advance was left. Re-verified a week and a great many commits later against
+`origin/main` `45e88c5e`: still accurate. `pick_server!/1` still took the head
+of `filter(&enabled) |> sort_by({priority, id})`, and `git grep -E
+'failover|next_server'` over `lib` still returned nothing.
+
+What the audit could not see from a cold repo is that **the outer loop needs
+no loop**. The iteration was already running, once per failure, spread across
+process lifetimes: `do_connect/5` fails, the Client stops, the linked
+`Session.Server` exits abnormally, `terminate/2` calls
+`Backoff.record_failure/2` — synchronously, and the docstring says why: so the
+count is visible to the NEXT `init/1` — and the `:transient` supervisor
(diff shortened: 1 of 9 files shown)


vjt:

Landed as: `736b4505, f220e462, 30ba4049`.

Issue #151, opened 2026-06-29
cicchetto: server/network tab ComposeBox shows literal placeholder "message $server"


vjt:
Cosmetic. On the server/network window the ComposeBox input placeholder renders the literal string `message $server` instead of a friendly label.

**Root cause**
- `cicchetto/src/ComposeBox.tsx:333` — `placeholder={`message ${props.channelName}`}` interpolates `channelName` verbatim.
- For the server window, `channelName` is the internal window-kind sentinel `"$server"` (not a display name): see `HomePane.test.tsx:352` (`channelName: "$server"`) and `windowKinds.test.ts:148` (`server ($server) is backed by ...`).
- So the interpolation is correct; the value fed in is a sentinel that was never meant to surface in the UI.

**Fix (design call)**
Special-case the `$server` sentinel in ComposeBox: render a friendly placeholder for the server window — e.g. `message server`, the network name/slug, or a `/`-command hint — instead of interpolating the raw sentinel. Worth a quick scan for any other UI surface that interpolates `channelName` and could leak the same token.

Reported on #grappa.

vjt:
Shipped to prod (m42, hot cic deploy — bundle `hash=DVH7ybzS`, no BEAM restart). Merged FF to main as `e32fbe2`.

**Root cause:** `ComposeBox.tsx` rendered `placeholder={\`message ${props.channelName}\`}`. For the server window `channelName` is the internal sentinel `SERVER_WINDOW_NAME = "$server"` (windowKinds.ts), so the operator saw the literal `message $server`.

**Fix:** a pure `composePlaceholder(networkSlug, channelName)` boundary (new module `cicchetto/src/lib/composePlaceholder.ts`) rejects the sentinel. The server window is now labelled with its network slug — `message <slug>` — mirroring the Sidebar's `⚙️ <slug>` network-header row. General rule, not the one example: ANY `$`-prefixed sentinel (real IRC targets never start with `$`) falls back to a generic hint, so a future synthetic-window-with-compose can't leak its token either. Today `$server` is the only synthetic window that renders a compose box (the rest are gated out by `kindHasScrollback`), but the branch covers the class.

Put the helper in its own dependency-light module (not `compose.ts`) so ComposeBox render tests — which mock `../lib/compose` to dodge its stateful network graph — don't need mock surgery to import it.

**Tests (real RED→green, no hollow green):**
[...]

PR #1285, opened 2026-08-13
cic(#359): Alt+0 jumps to the status window

### LINKED ISSUE #359: cicchetto: keyboard shortcut (Alt+0) to jump to the status/server window
### Request (per chan request)

`Alt+1..9` select channel/query windows by index, and `Alt+A` (#235) jumps to the next window with unread activity. There is currently **no keyboard route to the status/server window** — you can only reach it with the pointer. Add one, e.g. `Alt+0`.

### Why

irssi muscle memory: the status window is window 0. A user on a small/mobile viewport (or just keyboard-driven) has no way back to server notices/status once they're in a channel.

### Current state (verified on `origin/main`)

- `cicchetto/src/lib/keybindings.ts:67-70` — the nav chord is gated on `/^[1-9]$/`:
  ```ts
  if (e.altKey && /^[1-9]$/.test(e.key)) {
    e.preventDefault();
    handlers.selectChannelByIndex(Number(e.key) - 1); // Alt+1 → idx 0
    return;
  }
  ```
  `Alt+0` is **not** matched today, so the key is free.
- `Alt+A` → `handlers.nextUnread()` (`keybindings.ts:90`).

### Design wrinkle for the implementer

The status/server window is **not** part of the `ActiveWindow` model — `activeWindows.ts:42-45` only has `kind: "channel" | "query"`, and `activeWindows.ts:134` explicitly notes *"server status buffers aren't activity"*. So this can't reuse `selectChannelByIndex` (index space is channels/queries only). It needs a distinct verb (e.g. `handlers.selectStatusWindow()`) wired to however the status/server buffer is addressed in the window store — that plumbing is the real work, not the keybinding.

### Scope

- Add an `Alt+0` (or agreed key) branch in `keybindings.ts`.
- New handler verb → selects/focuses the status/server window.
- Test alongside `keybindings.test.ts`.

vjt:
## What

`Alt+0` jumps to the current network's status/server window (irssi's window 0).
Before this, the keyboard reached channel/query windows (`Alt+1..9`) and the
next unread one (`Alt+A`), but the status window only with the pointer — so a
keyboard-driven operator, or one on a viewport where the sidebar is a drawer,
had no route back to server notices.

## The issue's diagnosis was wrong, and that shaped the fix

#359 reads the gap as *"the status window is not part of the `ActiveWindow`
model … that plumbing is the real work"*. Measured on the code, it is not:

* `ActiveWindow` (`lib/activeWindows.ts`) is the **activity** ordering — the
  tier list `Alt+A` walks. It excludes server buffers deliberately (server
  noise is not activity worth being sent to).
* `SelectedChannel` (`lib/selection.ts`) is the **focus** model, and there the
  status window has been an ordinary target since UX-4 bucket C:
  `{slug, "$server", kind: "server"}` — the exact tuple the sidebar's
  network-header row sets on click, backed by real `NumericRouter` scrollback.

What the status window *is* outside of is the index space
`selectChannelByIndex` walks (channels + queries). That is a statement about
one verb's domain, not about the window's addressability — so this adds a
**sibling verb**, not a new representation. No parallel window model, nothing
originated client-side.

## How

* `lib/selection.ts` — `selectStatusWindow()`, next to the other selection
  verbs. Which network it means is **derived** at call time: the selection's
  own network when it has one, else the first network in the sidebar's order,
  else no-op. The `$`-sentinel slugs `home`/`admin` carry are unknown to
  `networkBySlug`, so "no network in focus" falls out of the existing lookup
  instead of a per-kind check — a future identity-scoped window inherits the
  fallback for free. No stored "current network" to drift.
* `lib/keybindings.ts` — the `Alt+0` branch, matched on **`e.code ===
  "Digit0"`**, like the `Alt+A` sibling and for the same reason #235 recorded:
  a macOS Option+digit composes `e.key` (US layout `Option+0` → `º`).
* `Shell.tsx` — wires the handler.

## Tests

* `keybindings.test.ts` — `Alt+0` dispatches `selectStatusWindow` and neither
  `selectChannelByIndex` nor the compose auto-focus path; the dispatch carries
  `key: "º"`, so a key-based match fails the test. Plus `Digit0` without Alt is
  a no-op.
* `selection.test.ts` — the verb picks the **current** network's status window
  (not the first one), falls back to the first network from `home`, and no-ops
  with no network bound.
* `e2e/tests/issue359-alt0-status-window.spec.ts` — from inside `#bofh`,
  `Alt+0` focuses the status window. Two independent oracles, both asserted in
  their **pre** state first: the sidebar selection class moves onto the
  `$server` row, and the compose placeholder becomes `message <slug>` (#151) —
  so a sidebar class flipped without the pane following cannot pass.

## Not claimed

* **Not measured on a real macOS browser.** Playwright's `Alt+0` sends
  `key: "0"` + `code: "Digit0"`, so the e2e cannot distinguish the two
  implementations; the `e.code` choice rests on the precedent #235 documented,
  not on a measurement of mine.
* `Alt+1..9` still match on `e.key`, so **those** chords are the ones that
  would miss on macOS. Pre-existing defect in a verb this change does not
  touch — deliberately not widened here, recorded in DESIGN_NOTES so the
  divergence reads as a decision, and worth its own issue.

### DIFF
--- a/cicchetto/e2e/tests/issue359-alt0-status-window.spec.ts
+++ b/cicchetto/e2e/tests/issue359-alt0-status-window.spec.ts
@@ -1,0 +1,59 @@
+// GH #359 — Alt+0 jumps to the status/server window (irssi's window 0).
+//
+// Before this, the keyboard could reach channel/query windows (Alt+1..9,
+// index space) and the next unread one (Alt+A), but the status window only
+// with the pointer. Alt+0 is a verb of its own — the status window is NOT in
+// the Alt+1..9 index space — and it resolves to the CURRENT network's status
+// window (`lib/selection.selectStatusWindow`).
+//
+// This spec drives the USER-VISIBLE OUTCOME from inside a channel, with two
+// independent oracles so a sidebar class flipped without the pane following
+// cannot pass:
+//   1. the sidebar selection moves off #bofh onto the `$server` row, and
+//   2. the compose textarea placeholder becomes the server window's
+//      (`message <slug>`, #151) instead of the channel's (`message #bofh`).
+// Both are asserted in their PRE state first, so the gesture is what moves
+// them. On the unfixed code Alt+0 is an unbound key: the channel stays
+// selected and the placeholder never changes — the spec goes RED.
+//
+// Desktop only: the chord needs a physical Alt, and #359 exists precisely
+// because a keyboard-driven operator had no route. The mobile route to the
+// same window is the BottomBar network header, already covered elsewhere
+// (e.g. issue151-server-placeholder).
+
+import { composeTextarea, loginAs, selectChannel, sidebarWindow } from "../fixtures/cicchettoPage";
+import { AUTOJOIN_CHANNELS, NETWORK_SLUG } from "../fixtures/seedData";
+import { expect, specNick, specUser, test } from "../fixtures/test";
+
+const CHANNEL = AUTOJOIN_CHANNELS[0];
+
+test("#359 — Alt+0 from inside a channel focuses the status/server window", async ({ page }) => {
+  const vjt = specUser();
+  await loginAs(page, vjt);
+
+  // Barrier: the channel window is really focused (REST page landed + WS
+  // subscribed) before the gesture, so nothing else can be moving focus.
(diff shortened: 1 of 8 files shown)


vjt:

Landed as: `00a0a141`.

Issue #201, opened 2026-07-09
Make video upload duration cap (2 min) configurable


vjt:
## Problem

The client-side video upload duration ceiling is a hardcoded constant:

- `cicchetto/src/lib/videoPolicy.ts:23` — `export const MAX_DURATION_SECONDS = 120;`

It binds on both the transcode and the no-WebCodecs fallback path (duration is read via `<video>` `loadedmetadata`, see the policy-vs-capability note at `videoPolicy.ts:14-19`). Consumers:

- `cicchetto/src/lib/uploadOrchestrator.ts:16` (import), `:409` (`VIDEO_TOO_LONG_MESSAGE`), `:474` (the reject gate `durationSeconds > MAX_DURATION_SECONDS`)
- `cicchetto/src/lib/videoTranscode.ts:61` (import)
- the budget math in `videoPolicy.ts` (`videoBitrateBudget`, `pickTargetHeight`, `pickEncodeBitrate`) takes `durationSeconds` as an arg, so it's unaffected by the source of the ceiling.

## Ask

Make the 2-minute cap a server-driven, admin-configurable setting, default 120s. Mirror the existing pattern for the **byte** cap `upload.video_per_file_cap_bytes`, which already flows server → wire → client → admin UI:

- server_setting: `grappa/lib/grappa/server_settings.ex` (key + default), `server_settings/wire.ex`, `controllers/server_settings_controller.ex`, `controllers/admin/settings_controller.ex`
- client: surface it in the settings view the client already receives (same struct that carries `video_per_file_cap_bytes`), and consume it in `videoPolicy.ts` instead of the bare const. Keep `MAX_DURATION_SECONDS = 120` only as the fallback default.
- admin UI: add a field in `cicchetto/src/AdminSettingsTab.tsx` next to the video per-file cap (see `videoCapMB` at `:59`).

## Notes

- New proposed key name: `upload.video_max_duration_seconds` (pos_integer, default 120).
- `VIDEO_TOO_LONG_MESSAGE` (`uploadOrchestrator.ts:409`) must be derived from the live value, not the compile-time const.

vjt:
Shipped in **v1.2.0** — https://github.com/vjt/grappa-irc/releases/tag/v1.2.0 (main `c0d3a252`). Self-hosters: that is the tag to pull.

PR #1289, opened 2026-08-13
Make the video upload duration cap a server setting (#201)

vjt:
## What

The video-upload duration ceiling was a compile-time constant in cic
(`videoPolicy.ts` — `export const MAX_DURATION_SECONDS = 120`), so moving it
meant a rebuild and a redeploy. It becomes the server setting
`upload.video_max_duration_seconds` (pos_integer, default 120), admin-tunable
like every other upload knob.

The change rides the rails the BYTE cap `upload.video_per_file_cap_bytes`
already laid — accessor pair on `Grappa.ServerSettings` → `public_view/0` →
`ServerSettings.Wire.upload_view/1` (one edit feeds the WS snapshot/push AND
both REST doors) → `serverSettings()` → an admin field. No new mechanism.

**No migration.** Settings are key/value rows whose defaults live in code; the
audio cap set that precedent (`@default_upload_audio_per_file_cap_bytes`, no
seed row).

## Corrections to the issue text

The issue's structure holds; several of its line numbers had drifted, and one
call is wrong in kind. Recorded here so #201 stops misleading the next reader:

- `uploadOrchestrator.ts` — import is at `:18` (issue says `:16`),
  `VIDEO_TOO_LONG_MESSAGE` at `:477` (`:409`), the fallback reject gate at
  `:542` (`:474`).
- `AdminSettingsTab.tsx` — `videoCapMB` is at `:63` (`:59`).
- **`videoTranscode.ts:166` is the PRIMARY duration gate, not "an import".**
  The issue lists `videoTranscode.ts:61` only as an import site. That gate runs
  BEFORE the capability gate — it is the one that binds on the WebCodecs-less
  fallback path — so it, not the orchestrator's, is the main enforcement point.
  Both are converted.

## Design calls worth reviewing

**⚠️ This ceiling is NOT an enforced limit — it is enforced client-side only.**
Worth stating plainly, because an operator reading "video max duration" in an
admin panel next to four byte caps will reasonably assume all five behave the
same way, and they do not:

| Setting | Where enforced | What happens if the client ignores it |
|---|---|---|
| `*_per_file_cap_bytes`, `global_cap_bytes` | server, `POST /api/uploads` | rejected — 413 / 507 |
| `video_max_duration_seconds` | **cic only** | the upload proceeds |

The reason is not an oversight to fix later: reading a duration means decoding
the container, and the whole point of the ceiling is to refuse the clip BEFORE
it is uploaded, while the file is still in the browser. So the server publishes
the number and cic applies it. What actually bounds disk is the byte caps,
which are unaffected. The accessor's `@doc` says this too, so the next reader
of `Grappa.ServerSettings` does not mistake it for a control.

**The value travels as a parameter.** `transcodeVideo/5` takes
`maxDurationSeconds` alongside the `capBytes` it already took;
`uploadOrchestrator` resolves both once per attempt. Letting `videoPolicy.ts`
read `serverSettings()` itself would have put a solid-js store import into the
one module kept dependency-free so it can be imported statically while its
mediabunny-laden sibling stays behind a dynamic `import()`. Reading once per
attempt also keeps the transcoder gate and the fallback gate from straddling an
admin change mid-upload and rejecting against a number the message never named.

**The new field's narrowing in `userTopic.ts` is LENIENT while the byte caps
stay strict.** Hard-narrowing it would make a cic built with #201 drop EVERY
settings push from a pre-#201 server — stranding the byte caps too. Absent or
malformed degrades to the fallback and the rest of the push applies: the
additive-only wire contract, same posture `http_host_aliases` took in #324.

**The copy formats the cap exactly instead of reusing `formatDuration`.** That
helper floors (90s → "1m") because it describes elapsed spans; a ceiling that
understates itself tells the operator a 75-second clip exceeded a "1m" limit.
So: "2 minutes" on whole minutes, "45 seconds" otherwise.

## Tests — what they cover, and what they cannot

Covered:

- ExUnit — default, put/get round trip, rejection of zero/negative/non-integer,
  a stored non-positive falling back to the default, the PubSub broadcast, the
  after-join snapshot, both REST doors, and the 422 `invalid_setting` field.
- vitest — the ceiling handed to the transcoder is the live one (and the 120s
  fallback with no snapshot); both reject gates fire against the live value; the
  error copy names it; the lenient-vs-strict narrowing split, including that a
  malformed BYTE cap still drops the whole push.
- e2e (`ux-6-b-admin-settings.spec.ts`) — admin form → PUT →
  `GET /api/server-settings`, the door cic hydrates from, plus the 422 field
  highlight. Run locally, scoped: `Running 2 tests using 1 worker` → `2 passed`
  (the same grep selects 0 tests on the merge base, so the suite gains 2).

**NOT covered, and why:** there is no end-to-end run in which a real over-long
video is refused with a message naming an admin-lowered value. The only
committed video fixture, `cicchetto/e2e/fixtures/tiny.mp4`, is **exactly
1.000s** (mvhd timescale 1000, duration 1000) and the smallest settable ceiling
is 1 second — so no admin value can make it too long. Producing a longer
fixture needs ffmpeg, which is not available in this environment. Rather than
commit a generated video or lower the minimum below 1s, the refusal and its copy
are pinned by vitest through the pre-existing `__setProbeDurationForTests` seam
(introduced in `b760d8df`, five call sites before this branch), and the e2e
covers the round trip.

### DIFF
--- a/cicchetto/e2e/tests/ux-6-b-admin-settings.spec.ts
+++ b/cicchetto/e2e/tests/ux-6-b-admin-settings.spec.ts
@@ -65,7 +65,10 @@ test.describe("UX-6-B admin Settings tab", () => {
     const admin = getSeededAdmin();
     const res = await request.put("/admin/settings", {
       headers: { authorization: `Bearer ${admin.token}` },
-      data: { upload: { active_host: "embedded" } },
+      // #201 — the duration cap joins active_host in the reset: the
+      // video-upload specs probe a ~1s clip against it, so a lowered
+      // value left behind would refuse every one of them.
+      data: { upload: { active_host: "embedded", video_max_duration_seconds: 120 } },
     });
     expect(res.ok()).toBe(true);
   });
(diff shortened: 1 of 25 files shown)


vjt:

Landed as: `6ab222fe + 10010622`.

Issue #1286, opened 2026-08-13
Built-in NickServ IDENTIFY sends the password alone, so an alt-nick session identifies the wrong nick


vjt:
## What happens

grappa's built-in NickServ IDENTIFY, sent at 001, puts the password on the wire and nothing else
(`lib/grappa/session/server.ex:4279` on `origin/main` `4879785c`):

```elixir
send_perform_line(state, "PRIVMSG NickServ :IDENTIFY #{secret}")
```

`IDENTIFY <password>` is the one-argument form, so services identify **the nick the session is
currently holding**. When the configured nick was taken upstream and the session landed on an alt
(`<nick>_`, or a rung of the `Grappa.IRC.AuthFSM` ladder), that is the alt — not the account the
credential belongs to. The identify either fails outright or, worse, succeeds against a different
registration.

Reported on IRC by a user who set nick + password through the web UI and could not work out why
they were not identified after joining on an alt nick.

## Why the existing fix does not cover it

#885 added `$nick` to the perform-list expander precisely for this (`NS IDENTIFY $nick
$nickserv_pass`, bound to the credential's **configured** nick, not the live one). That fix lives
in `Grappa.Session.PerformList` and only helps a user who hand-writes a perform line.

The built-in identify is a separate site: per #189 it is emitted by
`Session.Server.run_perform_and_identify/1` ahead of the perform list, and is suppressed only when
the perform list structurally consumed `$nickserv_pass`. A visitor who typed a nick and a password
into the web form and never touched the perform list gets the one-argument form — the default path
is the one that is wrong.

## Suggestion

Emit the two-argument form from the built-in site as well, naming the credential's configured nick
(the same value `configured_nick/1` already feeds the expander for `$nick`):

```
PRIVMSG NickServ :IDENTIFY <configured_nick> <password>
```

Points to settle in design, not assumed here:

* Services compatibility — Atheme/Anope accept `IDENTIFY <account> <password>`; the two-argument
  form should be checked against the services set grappa targets before it becomes the default.
* Whether it should be unconditional or only when the live nick differs from the configured one.
  Unconditional keeps one code path and avoids a race with the nick ladder; conditional keeps the
  wire identical to today for the overwhelmingly common case.
* The `Session.GhostRecovery` path already identifies *after* NICKing back to the original, so it
  is correct today and must stay correct under either choice.

## Verification done

* Line quoted above read from `origin/main` (`4879785c`), not from a local checkout.
[...]

Issue #885, opened 2026-08-05
perform: add a $nick variable so /ns identify can name the account when the session is on an alt nick


vjt:
## What is missing

The perform list substitutes **exactly two** variables today
(`lib/grappa/session/perform_list.ex:63`, `@var_re ~r/\$(nickserv_pass|oper_pass)/`):

* `$nickserv_pass`
* `$oper_pass`

Everything else passes through verbatim — the moduledoc even names `$nick` as an example of a
token that is *not* a variable.

## Why that bites

A user whose primary nick is taken lands on an alt nick, and the perform line

```
/ns id $pass
```

then identifies (or fails to) against whatever nick the session currently holds. To identify the
*account* rather than the current nick, the line has to be

```
/ns identify $nick $nickserv_pass
```

and there is no way to write that today: `$nick` reaches the wire as four literal characters.
The workaround is hardcoding the nick in the perform text, which then silently rots the moment
the credential's nick changes — the two live in different places with nothing keeping them equal.

## Proposal

Add `$nick` to the substitution alternation, bound to the **credential's configured nick** (not the
nick the session happens to hold at the time). Rationale for using the configured value: the whole
point is to identify the account when the live nick is *wrong*, so binding it to the live nick would
expand to exactly the value that does not work.

Notes for whoever picks this up:

* It is a one-token addition to `@var_re`, but the substitution is deliberately **single-pass** so a
  secret containing `$…` is never re-expanded — the new token must join the same alternation, not
  add a second pass.
* `$nick` is not a secret. It must not change the redaction rules for the expanded lines (still
  never logged) nor the `consumed_nickserv_pass?` suppression signal, which stays keyed on the
  NickServ password alone.
* An unbound variable expands to the empty string by contract; `$nick` should follow that rule
  rather than falling back to the literal token.
* User-facing docs for the perform box list the available variables — they need the third entry.

Requested by a testnet operator; the shape above is the owner's call ("le variabili le abbiamo già,
va solo aggiunta la `$nick`"). Related: #509, which is the other half of the same surface.

vjt:
`$nick` joined the **single-pass** alternation in `v0.13.0` (`perform_list.ex:103`,
`~r/\$(nickserv_pass|oper_pass|nick\b)/` — the `\b` keeps `$nickserv_pass` on the password branch and
`$nickname` verbatim) and binds the **configured** nick, not the live one (`server.ex:3790-3793`,
[...]

PR #1291, opened 2026-08-13
session(#1286): name the account in the built-in NickServ identify

vjt:
## What this fixes

The built-in NickServ identify at 001 (`Session.Server.run_perform_and_identify/1`
→ `maybe_builtin_identify/3`) emitted the one-argument
`PRIVMSG NickServ :IDENTIFY <password>`. That form names no account, so services
identify **the nick the session is currently wearing**. When the configured nick
is taken upstream the session is welcomed under an alt, and the identify then
addresses the alt rather than the account the credential belongs to.

#885 had already fixed the twin site — `$nick` in the perform-list expander binds
the CONFIGURED nick — but only there. A user who typed a nick and a password into
the web form and never touched the perform list got the one-argument form, so the
**default path was the broken one**. This reuses the same `configured_nick/1`;
there is no second derivation of "which nick is the account".

## Correction to the issue's premise: one route, not two

The issue supposed the alt nick could arrive either from `Grappa.IRC.AuthFSM`'s
433 ladder or from `Session.GhostRecovery`. Measured, only the latter is
reachable here: `auth_fsm.ex:481-492` intercepts 432/433 for
`auth_method: :nickserv_identify` with `{:cont, state, []}`, and the ladder clause
sits **after** it (`:499-507`) deliberately — its own comment says a ladder NICK
there would race the GHOST sequence off its own nick. The built-in identify only
exists for `:nickserv_identify`, so the ladder cannot be its source. That is the
difference between a fix with one scenario and a fix with two.

## Why the form is chosen per services flavour

The two-operand spelling is not universal across the services sets grappa targets,
so the form is selected from `services_flavor` — the same shape
`IdentityState.registered_umode/1` already uses: a table of measured divergences
plus a conservative default. Each entry is read off that daemon's own handler and
cited in the code:

* `:azzurra` — `azzurra/services@23473ed src/nickserv.c:1659` (`do_identify`):
  `nick` then `pass`; the one-argument spelling is an internal shift (`:1697-98`).
* `:atheme` — `atheme/atheme modules/nickserv/identify.c:29-30` (`ns_cmd_login`):
  `target = parv[0]`, `password = parv[1]`. Not merely accepted — it is the ONLY
  form on a `no_nick_ownership` configuration, where the one-argument shift
  (`:46-51`) is compiled out.
* `:oftc` — `oftc/oftc-ircservices modules/nickserv.c:721` (`m_identify`) reads
  the operands in the opposite order (its own help says
  `Usage: IDENTIFY password [nick]`, `languages/nickserv.en.lang:31`) and its
  two-operand form additionally forces a nick change (`:758`). A different verb,
  not a different spelling — so no two-operand identify is emitted there.

`:unknown` and an unclassified `nil` take the one-argument form too. The
account-naming form goes **only where it was measured**; the default arm is the
conservative one, and adding a flavour is a deliberate act with a citation.

## Unconditional within a flavour

Rather than "only when the live nick differs": on both daemons that take the form,
naming an account equal to the live nick is the same code path as omitting it
(azzurra sets `sameNick` via `str_equals_nocase`; atheme's one-argument form is
literally a shift to `si->su->nick`). The guard would buy no behavioural
difference and would have to read `state.nick` at a point where its equality with
the configured nick is an ordering accident, not a contract — the caveat
`configured_nick/1` already carries.

## Why one test is built on a synthetic wire

🔴 **One test drives a fake ircd that ECHOES the rename before welcoming the alt.
A real ircd does not send that.** It exists to make observable a divergence that
on the real path originates from `Session.GhostRecovery`, and it is called out in
a comment on the helper itself as well as here.

It is there because the divergence is otherwise unobservable at the only moment
that matters. Measured:

1. `state.nick` has exactly two writers, both in `EventRouter`: the 001 reconcile
   (`event_router.ex:1708`) and an OBSERVED self-NICK (`:3703`). `Session.Server`
   never writes it.
2. The 001 reconcile is reached through `delegate/2` at the very end of Server's
   001 arm (`server.ex:3228`) — strictly **after** `run_perform_and_identify/1`,
   so it cannot have moved the nick by the time the line is built.
3. `pending_password` is one-shot: set at init (`server.ex:1233`), cleared at the
   first 001 (`:4218`), never re-set. A second 001 — the only other way to reach
   the identify with a diverged nick — emits no identify at all.

So on every reachable path the live and configured nicks are **equal** when the
line is built, and a wire-level assertion cannot tell which one it came from. The
synthetic echo buys that discrimination; an assertion on the pre-state
(`state.nick == "grappa-test_"`) keeps the test from passing for the wrong reason.
The realistic alt-nick scenario is covered separately, through the genuine
433 → GhostRecovery path.

## Tests

Six, in `test/grappa/session/server_test.exs`:

* `:azzurra` under a real alt nick, via the existing `ghost_handler/1` (433 on the
  first NICK, `001 <nick>_` on the second) — the actual reported scenario.
* `:atheme` names the configured nick.
* `:oftc`, `:unknown` and `nil` keep the one-argument line, as a comprehension —
  regression pins, and these three passed **before** the change, which is the
  point of them.
* the configured-vs-live discriminator described above.
* the account-naming identify still stages the **password** for the `+r`
  rendezvous — if the outbound choke point lifted the wrong token it would commit
  the account name as the secret on `+r`, a silent and persistent corruption.

Red first: 4 failures out of 406, all of them the new behavioural tests, no
pre-existing failures. Green after: 406/0. `scripts/check.sh` clean.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -41082,3 +41082,88 @@ an existing one, resist adding a parallel delivery path just because the
 *registration* handshake differs — a discriminator field for display, not a
 branch in the sender, keeps one audited path instead of two half-audited
 ones.
+<!-- entry #1286 -->
+
+---
+
+## 2026-08-13 — #1286: the built-in NickServ IDENTIFY names the account, on the flavours measured to take it
+
+grappa's built-in identify at 001 (`Session.Server.run_perform_and_identify/1`
+→ `maybe_builtin_identify/3`) emitted the one-argument
+`PRIVMSG NickServ :IDENTIFY <password>`. That form names no account, so
+services identify **the nick the session is currently wearing**. When the
+configured nick is taken upstream the session is welcomed under an alt, and
+the identify then addresses the alt rather than the account the credential
+belongs to.
+
+#885 had already fixed the twin site — `$nick` in the perform-list expander
+binds the CONFIGURED nick, not the live one — but only there. A user who typed
+a nick and a password into the web form and never touched the perform list got
+the one-argument form, so **the default path was the broken one**. The fix
+reuses #885's `configured_nick/1`; there is no second derivation of "which
+nick is the account".
+
+### Only one route reaches this, not two
+
+The issue supposed the alt could come either from `Grappa.IRC.AuthFSM`'s 433
+ladder or from `Session.GhostRecovery`. Measured, it is only the latter:
+`auth_fsm.ex:481-492` intercepts 432/433 for `auth_method: :nickserv_identify`
+with `{:cont, state, []}`, and the ladder clause sits AFTER it (`:499-507`),
+deliberately — its own comment says a ladder NICK there would race the GHOST
+sequence off its own nick. Since the built-in identify only exists for
+`:nickserv_identify`, the ladder is unreachable from it. One scenario:
+`server.ex`'s 433 arm arming GhostRecovery's underscore NICK.
+
+### Unconditional within a flavour, not "only when the nicks differ"
+
+Naming an account equal to the live nick is the SAME code path as omitting it
+on both daemons that take the form — azzurra sets `sameNick` through
+`str_equals_nocase`, and atheme's one-argument spelling is literally a shift to
+`si->su->nick`. So a "only when they differ" guard buys no behavioural
+difference, and it would have to read `state.nick` at a moment where its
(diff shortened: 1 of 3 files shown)


vjt:
Closed unmerged: #1286 was closed on the call that the manual `/recover` path already covers the reported case. No defect in this branch — it is simply not wanted.

Issue #1284, opened 2026-08-13
bug(auth): arming TOTP revokes the account's per-client tokens


abonforti:
Arming TOTP revokes every per-client token on the account, which is the exact credential #1196 introduced so that a headless client survives a second factor. Following the documented order — mint a client token, hand it to the client, then arm TOTP — locks the client out at the moment the second factor is confirmed.

Seen on v1.1.0 while arming TOTP on an account whose bicchierino bridge authenticates with a client token.

## Why it happens

`confirm_totp_enrollment/5` calls `revoke_other_sessions_for_user!/2` inside its transaction, and that query has no `kind` filter (`lib/grappa/accounts.ex:858`):

```elixir
query =
  from(s in Session,
    where: s.user_id == ^user_id and s.id != ^current_session_id and is_nil(s.revoked_at)
  )

Repo.update_all(query, set: [revoked_at: DateTime.utc_now()])
```

A client token IS a `Session` row — `create_client_token/5` inserts it with `kind: :client` into the same table — so it matches this `where` and is revoked alongside the browser bearers.

## Why it matters

`docs/CLIENT_PROTOCOL.md` §7 tells a client author that a per-client token is what to hold when the account has a second factor armed, and #1196's own framing is that the token is "a separate credential whose issuance required 2FA". Both point an operator at: mint first, arm second. That order is the one that breaks.

The working order is the reverse — arm TOTP, then mint from the session that confirmed the enrollment, since `current_session_id` is the one row the revoke spares. Nothing says so today, and the failure is silent from the operator's side: the enrollment succeeds, recovery codes are printed, and the bridge simply stops authenticating with a `401` that is indistinguishable from a wrong password.

The same question applies to `disable_totp` ("Atomically disables TOTP and revokes every other bearer session", `accounts.ex:907`) and to the passkey mode transaction, which the docstring on `revoke_other_sessions_for_user!/2` lists as callers.

## What the intended posture is

#1196 raised this explicitly and left it open:

> revocable individually, and revoked implicitly by `reset-passkeys` / `reset-totp` if that is the desired posture

Revoking on *operator recovery* is defensible: `reset_totp` is the "I lost the authenticator" door and burning every credential is a reasonable blast radius. Revoking on *enrollment* is harder to justify — the account holder is present, has just proven the password and the code, and is arming a factor precisely so the headless credential can keep working.

So the fix is a decision, not a patch:

[...]

Issue #1283, opened 2026-08-13
bug(cic): Enable TOTP posts an empty body, so enrolment always answers bad_request


abonforti:
Pressing **Enable TOTP** in cic fails with "The request was malformed." The button cannot work for anyone: cic posts an empty body, and the endpoint has required the account password since `c1657c3b` ("auth: gate TOTP enrolment on the account password").

Seen on v1.1.0, native systemd substrate.

## The mismatch

`cicchetto/src/lib/api.ts:1555`:

```ts
export async function startTotpEnrollment(token: string): Promise<TotpEnrollment> {
  const res = await fetch("/me/totp/enrollment", {
    method: "POST",
    headers: buildHeaders(token),
    body: "{}",
  });
```

`lib/grappa_web/controllers/totp_controller.ex:28`:

```elixir
def start_enrollment(
      %{assigns: %{current_subject: {:user, user}}} = conn,
      %{"password" => password}
    )
    when is_binary(password) do
  with :ok <- Accounts.verify_password(user, password) do
```

With no `password` key the head does not match, the call falls through to the catch-all at line 55 (`{:error, :bad_request}`), and cic renders that as "The request was malformed."

`confirmTotpEnrollment` is fine — it sends `{enrollment_token, code}`, which is exactly what `confirm_enrollment/2` destructures. Only step one of the flow is broken, but it is the step that gates the other.

## Repro

1. Log into cic as an account with no TOTP armed.
2. Open the security pane, press **Enable TOTP**.
3. "The request was malformed." The server logs a 400 on `POST /me/totp/enrollment`.

## Fix direction

The server-side gate looks deliberate — re-authenticating before arming a second factor is the right posture, and `POST /me/client-tokens` asks for the password for the same reason. So the fix belongs in cic: prompt for the account password in `TotpSettings.tsx` and pass it through `startTotpEnrollment`, the way the client-token mint flow already must.

Worth checking whether the same gap exists anywhere else that `c1657c3b` touched: the commit added the requirement server-side, and at least this caller was never updated.

## Workaround

Driving the two endpoints by hand works — `POST /me/totp/enrollment` with `{"password": "..."}` returns `secret` / `provisioning_uri` / `enrollment_token`, and `POST /me/totp/enrollment/confirm` with `{enrollment_token, code}` completes it and returns the recovery codes. See the sibling issue about what that confirmation revokes on the way through.

vjt:
[...]

Issue #739, opened 2026-08-03
cic: three defects in the new auth client — token wipe, forked type, dead mode


vjt:
Found by the codebase-review campaign on `main` (`v0.10.0..388f883b`), `cicchetto/src` slice. Verified against the files at `388f883b`.

Three small defects in the new auth client surface. Grouped because they are one file and one sitting.

## 1. A wrong TOTP code wipes the shared bearer token — `cicchetto/src/lib/api.ts:1440`

`verifyTotpLogin` calls `readError(res)` with the default `fireDeadTokenHandler = true`, so a 401 invokes `on401Handler` → `setToken(null)` → the bearer is removed from `localStorage`. Every passkey helper correctly passes `false` (`:1470`, `:1480`, `:1489`).

**Failure scenario:** a user with a live session in another tab opens `/login`, mistypes their six-digit code once, and the shared-`localStorage` bearer is wiped — the other tab loses its session on next reload.

**Fix:** pass `false` on the unauthenticated login-completion endpoints, matching the passkey helpers.

## 2. `PasskeyOptions` is declared twice — `cicchetto/src/lib/api.ts:165` and `cicchetto/src/lib/passkeys.ts:1`

Two byte-identical declarations added in the same commit. Values flow between them (`auth.ts:111` passes the `api.ts` type into `passkeys.getPasskey`, which annotates its own), so they typecheck only by structural coincidence.

**Failure scenario:** the server adds a required sibling field or renames `challenge_id`. Updating one declaration leaves the other stale; because the two stay structurally compatible for the unchanged subset, `tsc` is green and the mismatch surfaces as a runtime WebAuthn failure on device.

**Fix:** delete the `passkeys.ts` copy and import the type from `api.ts`.

## 3. `startPasskeyModeChange` accepts a mode the server always rejects — `cicchetto/src/lib/api.ts:1536`

The helper types `mode: "second_factor" | "passwordless"`, but `PasskeyController.mode_options/2` has a single clause guarded `when mode == "second_factor"` and a catch-all returning `{:error, :bad_request}`. Passwordless activation is a different two-step route (`/me/passkeys/passwordless/recovery` → `/me/passkeys/passwordless/options`).

**Failure scenario:** a future caller passes `"passwordless"`, gets a bare 400 `bad_request`, and has no hint that the route is simply wrong for that mode — a bug the type invited and the compiler could have prevented.

**Fix:** narrow the parameter to the literal `"second_factor"`.

[...]

Issue #729, opened 2026-08-03
cic: four privileged passkey actions silently consume the add-form password field


vjt:
Found by the codebase-review campaign on `main` (`v0.10.0..388f883b`), `cicchetto/src` slice. Verified against the files at `388f883b`.

**File:** `cicchetto/src/PasskeySettings.tsx:40`, `:55`, `:69`, `:108` (all read `password()`); the input lives at `:155`
**Severity:** MEDIUM

## Problem

Four privileged actions read the `password()` signal: add a passkey (40), enable second-factor mode (55), prepare passwordless mode (69), and delete a passkey (108). Only the first is inside the form that visually owns the "Account password" input; the other three are buttons rendered elsewhere in the pane with no password prompt of their own.

The password is also never cleared after a successful operation — only `setName("")` at line 42 — so it stays live in the signal and in the DOM for the whole drawer session.

## Failure scenario

A user scrolls to their passkey list and clicks "remove" without ever having filled the add-passkey form. `deletePasskey(token, id, "")` → 401, surfaced as a raw wire token (see the sibling issue on `friendlyApiError`), with nothing indicating that a password was expected.

Conversely: a user who typed their password to ADD a passkey and then clicks "use passkey as primary login" has that password silently reused to mint a passwordless recovery set — no re-confirmation for a change to how the account authenticates.

## Suggested fix

Give each privileged action its own confirmation step, or one shared and explicitly-labelled re-auth field that is clearly attached to all of them, and clear the password signal in the `finally` of every action that consumed it.

vjt:
Shipped in **v0.12.0** — https://github.com/vjt/grappa-irc/releases/tag/v0.12.0

Pull that tag to get this. Note it is a tag-only release: irc.sindro.me is deliberately held on the previous release while resource usage is measured over several days, so "shipped" here means "in a named release", not "running on our instance".

PR #1294, opened 2026-08-13
cic(#1283): ask for the account password before starting a TOTP enrolment

vjt:
Fixes the **Enable TOTP** button, which could not work for anyone.

`POST /me/totp/enrollment` has required the account password since `c1657c3b`.
`startTotpEnrollment` went on posting a literal `"{}"`, so the controller head
`%{"password" => password}` never matched, the catch-all answered
`bad_request`, and the pane rendered "The request was malformed." on every
press, for every user, on every substrate.

The server gate stays. Re-authenticating before arming a second factor is the
posture `c1657c3b` argued for: confirming an enrolment revokes every other
bearer session and hands out the recovery codes, so a borrowed token alone
must not arm a factor the real owner does not hold. The client learns to
knock.

## The census

The issue asks whether the same gap exists anywhere else `c1657c3b` touched.
It does not — and here is the enumeration behind that negative, because a
gate is a sample and not a list.

Seven authenticated endpoints require the account password:

| Endpoint | Controller | cic caller | Sends the password? |
| --- | --- | --- | --- |
| `POST /me/totp/enrollment` | `TotpController.start_enrollment` | `startTotpEnrollment` | **no — this defect** |
| `DELETE /me/totp` | `TotpController.delete` | `disableTotp` | yes |
| `POST /me/client-tokens` | `ClientTokenController.create` | **none** | n/a |
| `POST /me/passkeys/registration/options` | `PasskeyController.registration_options` | `startPasskeyRegistration` | yes |
| `POST /me/passkeys/mode/options` | `PasskeyController.mode_options` | `startPasskeyModeChange` | yes |
| `POST /me/passkeys/passwordless/recovery` | `PasskeyController.prepare_passwordless` | `preparePasswordless` | yes |
| `DELETE /me/passkeys/:id` | `PasskeyController.delete` | `deletePasskey` | yes |

`c1657c3b` made **exactly one** of them newly password-gated — the enrolment
start. It also moved `disable_totp/3` onto the new `Accounts.verify_password/2`
without changing that endpoint's contract, which is why nothing else moved.

Two things the census turned up that the issue could not:

* **`POST /me/client-tokens` has no cic caller at all.** The issue's fix
  direction says to prompt for the password "the way the client-token mint
  flow already must" — that flow is API-only. `client_token` appears in
  `cicchetto/src` solely as wire-error copy (`friendlyApiError.ts`); there is
  no mint UI to imitate. The real in-cic precedents are #729's passkey section
  and the TOTP disable form, and those are what this follows.
* **The five passkey callers all route through one gate**, `withPassword`
  (#729), which refuses an empty field locally and clears it afterwards. The
  TOTP section needed the same behaviour, so the gate moved to
  `lib/accountPassword.ts` rather than being copied.

## The half a body-only fix would have broken

A password-gated door answers 401 `invalid_credentials`, and this caller read
its errors with the dead-token handler **armed** — `readError`'s default.
Sending the password without disarming it turns one typo into a logout of
every tab sharing the bearer: the #739 class, arriving this time through an
*authenticated* endpoint. `disableTotp` and every `passkeyRequest` already read
theirs disarmed for exactly this reason. On a re-auth door a 401 means "wrong
password", never "dead bearer".

## Also in here

* The disable field loses its `required` attribute, so the code gate is the
  ONE mechanism refusing an empty password across this pane. A native
  constraint makes the gate's arm unreachable in a browser while leaving it
  live in the section next door: two mechanisms for one rule, only one of them
  testable.
* The #729 comment claimed an empty-field request also spends "a slot in the
  server's login-throttle window". Measured, it does not:
  `Grappa.RateLimit.FailureWindow` guards the login, TOTP-login,
  passkey-login-options and passkey-recovery doors, and none of the
  authenticated re-auth doors records a failure there. The claim is dropped
  rather than carried across.

## Tests

Written before the fix; the component and boundary tests were watched red
first (11 failures), and the e2e — which could only be written after — was
proven by measured mutants instead.

* `api.test.ts` — the request body carries the password; a 401 from this door
  does not fire the dead-token handler.
* `TotpSettings.test.tsx` — the typed password reaches the call; an empty
  field is refused locally with no request; the field is cleared after a
  refusal; the field is `type=password` / `autocomplete=current-password`.
* `issue1283-totp-enrolment-password.spec.ts` — real pane, real endpoint. A
  wrong password gets a refusal that is **not** the malformed one, on a pane
  that is still there; the right password reaches the QR and the manual key.
  Enrolment is started and never confirmed, so the spec arms nothing and
  mutates no account state.

Mutants, scoped `--grep "#1283"` on the real stack:

| Mutant | Outcome |
| --- | --- |
| body key `password` → `passwd` (the production defect, in a form that still typechecks) | test 1 dies on `not.toHaveText(/malformed/i)`, `Received string: "The request was malformed."`; test 2 dies with no enrolment form |
| `readError(res, false)` → `readError(res)` | test 1 dies waiting for the alert — the pane is gone, the app logged out; test 2 stays green, correctly, since a right password never reaches a 401 |
| none | 6/6 green under `--repeat-each 3` |

The first mutant is worth a note: reverting the body to the literal `"{}"`
does not reach the browser at all, because `password` then goes unread and
`tsc` fails the bundle build. The mutant had to be re-cut as a wrong KEY to
test what the spec actually watches.

## Gates

* `bun run test` — 5372 passed / 282 files, rc 0
* `bun run check` — biome + `tsc` over `src` and `e2e`, rc 0
* `scripts/design-notes-gate.sh` — rc 0
* `scripts/integration.sh --grep "#1283"` — 2 passed, and 6/6 under
  `--repeat-each 3`

## What is NOT claimed

The **full** integration suite has not been run on this branch — only the
scoped `--grep`. No Elixir gate was run either: the diff touches no `.ex`,
and the only non-cic file is the DESIGN_NOTES entry. The blast radius on the
rest of the e2e suite is a one-line widening of the `SettingsSection` union
in a shared fixture, which no other spec reads.

The sibling issue #1284 is untouched.

### DIFF
--- a/cicchetto/e2e/fixtures/cicchettoPage.ts
+++ b/cicchetto/e2e/fixtures/cicchettoPage.ts
@@ -934,7 +934,8 @@ export type SettingsSection =
   | "watchlists"
   | "aliases"
   | "perform"
-  | "vhost";
+  | "vhost"
+  | "security";
 
 // Open the settings drawer (viewport-aware) and navigate into `section`'s
 // sub-page, returning the sub-page section locator so callers scope assertions
--- a/cicchetto/e2e/tests/issue1283-totp-enrolment-password.spec.ts
+++ b/cicchetto/e2e/tests/issue1283-totp-enrolment-password.spec.ts
@@ -23,7 +23,9 @@
 // read its errors with the dead-token handler ARMED. Sending the password
 // without disarming it would have turned one typo into a logout of every tab
 // sharing the bearer — a regression the fix itself would have introduced.
-// "Still on the security pane, not bounced to /login" is how that is visible.
+// "the refusal is on a pane that is still there" is how that is visible, and
+// it is deliberately ONE assertion rather than two: a separate
+// `pathname !== "/login"` check never fires, because the pane vanishes first.
 //
 // Then the RIGHT password, which has to reach step two: QR and manual key on
 // screen. Enrolment STARTS here and is never confirmed — the returned secret
@@ -59,8 +61,20 @@ test.describe("#1283 — TOTP enrolment asks for the account password", () => {
     await form.getByLabel("Account password").fill("definitely-not-the-password");
     await form.getByRole("button", { name: "enable TOTP" }).click();
 
+    // Two claims in one wait, and the message says so because they fail the
+    // same way. There IS a refusal on the pane — and the pane is still THERE,
+    // which is the survival claim: the door is authenticated, so a wrong
+    // password answers 401, and reading that 401 with the dead-token handler
+    // armed clears the bearer and bounces RequireAuth to /login. Measured:
+    // flipping `readError(res, false)` back to the default kills exactly this
+    // assertion, with the pane gone and nothing to find.
     const alert = pane.getByRole("alert");
-    await expect(alert).toBeVisible();
+    await expect(
+      alert,
+      "no refusal on the TOTP pane — if the pane itself is gone, the 401 was " +
+        "read as a dead bearer and the app logged out (readError must stay " +
+        "disarmed on a re-auth door)",
+    ).toBeVisible();
     // The regression pin. Before the fix EVERY press landed here, whatever
     // was typed, because the request carried no password at all.
     await expect(alert).not.toHaveText(MALFORMED);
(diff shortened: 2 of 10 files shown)


vjt:

Where the work landed: `d421f008`, `143e39a5`, `4bc9eaf2`, `f901c364`, `73f7e415`

Issue #1196, opened 2026-08-10
Per-client tokens: 2FA on an account locks out every headless client


abonforti:
## Problem

Enabling TOTP or a passkey on an account locks out every headless client that authenticates with account name plus password, because `/auth/login` is the only door and it has no non-interactive path.

`AuthController.account_login/3` branches like this after the password check:

| Account state | Response |
|---|---|
| `passkey_mode: :passwordless` | `{:error, :invalid_credentials}` |
| `passkey_mode: :second_factor` | `202`, `two_factor_required: true` + passkey options |
| TOTP enabled | `202`, `two_factor_required: true` + `challenge_token` |
| none of the above | bearer minted |

A browser resolves any of the first three. A headless client cannot:

- **TOTP** needs a code that rotates every 30 seconds. A client that stores its credential in a config file has nowhere to put one, and no channel to be prompted on: it authenticates once at startup and again on every reconnect, unattended.
- **WebAuthn** needs an authenticator and an origin binding. `OPERATIONS.md` already states this for shottino.
- **Recovery codes** are single-use by design. Burning one per reconnect is not an access mechanism.

The practical consequence today is a hard either/or: 2FA on the account, or a headless client. For an always-on deployment that is a real cost, because the account most worth protecting is exactly the one a client needs.

## Proposal

A per-client token, minted from a session that has **already** passed the second factor, accepted in place of the password at login. This is the app-password model (Gmail) or the personal access token model (GitHub): the token is not a 2FA bypass for the account, it is a separate credential whose issuance required 2FA.

The primitive already exists. `Grappa.Accounts.Session` is a bearer-token row whose `:id` *is* the token, and it already carries `client_id`, `last_seen_at` and `revoked_at`. A client token is close to that row with a different lifecycle and a label.

Shape:

- created only from an authenticated session that cleared the second factor
- carries an operator-supplied label, so the device list is readable
- shown once at creation, never retrievable again
- revocable individually, and revoked implicitly by `reset-passkeys` / `reset-totp` if that is the desired posture
- listed with `last_seen_at` and last source IP, so an unused or unexpected one is visible
- rate-limited on use exactly like a password

### Where it is accepted

The cheapest integration is for `/auth/login` to accept the token in the `password` field and, on a match, mint the session directly, skipping the `cond` above. A client keeps sending the credentials it already sends, so **no client-side change is required** for anything that talks to grappa today.

[...]

PR #1295, opened 2026-08-13
auth(#1284): stop the other-sessions sweep from killing client tokens

vjt:
Refs #1284. Ruling: **exclude** (vjt, on #grappa 2026-08-13 18:37), extended to all three call-sites (19:18).

## The defect

`revoke_other_sessions_for_user!/2` swept an account's live bearers by `user_id` alone. A per-client token (#1196) is a `Session` row like any other, so arming TOTP revoked the exact credential the feature exists to provide — and the order `CLIENT_PROTOCOL.md` §7 puts an operator on (mint the token, hand it to the headless client, then arm the factor) was the order that broke. Silently: the enrolment succeeds, the recovery codes print, and the bridge starts answering a `401` indistinguishable from a wrong password.

## The fix

The sweep now carries `s.kind != :client`. The filter lives in the function, not at the call-sites: an opt-out argument would put a security posture in three places where it can drift, and a default argument would make the safe answer the one nobody typed.

`!= :client` rather than `== :web` so a future third kind is still swept — fail-closed is the right default for a revoke, and `:client` is the only kind with an argument for surviving.

## Census

Three call-sites, all covered, because they are one class rather than three cases — each is reached by an account holder who has just proven they hold the account and is changing their factors precisely so the headless credential keeps working:

| call-site | door |
|---|---|
| `accounts.ex` `confirm_totp_transaction` | `POST /me/totp/enrollment/confirm` |
| `accounts.ex` `disable_totp_transaction` | `DELETE /me/totp`, under password |
| `webauthn.ex` `set_mode_transaction` | `POST /me/passkeys/mode`, under assertion |

Operator recovery (`reset_totp` / `reset_passkeys`) and admin password rotation are the opposite door — nobody proved anything there — and they already go through a **different** function, `revoke_sessions_for_user!/1`: whole-account, no `other`, no `kind` filter. Untouched, and a test pins that the filter did not leak into it.

The census also covers the write side, not only the call side: four sites in `lib` write `revoked_at`, and the other three (`revoke_session/1`, `revoke_sessions_for_visitor/1`, `revoke_sessions_for_user!/1`) are all correct as they stand.

## Evidence

Red before green, then each assertion bought with a measured mutant:

| mutant | kills |
|---|---|
| no `kind` filter (the pre-fix code) | `the token survives` — 3 failures |
| filter that matches nothing | `the browser bearer still dies` — 3 failures |
| `kind` filter leaked into `revoke_sessions_for_user!/1` | the two recovery-side guards — 2 failures |

One mutant, one assertion: none of them is decorative.

`scripts/check.sh` green on this tip — 8 doctests, 59 properties, 6076 tests, 0 failures; Sobelow `Total errors: 0`; Credo clean; `design-notes-gate.sh` reports the entry's separator and marker present.

## Docs

- `docs/DESIGN_NOTES.md` — entry `#1284`, separator and marker in the same commit.
- `docs/CLIENT_PROTOCOL.md` §7 — the paragraph the issue names said revocation by the owner was "the only thing that ends it", which was false before this change and still not exhaustive after it. It now names the operator doors and states that arming a factor does not revoke the token, so the mint-then-arm order is documented as safe.
- `docs/OPERATIONS.md` needed no edit: its client-token paragraph already claimed arming a factor "does not have to lock shottino out any more", which this change makes true rather than aspirational, and its "the two reset verbs revoke client tokens along with everything else" is unaffected.

No wire change, no `cicchetto/` file touched, no e2e spec covers this surface.

### DIFF
--- a/docs/CLIENT_PROTOCOL.md
+++ b/docs/CLIENT_PROTOCOL.md
@@ -356,9 +356,12 @@ Three properties worth designing around:
   may also skip `/auth/login` entirely and present it directly as
   `Authorization: Bearer <token>` / the WS bearer subprotocol (§3a).
 - **It does not expire while idle.** A browser session dies after seven
-  days of silence; a client token does not. Revocation by its owner is
-  the only thing that ends it — expect a `401`, and surface it as "this
-  token was revoked", not as a transient network error.
+  days of silence; a client token does not. Only revocation ends it —
+  by its owner, or by an operator resetting the account's factors or
+  rotating its password. Its owner arming, disarming or changing a
+  second factor does NOT (#1284), so minting the token first and arming
+  the factor afterwards is a safe order. Expect a `401`, and surface it
+  as "this token was revoked", not as a transient network error.
 - **It is scoped.** A client token can read and send as the account, and
   that is all. The account's own credential surfaces — `/admin/*`,
   `/me/totp*`, `/me/passkeys*`, `DELETE /me`, and the token routes
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -41195,3 +41195,54 @@ exemptions and should not grow any.
 **Apply:** when a limit exists to mirror an external system's limit, key it
 on state that system publishes, and let the letter table say what was READ
 at source rather than what is probably true elsewhere.
+<!-- entry #1284 -->
+
+---
+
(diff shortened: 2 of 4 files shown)


vjt:

Where the work landed: `618d0ef8`, `569ad44a`

Issue #1288, opened 2026-08-13
cic: refreshScrollback ingests a 200-row catch-up page one store write at a time — O(page x pane) copies and 200 reactive passes per channel, ~1s of dead main thread on open (#408 follow-up, now with JS stacks)


vjt:
Follow-up to #408, which was closed on 2026-08-07 ("no more user reports so far") after a Perfetto system trace confirmed hypothesis (a) — main-thread burst, not a blocking fetch — but **could not name the code**, because that capture had no Chrome tracing category and therefore no JS stacks.

A new report came in today, this time on **desktop** (Firefox 153, Linux, self-hosted v1.1.0 install), with the same shape: ~1 s of dead buttons shortly after opening cicchetto. The reporter captured a **Firefox Profiler trace with full JS stacks**, which is exactly the granularity #408 was missing. The profile is a whole-browser capture carrying the reporter's own tabs and hostnames, so it is **deliberately not attached** — only aggregate numbers are published here.

## Measurement (reporter's trace, tab main thread, 9.0 s capture)

Event delay on the content-process main thread:

| window | length | peak event delay |
|---|---|---|
| 1 | 1019 ms | 1107 ms |
| 2 | 1047 ms | 1137 ms |
| 3 | 1024 ms | 1114 ms |
| 4 | 270 ms | 367 ms |

Three back-to-back ~1 s ramps: the main thread is effectively saturated for ~3.9 s, which the user experiences as "the buttons freeze for a second".

Where that time goes, sampled over the 3870 samples (1 ms interval) of the busy window:

| bucket | samples | share |
|---|---|---|
| cicchetto bundle on the stack | 2002 | 52% |
| browser-extension content scripts | 1157 | 30% |
| browser internals with neither on the stack (cycle collector, GC, allocator) | 711 | 18% |

Inside the cicchetto share, the attribution is not ambiguous:

- **1917 of 2002 samples (96%) have `refreshScrollback` as the outermost application frame**, entered via `AsyncFunctionNext` (it is the awaited async verb, not a render effect).
- Over the whole 9 s capture, `refreshScrollback` is on the stack for **2169 samples (~2.2 s)**, split into **21 separate bursts of 20–180 ms** — consistent with one burst per channel being caught up, not one long task.
[...]

Issue #408, opened 2026-07-26
cicchetto (mobile): scrollback freezes input for seconds, then jumps, then behaves — hypothesis unresolved (needs perf trace)


vjt:
Reported on #grappa by a user (per channel report, 2026-07-26). **Not yet reproduced by me** — filing so the repro isn't lost, deliberately **not queued** until a trace exists.

## Symptom

On mobile, shortly after opening cicchetto, scrolling the scrollback:

1. UI freezes for several seconds — dragging with the finger does nothing at all.
2. Then it suddenly scrolls by an unpredictable amount.
3. Freezes again for a similar interval.
4. After that, everything behaves normally.

## Environment

- Samsung Galaxy S21 Ultra 5G (2021 flagship), Android.
- Chrome, installed as a home-screen app (default "save site as app" flow → PWA/WebAPK shell).

Worth noting: an S21 Ultra is not a slow device. "Not enough CPU" on its own does not explain this, though a transient burst of work saturating the main thread would be consistent with the reporter's own hypothesis.

## Two competing hypotheses

- **(a) Client-side main-thread burst.** The initial history batch is laid out in one go; the list is not virtualized, so measuring/painting a large batch blocks input until it settles. Consistent with "freeze, jump, freeze, then fine".
- **(b) Blocking history fetch.** The first scroll triggers a fetch of older messages and the UI waits for the response before it can move.

Both produce the same user-visible shape, so the report alone cannot separate them.

## Server-side observation (not proof)

History is fetched in batches of 200:

```
GET /networks/<network>/channels/<channel>/messages?after=<id>&limit=200
```

200 message rows laid out in a single pass is a plausible main-thread burst, which is why (a) is not dismissible in favour of (b).

## What settles it — do this before touching code

Performance trace while scrolling a channel with deep scrollback, either on a real Android device over remote debugging, or desktop Chrome with mobile emulation plus 4–6× CPU throttling. Then check:

- Long tasks during the freeze windows: does the blocked interval line up with layout/recalculate-style work, or with a pending network request?
- Whether the batch size (200) correlates with the freeze duration — try a smaller batch and see if the freeze shrinks proportionally.
- Whether the jump-by-an-unpredictable-amount is scroll-anchoring correcting itself after the batch lands.

Only after that does it make sense to decide between virtualizing the list, chunking the render across frames, or making the fetch non-blocking.

vjt:
## Capturing the trace from the phone alone (no PC)

[...]

PR #1296, opened 2026-08-13
cic(#1288): ingest a catch-up page in one store write, not two hundred

vjt:
Re #1288.

`refreshScrollback` fetched up to `PAGE_LIMIT = 200` rows on every WS join-ok and
ingested them one `appendToScrollback` per row. Each of those calls scans the pane
for a duplicate id, copies the row array, copies the channel map and writes the
Solid signal — so a full page did O(page × pane) array work and ran the pane's
reactive graph **200 times**. It now ingests the page in ONE store write.

## The issue's literal fix diverges — measured, not argued

The issue proposed reusing `mergeIntoScrollback`. I checked it against the loop
before adopting it, and it is **not the same verb**: it applies no S20 ring cap
(deliberately — a scroll-up prepend must not be truncated mid-scroll) and so never
invalidates the `loadMoreExhausted` latch an eviction implies.

I implemented that swap and ran the new specs against it. It **passes both cost
oracles** and breaks four invariants:

| assertion | result under the literal swap |
|---|---|
| ring cap holds at 1000 | ✗ pane grows to 1200 and stays |
| no eviction below the read cursor | ✗ nothing is evicted at all |
| eviction clears the loadMore latch | ✗ latch survives |
| an id repeated within one page lands once | ✗ lands twice |

It would have removed the ring cap from the reconnect catch-up path — one of the
two paths S20 names — and **every pre-existing cap test would have stayed green**,
because they all drive `appendToScrollback` directly. A cost-only test suite ships
this bug.

So `appendPageToScrollback` is the page-shaped verb carrying the cap and the latch,
and `appendToScrollback` becomes its P=1 case: one implementation, both arities.
One row scans and a page indexes — building an id `Set` per live WS message would
add allocations to the hot path this change exists to unburden.

## Measurement — same harness both sides

`scrollbackIngestCost.test.ts`, jsdom, reactive passes / id reads / ms (min of 7):

| pane | page | reactive passes | id reads | ms |
|---|---|---|---|---|
| 200 | 50  | 50 → **1** | 22 450 → **300** | 0.17 → 0.06 |
| 200 | 200 | 200 → **1** | 119 801 → **601** | 0.47 → 0.10 |
| 500 | 50  | 50 → **1** | 52 450 → **600** | 0.17 → 0.08 |
| 500 | 200 | 200 → **1** | 239 801 → **901** | 0.63 → 0.09 |
| 900 | 50  | 50 → **1** | 92 450 → **1 000** | 0.27 → 0.09 |
| 900 | 200 | 200 → **1** | 389 901 → **1 301** | 1.06 → 0.15 |

**Displacement**, page axis at pane 900: ×4 the page (50→200) multiplies the old id
reads by 4.22 (super-linear — each ingested row joins the pane the next row must
scan) and the new ones by 1.30. That is the complexity class changing, not a
constant factor.

Every row is handed to the store with `id` behind a counting getter, so "id reads"
is exact and reproducible rather than sampled.

## What the committed guard asserts, and what it does not

The guards assert the **reactive-pass count** and an **id-read bound**. Neither is
wall clock: a time threshold in a shared-runner suite goes red on host load rather
than on a regression. The ms column above is reported, never asserted — and it is
jsdom, so it excludes the DOM construction the reporter's trace shows as the actual
leaf cost. It understates the win rather than proving it.

A **pane-axis displacement assertion was written and deleted.** Both shapes are
linear in the pane (the loop scans it P times, the batch indexes it once), so
quadrupling the pane moves both numbers — measured ratios 2.86 vs 2.67 — and the
S20 cap pins the pane at 1000, so a "4× pane" is not one. It passed against the
unfixed code, which is the definition of an assertion that is scenery rather than
coverage.

## No e2e, and the number that settles it

No e2e spec, and no `MutationObserver` batch-count oracle either. The right guard
already exists and is cheaper: the unit assertion that a catch-up page costs
**1 reactive pass instead of 200**. That assertion is not decorative — restoring
the per-row loop was run as a mutant and it fired, together with both id-read
bounds. A regression to the loop is caught.

What an e2e would have to threshold is wall clock, and the wall clock does not
survive its own noise. Measured on the **fixed** path, pane 900 / page 200, 40
trials, jsdom, idle host:

| min | p50 | p95 | max | mean | sd |
|---|---|---|---|---|---|
| 0.111 | 0.171 | 0.276 | **0.917** | 0.187 | 0.123 |

The entire improvement being claimed is **0.91 ms** (1.06 → 0.15 on minima). The
fixed path's own trial-to-trial range is **0.81 ms**, and one fixed trial reached
**0.917 ms — within 14% of the unfixed best case of 1.06 ms**. A threshold placed
between the two flips on jitter alone.

And that is the *friendly* measurement: jsdom, no browser, no network, no ircd —
and **no rendering at all**. The DOM construction/teardown and cycle collection
that the reporter's trace names as the leaf cost appear in none of the numbers
above, so an e2e threshold would be set against a quantity nothing measured here
bounds. The pass count is the honest proxy for that half, and it is already
pinned at 1.

## Claims I am refusing to make

- **This does not "fix the user's freeze".** It removes a measured O(page × pane)
  cost on the verb that owns 96% of the cicchetto samples in the trace. Whether the
  remaining time disappears is not something this PR measured.
- **The #408 sleep-duration correlation stays consistency, not measurement.** The
  issue says so itself — nobody has profiled across sleep durations — and a
  plausible mechanism is not a reproduction.
- **The 30% in extension content scripts is not a defect of theirs.** It is
  amplification of our own DOM churn, measured on one reporter's browser.
- **Not a v1.1.0 regression.** `scrollback.ts` is byte-identical between the tags;
  nothing was bisected.

## Gates

- `bun run check` (biome + tsc + e2e tsc) — RC 0
- `bun run test` — 283 files, **5392 tests**, RC 0, run on the rebased tip so the
  gate covers the union with the `#201`/`#359` cic commits that landed underneath
- `scripts/design-notes-gate.sh` — 1 new entry, separator and marker present
- Server-side tree untouched: the diff is `cicchetto/` plus the decision log

### DIFF
--- a/cicchetto/src/__tests__/scrollbackIngestCost.test.ts
+++ b/cicchetto/src/__tests__/scrollbackIngestCost.test.ts
@@ -1,0 +1,347 @@
+import { createEffect, createRoot, on } from "solid-js";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import type { ScrollbackMessage } from "../lib/api";
+import { channelKey } from "../lib/channelKey";
+
+// #1288 — the COST of ingesting a catch-up page, and the invariants that cost
+// is not allowed to buy.
+//
+// `refreshScrollback` used to ingest its REST page one `appendToScrollback`
+// per row. Each of those is a duplicate SCAN over everything the pane already
+// holds, a copy of the whole row array, a copy of the channel map, and its own
+// Solid signal write — so a 200-row page ran the pane's reactive graph 200
+// times and did O(page x pane) work. A reporter's Firefox Profiler trace put
+// 1917 of 2002 cicchetto samples under that verb, ~1 s of saturated main
+// thread per burst.
+//
+// Two oracles here, both DELIBERATELY host-independent — a wall-clock
+// threshold in a shared-runner suite goes red on load rather than on a
+// regression, so the committed guard measures WORK, not time:
+//
+//   * reactive passes — an effect on the store signal counts one tick per
+//     store write. The honest user-visible unit: each pass is a render the
+//     browser pays for. 200 -> 1.
+//   * id reads — every row is handed to the store with `id` behind a getter
+//     that counts. The dedupe scan reads every loaded row's id once per
+//     ARRIVING row in the per-row shape (O(P*N)) and once in total in the
+//     batched one (O(P+N)). The number is exact, reproducible and moves with
+//     the mutation: restore the loop and it goes back up by two orders of
+//     magnitude on a full page.
+//
+// Both are asserted with DISPLACEMENT — the same measurement at several page
+// and pane sizes — because a single before/after pair cannot tell a constant
+// factor from a changed complexity class.
+//
+// The third block asserts what the fix must NOT cost. The issue proposed
+// reusing `mergeIntoScrollback` (already batched, in this same module); it is
+// NOT the same verb. It deliberately applies no S20 ring cap (a scroll-up
+// prepend must not be truncated mid-scroll) and so never invalidates the
+// loadMore exhausted latch either. Swapping it in would have removed the ring
+// cap from the reconnect path — the very path S20's own comment names — with
+// every existing cap test still green, because they all drive
+// `appendToScrollback` directly.
+
+vi.mock("../lib/api", () => ({
+  listMessages: vi.fn(),
+  listMessagesAfter: vi.fn(),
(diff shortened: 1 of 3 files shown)


vjt:

Where the work landed: `dac00558`

PR #1324, opened 2026-08-14
cic(#1229): bound the unread exemption so scrollback retention stops growing

### LINKED ISSUE #1229: cic(pwa): returning to the installed PWA restores the shell at half height — always on staging, never on prod
Reported by vjt (#grappa, from an iPhone on staging `b0d5342d`): «a barra settings aperta faccio app switch e torno indietro, trovo l'app a metà altezza come se fosse attivata la keyb ma la keyb non c'è».

That is the **exact symptom of #649** (closed 2026-08-02, "returning from app-switch leaves the shell at half height, keyboard closed"), with one detail that was not in the original report: **the settings drawer was open** across the app-switch.

## What is already in place (origin/main `b0d5342d`)

`cicchetto/src/lib/viewportHeight.ts` is the single writer of `--vh` / `--viewport-height`, and #649's fix was *more triggers into that one writer*:

- `visualViewport` `resize`
- `visibilitychange` (→ visible)
- `pageshow`
- window `focus`

each running `writeAndSettle`, i.e. an immediate write plus re-reads at **100 / 400 / 900 ms** (`SETTLE_REREAD_DELAYS_MS`).

So the shell is stuck at the keyboard-open height *despite* those four triggers. Whatever is happening is outside what #649 fixed.

## Not measured — what to establish first

Do not patch on a guess; this surface has already produced #79, #209, #649 and an epic (#654), and its own comments warn that a second writer of the vars is the recurring failure. Open questions, in order:

1. Is this Safari-tab or the installed PWA? #649 was reported on the installed PWA; the return path (and which of the four triggers fires) differs.
2. Does *any* trigger fire on this return? `AdminDebugTab` already renders `--viewport-height` live (`AdminDebugTab.tsx:209`) — read it while reproducing, before and after the switch.
3. If a trigger does fire, is the settle window simply too short — i.e. iOS restores the visible viewport **after** 900 ms, so all three re-reads sample the shrunk height and the last write wins forever?
4. Does the open drawer matter, or does it only make the bug *visible* (the drawer is `height: var(--viewport-height, 100dvh)`, `default.css:3971`, so it renders the stale value in the most obvious way)? Try the same app-switch with the drawer closed.

## Wanted

Returning from an app-switch restores full height with no keyboard on screen, drawer open or closed. If the answer is the settle schedule, extend/repeat it inside the existing writer — the file's own rule: more triggers into the one writer, never a second writer.

vjt:
## What this is

S20's per-channel ring cap (`SCROLLBACK_RING_CAP = 1000`) exempts every row at
or after the read cursor, because the in-pane `── N unread ──` divider anchors
on that boundary. The exemption had no ceiling of its own, and the module said
so in its own comment: a channel the operator does not read holds ALL of its
unread rows, so the cap does nothing at all. This gives the exemption a
ceiling.

## Why it is worth doing — measured, not estimated

Measured on `origin/main` before choosing anything:

- **The pane renders EVERY retained row.** `<For each={rows()}>`
  (`ScrollbackPane.tsx`), no windowing, no slice, no `IntersectionObserver`
  anywhere in the render path: 1084 retained rows produce **1084**
  `.scrollback-line` nodes, ratio **1.0000** (day separators only ever ADD
  rows). There is no screen-bounded ceiling on the DOM.
- **A rendered row costs 32.3 KB of renderer RSS over the first 500, 18.8 KB
  marginal past that** — chromium, the real theme CSS, 390×844, production row
  markup, `VmRSS` of the `--type=renderer` process, fresh browser per data
  point (0 rows 85 647 KB → 500 rows 101 818 KB → 1000 rows 111 202 KB). The
  same row costs **249 B** as a JS object, measured in JSC.
- So a 1084-unread window is **~26.5 MB in one window**, ~100× the stored
  bytes, against a memory ceiling iOS applies **per web process**.

Retention and rendered DOM are one curve, so bounding the store is the only
lever that bounds the DOM — and the only one of the options considered that
turns an unbounded term into a bounded one.

## The design

**The number is `PAGE_LIMIT`, not a new constant.** `isFarBehind(gap) = gap >
PAGE_LIMIT` already draws the line at "more than one page behind", so past this
bound the client *already* classifies the window as far behind and already owns
the whole apparatus for it: `anchorAtTail` keeps exactly one page, the divider
is suppressed, and the "N unread — jump back" banner plus `jumpToUnread`
rebuild the region from the server that owns it. A second threshold here would
be a second answer to the same question, free to drift from the first. The work
was to REACH that state sooner, not to invent a pruned-window UX.

**A design correction I made to myself halfway, and the most useful part for a
reviewer.** The obvious implementation is one slice: keep the newest page, drop
the rest. It is simpler and it is wrong. Consider the operator scrolled UP
reading history while a busy channel runs on below them: their read cursor sits
at the last row they can see, so everything arriving below is unread, and when
the ceiling bites the one-slice version **deletes the screen they are reading**
and the pane jumps to the tail. This drops the OLDEST unread instead — the rows
just under the divider — which are off-screen BELOW, so `scrollTop` and
everything above it survive untouched. What remains of the read context is then
trimmed by the ring cap exactly as before. It costs one extra slice.

**`missed` accumulates.** After the first bite the store only ever holds one
page, so recomputing the banner's count from the rows would report 200 for ever
while the operator is thousands of rows behind.

**Below the bound nothing changes.** The two tests that pinned the S20
exemption now state their scenario within the ceiling — where that promise
still holds in full — and each gained an assertion that the window did NOT tip
into far-behind.

## The e2e failed three times, and it was right every time

The first red was read as "the spec is wrong". That was only half true, and
saying so up front is cheaper than having a reviewer find it. The rounds were
**different defects at different layers, and every correction stays** — none
made another redundant.

### Round 1 — a FIXTURE defect: the gesture was two rows, not one

The bar read 201 where the spec wanted 200. A peer's JOIN is a scrollback row
like any other: the server emits `:persist :join` for an other-user JOIN
(`event_router.ex`, pinned by `event_router_test.exs:1727`), `subscribe.ts`'s
`routeMessage` appends every message with no kind gate, and `unreadHeld` counts
rows by id with no kind gate either. Issuing the JOIN AFTER the cursor was
planted therefore put a row above it: the protected region crossed the ceiling
on the JOIN, and again on the PRIVMSG, and `missed`'s deliberate accumulation
reported `BOUND + 1`. The number was an artefact of the setup.

Fixed in `57e0408f` by joining during SETUP, behind a barrier that waits for
the persisted JOIN row, so the gesture really is the single PRIVMSG the spec
says it is.

### Round 2 — a PRODUCT defect: the bound pruned the wrong row

With the fixture honest, the spec failed on the assertion that matters:
`.scrollback-line:has-text("seed line #224")` expected 0, received 1. The row
the bound was supposed to prune was still mounted.

**Measured** from the DOM snapshots either side of the gesture in the trace of
CI run `31818106842`:

| | seed rows in the DOM | range | contiguous |
|---|---|---|---|
| before the gesture | 247 | 174..420 | yes |
| after the gesture | 246 | 174..420 | **no** |

Exactly **one** row was pruned, and the only number missing from the range is
**223** — the **cursor row, which is READ**. `#224`, the oldest unread, is
still mounted; so is the row that crossed the ceiling; the total node count is
unchanged (one out, one in).

So the bar was **honest**: `unreadHeld` counts `id > cursor` before the drop,
which was 200, and the DOM held exactly those 200 unread rows. **The
incoherence was never in the count — it was in WHAT left.** `capScrollbackRing`
measured the ceiling against `protectedCount` (`id >= cursor`) and sliced from
`firstProtected`, so the region it capped was one row wider than the unread it
names and the first row it evicted was the boundary row: the newest READ row,
the one the divider anchors on, and the one row this function's own rule
declares unevictable. It was then reported as `unreadDropped` — the signal that
arms the far-behind banner.

### The second defect, which is the one that matters most

The same off-by-one moved the THRESHOLD, and that is worse than the wrong row
leaving.

Biting at `protectedCount > CAP` means biting at **200 unread**. But
`isFarBehind(gap) = gap > PAGE_LIMIT` calls a 200-unread window **near**. So at
exactly 200 unread the two disagreed, observably:

- **live session** — the bound fires, `unreadDropped > 0` arms the far-behind
  state, the banner goes up and the divider is suppressed;
- **after a reload** — `loadInitialScrollback` probes the gap, `isFarBehind(200)`
  is false, the window is classified as near, the anchored fetch runs and **the
  divider comes back**.

Same window, same cursor, two answers. The constant is shared between the bound
and `isFarBehind` **precisely so that cannot happen** — it is written into the
comment at `scrollback.ts:130-137`, that a second threshold would be "a second
answer to the same question, free to drift from the first". The implementation
drifted from it by one row, and only the e2e was positioned to notice.

### The correction

`52103413`. The ceiling is now measured on the **unread region** (`id >
cursor`), the trim slices from `firstUnread`, and the bite happens at
`> CAP` — the same comparison `isFarBehind` makes. The boundary row is never
this bound's to drop.

**Why the unit suite missed it:** every case in it sat a page or more from the
ceiling, where a one-row phase error in the trim is invisible — and the S20
boundary test had been narrowed to `CAP - 1` unread to stay under it. Two new
cases pin the first bite by one: which end the trim takes, and that `=== CAP`
does not bite at all.

**The spec's targets moved with the measured numbers, not its thresholds.** The
crossing now needs 201 unread, so the cursor is planted at exactly one page;
the banner asserts 201, which is the count before the bite and the number the
operator is actually behind. `toHaveCount(0)` was not touched.

### Round 3 — the viewport jumps to the tail, and it is NOT drift

With the right row pruned, the spec got far enough to reach its last assertion:
`scrollTop` must not move (tolerance 3 px, the one the sibling scroll specs
use). It moved **3630 px**.

**The obvious reading — "rows left from above the fold and the content
slid up" — is false, by a factor of 158.** Measured from the trace of run
`31822719344`: exactly ONE row left the DOM (`seed line #223`, the oldest
unread — the fix works), and the container's `scrollHeight` went **5296 → 5273,
i.e. 23 px**. A drift of 3630 px would have needed ~158 rows to leave. **The
pane was not DRIFTED. It was WRITTEN.** `overflow-anchor` appears nowhere in
`src/`, so Chromium's native scroll anchoring is at its default and absorbs a
23 px removal above the fold; there is nothing to compensate for.

The recorded scroll positions give the whole sequence: `1324 → 1078 → 4954`
(7 px short of the bottom). `1078` is the marker's position from mount. The
mechanism, in `ScrollbackPane.tsx`:

1. the marker-activation latch is armed and nothing in this flow clears it (it
   clears on operator input or an own send: `:2578, :3279, :3568, :3627`);
2. on the rows change, the intent gate at **`:3025`** finds a **rendered**
   `unread-marker` and admits `marker-activation`;
3. the dispatch at **`:2974-2976`** writes twice — a SYNC
   `marker.scrollIntoView({block:"start"})` (→ 1078) and then
   `scrollToActivation("marker-or-tail", false)`, deferred to rAF×2;
4. in between, the far-behind state armed by that same rows change **suppresses
   the divider** (`injectMarker`, **`:1512`**);
5. the deferred leg finds no marker, falls into the else arm at
   **`:2290-2298`**, does `tail.scrollIntoView({block:"end"})` → 4954, and sets
   `followMode(true)`.

**The gate and the write read the marker two frames apart.** When one rows
change both admits the intent and suppresses the divider — which is precisely
what crossing this bound does — the intent is honoured against a node that no
longer exists, and the fallback is the tail.

**Attribution: this belongs to the bound as first shipped in `0043fcae`, not to
the correction in `52103413`.** In the previous red, on the pre-correction
product, the same scroll went **1318 → 4933** — the same tail snap. The spec
never reported it because it died earlier, at the pruned-row assertion. The
second correction introduced nothing; it **uncovered** this.

**And it settles a design question this PR raised:** pruning from `firstUnread`
IS compatible with "the viewport does not move" for a reader parked above the
unread region. The prune costs 23 px and native anchoring absorbs it. The two
properties do not conflict — the 3630 px came from the activation, not from the
bound's slice.

### The cure — one `batch`, and two candidates measured and discarded first

The contract taken is that **entering far-behind IS the scroll contract**: the
window promises "you stay where you are, here is jump-back", and the pane must
not contradict the affordance it just rendered. Three ways to get there were
considered; two were measured and thrown away.

**Discarded — re-aim the fallback.** Passing `marker-or-preserve` instead of
`marker-or-tail` at the deferred leg kills the tail snap but **leaves a 246 px
yank**: the dispatch writes TWICE, and the SYNC leg has already pulled the
reader from 1324 to the divider at 1078 before the deferred leg runs. Measured,
not argued — that intermediate position is in the recorded scroll timeline. It
would still fail the spec's 3 px assertion.

**Discarded — deny the intent on the far-behind signal.** Adding
`farBehindByChannel()[k] === undefined` to the gate looks like the surgical fix,
and it is dead code: with the batch in place, **removing that conjunct leaves
266 of 266 tests green.** It survives only by resembling a fix.

**Shipped — publish the transition whole.** `appendPageToScrollback` was writing
the pruned rows and the far-behind flag as **two separate signal writes**
(`setScrollbackByChannel`, then `setFarBehindByChannel`), so Solid ran every
effect of the rows change in a world where the window was pruned but not yet far
behind — a state that never logically exists. That is the state the pane's gate
read. Both writes now go in one `batch`, so by the time the pane's length-effect
runs, Solid's render effects have already taken the divider out of the DOM and
the probe that was always in that gate answers correctly by itself.

**Mutation, one-sided and exact:** neutralising the batch kills exactly one
assertion — "publishes the pruned rows and the far-behind flag in ONE flush",
`[false]` where `[true]` was expected — and leaves the other 265 green.

**A correction to something stated earlier in this PR's history:** the cure was
reported as necessarily living in the scroll-intent machine, with nothing
available inside the bound. That was wrong. The scroll machine reads its inputs
correctly; the store was handing it a state that never existed.

The fix does not touch the shared fallback. `applyActivation` (cold mount,
channel switch) reaches `scrollToActivation` directly and never passes through
`dispatchScrollWrite`, so the fallback-to-tail those two depend on — pinned by
`ScrollbackPane.test.tsx`'s switch and cold-mount cases — is untouched and
green.

## 🔴 This PR does NOT close #1229, and must not be read as doing so

The maintainer's newest reproduction of the reported symptom (the shell coming
back at half height after an app-switch) was captured on staging with the
**ADMIN pane focused, keyboard closed, in a fresh session**. `AdminPane` and
`ScrollbackPane` are **alternative branches of the same `<Switch>`**
(`Shell.tsx:936` vs `:943`), so in that frame there are **zero
`.scrollback-line` nodes mounted**. You cannot prune a list that is not
mounted.

So: this is a legitimate, measured **memory** fix for unbounded scrollback
retention. It is **not** the fix for the half-height frame, and no scrollback
bound can be. The live hypothesis for that frame is the **height recalculation
when a pane is unmounted and remounted by that `<Switch>` on return from
background**, which is a separate investigation. Deliberately no auto-closing
keyword.

## Tests

- **Unit, TDD.** Five cases (`scrollback.test.ts`, `scrollbackIngestCost.test.ts`):
  the all-unread channel is bounded at one page instead of growing without
  limit; the pruned window lands in the far-behind state with an honest
  `missed` and `resumeFrom`; the divider contract is untouched while the unread
  region fits in one page; plus the two boundary cases above — the first bite
  drops the oldest unread and never the boundary, and one page of unread does
  not bite.
- **Mutation-measured, on the revision before the correction.** Disabling the
  bound (one `false &&`) turned exactly the first three red and **nothing
  else** — 70 of 73 in the two files stayed green, so the preserved-contract
  tests demonstrably do not depend on the new code. The two boundary cases were
  proven the other way instead, since the same mutant would also kill them:
  they were run RED against the uncorrected product (`ids` did not contain the
  boundary row; 209 rows where 210 were expected) and green after.
- **e2e** (`issue1229-unread-retention-bound.spec.ts`): one live PRIVMSG
  crosses the ceiling (the cursor is planted at exactly one page of unread, so
  a burst — and the ircd's flood kill with it — is unnecessary). It pins the
  state transition (bar up with the honest total, divider gone), that the
  OLDEST unread row left the DOM while the rows the operator is looking at did
  not, and that `scrollTop` is preserved. That absence assertion is only
  meaningful because the list is not virtualised — the same measurement that
  motivated the fix, and the assertion that caught both defects above.

## What I have NOT verified

- **I have never run the e2e myself.** It has been executed three times, all by
  CI and all red for the reasons dissected above; I hold no stack lane, so the
  round-3 cure is proven by unit test and mutation only — CI is the first run
  that will exercise it end to end.
- **Every measurement in rounds 2 and 3 is read from CI trace artifacts** — DOM
  snapshots, recorded scroll positions, and the geometry the spec itself
  evaluated — plus source reading. None of it was reproduced locally, and no
  cure for round 3 has been written or measured.
- Nothing here is measured on iOS: the per-row figures are desktop chromium and
  the JS-object figures are JSC via bun. Engine families match, builds and
  allocators do not — treat them as magnitudes and ratios.
- I do not claim memory pressure explains any reported symptom. This PR bounds
  a measured unbounded structure; that is all it claims.

Local gates run: `bun run check` (biome + tsc + e2e tsc) green, full cic unit
suite green (5464), `scripts/design-notes-gate.sh` green.

### DIFF
--- a/cicchetto/e2e/tests/issue1229-unread-retention-bound.spec.ts
+++ b/cicchetto/e2e/tests/issue1229-unread-retention-bound.spec.ts
@@ -44,6 +44,29 @@
 // pane left at the tail would advance the cursor to the newest row, unread
 // would fall to zero, and the ceiling would never be approached.
 //
+// ── Why the peer joins during SETUP, and not as part of the gesture ──────
+//
+// A peer's JOIN is a scrollback row like any other. Server side, an
+// other-user JOIN emits `:persist :join` (`lib/grappa/session/event_router.ex`,
+// pinned by `test/grappa/session/event_router_test.exs:1727` — "JOIN-other
+// adds nick to state.members[channel] + emits :persist :join"). Client side
+// nothing filters it back out: `subscribe.ts`'s `routeMessage` appends every
+// message with no kind gate, and `capScrollbackRing`'s `unreadHeld` counts
+// rows by id with no kind gate either.
+//
+// So a JOIN issued AFTER the cursor is planted lands above it, and the
+// "one row" crossing is really TWO: the JOIN takes the protected region past
+// the ceiling and the bound bites once (banner at `unreadHeld`), then the
+// PRIVMSG bites it again — and `scrollback.ts`'s accumulation on the second
+// bite (`current.missed + unreadDropped`, deliberate: after the first bite the
+// store only ever holds one page, so a recount would report the bound forever
+// while the operator is thousands behind) reports `UNREAD_BOUND + 1`.
+//
+// That number is an artefact of the setup, not the behaviour under test. The
+// peer therefore joins BEFORE anything is measured, so its row is part of the
+// read history the cursor is planted into, and the gesture is the single
+// PRIVMSG this spec says it is.
+//
 // ── NOT covered here, deliberately ───────────────────────────────────────
 //
 // The far-behind banner's own behaviour (jump, dismiss, badge) belongs to
@@ -79,6 +102,7 @@ const UNREAD_BOUND = 200;
 // above the bound so the planted cursor has somewhere to sit.
 const SEED_COUNT = 420;
 const SEED_SENDER = "seed-bot";
+const PEER_NICK = "bound1229";
 
 // One short of the ceiling: the protected region is the cursor row plus the
 // rows after it, so `UNREAD_BOUND - 1` unread rows leave exactly `UNREAD_BOUND`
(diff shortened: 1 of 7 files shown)


vjt:
Rebased onto `35333266` and pushed together with the fix for the one red assertion, as a single push.

**The red, and which side was wrong.** The far-behind bar read `201` where the spec expected `200`. The product is right; the spec was wrong, and the reason is not visible from the spec as it stood.

A peer's JOIN is a scrollback row like any other:

- the server emits `:persist :join` for an other-user JOIN — pinned by the pre-existing `test/grappa/session/event_router_test.exs:1727`, *"JOIN-other adds nick to state.members[channel] + emits :persist :join"*;
- `subscribe.ts`'s `routeMessage` appends every message with no kind gate;
- `capScrollbackRing`'s `unreadHeld` counts rows by id, with no kind gate either.

Joining *after* the cursor was planted put that row above it, so the "one row" crossing was really two: the JOIN took the protected region past the ceiling and the bound bit once, then the PRIVMSG bit it again — and the deliberate accumulation on the second bite (`current.missed + unreadDropped`) reported one more than the bound. `201` was an artefact of the fixture, not of the behaviour under test, so pinning it would have left the next person to touch the setup with an unexplained constant. The peer now joins during setup, behind a barrier on the persisted row, and the assertion stays exactly as strict as it was.

**How this diagnosis was reached, stated plainly: by reading code and one existing regression test. Nothing was executed.** The full spec was not run locally — this CI run is the first execution of the correction.

**Prediction, recorded before the run: the bar reads `200` and the spec passes.** If CI reports anything else, the chain above is incomplete and the next step is to resume measuring — not to adjust the assertion.

vjt:
For the record, since GitHub closed this on its own when the union landed: the content of this PR went in via #1333, the 1.2.0 union, on main as `716eb672` (range `6e555382..716eb672`, 10 commits). Its commits were replayed onto a fresh branch off `origin/main` at their own merge-base and gated once together with #1329 and #1332 — each of the three was green only against its own base, and a green attests to branch + base, never branch + the other branches.

Issue #1325, opened 2026-08-14
push: reap a subscription after 5 consecutive non-retryable 4xx when it has never delivered


vjt:
## Problem, measured on staging

A subscription created with a stale VAPID key (#1323) is rejected forever with `400`, and nothing
ever removes it. Measured on the staging DB today: `push_subscriptions` holds two rows for the same
user and device — the one created at 14:55Z, which returns `400` from `web.push.apple.com` on every
single send, and the one created at 15:12Z after a PWA reinstall, which delivers. The dead row costs
one wasted HTTP/2 request to Apple per notification, forever.

Today the only terminal is `{:error, :gone}` — `lib/grappa/push/sender.ex` sweeps via
`Push.delete_dead/1` on `410`/`404`. Every other failure lands in the `{:error, {:http_error,
status}}` branch at `lib/grappa/push/sender.ex:323-331`, which logs and returns. Nothing counts, so
nothing ever reaps.

## Ruling from the maintainer

Decided in channel, and it constrains the shape of the fix:

- **No parsing of vendor strings.** The `reason` body (`VapidPkHashMismatch` and friends) is for
  logs and human diagnosis only — see #1321 and #1323. Reaping must key off the numeric status.
- **Not on the first `4xx`.** A `4xx` is not proof the subscription is dead: `401`/`403` fire when
  our own VAPID JWT is wrong (bad subject, clock skew), `413` when the payload grows, `429` means
  explicitly *retry later*. Reaping on the first failure means one bad deploy of ours silently
  unsubscribes every healthy user — and they do not come back on their own, they have to redo the
  reinstall dance by hand.
- **N = 5**, chosen by the maintainer.

## Proposed rule

Drop a subscription when **both** hold:

1. it has **never** delivered — `last_used_at IS NULL` (the field already exists on
   `Grappa.Push.Subscription`, and `Push.touch_last_used/1` sets it on every `:ok` delivery), and
2. it has accumulated **5 consecutive** `4xx` responses, **excluding `408` and `429`**, which are
   retryable by definition. Excluded by number, not by string.

Clause 1 protects a subscription that works from being reaped during an outage of ours. Clause 2
protects a fresh subscription from a transient failure. Together, the dead row above dies on its
fifth notification and no live user is ever at risk.

## What this needs

- A counter on `push_subscriptions` (migration): consecutive qualifying failures, reset to zero on
  any successful delivery — the reset belongs next to the existing `touch_last_used` call.
[...]

Issue #181, opened 2026-07-04
cic: push subscription silently re-disables itself (likely on hot-reload / SW update) — no auto re-subscribe


vjt:
## Report

Push notifications are broken in cicchetto (PWA). Confirmed in-channel by vjt: "in qualche modo non meglio noto abbiamo rotto le push" — **root cause unknown**. Originally surfaced as: channel mentions not push-notified on iOS PWA.

## Suspects / where to look first

- PWA push depends on **service-worker registration**. If the SW fails to register, push silently dies with no user-visible signal. **Check the SW-registration path first** — companion #120 (surface `registerSW` failure) may directly reveal the cause.
- Possible **regression from a recent cicchetto batch** — establish the regression window (when did push last work vs. which deploy).
- Also check: push subscription, notification permission state, VAPID key, and server-side push delivery.

## Open questions (to confirm with reporter)

- iOS-only (as originally reported) or all platforms?
- When did push last work? Correlates with a recent deploy?

## Affected

- `cicchetto/src/main.tsx` (`registerSW`), push subscription / permission flow, server-side push delivery.

Per chan request. Companion to #120 (SW-registration failure surfacing).

vjt:
**Converged from live channel/device testing** (iOS 26.5.2, reproduced on 2 independent devices — vjt + morph).

**Reframe — push delivery is NOT broken.** The subscription silently becomes *deactivated*: the app shows notifications as disabled. Re-enabling in-app re-subscribes and delivery is restored. So the defect is a **silently re-disabling push subscription**, not a delivery failure.

Symptom chain:
1. Push subscription silently drops → UI reflects "disabled".
2. No auto re-subscribe → notifications stop, no error surfaced.
3. Manual re-enable re-subscribes (VAPID) → delivery restored — **until it drops again** (observed re-disabling on a second device).

**Recurrence: confirmed** — one tester saw notifications re-disable themselves after re-enabling.

**Trigger — leading hypothesis (from tester): hot reloads / bundle refresh kill the subscription.** This fits the SW-registration path (#120): a SW re-registration / `skipWaiting` swap that drops the `pushManager` subscription without renewing it kills push silently until the user re-subscribes.

**Investigation direction (answering vjt's "why do they deactivate?"):**
- Does `pushManager.getSubscription()` return null after a SW update / bundle refresh (hot reload), and do we auto re-subscribe (VAPID) when it does?
[...]

PR #1327, opened 2026-08-14
cic(#1323): reconcile the VAPID key instead of recalling it

### LINKED ISSUE #1323: push: cached VAPID public key never revalidated — a key rotation silently kills push on every installed client
## Symptom

Web push from staging (`1.1.0-1e36a5fa`) is never delivered. Every attempt is rejected by the push
service: `400` from `web.push.apple.com`, and earlier the same day `403` from FCM for an Android
subscription.

## Root cause, measured

Probing the **real** Apple endpoint by hand with a freshly signed VAPID JWT returns:

```
400 {"reason":"VapidPkHashMismatch"}
```

Apple binds a subscription to the hash of the VAPID public key that created it. The mismatch means
the browser subscribed with a **different** application server key than the one the server now signs
with. FCM's `403` on the Android subscription is the same failure expressed in Google's vocabulary.

The server is not at fault, and this was checked rather than assumed:

- `GET /push/vapid-public-key` serves **exactly** the key the server signs with (compared byte for
  byte against the running container's `VAPID_PUBLIC_KEY`).
- The keypair is internally consistent — deriving the public point from the private scalar
  reproduces the advertised key.

## Where the stale key comes from

`cicchetto/src/lib/push.ts` caches the key in `localStorage["cic.vapidPublicKey"]` and
`getVapidPublicKey()` returns the cached value unconditionally when present. The only path that
refreshes it is `forceRefresh`, documented as "used after an `InvalidApplicationServerKey`
exception".

That exception is the wrong trigger to rely on. The browser raises it when an *existing* subscription
conflicts with a *new* key passed to `subscribe()`. In the failure we have, the client reads the
**old** key out of localStorage and subscribes with it — so the browser sees nothing inconsistent,
raises nothing, and produces a perfectly valid subscription that the server can never sign for. The
client believes push is enabled. The server logs a rejection the user never sees. Nothing reconciles
the two, ever.

This makes any VAPID rotation a silent, permanent break for every already-installed client, and the
worse property is that it is invisible from the client side: there is no state in which the UI says
anything is wrong.

## Fix

Reconcile the cached key with the served one instead of trusting the cache:

- On subscribe (and ideally on app start when push is enabled), fetch `/push/vapid-public-key` and
  compare it with the cached value. On difference: clear the cache, `unsubscribe()` the existing
  `PushSubscription`, subscribe again with the new key, and re-register with the server.
- Consider dropping the cache entirely — it saves one small GET against a class of bug that is silent
  and permanent. If it stays, it must be validated, not merely populated.
- Keep the `InvalidApplicationServerKey` recovery path; it is correct, it is just not sufficient.

Worth checking while in there: the server currently deletes subscriptions on `410`/`404`
(#1290 made `404` terminal too). A `VapidPkHashMismatch` is also permanently dead for that
subscription — it can never succeed — but it arrives as a `400`, which is not swept. Decide whether
that specific reason should sweep as well; reading the reason requires #1321.

## Immediate operator workaround

Clear `cic.vapidPublicKey` from localStorage (or the site data / reinstall the PWA) and re-enable
push. Verified as the mechanism, not yet verified end to end as a cure.

## Not verified

That a re-subscription with the current key actually delivers. The mismatch is measured; the cure is
inferred from it.

vjt:
## What was broken

`cicchetto` cached the VAPID public key in `localStorage["cic.vapidPublicKey"]` and returned it
unconditionally forever. The only refresh path was an `InvalidApplicationServerKey` exception — which
**cannot fire on this failure**: the browser raises it when an *existing* subscription conflicts with
a *new* key handed to `subscribe()`, and here the client handed it the *old* key it had just
recalled. Nothing looks inconsistent from the browser's seat, so it mints a perfectly valid
subscription the server can never sign for.

Measured on staging (#1323): `400 {"reason":"VapidPkHashMismatch"}` from `web.push.apple.com`, and
`403` from FCM for the same failure in Google's vocabulary. The server was cleared by measurement —
`GET /push/vapid-public-key` serves byte-for-byte the key the container signs with.

## What changed

- **`cic.vapidPublicKey` is re-meant, not deleted**: it now records *the key the LIVE subscription was
  created with*, written **only after a subscribe succeeds** and cleared with the subscription. A
  comparison anchor, never a source.
- **`getVapidPublicKey` → `fetchVapidPublicKey`**: always a round-trip, no recall path left to misuse.
- **Reconciliation on the ensure seam**, not only at the toggle. `enablePush` needs a user gesture a
  working-looking client never gets, so the check sits in `ensurePushSubscription` (boot /
  `controllerchange` / `visibilitychange`, wired by #181). On `recorded !== served` it unsubscribes and
  re-subscribes with the fresh key, reporting `rekeyed` — a new `EnsurePushOutcome`, distinct from
  #181's `renewed`.
- **`InvalidAccessError` recovery corrected**: re-fetching the key cannot clear a subscription bound to
  a *different* key. The retry now unsubscribes the incumbent.

**No migration needed**, and that is why the storage key was re-meant rather than replaced: the old
cache was only ever populated at subscribe time, so a client broken by the rotation already holds
exactly the key it subscribed with. The installed fleet heals on its first seam.

**Ordering matters**: recording the served key at *fetch* time (as the old cache did) would rebuild
the silence one seam later — a failed re-subscribe would leave `recorded == served` and the next pass
would read that as health. A test pins it.

## The replaced row is named, and that is not a reap

The `rekeyed` POST carries **`supersedes: <old endpoint>`, #181's field** — on the wire since then,
implemented server-side by `Grappa.Push.insert_superseding/3`, and already sent by the `renewed`
branch it shares a tail with. Same problem, same solution; no new field, no server change.

This is the **opposite** of the reap closed as `not planned` in #1325. There the *server* would infer
a row's death from a run of 4xx, which requires parsing a vendor error string to mean anything. Here
the *client* declares the single row it is itself replacing, at the moment it replaces it. **No status
is matched anywhere, and a `400` still sweeps nothing.**

## Cost accepted, stated so it is not read as an oversight

**One `GET /push/vapid-public-key` per resume seam** for opted-in users, single-flighted by
`pushResubscribe`. **Deliberately unthrottled**: a throttle would be duplicated state that drifts,
guarding a request smaller than the headers carrying it. Size is an **estimate, not a measurement** —
~150 bytes, from the ~88-byte base64url key plus the JSON envelope. Nobody has counted the bytes on
the wire.

## Verification debt — read this before closing

The defect is **measured**; the automatic cure is **deduced**.

What #1323's second comment proves is narrower than it looks: a maintainer **reinstalled the PWA by
hand** and the fresh subscription delivered. That establishes *a re-subscription with the current key
delivers* — it does **not** exercise this code path. **Nothing here has run on a real device.** What
remains unproven is that the unattended `unsubscribe()` → `subscribe(fresh key)` on the ensure seam
produces a subscription the push service accepts, on iOS in particular, where the SW lifecycle is
where #181's silent drop came from in the first place.

**A real-device verification is required before this closes.** No `Closes #1323` on purpose.

## Test coverage

vitest, TDD (red first — the rotation cases returned `'present'`, the exact silence):

- rotation on the ensure seam → `rekeyed`, live subscription unsubscribed, `subscribe()` called with
  the *served* key bytes, record updated
- **the `rekeyed` POST carries `supersedes: <old endpoint>`** — proven non-vacuous by mutation:
  deleting the field from the shared tail fails exactly two assertions, this one and #181's
- matching key → `present`, revalidated (one GET) but nothing torn down
- **re-subscribe fails → the recorded key is left alone** (the ordering trap above)
- `enablePush` drops a subscription bound to a superseded key
- `InvalidAccessError` → incumbent unsubscribed, retry succeeds
- `fetchVapidPublicKey` never reads or writes storage

`bun run test` 5457/5457 green, `bun run check` clean. **No e2e**: an honest one needs the stack
restarted with a *different* `VAPID_PUBLIC_KEY` mid-spec, and even then it cannot assert delivery
(Apple/FCM are unreachable from CI). An e2e that does not rotate the key would assert nothing.

### DIFF
--- a/cicchetto/src/__tests__/push.test.ts
+++ b/cicchetto/src/__tests__/push.test.ts
@@ -1,11 +1,11 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import {
-  clearVapidPublicKeyCache,
   deletePushSubscription,
   disablePush,
+  enablePush,
   ensurePushSubscription,
+  fetchVapidPublicKey,
   formatDeviceActivity,
-  getVapidPublicKey,
   listPushDevices,
   type PushDeviceSummary,
   postPushSubscription,
@@ -31,9 +31,12 @@ const sample = {
   publicKey: "BJk1234567890abcdefghijklmnopqrstuv-_wxyzABC",
 };
 
+// #1323 — the key the operator rotated TO. Same length class as `sample`
+// so both decode cleanly through `vapidKeyToUint8Array`.
+const ROTATED_KEY = "BRotated9876543210zyxwvutsrqponmlkjihgfe-_AB";
+
 beforeEach(() => {
   localStorage.clear();
-  clearVapidPublicKeyCache();
 });
 
 afterEach(() => {
@@ -46,6 +49,9 @@ afterEach(() => {
 // (pushManager / Notification / fetch); the real push.ts handlers run.
 const SUB_ID_KEY = "cic.pushSubscriptionId";
 const SUB_ENDPOINT_KEY = "cic.pushSubscriptionEndpoint";
+// #1323 — the key the LIVE subscription was created with (not a read-through
+// cache: nothing is ever served from it).
+const SUBSCRIBED_KEY = "cic.vapidPublicKey";
 
 function fakeSub(endpoint: string): PushSubscription {
   return {
@@ -80,14 +86,34 @@ function stubPushEnv(opts: {
   return { getSubscription, subscribe };
 }
 
+// The body of the `POST /push/subscriptions` the run made, or null when it
+// registered nothing.
+function postedSubscription(
+  fetchMock: ReturnType<typeof vi.spyOn>,
+): { endpoint?: string; supersedes?: string } | null {
+  const call = fetchMock.mock.calls.find(
+    (args: unknown[]) =>
+      args[0] === "/push/subscriptions" && (args[1] as RequestInit | undefined)?.method === "POST",
+  ) as [string, RequestInit] | undefined;
+  return call === undefined ? null : JSON.parse(call[1].body as string);
+}
+
+// The application server key bytes handed to the FIRST `pushManager.subscribe`
+// call — the assertion that says WHICH VAPID key a subscription was born with.
+function firstSubscribeKeyBytes(subscribe: ReturnType<typeof vi.fn>): number[] {
+  const [options] = subscribe.mock.calls[0] as [PushSubscriptionOptionsInit];
+  return Array.from(options.applicationServerKey as Uint8Array);
+}
+
 // Route fetch by URL: VAPID key GET, subscription POST, subscription DELETE.
-function stubPushFetch(): ReturnType<typeof vi.spyOn> {
+// `servedKey` is what the server currently signs with — the axis #1323 turns on.
+function stubPushFetch(servedKey: string): ReturnType<typeof vi.spyOn> {
   return vi.spyOn(globalThis, "fetch").mockImplementation((input: RequestInfo | URL, init?) => {
     const url = String(input);
     const method = init?.method ?? "GET";
     if (url.includes("/push/vapid-public-key")) {
       return Promise.resolve(
-        new Response(JSON.stringify({ public_key: sample.publicKey }), { status: 200 }),
+        new Response(JSON.stringify({ public_key: servedKey }), { status: 200 }),
       );
     }
     if (url === "/push/subscriptions" && method === "POST") {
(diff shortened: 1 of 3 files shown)


vjt:
Superseded by #1330, which landed on main as `6e555382` (range `35333266..6e555382`, 5 commits). #1330 is the union branch: this PR's own commits were replayed onto a fresh branch off `origin/main` at their own merge-base, gated once and merged once. The content of this PR is on main — the replay gave the commits new SHAs, so GitHub cannot close this automatically. Closing by content.

Issue #1321, opened 2026-08-14
push: log the rejection reason, not just the HTTP status


vjt:
## What happened

Web push was exercised against staging (`1.1.0-1e36a5fa`, the #1290 contract live —
`api/config` advertises `push_content_encoding: aes128gcm`). Deliveries were attempted and
**rejected by the push services**: four attempts in two minutes, `403` from FCM and `400`
from `web.push.apple.com`, repeated identically.

The status codes are all we have. `lib/grappa/push/sender.ex:327` logs:

```elixir
{:error, {:http_error, status}} ->
  Logger.warning("push.send http error", status: status, endpoint: sub.endpoint)
```

The response **body is discarded** — and that is exactly where both services write the reason.
FCM returns a JSON error describing whether the VAPID key does not match the one the
subscription was created with, whether the JWT is malformed, or whether the claim set is
rejected; Apple returns a `reason` string. Without it, a rejected delivery is indistinguishable
from any other rejected delivery, and diagnosis degrades to guessing from the number alone.

Concretely, today: the `400` has a plausible cause (`VAPID_SUBJECT` is unset in this
deployment, so `config/runtime.exs:378-383` falls back to `mailto:admin@example.org`, and
Apple validates the `sub` claim strictly per RFC 8292). The `403` has **no** attributed cause
at all, and cannot get one from the current log line.

Ruled out already, measured, so the work does not start in the wrong place: the server's VAPID
keypair is internally consistent — deriving the public point from the private scalar
(secp256r1, uncompressed) reproduces the advertised public key exactly. This is not a broken
key on our side.

## What to do

Carry the response body into the failure log and the error term:

- Log `status` **and** a length-capped body (these are small error documents; a cap in the low
  hundreds of bytes is plenty) plus `retry-after` when present.
- Consider widening `{:error, {:http_error, status}}` to carry the body so callers and tests can
  assert on the reason, not just the number. Check the `ExNudge` return shape first — if it
  already surfaces the body, this is a plumbing change in `sender.ex` only.
- Keep the existing `410`/`404` sweep semantics untouched (#1290 made `404` terminal alongside
  `410`); this issue is about observability on the failures that are **not** terminal.

## Scope note

Setting `VAPID_SUBJECT` on the deployment is a separate, one-line config fix and not part of
this issue. This issue is the code gap that made the cause unknowable in the first place — it
will pay for itself the next time a push service rejects a delivery for a reason nobody guessed.

[...]

PR #1328, opened 2026-08-14
push(#1321): log the rejection reason, not just the HTTP status

vjt:
## What

A rejected web push logged a bare status code. A `400` or a `403` on its own
is indistinguishable from a bad JWT, a stale application server key, a clock
skew or an expired subscription — the services write the reason into the
response body, and reading it turns an afternoon of guessing into a
one-field diagnosis.

`Grappa.Push.VendorLog` is a supervised telemetry sink that logs that reason
next to the status, in the shape the tree already uses for
`Grappa.SessionLog` and `Grappa.DbLatency`.

## Why a sink and not plumbing in the sender

The issue suggested checking whether the library already surfaces the body,
in which case this would be a change in `sender.ex` alone. It does not.
`ExNudge.send_notification/3` binds the response on the failure branch and
returns `{:error, {:http_error, status}}` — the response never leaves the
library by the call path, and its own `@spec` for `send_notifications/3`
promises a response it does not hand back (the same unreliability
`Grappa.Push.Sender` already documents about the library's declared types).

The body leaves by exactly one route: the library's own
`[:ex_nudge, :send_notification]` event carries it as `error_reason` for
every non-2xx it did not classify itself. That is where this attaches.

## 🔴 The ask needs amending on one point: `retry-after` is unreachable

The issue asks for the status, the body **and** `retry-after`. The first two
are delivered; the third cannot be, and not for want of trying: response
HEADERS never leave `ExNudge` by any route it exposes, telemetry included.
There is no bump to wait for either — 1.0.2 (2025-07-28) is the latest
release. Getting `retry-after` needs a change upstream or a different
client, which is not this issue.

## What it deliberately does not log

Terminal and transport failures. The library classifies 410 / 413 and
transport errors itself and passes an atom in place of a response, so those
events carry a nil `http_status_code` and no body at all. An integer
`http_status_code` is therefore the exact discriminator for "a vendor body
exists", and the handler keys on it; those paths already log at the sender
with everything there is to say, so a second line would be noise. The
silence is "no body was carried", not "work was skipped".

## What reaches stdout

stdout persists across restarts and ships out with any log forwarder, so two
rules constrain the line:

- **The host, never the path.** The line records the endpoint's host alone,
  because a path segment can itself be a credential. An operator correlates
  with the sender's own `push.send http error` line for the same delivery.
- **The reason is length-capped and folded onto one line.** It is vendor
  text of unbounded size and arbitrary bytes, so it is made valid UTF-8
  before folding; the cap is what stops an error page from becoming a log
  flood.

Per the maintainer's ruling on #1323 the reason is for human diagnosis only:
no code matches on it, and the 404 / 410 sweep semantics are untouched.

## Test plan

- [x] `test/grappa/push/vendor_log_test.exs` — 10 tests, written before the
      module existed and observed red (4 genuine failures; the other 6
      assertions passed on the telemetry crash report the missing handler
      produced, so they were not trusted and were re-verified by mutation).
- [x] **Mutation**: 7 mutants applied one at a time to the production module
      — drop the status, drop the host, freeze the reason, make the cap a
      no-op, drop the `is_integer/1` guard, match any status, drop
      `String.replace_invalid/1`. All 7 killed, each with clean attribution
      (the reason mutant kills the four reason assertions, the cap mutant
      kills its two; the rest are 1:1).
- [x] `scripts/check.sh` green (format, Credo, Dialyzer, Sobelow, 6190
      tests).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -31,6 +31,7 @@ Grappa.Application
 ├── Grappa.AdminEvents                 (M-11 admin-event ring buffer + telemetry sink)
 ├── Grappa.SessionLog                  (#215 IRC session-lifecycle log sink)
 ├── Grappa.DbLatency                   (#357 SQLite write/query-latency telemetry sink)
+├── Grappa.Push.VendorLog              (#1321 push rejection-reason telemetry sink)
 ├── Grappa.ShareTokens                 (ETS one-shot share-link tokens, both subject kinds)
 ├── Grappa.RateLimit.DailyQuota        (#75 per-(bucket, subject, day) creation quota)
 ├── Grappa.RateLimit.FailureWindow     (S6 per-(bucket, key) login-throttle window)
--- a/config/config.exs
+++ b/config/config.exs
@@ -534,6 +534,12 @@ config :logger, :console,
     :endpoint,
     :status,
     :count,
+    # #1321 — `Grappa.Push.VendorLog`'s rejection line. `:vendor` is the
+    # HOST of the push service that rejected a delivery: the line records
+    # the host only, never the path, because a path segment can itself be
+    # a credential. The reason string itself rides the pre-existing
+    # `:reason` key, length-capped at the sink.
+    :vendor,
     # M-11 admin-events: `:topic` is the PubSub topic that failed (string,
     # e.g. `"grappa:admin:events"`); `:kind` is the typed event-kind atom
     # (`:visitor_deleted`, `:circuit_reset`, etc.) that was about to be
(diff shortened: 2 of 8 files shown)


vjt:
Superseded by #1330, which landed on main as `6e555382` (range `35333266..6e555382`, 5 commits). #1330 is the union branch: this PR's own commits were replayed onto a fresh branch off `origin/main` at their own merge-base, gated once and merged once. The content of this PR is on main — the replay gave the commits new SHAs, so GitHub cannot close this automatically. Closing by content.

Issue #1331, opened 2026-08-14
Reconnect is only reachable from HomePane, not from the window where the failure is seen


vjt:
## The gap

Split out of #406 (closed), whose Q3 was the one question shipped code has not answered.

When a network is `parked` or `failed`, the only control that reconnects it is the `[Reconnect]` chip in `HomePane.tsx:476`. Measured on `origin/main` `35333326`:

- `Sidebar.tsx:350` renders a `reconnecting…` badge — a status overlay from `reconnectingStatus.ts`, with no action attached.
- `PaneTopBar.tsx` carries only the rail hamburger (#881); nothing state-aware.
- `ComposeBox.tsx:112` greys the seam for `parked`/`failed` networks, so the state IS visible where the user is — but the way out is not there.

So the user observes the failure in the channel window (greyed compose, nothing sends) and has to know to navigate to the home pane to recover. In the session that produced #406, that navigation is exactly what the user did not find; they only reached the reconnect control by hunting on another page.

## What this issue asks

Where the reconnect affordance belongs, given the failure is observed in a channel window. Options, not a decision:

1. Attach an action to the state that already renders in the compose seam — the greyed `ComposeBox` says why it is greyed and offers the reconnect inline.
2. Attach it to the sidebar network row, next to the state it already shows.
3. Leave `HomePane` as the only actionable surface, and make the greyed seam point at it explicitly.

Option 3 is the cheapest and still removes the dead end; 1 puts the cure where the symptom is.

## Not in scope

The reconnect mechanism itself (`onReconnect` / the PATCH) works — this is placement and discoverability only.

vjt:
## Ruling

Maintainer call on IRC: **option 2 is out**, options **1 or 3** are both acceptable ("si va bene 1 o 3").

So the sidebar network row is off the table as a placement, and what remains is:

1. inline action on the greyed `ComposeBox` seam — the cure where the symptom is;
3. `HomePane` stays the only actionable surface, and the greyed seam points at it explicitly.

**How the remaining two get decided** (my reading, stated in channel and not contested): the choice is a cost question, not a taste one — if attaching the action to the compose seam does not cost serious prop-drilling, take **1**; if it does, take **3**, which removes the dead end either way. The recon measurement already in flight (prop-drilling hops, existing witnesses, cost per option) is what settles it.

Nothing here decides the implementation; it narrows the placement to two and names the tiebreaker.

vjt:
[...]

Issue #283, opened 2026-07-17
cicchetto: Disconnect button for visitors in main panel, parity with users


vjt:
## Summary

The main panel exposes a Disconnect control for **users** but not for **visitors**. Add a Disconnect button for visitors in the main panel, behaving exactly like the users one (same action, same placement, same confirmation/feedback).

## Decision

- Visitor entries in the main panel get a **Disconnect** button, at parity with users — same control, same look, same behavior.
- Reuse the existing users Disconnect path; do not fork behavior.

## Notes / relations

- Complements the admin per-network Disconnect+Reconnect for visitor sessions added in #269 (Visitors tab). This request is about the **main panel** parity, not the Visitors tab.
- Sibling to #282 (explicit Reconnect button in the vhost/user panel). Independent of #281 — this is a plain UI parity change and can ship on its own.

## Acceptance

- Visitor rows in the main panel show a Disconnect button identical to the users one.
- Pressing it disconnects the visitor session via the same code path used for users.
- No behavioral divergence between the users and visitors Disconnect controls.

vjt:
## Spec-challenge before building (investigation, not a guess)

Scoping #283 against the current code, the premise — *"the main panel exposes a Disconnect for **users** but not **visitors**"* — does not match what's in the tree. Halting to confirm intent before a full TDD+e2e cycle.

### Evidence

**Disconnect is already at full parity for both subjects — there is no user-only Disconnect anywhere.**

- `Sidebar` × and `BottomBar` × (both `ariaLabel="Disconnect <slug>"`) route through the *same* `windowClose.confirmDisconnectNetwork → disconnectNetwork` (`cicchetto/src/lib/windowClose.ts:69-82`). Since **#211 phase 6 ruling D** that verb is subject-agnostic: it `PATCH`es the one network to `:parked` for user **and** visitor alike.
- The "visitor = quitAll / user = park-one" split still written in the *comments* (`Sidebar.tsx:243-248`, `Sidebar.tsx:433-439`, `BottomBar.tsx:36-37`) is **stale** — it describes pre-phase-6 behavior; the code no longer diverges by subject.
- The only subject-gated UI is HomePane welcome-copy + available-networks, and SettingsDrawer detach/delete/vhost/share — **none of them is a Disconnect**.

**The real gap is in the actual main panel, and it is symmetric (missing for both subjects), not user-vs-visitor.**

[...]

Issue #282, opened 2026-07-17
cicchetto: explicit Reconnect button (bottom of vhost panel), same clean teardown as Switch identity


vjt:
## Summary

The vhost / user panel in cicchetto lets a user edit their vhost, but the change is inert until the session reconnects. There is currently no first-class way to apply it. Add an explicit **Reconnect** control that performs the same clean teardown/rebuild as **Switch identity**.

## Decision

- **Explicit button, not implicit.** Do **not** trigger the reconnect implicitly on "back"/panel-close. Reconnect is a heavyweight, externally-visible action (drops the connection, re-JOINs every channel, emits QUIT/JOIN others can see); hiding it behind navigation violates least-astonishment and makes "I only looked" indistinguishable from "apply". Dirty-detection on back is fragile and is exactly the kind of unintended reconnect that caused the self-ban in #281.
- **Placement:** a single sticky button at the **bottom of the list** (footer). One instance only — not duplicated top+bottom (a duplicated disruptive action invites accidental double-fire).
- **State:** disabled until there is a pending change; label should communicate intent (e.g. "Reconnect to apply") so the user understands the vhost only takes effect on reconnect.
- **Consistency:** mirrors the existing "Switch identity" button — same pattern, same visual weight.

## Hard constraint (depends on #281)

The Reconnect action **must** go through the same teardown/rebuild path as Switch identity: a full reset of client-side session state (networks, channels, cached history-fetch state). If it reuses the stale path, it re-triggers the #281 404-storm self-ban. This button is the user-facing trigger for the clean path #281 establishes — land #281 first or in the same change.

## Acceptance

- Explicit Reconnect button, sticky at bottom of the vhost/user panel list.
- Disabled/greyed until a pending vhost change exists.
- On press: clean teardown + rebuild identical to Switch identity; no replay of the previous session's JOINs/history-fetches.
- No implicit reconnect on back/close.

vjt:
**Design decisions (2026-07-18, orchestrator auto-default under autopilot — vjt non-responsive ~10min to the #grappa ping).**

Investigation finding: there is **no literal "Switch identity" button**; the real clean same-account path is the home-page per-network park→reconnect.

[...]

PR #1332, opened 2026-08-14
cic(#1331): offer the reconnect where the operator meets the parked state

vjt:
## The gap

The greyed compose seam already stated that a network was parked. The only way *out* of it was the chip on `HomePane`, which the operator had to know to go and find.

`lib/selection.ts:702-740` softens that but does not close it: it redirects to Home on the park **transition** only, and deliberately so (`:700-703`, *"so the operator can still navigate BACK to a parked window (to view history) without bouncing back to home"*). So three entries still land on a greyed compose with no exit — a cold load with the network already parked (`prev === undefined`, `:726`), a walk back into the window, and another device parking it while this tab was closed. The first two are **pinned as intentional** by `src/__tests__/selection.test.ts:828` and `:853`, and this change leaves both green.

## The number that decided the placement

The ruling was: inline **if** it costs no serious prop-drilling — threshold **≤2 hops and no new plumbing**, otherwise point the seam at HomePane instead.

**Measured: ZERO hops.**

- `ComposeBox` props are `{ networkSlug, channelName }` and nothing else (`ComposeBox.tsx:105-108`); both call sites (`Shell.tsx:708`, `:989`) already pass the slug.
- The state comes from `networkBySlug`, a **module-scope store accessor**, not a prop (`ComposeBox.tsx:440`).
- The verb needs `patchNetwork` + `token`, both plain module imports.

No new context, no new store, no new global, no `Shell.tsx` edit. Zero is below the threshold, so: inline.

## What changed

**The action is gated on the NETWORK cause, not on `greyed()`.** A failed or kicked *window* under a live network has nothing to unpark — its way back in is `/join`. That required splitting `networkGreyedState()` out of the collapsed boolean. The greyed *visual* is unchanged, and the `(not joined)` copy is untouched.

**`lib/networkReconnect.ts` is an extraction, not a second copy.** The PATCH is one line, but the UX around it is not — awaited request, pending latch, friendly-mapped failure — and that trio now has one home instead of two that drift apart. `HomePane`'s `DisconnectedRow` was moved onto it in the same commit.

Each caller keeps its own error **sink**, because both surfaces already own an error line and a second one beside it would contradict the first: HomePane keeps its `role="alert"` span, the seam uses the existing #356 feedback line (sticky, `role=alert`).

Deliberately **not** reused: `lib/reconnect.ts`, which despite its name is the #282 vhost bounce — it selects `connected` networks and does park-**then**-connect. Wrong verb on the wrong set.

**No confirmation modal**, matching the HomePane chip exactly. #283 (2026-07-20) settled that Reconnect is the awaited-PATCH UX while DISCONNECT is the one behind the #195 confirm modal — recorded in the comment on `HomePane`'s `onDisconnect` (`:299-307`: *"match the ×, not Reconnect's awaited-PATCH UX"*). Reconnect is also trivially reversible, which is the property that modal exists to protect. Adding one here would have been a new UX for the verb rather than a reuse of the chip's, and it would have broken the five specs below, none of which has a confirm step after the click.

## New witnesses, and proof they are not decoration

Six tests in `src/__tests__/ComposeBox.test.tsx`, under `#1331 — inline reconnect in the greyed seam`: offered on a parked network, offered on a failed one, **not** offered for a kicked window on a live network, **not** offered when nothing is greyed, the click issues the unpark PATCH for *this* slug, and a rejected reconnect surfaces in the existing error seam.

TDD: the four positives were read RED first, by name — `Unable to find an accessible element with the role "button" and name /reconnect freenode/i`.

The two **negative** tests would be decoration if nothing could kill them, so both were checked by mutation:

| mutant | result |
|---|---|
| gate the button on `greyed()` instead of `networkGreyedState()` | **1 failed / 114 passed** — only the kicked-window test, by name |
| hoist the button out of the greyed seam (render unconditionally) | **2 failed** — both negatives |

The first mutant kills exactly one assertion, and it is the one that encodes the network-vs-window distinction.

## Gates

`scripts/bun.sh run check` RC=0 (biome + `tsc` over `src` and `e2e`). `scripts/bun.sh run test`: **283 files / 5466 tests**, all green — the HomePane suite included, which is the one that would have caught the extraction changing its row's behaviour.

The five e2e specs that press the HomePane chip as their park→reconnect *fixture* (`cp15-b6-parked-disconnect-reconnect.spec.ts:181`, `issue100-reconnecting-badge.spec.ts:92`, `issue248-lusers-no-auto-surface.spec.ts:100`, `issue38-keyed-tab-dismiss-after-rejoin-fail.spec.ts:113`, `issue511-failed-autojoin-dismiss-durable.spec.ts:133`) all resolve it **scoped** to a `.home-pane-network-row-parked` locator, so the seam's identically-named control cannot make them ambiguous. The two never render together in any case — HomePane renders no compose box.

## What this does NOT claim

- **No e2e was written and none was run.** The new affordance is pinned at unit level only. The scoping argument above is read from the five specs' source, not observed in a browser.
- **Nothing was verified in a real browser at all** — no visual check, no mobile viewport, no tap-target measurement. The chip sizes its tap target with `--tap-min` (44px absolute, not rem) but that is a reading of the token, not a measured box.
- The `(not joined)` copy stays wrong for a parked network — the operator never failed to join anything. Flagged during the recon as a separate, smaller change; deliberately not bundled here.
- Bare `/connect` still errors (`slashCommands.ts:561-565`) while bare `/disconnect` defaults to the active network. That asymmetry is untouched.
- No measurement of how often the dead-end entries actually occur; the issue's report is the only evidence that they do.

### DIFF
--- a/cicchetto/src/ComposeBox.tsx
+++ b/cicchetto/src/ComposeBox.tsx
@@ -27,6 +27,7 @@ import { composePlaceholder } from "./lib/composePlaceholder";
 import { diagPush } from "./lib/diagLog";
 import { frameBudgetForTarget } from "./lib/frameBudget";
 import { frameBudgetBaseForNetwork } from "./lib/isupport";
+import { createNetworkReconnect } from "./lib/networkReconnect";
 import { networkBySlug } from "./lib/networks";
 import { routeClipboardPaste, routePastedInput } from "./lib/pasteRoute";
 import {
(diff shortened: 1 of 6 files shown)


vjt:
Superseded by #1333, the 1.2.0 union, which landed on main as `716eb672` (range `6e555382..716eb672`, 10 commits). This PR's own commits were replayed onto a fresh branch off `origin/main` at their own merge-base, gated once as a union and merged once — each of the three was green only against its own base, and a green attests to branch + base, never branch + the other branches. The replay gave the commits new SHAs, so GitHub cannot close this automatically. Closing by content.

Issue #1338, opened 2026-08-15
review: restore unknown-is-never-fatal at three boundaries (M-S2, W-S1, X-S14)


vjt:
Bucket issue from the 2026-08-15 code review (`docs/reviews/codebase/2026-08-15-codebase-review.md`, merged as `3a8eddec`). One theme: a supervised process or a boundary that treats an unrecognised-but-legal input as fatal, against the additive-only contract in CLAUDE.md.

- **M-S2 (HIGH)** — `Session.Server` (`lib/grappa/session/server.ex:1383`, `:1392`, `:3734`) subscribes to `Topic.ws_presence/1` and `Topic.user_settings/1` with 40 `handle_info` clauses and **zero catch-all**. A second tuple broadcast by `Grappa.UserSettings` takes every `Session.Server` for that subject down with a `FunctionClauseError`, dropping the user's IRC connection — caused by an unrelated context, with no compile error. Same for `lib/grappa_web/session_revocation_listener.ex:39`, `:44`, which has exactly one clause. `lib/grappa/irc/client.ex:1219-1222` already carries the loud, logging catch-all to copy.

- **W-S1 (HIGH)** — `MessagesController.parse_int/1` (`:477`) and `parse_limit/1` (`:490`) are defined only for binaries, so `?before[]=1` / `?after[]=1` / `?around[]=1` / `?limit[]=1` decode to a list and raise, producing a 500 with a stacktrace that bypasses `FallbackController` — on a route whose documented contract is "present and unparseable = 400". `parse_after/1` (`:475`) one function away already has the catch-all, as do the sibling controllers.

- **X-S14 (MEDIUM)** — the client hard-drops `web_session_severed` on an unrecognised code, inverting unknown-is-never-fatal on the client side of the same contract.

Each fix inherits an existing, documented behaviour rather than inventing one. Failing test first for each.

vjt:
Closed by the pipeline change ruled by vjt on 2026-08-20: **`status:soon` is retired**. An issue is now closed when its work **merges**, and the **milestone** records which release carries it — the label duplicated what the milestone already said.

This issue was verified individually before closing, not swept: its work is on `origin/main` under commits that are not documentation-only. No behaviour changes here, only bookkeeping.

<sub>Applied by vjt-claude on vjt's instruction.</sub>

PR #1342, opened 2026-08-15
Restore unknown-is-never-fatal at three boundaries (#1338)

vjt:
Bucket from the 2026-08-15 codebase review (issue #1338): three findings, one
class — a supervised process or a boundary that treats an unrecognised but
legal input as FATAL, against the additive-only wire contract. Each fix
inherits a shape the codebase already carries rather than inventing one.

## M-S2 — `Session.Server` and `SessionRevocationListener` had no catch-all

40 `handle_info` clauses, zero catch-all, on a process subscribed to the
ws-presence and user-settings topics. The first extra tuple `Grappa.UserSettings`
broadcasts would take down every session of that subject with a
`FunctionClauseError` and drop a live IRC connection — caused by an unrelated
context, with no compile error. `IRC.Client` has carried the loud, logging
catch-all for the lower-stakes process all along; this copies it.

**The EXIT class is deliberately carved out and stays fatal.** `init/1` traps
exits, so a linked process dying abnormally arrives as a message, and answering
a dead dependency with a log line is exactly the safety net that absorbs the
next class of bug. An explicit `{:EXIT, _, reason} -> {:stop, reason, state}`
sits ahead of the catch-all and preserves what the `FunctionClauseError` did by
accident: `terminate/2` records the Backoff failure and the `:transient`
supervisor rebuilds. REV-D's H12 regression synthesized its non-Client crash
*through* the absence of a catch-all, so it now drives `Process.exit/2` through
that clause — which makes it the alarm on the carve-out.

**`SessionRevocationListener` reverses its own moduledoc**, which argued that a
crash surfaces a second publisher. It buys nothing the warning log does not,
and costs the mailbox behind the unknown message plus — since a second
publisher repeats — the process itself once restart intensity runs out. The
moduledoc now states why the catch-all does not weaken the teardown guarantee:
that guarantee is per-message and stateless, so dropping what it cannot read
and serving the next is strictly more teardown, never less.

## W-S1 — `MessagesController` 500s on list- or map-shaped pagination params

`parse_int/1` and `parse_limit/1` were binary-only, but `Plug.Conn.Query`
decodes `?before[]=1` to a list and `?before[k]=1` to a map with no cooperation
from the client. Four params answered a malformed request with a stacktrace
past the `FallbackController`, on a route documented as "present and
unparseable = 400". `parse_after/1`, one function away, already carries the
catch-all and the count endpoint already has the test.

## X-S14 — the client hard-dropped an unrecognised sever code

The sever's action (latch, clear local auth, drop to login) does not depend on
the code; the code only selects copy. Rejecting an unrecognised one left cic on
a dead shell holding a revoked bearer, waiting for the WS reconnect-failure
path — #630's own stated must-not-happen. `code` is now any string, carried
verbatim; clearing auth is unconditional, only `rate_limit_flood` latches the
dedicated banner, anything else lands on the generic logged-out screen
`Grappa.RateLimit.Wire` already names as the fallback. Widened on the CLIENT
only: the server typespec keeps the closed literal, an honest statement of what
it emits today, and `api.ts` hand-widens the arm the way `recover_result.reason`
already does two arms up.

## Tests

Failing test first for each. M-S2's sends an unmodelled message and asserts the
process survives AND keeps serving (`connection_info/2` behind the send, not
`casemapping/2`, which answers `:ascii` for a dead session and would pass
against the very crash under test); the listener's asserts a REAL revocation
still tears down after the unknown message, because survival alone is satisfied
by a listener that survived by no longer listening. A second server test pins
the EXIT carve-out.

## Verification

- cic gates green: 283 files / 5476 tests, biome + tsc, `RC=0`.
- Three client-side mutants, each killed by exactly the assertions meant for it
  and nothing else: restore the closed-set narrowing (2 failures), latch the
  banner unconditionally (1), drop the string guard (3).
- `scripts/design-notes-gate.sh` green.
- **Not run locally:** the Elixir gates — the COMPILE lane was contended, so
  the server half is gated by this PR's CI rather than by a local run.
- **Not run:** e2e. `issue630-flood-protection.spec.ts` exercises the known code
  only and that path is unchanged by construction, but nothing was observed in
  a browser.

### DIFF
--- a/cicchetto/src/__tests__/userTopic.test.ts
+++ b/cicchetto/src/__tests__/userTopic.test.ts
@@ -1981,18 +1981,33 @@ describe("userTopic", () => {
       expect(flagOrder).toBeLessThan(clearOrder);
     });
 
-    it("drops the event when code is not rate_limit_flood (narrower rejects)", async () => {
-      const fs = await import("../lib/floodSever");
+    // #1338 X-S14 — the sever ACTION (drop to login) does not depend on the
+    // code; the code only selects the copy. An additively-added sever reason
+    // must therefore still bounce the user to the login screen, generic
+    // banner and all, instead of leaving them on a dead shell holding a
+    // revoked bearer waiting for the WS reconnect-failure path (#630's
+    // stated must-not-happen).
+    it("clears auth on an unrecognised code (unknown-is-never-fatal)", async () => {
       const authMod = await import("../lib/auth");
-      channelMock.fireEvent({ kind: "web_session_severed", code: "something_else" });
-      expect(fs.setSeveredForFlood).not.toHaveBeenCalled();
-      expect(authMod.clearLocalAuth).not.toHaveBeenCalled();
+      channelMock.fireEvent({ kind: "web_session_severed", code: "operator_kill" });
+      expect(authMod.clearLocalAuth).toHaveBeenCalled();
     });
 
+    it("does NOT latch the flood copy for an unrecognised code", async () => {
+      const fs = await import("../lib/floodSever");
+      channelMock.fireEvent({ kind: "web_session_severed", code: "operator_kill" });
+      expect(fs.setSeveredForFlood).not.toHaveBeenCalled();
+    });
+
+    // The tolerance is over the VALUE of a required field, not over its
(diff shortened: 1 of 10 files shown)


vjt:

Issue #1341, opened 2026-08-15
cic: after a nick rename the DM listener holds both own-nick topics


vjt:
Found by w1 while building the RED case for #1340 (C-S1), and deliberately left out of that branch's scope.

After a rename, the DM listener ends up holding **both** own-nick topics at once (the pre-rename one and the post-rename one) instead of releasing the old one. Same family as the #200 handler leak: a subscription keyed on an identity that moved, never torn down.

Not yet characterised: whether it leaks once per rename or unbounded across repeated renames, and whether the stale topic still delivers (a duplicate-delivery symptom would look like double notifications rather than a pure resource leak). w1 observed it as a by-product of the #1340 RED run and did not measure it — treat the mechanism as read, not established.

---

## Now characterised — measured by w1, 2026-08-16 01:49

The paragraph above is superseded on both open points:

- **Bounded, not unbounded.** The leak is **one held topic per _distinct_ nick**: 4/5/6/7 keys at 0/1/2/3 renames. Renaming back to a nick already seen adds nothing — reuse does not grow the set. It is never released.
- **No duplicate delivery — but not a pure resource leak either.** The stale listener does not double-deliver; it **misroutes**, dropping the line into the self window (orch's reading of the same measurement, `#grappa-live` 01:52). Calling it a plain leak understates it: a message lands in the wrong window, which is a correctness symptom, not just held memory.

**Fix shape ruled by orch (01:52): migrate the subscription, do not re-key `ownNick`.** Both of orch's prior hypotheses were refused by the measurement.

**Blocking precondition, measured by w1 (01:53).** The own-nick key and a *real query window* can collide, so a blind leave would tear down a live query subscription. The release must therefore be **gated on ownership**, not unconditional.

**Landed shape:** PR #1366, `+219/-0` across `cicchetto/src/lib/subscribe.ts`, its test file and `DESIGN_NOTES.md` — "release the DM-listener own-nick topic, gated on ownership". Deploy classed HOT, cicchetto-only.

**Left open by w1, not part of this issue:** there is no server-side pin for the outbound folded-channel invariant. Unmeasured here — it needs its own issue once someone reproduces it.

*Body updated by vjt-claude from vjt's `gh` account so the issue carries the current state. The measurement is w1's, reported on `#grappa-live`; I did not run it.*

vjt:
Characterised. The mechanism is read as follows, and each claim below was measured rather than reasoned.

[...]

Issue #1340, opened 2026-08-15
review: finish the nick-identity migration set — stale ownNick in per-channel handlers (C-S1), muted_targets outside the #373 set (K-S2)


vjt:
Bucket issue from the 2026-08-15 code review (`docs/reviews/codebase/2026-08-15-codebase-review.md`, merged as `3a8eddec`). One theme: a store or a closure that holds a nick captured before the rename, so a `/nick` (or a NickServ ghost recovery) silently forks identity — the exact class #373 closed for three stores and left open for these two.

- **C-S1 (HIGH)** — `installChannelHandler(phx, slug, name, key, ownNick)` (`cicchetto/src/lib/subscribe.ts:389`, call sites `:914`, `:979`, `:1031`, guarding effects `:871`/`:937`) closes over `ownNick` as a snapshot taken at topic-join, and every install loop is guarded by `if (joined.has(key)) continue;`, so the handler is never rebuilt. `own_nick_changed` → `mutateNetworkNick` (`userTopic.ts:1210`) re-runs the join effects, but the `joined` guard short-circuits every already-joined channel key. Only the DM listener recovers, and only by accident of keying. Consequences after any `/nick`: alerts follow the old nick, own-PART stops being detected, and the #200 handler leak returns.

- **K-S2 (MEDIUM)** — `muted_targets` became a nick-keyed per-conversation store in #1038 (`conversationMute.ts:52`, `:66`) but was never added to the peer-rename set at `lib/grappa/session/server.ex:6111-6147`; `grep -rn "rename" lib/grappa/user_settings.ex` returns zero hits. Mute `guest`, they `/nick Guest2`: the window, the DM scrollback and the cursor all migrate, the mute key does not, the peer starts notifying again with no UI event explaining why, and the orphan renders forever in `SettingsDrawer.muteCandidates`. Fix shape per the review: a `UserSettings.rename_muted_target/4` on the `:renamed` arm **before** the barrier broadcast, plus the `:own_nick_renamed` self-window arm, with the same fold-collision merge the other three stores do — and extend CLAUDE.md's "a NEW nick-keyed store MUST be added to this migration set" list so the set is enumerable.

Both need a real test that exercises the rename, not a unit that asserts the helper exists.

vjt:
### Unlanded work already on `origin/w1-1340-nick-rename-set`

[...]

Issue #498, opened 2026-07-27
Unread/badge counts keep matching the pre-/nick nick: they read the configured credential nick, never the live one


vjt:
## Symptom

Reported on #grappa by two people (second-hand from a third): after changing nick,
highlights/notification counts still fire for the **old** nick — "if someone says the
old nick, the highlight goes off".

## Root cause (read on `origin/main`, not a local checkout)

There are two doors into "is this message notify-worthy", and they resolve the own-nick
from **different sources**:

1. **Live push dispatch** — `lib/grappa/session/persistor.ex:129` passes `own_nick: ctx.nick`,
   and `ctx` is the live `Session.Server` state, whose `nick` is reconciled at 001 and updated
   on self-NICK (`lib/grappa/session/server.ex:3481` `maybe_broadcast_own_nick_changed/2`).
   This door is correct after a rename.

2. **Badge count + `/me` unread seed** — `lib/grappa/push/badge_count.ex:144` calls
   `configured_nick_windows/1`, which is `Networks.configured_nick_index/1`
   (`lib/grappa/networks.ex:436-448`) and selects **`c.nick` — the credential (configured)
   nick**, deliberately never the live one. Its own docstring states the trade-off:

   > Accepted staleness: after a `/nick` the count uses the configured nick until the next
   > reconnect rewrites the credential.

So after `/nick newnick`, every unread/badge computation keeps matching mentions of the
**old** nick, and stops counting mentions of the new one. Both halves are wrong, and the
second half (new nick no longer counted) is arguably the worse one.

## Why the documented escape hatch does not fire (for registered users)

The docstring assumes "the next reconnect rewrites the credential". Nothing appears to do
that for user credentials:

- `lib/grappa/networks/credentials.ex:404` has `update_visitor_credential_nick/3` — **visitors only**.
- `Credential.identity_changeset/2` (`lib/grappa/networks/credential.ex:594`) can write `nick`,
  but the self-NICK path in `Session.Server` does not route through it; it updates in-memory
  state and broadcasts `own_nick_changed`.
- The credential schema (`credential.ex:171`) has a single `nick` field — there is no
  separate "live nick" column the count could read.

A reconnect uses the configured nick to connect, so it has no new value to persist. Net
effect for a registered user: the staleness is **permanent** until they edit the nick in
network settings by hand.

## Not verified

- No live reproduction was run. This is static reading of `origin/main` plus the report.
  A repro needs a second account on staging and staging is in use tonight.
[...]

PR #1344, opened 2026-08-15
#1340: finish the nick-identity migration set (C-S1 stale ownNick, K-S2 muted_targets)

vjt:
Bucket from the 2026-08-15 codebase review. One class: something holds a nick
captured before the rename, so a `/nick` forks the identity in silence — the
class #373 closed for three stores and left open for a closure and a map.

## C-S1 (HIGH) — the per-channel WS handler held a snapshot

`installChannelHandler` took `ownNick` as a VALUE resolved at topic-join, and
every install loop is guarded by `joined.has(key)`, so an already-joined
channel never rebuilds its handler. `own_nick_changed` re-runs the join
effects and the guard short-circuits all of them. After any `/nick` the
handler answered every own-identity question against the retired nick: the
beep followed the old nick, own-PART stopped being detected (reopening #200's
subscription leak), own presence echoes stopped being recognised, and our own
next rename looked like a peer's.

The parameter is now an accessor read per event.

**The review's proposed fix would not have worked, and this is measured.** It
suggested `() => ownNickForNetwork(net, u)`. `mutateNetworkNick` maps the list
into fresh row objects (`prev.map(n => n.id === id ? {...n, nick} : n)`), so a
closure over the row is exactly as stale as the string was. The accessor
resolves from the SLUG through the live `networkBySlug` memo instead.

Rejoining the topics on a rename was rejected: heavier than the problem, and
it races the JOIN echo the way #200's revert did.

## K-S2 (MEDIUM) — `muted_targets` was never in the migration set

#1038 re-keyed the mute on `(network, target)` where, for a DM, the target IS
the peer nick. `UserSettings.rename_muted_target/4` adds it to the set, on
both rename arms.

**Placement diverges from the review's text, deliberately.** In the peer arm
the migration is UNCONDITIONAL rather than gated on `:renamed`: a mute
outlives the window it silenced — closing a tab does not unmute it — so
gating on the window row would strand exactly the mute nobody can then see to
fix. The precedent is three lines below, where the presence reset is
unconditional for the same reason. The self arm is the opposite and stays
INSIDE `migrate_self_window`'s gate, because there the row count is the only
evidence the window at our old nick is ours and not a peer's who bore it
before us.

## Evidence

Every gate below was run and read locally before Docker Desktop died on the
worker host. **The full `scripts/integration.sh` was NOT run** — it died at
container bring-up with zero specs executed. CI is the arbiter for it.

| Gate | Result |
|---|---|
| `bun run test` (cic) | 283 files, 5477 tests, RC=0 |
| `bun run check` (biome + tsc + e2e tsconfig) | RC=0 |
| `scripts/check.sh` (clean tree, at `b4b278ed`) | 8 doctests, 60 properties, 6201 tests, 0 failures; dialyzer 0 errors; credo clean |
| e2e `--grep "#1340"` | RED with the fix reverted, GREEN with it |
| Elixir targeted, K-S2 | RED with both hooks removed (2 failures), GREEN restored (536 tests, 0 failures) |
| `scripts/integration.sh` (full) | **NOT RUN** — Docker Desktop down |

Both halves are proven by displacement, not by a green alone.

### Why the e2e witness asserts the teardown

Three of C-S1's four consequences are masked end-to-end: the beep and the
title bump are shadowed by the server's own push, which already resolves the
live nick since #498, and the stranded window state draws no sidebar row
(`pseudoChannelsForNetwork` skips `"joined"`, and the server has already
dropped the channel from `channels_changed`). What only cic can decide is
whether a PART was OURS. Read through `__cic_joinedTopicKeys`; RED output was
`["azzurra i498-user", "azzurra $server", "azzurra #i498", "azzurra
i1340-4a65ec10"]` — the topic key still present after the sidebar row had
already gone.

It lives in the #498 spec file because it needs the one thing that file
already provides: a session whose live nick can be destructively renamed
without poisoning the shared vjt identity.

### A comment on record is corrected

That file's header claimed the client-side foreground increment was "also
already correct" after a rename because `own_nick_changed` updates
`ownNickForNetwork`. It does — but the per-channel handler never saw the
update, so a mention of the new nick did not beep. Measured in
`subscribe.test.ts`; the header now says so.

## Scope notes

- The scoped e2e green is attributable to the cic commit; it was measured
  before the K-S2 server change existed. That spec does not touch the mute.
- Residual, named in DESIGN_NOTES rather than papered over: notification prefs
  have no broadcast, so cic keeps the pre-rename mute map until its next
  user-topic join. The server push predicate — the one that reaches the phone
  — honours the migrated key immediately.
- The second leak the RED run exposed (after a rename the DM-listener holds
  BOTH own-nick topics) is #1341 and is deliberately NOT in this branch.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -151,13 +151,18 @@ Key invariants — break only with deliberate cause + DESIGN_NOTES entry:
   a fold-collision with an existing `new` window), the DM scrollback
   (`Scrollback.rename_dm_peer/4` — `dm_with` + outbound/orphan `channel`),
   the DM read cursor (`ReadCursor.rename_dm_peer/4` — else the migrated
-  history reads fully unread), and cic's own caches
+  history reads fully unread), the per-conversation MUTE
+  (`UserSettings.rename_muted_target/4` — nick-keyed since #1038 keyed it
+  `(network, peer)`; #1340), and cic's own caches
   (`scrollback.renameScrollbackKey` + `readCursor.renameReadCursorChannel`
   + `selection.followQueryNick`, driven by the per-channel `nick_change`,
   mirroring `members.ts`). Server-driven: `EventRouter` emits
   `{:peer_nick_renamed, old, new}`, `Session.Server.apply_effects/2`
   renames the row (`QueryWindows.rename/4`, no broadcast), migrates the
-  DM history + read cursor on `:renamed` only, and THEN broadcasts
+  DM history + read cursor on `:renamed` only, migrates the mute
+  UNCONDITIONALLY (a mute outlives the window it silenced, so gating it on
+  the window row would strand it — same posture as the `:unknown` presence
+  reset in that arm), and THEN broadcasts
   `query_windows_list` (`QueryWindows.broadcast_windows_list/2`) — the
   broadcast is a truthful "rename fully applied" barrier, so a consumer
   reacting to the event is guaranteed the history has already moved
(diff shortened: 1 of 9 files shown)


vjt:

Issue #1345, opened 2026-08-15
review: 477/476 never reach the join-failure set, so the window stays :pending forever (I-S1, I-S2)


vjt:
Bucket issue from the 2026-08-15 code review (`docs/reviews/codebase/2026-08-15-codebase-review.md`, merged as `3a8eddec`). Theme: a numeric set enumerated in one place and truncated in another, so a legitimate server reply falls through and leaves state stuck.

- **I-S1 (HIGH)** — `lib/grappa/session/event_router.ex:1739` declares `@join_failure_numerics [471, 473, 474, 475, 403, 405]`, while the channel-param canonicalisation one screen above (comment `:326`, guard `:330`) enumerates the join-failure set as "403/405/471/473/474/475/**476/477**". `NumericRouter.@delegated_numerics` mirrors the truncated six. A JOIN rejected with **477 ERR_NEEDREGGEDNICK** (a `+R` registered-only channel — common on the networks we serve) or 476 emits no `{:join_failed, …}` effect at all. Consequences the review verified: `state.in_flight_joins` holds the entry until the lazy 30s sweep at the next `record_in_flight_join/2` (`server.ex:6458`); `window_state[chan]` stays `:pending` **permanently**, because `WindowState.set_failed/4` is reachable only from the `{:join_failed, …}` arm (`server.ex:5615`) and nothing else sweeps `:pending`; cic renders a greyed tab that never resolves and never learns the reason or the numeric.

- **I-S2 (MEDIUM)** — the WHOIS-leg correlation reads the accumulator with a narrower fold than the one that wrote it, so on an rfc1459 network the write and the read disagree.

The fix for I-S1 is not just extending the literal: the two enumerations must stop being independently maintained, or the next numeric repeats this exactly. A test should prove a 477 reply drives the window to `:failed` and surfaces the numeric to the client, and should fail if either list is truncated again.

vjt:
### Unlanded work already on `origin/w1-1345-join-failure-numerics`

Branch census of 2026-08-20 (measured against `origin/main` = `41190db8`). This branch never went through a PR, so the work below is **not on `main`** and is easy to redo by accident. Recording it here so whoever picks this issue up starts from it instead of from zero.

Branch head: `cc0314c7` — last commit 2026-08-15.

Commits with no equivalent in `main` (`git cherry`, patch-id):

- `fbffff38` irc(#1345): fold the WHOIS-leg read the way the write folded it (I-S2)

Files those commits touch:

[...]

PR #1346, opened 2026-08-15
irc(#1345): one measured join-failure numeric set, delegated only when it correlates

vjt:
Issue #1345 (bucket from the 2026-08-15 code review): I-S1 (HIGH) and I-S2 (MEDIUM).

## The defect

`477 ERR_NEEDREGGEDNICK` was named by the channel-param canonicaliser
(`event_router.ex:326/330`) and missing from both lists that act on the
numeric (`@join_failure_numerics`, `NumericRouter.@delegated_numerics`). A
JOIN rejected by a `+R` channel therefore produced no `{:join_failed, …}`
effect: the in-flight entry waited for the lazy 30 s sweep, the window stayed
`:pending` permanently (`WindowState.set_failed/4` is reachable from that
effect alone), and cic drew a greyed tab that never resolved.

## Shape of the fix

Not a longer literal. `Grappa.IRC.JoinFailure` is the single enumeration and
all three consumers derive from it, so there is nothing left to truncate
independently. Delegation is **correlation-gated** (`join_failure_leg?/3`):
a member of the set is delegated only when its `params[1]` folds to a channel
whose JOIN is in flight — the same match `EventRouter` performs — so the two
decisions cannot disagree.

## The set comes from the ircd sources, not from the issue text

Read out of azzurra/bahamut `5c41c8b` and solanum `2ce64de`, one citation per
code in the moduledoc: **403 405 471 473 474 475 476 477 479 480 485**.

Corrections to the issue body, all measured:

* **481 is excluded**, although it IS a `can_join` exit (`+O`,
  `channel.c:1967`). Its format string carries **no channel** —
  `":%s 481 %s :Permission Denied, …"` (`s_err.c:550`) has two `%s` and drops
  the `name` m_join passes. Nothing on the wire to correlate, so a `+O`
  refusal cannot reach its window and that window still stalls at
  `:pending`. 481 is also emitted 29× from `s_serv.c` alone for ordinary
  oper-command refusals. **The cure for the uncorrelatable class is a timeout
  on the in-flight JOIN, not another numeric** — deliberately not built here.
* **437 is excluded, and it is the interesting one.** It IS a JOIN exit on
  solanum (ERR_UNAVAILRESOURCE), but on bahamut — all of prod — 437 is a
  different numeric, ERR_BANNICKCHANGE (`numeric.h:333`), and `m_nick.c:525`
  passes it `chptr->chname`: `params[1]` is a **channel** while the failure
  is a nick change. It would correlate against a live in-flight JOIN and flip
  that window to `:failed` off a `/nick` — the ghost-correlation hazard the
  self-JOIN strip (`server.ex:5600`, lifecycle review HIGH S1) already guards
  for the real failure numerics. A juped channel on Libera keeps sitting at
  `:pending` instead; prod wins the trade.
* **479, 480, 485 are added**: JOIN error exits carrying the channel at
  `params[1]` on one of the two ircds. 479 is emitted by bahamut too, from
  `check_channelname` (`channel.c:2002`, called on the JOIN path at `:2204`);
  480 is absent from bahamut's `numeric.h` entirely. 443 (channel at
  `params[2]`) and 470 (a redirect, not a refusal) stay out for shape.

**A cross-flavour collision is inert only when the other meaning cannot
present the same shape.** That holds for 476 (ONLYSSLCLIENTS vs BADCHANMASK)
and 485 (CHANBANREASON vs BANNEDNICK) — solanum defines both in `numeric.h`
and gives neither a `NUMERIC_STR_` format string, so core never emits them —
and fails for 437, which is why the guard comment tells the next reader not
to loosen `params[1]`.

Nothing in the change maps a numeric to a phrase: `reason` is the ircd's own
trailing text and the code rides beside it as `meta.numeric`, so no
flavour-blind sentence is invented.

## A silent drop the old comment denied

The handler's comment promised that an uncorrelated 403/473/… "falls through —
the caller's NumericRouter `$server` route persists it as a server message".
Measured, that was false while the codes were unconditionally delegated:
`Server.delegate/2` hands the numeric to `EventRouter` alone, whose numeric
catch-all (`event_router.ex:2483`) returns `[]`, and nothing persisted it
anywhere — an unsolicited 403 answered a `/topic` with silence. The gate puts
those back on the param scan; a test pins the landing.

## I-S2

`whois_leg?/3` read `whois_pending` with the arity-1 ASCII fold while
`EventRouter` writes those keys with `canonical_target/2` at the network
casemapping. On rfc1459 a target holding `[ ] \ ~` is written `foo{1}` and was
looked up as `foo[1]`. The casemapping now travels in `router_state()` (the
join gate needs the same fold) and both reads use it. On `:ascii` — all of
prod — nothing moves.

## Gates

| gate | result |
|------|--------|
| `scripts/check.sh` | **NOT RUN** — Docker Desktop down on the worker host for the whole change |
| `scripts/test.sh` | **NOT RUN** — same |
| `scripts/integration.sh` | **NOT RUN** — same |
| `scripts/design-notes-gate.sh` | **PASS** (pure bash, run on the branch tip) |

Every test here is written and **unobserved**: no red was seen before the fix
and no green after it. CI is the only arbiter — the first run was red on
Credo (a named unused variable in a test helper, now deleted along with the
helper) and on Dialyzer (`numerics/0`'s contract was wider than its success
typing; the return type is the measured union via a new `@type t`, so the
union and the list now pin each other in both directions). `mix format` was
unavailable too, so the formatter's joins are applied by hand.

## Not established / adjacent findings

* The `+O` (481) window still stalls at `:pending`, and so does a juped
  channel on solanum (437) — both need the in-flight JOIN timeout.
* Not touched, pre-existing: `server.ex:3293` consumes **any** 433/437 as
  `{:nick_error, code}` while a recover-identity is armed, without checking
  the param shape.
* Not touched, pre-existing: the self-JOIN strip at `server.ex:5605` deletes
  the in-flight entry with `Identifier.canonical_target/1` (ASCII) while
  `record_in_flight_join/2` wrote the key network-folded — the same I-S2
  mismatch, one layer over, so on rfc1459 a national-char channel leaves the
  ghost behind.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -42725,7 +42725,7 @@ sweeps `:pending`), and cic drew a greyed tab that never resolved.
 `Grappa.IRC.JoinFailure` is now the single enumeration and every consumer
 derives from it. The set is read out of the two bound ircds rather than an
 RFC (azzurra/bahamut `5c41c8b`, solanum `2ce64de`), one citation per code in
-the moduledoc: **403 405 437 471 473 474 475 476 477 479 480 485**.
+the moduledoc: **403 405 471 473 474 475 476 477 479 480 485**.
 
 **481 is excluded, and that corrects the two texts that put it in.** It IS a
 `can_join` exit on bahamut (`+O` oper-only, `channel.c:1967`), but its format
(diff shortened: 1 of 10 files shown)


vjt:

Issue #1343, opened 2026-08-15
review: put a gate under three hand-maintained lists the docs call authoritative (M-S1, D-S1, D-S5)


vjt:
Bucket issue from the 2026-08-15 code review (`docs/reviews/codebase/2026-08-15-codebase-review.md`, merged as `3a8eddec`). One theme: a list the documentation describes as authoritative, maintained by hand, with nothing holding it to that claim. Each fix is a gate, not a behaviour change.

- **M-S1 (HIGH)** — the migration-version rule CLAUDE.md marks 🔴 has zero mechanical enforcement. Measured: duplicate versions today 0, but **40 of 86 stamps are hand-typed round numbers**, and `20260717120000`→`20260811130000` is 24 consecutive hand-typed files. `Preflight.contract_migrations/2` (`lib/grappa/deploy/preflight.ex:480`) filters changed git paths, not versions against `schema_migrations`, so it cannot see the dangerous regime — the one where a version already applied makes both files drop out of pending and neither ever runs, with the run reporting success. Fix per the review: a small `async: true` test asserting `versions == Enum.uniq(versions)` over parsed filenames; a "no round stamp" assertion for newly added files only needs an allowlist for the 40 historical ones, so uniqueness alone is acceptable if the allowlist is not worth it.

- **D-S1 (HIGH)** — `lib/grappa/hot_reload/long_lived_modules.ex:92` enumerates 9 of the supervised stateful set; four children carrying exactly the state shapes the extractor reads are missing: `Grappa.Accounts.Reaper` (`accounts/reaper.ex:64`), `Grappa.DbLatency` (`db_latency.ex:123`), `Grappa.SessionLog` (`session_log.ex:82`), `Grappa.Net.PtrCache` (`net/ptr_cache.ex:142`). `Preflight` reads this list to decide whether a state-shape change forces COLD, so a `defstruct` field-add to any of the four classifies HOT on the jail — the exact class the preflight exists to refuse. `Accounts.Reaper` is the sharpest case: same `interval_ms` struct as its two listed sibling reapers. Note the asymmetry the review found: `test/grappa/application_supervision_tree_test.exs:33` already pins every supervised child against the CLAUDE.md tree, so a new GenServer is caught by that gate and silently skipped by this one — the membership itself is untested (`grep -rln LongLivedModules test/` hits only `preflight_test.exs`, which tests the extractor).

[...]

PR #1347, opened 2026-08-15
Gate three hand-maintained lists the docs call authoritative (#1343)

vjt:
## What

The 2026-08-15 review found one shape three times: a list the documentation
calls authoritative, maintained by hand, with nothing holding it to the claim.
Each change here is a gate, not a behaviour change.

**M-S1 — migration versions.** The 🔴 CLAUDE.md rule ("generate the stamp,
never hand-type it") had zero enforcement; 40 of 86 stamps on main are round
hand-typed numbers, and that is how #1044 and #1038 came to claim one version
under two basenames. `test/infra/migration_version_test.bats` asserts parseable
names, unique versions, and a ratchet on the round-stamp count — a ratchet
rather than an allowlist of the 40, because an allowlist is the same
hand-maintained shape being removed.

**D-S1 — `LongLivedModules`.** `Preflight` state-shape-checks only the files
behind that list, and nine supervised GenServers carrying exactly the shapes
the extractor reads were absent. The membership is now asserted, not the names:
the candidate set is derived from the app's GenServer behaviour, so there is no
second enumeration to drift.

**D-S5 — the `oven/bun` digest.** Transcribed on four surfaces, asserted
well-formed but never equal. Every `image:tag` family must now carry one digest.

## Two things the review did not have

- `Preflight.module_to_path/1` hardcoded `lib/grappa/`, so no `GrappaWeb` module
  could ever be state-shape checked — and the failure is silent-HOT, because a
  path absent at both revs compares equal. `GrappaWeb.SessionRevocationListener`
  is already a supervised child.
- Adding `Net.PtrCache` as the review asked would have been **inert**: it bound
  its six-field map to a variable before returning it, and the extractor only
  reads a literal. Hence its `init/1` now returns the literal, and hence the
  membership test's second assertion — a listing the extractor sees nothing in
  reads as coverage while buying no check.

Also: `Grappa.Net.SourceAliasManager` was missing from the CLAUDE.md tree, and
the pin that exists to catch that cannot, because it compares against the
RUNNING children and the manager does not boot in the test env.

## Proof

Every gate was to be proved by introducing the drift it must catch.

- M-S1 — proved, one mutant per assertion: a duplicate of an existing non-round
  version reddens only the uniqueness case, a new round stamp only the ratchet,
  a malformed name only the naming case.
- D-S5 — proved: rewriting `compose.yaml`'s digest reddens only the new case.
- Full `scripts/bats.sh`: 481 ok, 0 failures.
- **D-S1 — NOT proved.** Docker Desktop is down on the authoring host, so no
  ExUnit run of any kind was possible. The mutation to run: remove
  `Grappa.Accounts.Reaper` from `@modules` and the first assertion alone must
  redden. Nothing in this PR's Elixir has been compiled locally — CI is the
  first thing to see it.


vjt:

Issue #1354, opened 2026-08-15
review: make the docs say what the code does (K-S1, X-S3, L-S14) — 51 stale references, two of them inverted


vjt:
Bucket issue from the 2026-08-15 code review (`docs/reviews/codebase/2026-08-15-codebase-review.md`, merged as `3a8eddec`). Theme: documentation that has outlived its subject. Harmless-looking until someone builds on it — and two members here state the **opposite** of what the code does.

- **K-S1 (HIGH)** — **51 live references** to `canonical_channel/1` and `canonical_nick/1`, functions #537 deleted when it collapsed both into `Identifier.canonical_target/1`. Four scope agents found the same rot independently (irc I-S6, lifecycle L-S11, persistence P-S10, consistency K-S1); K-S1 is the one that identifies the two members which now assert the inverse of the current fold, **one of them in a published `@doc`**. Sweep all 51 in one pass, and fix the two inverted statements first — a wrong doc is worse than a missing one.

- **X-S3 (HIGH)** — the published carrier topic for `joined` / `join_failed` / `kicked` is wrong in three places at once: CLAUDE.md, the server-side comment, and the client-side comment. The helper the text names **does not exist**. A client author following the documented topic subscribes to nothing.

- **L-S14 (LOW, but it disables a gate)** — CLAUDE.md's supervision-tree diagram omits a child whose placement is documented as load-bearing. Related, found separately while building #1343's gates: `Grappa.Net.SourceAliasManager` is missing from the same diagram, and the tree-pin test compares against the children that actually start — so the pin that should catch an unlisted child cannot.

Where a doc and the code disagree, the code is the truth: verify each statement against the source before rewriting it, and say in the PR body which ones you found to be inverted rather than merely stale.

vjt:
Closed by the pipeline change ruled by vjt on 2026-08-20: **`status:soon` is retired**. An issue is now closed when its work **merges**, and the **milestone** records which release carries it — the label duplicated what the milestone already said.

This issue was verified individually before closing, not swept: its work is on `origin/main` under commits that are not documentation-only. No behaviour changes here, only bookkeeping.

<sub>Applied by vjt-claude on vjt's instruction.</sub>

Issue #112, opened 2026-06-26
Tech-debt backlog: REV carry-forwards + Low/Observation nits


vjt:
Consolidated code-quality backlog migrated from the retired `docs/todo.md` (2026-06-26). Each item was re-verified against the current tree before inclusion; dead nits were dropped (Identifier.services_sender? clauses are all reachable; nginx `keepalive` already has `Connection ""` on the API location). Promote an item to its own issue when someone picks it up.

### cicchetto
- [ ] `ChannelPushError` stores the full reply in an `info` field that duplicates the `error` key — dedup. `cicchetto/src/lib/api.ts:1150`
- [ ] `Push.subscription.id` typed as `string` rather than a branded UUID. `cicchetto/src/lib/push.ts:114`
- [ ] `linkify` regex uses an unbounded `\S+` (a post-pass strips trailing punctuation) — bound it. `cicchetto/src/lib/linkify.ts:50`
- [ ] `uploadHost.ts` reads the bearer from localStorage directly instead of the `token()` signal. `cicchetto/src/lib/uploadHost.ts:381`
- [ ] Channel-directory search has no `AbortController` → last-arrival-wins race on rapid typing (low risk: GETs hit a cached snapshot). `cicchetto/src/lib/channelDirectory.ts:35` (#84 E3 follow-up)
- [ ] Compose `left` swipe is classified but unmapped — wire to reverse `tabComplete` if a back-cycle is ever wanted. `cicchetto/src/ComposeBox.tsx:116`
- [ ] Modal-chrome CSS dedup: `.media-viewer-close` duplicates `.archive-modal-close`, and `.media-viewer-backdrop` is a 4th `inset:0` dim backdrop — extract a shared `.modal-backdrop`/`.modal-close` base. `cicchetto/src/themes/default.css:2951,3966`

### Elixir / server
- [ ] AwayState is session-GenServer-only and is lost on crash/reconnect — the operator must re-issue `/away`. Surface an "AWAY lost on reconnect" hint or persist it. `lib/grappa/session/away_state.ex:53`
- [ ] `validate_target_name/1` runs on the pre-canonical `target` (bytes-equivalent today; minor drift risk). `lib/grappa_web/controllers/archive_controller.ex:109`
- [ ] `maybe_log_send_failure/2` takes a `String.t()` label that should be a closed-set atom (CLAUDE.md). `lib/grappa/session/server.ex:3151`
[...]

PR #1360, opened 2026-08-15
Make the docs say what the code does (#1354): 61 stale refs, five inverted

vjt:
## What this is

The doc-reality bucket from the 2026-08-15 codebase review — K-S1, X-S3, L-S14.
Documentation that outlived its subject. Every statement was verified against
the source before being rewritten; where a doc and the code disagreed, the code
won.

## Counts, recounted rather than inherited

| | review said | measured here |
|---|---|---|
| `canonical_channel/1` / `canonical_nick/1` in `lib/` | 51 | **51** |
| same, in `test/` | 11 | **10** |
| surviving after the sweep | — | **15**, every one an explicit historical contrast |

## Found INVERTED, not merely stale

Five, not two — and they are one family, each asserting that the identifier
fold is shape-aware, which is exactly the property #537 removed:

1. **`Grappa.PubSub.Topic.channel/3` — published `@doc`, ExDoc-visible.** Said
   the segment is folded "sigil-aware — nicks for DM windows pass through
   unchanged"; the body two lines down calls `canonical_target/1`, which folds
   `A-Z` unconditionally. Build a DM topic from that doc and you subscribe to
   nothing.
2. **`ArchiveController`** — "Shape-aware (delegates to the sigil-gated
   `canonical_target/1`)". `canonical_target/1` is not sigil-gated.
3. **`ReadCursor.get/3`** — claims a per-shape dispatch that no longer exists.
4. **`Scrollback.delete_for_dm/3`** — "The call is a no-op on nick-shaped
   input". It is not.
5. **`EventRouter.normalize_channel/2`** — named `canonical_channel/1` as the
   sigil predicate's "single source of truth" three lines above a comment
   correctly calling that function "the old" one. The block contradicted
   itself; the gate is the `#&!+` guard clause on the function's own head.

Plus the X-S3 set: seven statements on four surfaces putting live `joined` /
`join_failed` / `kicked` on the per-channel topic when
`broadcast_window_state/2` is `Broadcaster.to_user/2` and nothing else. Three
of them name `Session.Server.broadcast_window_state_dual/3`, which exists
nowhere in `lib/`.

## One review claim RETRACTED

`ReadCursor.Cursor`'s comment was cited as an inversion. It is not: it reads
"Was `canonical_channel/1`, which is a no-op for a DM-peer nick" and the very
next sentence names `canonical_target/1` as what the code now calls — matching
the call two lines below. Unchanged since #532-D (`f8f5d7eb`). It is a
historical contrast carrying a present-tense verb on a dead function, fixed to
past tense rather than rewritten.

## The root cause of X-S3's spread, which the review did not have

`grep -n 'broadcast_window_state' docs/DESIGN_NOTES.md` returns **nothing**
before this PR. F1 changed the carrier in May 2026 and recorded it only in code
comments. The sole statement in the decision log is CP15's original per-channel
claim — true when written, superseded ever since, never annotated. Later
authors checked the authority and the authority was three months stale. The
CP15 entry is left exactly as written (the log is chronological; another
author's entry is not rewritten); the correction lives in the new entry.

## The one non-comment change

`Scrollback.delete_for_dm/3` folded twice —
`canonical_target(canonical_target(peer))` — a fossil of the two-function world
where step one was the sigil-gated no-op and step two the nick fold. With one
shape-blind fold the second application is the identity, so the comment could
not be made true while the code stood. The fold is a plain ASCII `A-Z`
lowercase and is idempotent; behaviour-preserving.

## Deliberately NOT done

- **The `check.sh` grep gate** K-S1's fix (c) proposes. It would need a
  hand-maintained vocabulary of English prose markers ("the former",
  "pre-#537", "was", "the old") — the hand-list shape #1343 just spent a diff
  removing. And it cannot be line-based: **7 of the 15 legitimate survivors
  carry their marker on the PREVIOUS line** (wrapped comments), so a
  line-scoped grep false-positives on all seven. Measured, not assumed.
- **Closing the supervision-tree pin gap.**
  `application_supervision_tree_test.exs` asserts doc ⊇ RUNNING children, so a
  child the test env does not start is invisible to it —
  `Net.SourceAliasManager` (`:start_source_alias_manager` false in test) is the
  second such after `Bootstrap`, whose absence the pin documents as deliberate.
  Closing it needs either that same hand-list or a source parse the pin's own
  comment rejects on #112 grounds. It wants a design call.
- **A test pinning the X-S3 topic claim.** A test asserting that `joined` does
  NOT arrive on the per-channel topic would make §4 enforceable. It does not
  exist, and this PR does not add one.

## A defect of mine, caught and fixed in-branch

The first draft of the two new paragraphs enumerated the per-channel event
kinds, and both enumerations were wrong the moment they were written: they
named `members_seeded` / `topic_changed` / `channel_modes_changed` and omitted
`channel_created` (`server.ex:5494`) and `read_cursor_set`. A partial list read
as exhaustive is the exact failure this bucket exists to repair, so restating
it "correctly this time" would only reset the clock. CLAUDE.md now names the
`Broadcaster.to_channel/3` call sites as the SSOT and says not to restate the
list in prose; `CLIENT_PROTOCOL.md` gives the rule instead of a roster.

## Gates

Full `scripts/check.sh` green at `f3c196a7`, `RC=0` read from the log file:
`compile --warnings-as-errors` (+ Boundary), `format --check-formatted`,
`credo --strict` (5869 mods/funs, no issues), Sobelow, `mix_audit`, doctor,
`mix test --warnings-as-errors` (8 doctests, 60 properties, **6289 tests, 0
failures**), dialyzer (**Total errors: 0**), 481 bats.

cic gates green: `bun run check` (biome + `tsc --noEmit` ×2) and `bun run test`
(**284 files, 5484 tests**). `scripts/design-notes-gate.sh` green.

Above that green sit only the two markdown edits of the last commit, in files
no compiler reads. The single test that opens `CLAUDE.md`
(`application_supervision_tree_test.exs:101`) extracts the fenced tree block at
lines 21-51; that edit is at line 258. No test opens `CLIENT_PROTOCOL.md`. The
integration/STACK lane was not requested and not run: no wire change, no
client-facing behaviour change, and the one executable change is a removed
idempotent call covered by the unit suite.


vjt:
Superseded by #1362, which landed as the union of this PR and #1361 on main at 9b79329c.

Both PRs touched `docs/DESIGN_NOTES.md` (+121 and +52). Merging either one first would have made the other CONFLICTING, and a conflicting PR runs no CI at all — so the sequential path cost the same as the union and never gated the union. All six commits of this branch were replayed onto a fresh branch cut from current main, gated once, and merged once.

Landing verified by patch-id, not by title: all 6 of this branch's commits have identical `git patch-id --stable` values to 6 of the 10 commits in the merged range 609261ab..9b79329c. The DESIGN_NOTES entry survived intact (additions 173, deletions 0, both `<!-- entry -->` markers first-line, separator shape confirmed on main).

Closing by content — the replay rewrote the SHAs, so GitHub cannot close this automatically.

Issue #1357, opened 2026-08-15
cic(reply): a second reply buries the first ` << ` tail mid-line


vjt:
Replying twice without clearing the compose box buries the first ` << ` tail in the middle of the line. vjt hit it live on #grappa (22:03) and sent the result: four `<vjt-claude> … << ` blocks on one line, the last one with his own text typed after it.

## Measured on `origin/main` (`0376738a`)

`cicchetto/src/lib/replyQuote.ts:69` — `replyToMessage` hands the whole quote, tail included, to `appendToCompose`, which at `cicchetto/src/lib/composeAppend.ts:22` does `getDraft(key) + text` and never looks at the draft. So the second reply lands as:

```
<a> primo <<  <b> secondo << 
```

`REPLY_QUOTE_TAIL` is documented at `replyQuote.ts:14` as the tail whose trailing space exists "so the answer is typed straight after the caret" — an invariant about the END of the line. Appending a second quote breaks it: the earlier tail is now a mid-line `<<` that reads as part of the quoted text, and the answer the operator types belongs to the last quote only.

Same root as #1356 (blind concatenation, draft never inspected), different door and different remedy, so it is filed separately: #1356's `!addquote` **must** accumulate payloads by design (one verb, N quotes). Reply must not accumulate tails.

## Decision — keep both quotes, one tail

Ruled by vjt on #grappa, 2026-08-15 23:33-23:34. The second reply **keeps** the existing quote and **strips the trailing ` << `** before appending its own, so the line comes out:

```
<a> primo <b> secondo << 
```

Consistent with the `!addquote` ruling on #1356: **the compose line accumulates, the marker does not repeat.** The tail stays where it is documented to be — at the end of the line, with its trailing space, so the answer is still typed straight after the caret.

Alternatives, for the record: refusing the second swipe while a quote is in the draft is out. Replacing the draft's quote (last swipe wins) is **not** dismissed — vjt said it arguably makes more sense, and settled on "start with 2, we change it later if needed". So build 2, but keep the append path narrow enough that swapping to replace-semantics is a one-function change, not a refactor.

### Acceptance

- Second reply on a draft that already ends with ` << ` produces exactly one tail, at the end.
- N replies produce N quotes and one tail.
- A draft with operator text already typed after the tail is not silently mangled — the appended quote must not land before that text.
- The single-reply path is unchanged.

---
[...]

Issue #1356, opened 2026-08-15
cic(!addquote): a second long-press appends a second command to the same compose line


vjt:
`!addquote` from the message menu must be able to build ONE quote out of SEVERAL messages. Long-pressing a second message while a quote is already in the compose box must append that message's `<nick> body` to the line — **without a second `!addquote` verb, and without any separator token**. One verb, N payloads:

```
!addquote <ska> io oggi ho messo un reggiseno <alk> se lo allacciano davanti e lo girano
```

Spec by vjt on #grappa (22:01–22:03).

## What happens today

Measured on `origin/main` (`0376738a`) — there is no guard on the path, so the verb repeats:

- `cicchetto/src/MessageContextMenu.tsx:51` — the menu action calls `addQuoteToCompose` unconditionally; the only predicate is `enabled: addQuoteCommand(target.msg) !== null`, which asks about the row, never about the draft.
- `cicchetto/src/lib/addQuote.ts:48` — hands the full command, verb included, to `appendToCompose`.
- `cicchetto/src/lib/composeAppend.ts:22` — `const next = getDraft(key) + text`: blind concatenation, the draft is never inspected.

Result: `!addquote <ska> foo!addquote <alk> bar`. The bot reads one line, so the second quote is lost or swallowed into the first body.

## The fix

In the `!addquote` door (**not** in `appendToCompose`, which is shared with the reply quote of #1067 and the off-compose printable-key handler — appending to an existing draft is legitimate for those):

- draft already contains `ADDQUOTE_COMMAND` → append `attributionHead(msg) + " " + body`, no verb;
- otherwise → append the whole command, as today.

The menu item stays enabled: a second press is a legitimate append, not a duplicate to refuse.

**Reading I am declaring rather than assuming silently:** "no separator" is taken as *no separator token* (`|`, `—`, newline). One space still goes between the previous payload and the next `<nick>`, because `getDraft(key) + text` adds nothing of its own and `…reggiseno<alk> se lo…` would glue the two attributions into one word. Same applies with a half-typed draft: `ciao` + a long-press gives `ciao!addquote <alk> bar` today. If a bare space is also unwanted, say so and it goes in verbatim.

---
*Filed by vjt-claude from vjt's `gh` account, on his instruction on #grappa. The requirement is his; the code measurements are mine, read from `origin/main` at `0376738a`. The two-verb line was not reproduced in a running client — it is read off the concatenation, not observed.*

vjt:
[...]

PR #1361, opened 2026-08-15
cic: stop the compose box concatenating quotes blindly (#1356, #1357)

vjt:
`appendToCompose` does `getDraft(key) + text` and never looks at the draft. Three doors ride it; two of them were producing garbage from that one blindness. Same root, **opposite** remedies — which is why they stayed two issues, and why both fixes live in their own door rather than in the shared verb (the off-compose printable-key handler appends to a live draft legitimately, and a guard there would break it).

## #1356 — `!addquote` accumulates: one verb, N payloads

A second long-press wrote `<ska> primo!addquote <alk> secondo`. The bot reads one command per line, so the second quote was not malformed — it was eaten.

Now a draft that already holds the command gets the bare `<nick> body` appended. The predicate is `includes`, not `startsWith`, so a half-typed draft still counts as holding a command the bot will read as one; the accepted cost is that a draft merely mentioning the verb in prose suppresses it too, which the bot could not tell apart either. Separator is one space and no token, added only when the draft does not already end in whitespace, so the verb's own trailing space is not doubled. The menu item stays enabled: a second press is a legitimate append.

## #1357 — reply keeps both quotes and one tail

A second reply buried the first ` << ` mid-line, where the marker reads as quoted text and the answer typed at the caret belongs to the last quote only.

Per vjt's ruling, the line comes out `<a> primo <b> secondo << `. The marker is stripped only when the tail is at the END; the tail's leading space (#1235) is then exactly the separator the next quote needs. Once the operator has typed their answer the tail is mid-line and is not ours to cut — that would rewrite a human's sentence to fix our own duplication — so that line keeps two markers on purpose, per the issue's third acceptance criterion.

Replace-semantics (last swipe wins) was explicitly not dismissed, so the whole decision sits in `draftBeforeReplyQuote`: switching is that one function, not a refactor.

## `updateCompose`

Reply must REMOVE characters, which append cannot express. A `setDraft` beside the append would split one edit into two writes that disagree exactly when no composer is mounted — the strip lands, the quote does not. So the focus + iOS-scroll-short-circuit + caret-reveal dance is now parameterised by a draft transform, and `appendToCompose` is one line on top of it. `!addquote` stays on the append verb deliberately: it must not touch the composer at all for an unquotable row, and a transform callback would focus one.

## Gates

- `bun run test` — 5491 passed, 284 files (7 new arms: 4 on `!addquote`, 3 on reply).
- `bun run check` — biome + tsc + e2e tsc, rc 0.
- `scripts/design-notes-gate.sh` — 1 new entry, separator and marker present.
- Every new arm was watched RED first: the `!addquote` four reproduced the doubled verb, the reply two reproduced the buried tail. The third reply arm (operator has typed past the tail) is green from the start by design — it is the mutant killer for an implementation that strips a tail anywhere.

## Not established

jsdom only. Both behaviours are asserted on the compose store; that a real engine shows the caret after the surviving tail is the e2e's job, and no e2e was run for this pair.

Fixes the two issues #1356 and #1357.

### DIFF
--- a/cicchetto/src/__tests__/addQuote.test.ts
+++ b/cicchetto/src/__tests__/addQuote.test.ts
@@ -1,6 +1,6 @@
 // @vitest-environment jsdom
 import { beforeEach, describe, expect, it } from "vitest";
-import { addQuoteCommand, addQuoteToCompose } from "../lib/addQuote";
+import { ADDQUOTE_COMMAND, addQuoteCommand, addQuoteToCompose } from "../lib/addQuote";
 import type { ScrollbackMessage } from "../lib/api";
 import { channelKey } from "../lib/channelKey";
 import { getDraft, setDraft } from "../lib/compose";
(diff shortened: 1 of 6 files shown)


vjt:
Superseded by #1362, which landed as the union of this PR and #1360 on main at 9b79329c.

Both PRs touched `docs/DESIGN_NOTES.md`. Merging either first would have made the other CONFLICTING, and a conflicting PR runs no CI — so the union was the honest gate, not merely the cheap one.

Landing verified by content: diffing this branch's touched files against main leaves zero lines that the branch has and main lacks.

One commit in the union is not from either source branch: the `issue1105` reply-quote e2e pin was asserting the superseded semantics (`Q << Q << `, one tail per quote) and was updated to the one-tail ruling recorded on #1357. The single-reply path is unchanged — at one quote the two shapes coincide.

Closing by content — the replay rewrote the SHAs, so GitHub cannot close this automatically.

Issue #1353, opened 2026-08-15
hardening: close the three highest-severity items from the 2026-08-15 security review


vjt:
Tracking issue for the three highest-severity items of the 2026-08-15 full-project security review.

**Details are deliberately out of band.** The review output is kept in a local, gitignored file and is not reproduced here, in the commits, or in the PR: fixing these is necessarily public, describing them is not, and self-hosters need time to update before the shape of a defect is legible from the diff.

Scope: three items. Counts for the whole review, which are all that belongs in public: 0 critical, 3 high, 13 medium, the rest low or informational.

Working rules for this issue:
- Commit and PR messages state the guard that now exists and the behaviour that is now correct — never the consequence of its absence, no reproduction, and no file:line that would serve as a map.
- Each fix needs a test that fails without the guard.
- One of the three has a measured caveat recorded out of band that narrows what it actually implies; read that before assuming the declared severity.

vjt:
Closed by the pipeline change ruled by vjt on 2026-08-20: **`status:soon` is retired**. An issue is now closed when its work **merges**, and the **milestone** records which release carries it — the label duplicated what the milestone already said.

This issue was verified individually before closing, not swept: its work is on `origin/main` under commits that are not documentation-only. No behaviour changes here, only bookkeeping.

<sub>Applied by vjt-claude on vjt's instruction.</sub>

PR #1364, opened 2026-08-15
accounts(#1353): make the account name an identity key that folds

vjt:
## What changes

`users.name` names an account, and this schema folds identity keys at the MATCH
while storing them raw for display (#121/#525) — the pattern `network_credentials`,
`query_windows` and `notify_entries` already carry. The account name now carries it
too: `vjt` and `VJT` name ONE account, exactly as they name one nick on the wire.

Both halves move together:

- `users_folded_name_index` is a UNIQUE expression index on `lower(name)` — the same
  byte-pinned `lower()` that `Identifier.nick_fold_sql/1` emits, so the #525 fold pin
  now covers this migration too and one byte of drift reddens there rather than
  silently costing the index.
- The three account readers share one folded query, so the table and its readers
  cannot drift apart.
- The changeset names the new index explicitly, since ecto derives a constraint name
  from the column and would otherwise let the folded violation escape as an
  `Ecto.ConstraintError` instead of a changeset error.
- The column stays raw: the login envelope still returns the spelling the account was
  created with.

The two byte-exact readers in `test_support` move in the same commit — half a
migration leaves two patterns in the tree, and the next reader copies whichever is
closer.

## A pre-existing collision refuses the migration

The sibling folded-index migrations delete the losers, which is right for a visitor
credential and wrong for an account: two rows here are two accounts with their own
passwords, sessions, credentials and scrollback, and choosing a survivor is not a
decision a migration is entitled to make. This one raises and names the groups; the
operator renames one side and runs the deploy again.

We measured **7 accounts and zero collisions** — but that is measured on our
deployment, not on all of them. A self-hosted instance has never been measured by
anyone, which is exactly why the migration refuses loudly instead of resolving.

## Deploy class

🔵 **COLD.** This carries a migration, so it does not ride the hot batch.

## Verification

- Full `scripts/check.sh` green: Credo strict, Dialyzer, Doctor, Sobelow,
  `deps.audit`, the ExUnit suite and the bats set.
- The migration test pins the index two ways — first the SQL SQLite actually kept for
  it, read back from `sqlite_master`, then the query plan — and fails loudly if the
  index is absent. A plan assertion alone is not diagnostic: it reads the same whether
  the expression drifted or the planner simply declined a correct index. Run 3/3
  stable.
- Mutation-tested: drift in the index expression kills the source pin and the planner
  pin as two disjoint deaths; the folded `unique_constraint` and the migration's
  refusal each have an exact killing assertion; the two `test_support` readers are 1:1.

## Not established here

The migration has only ever run against a temp-file smoke repo, never against a
database of production size.

### DIFF
--- a/lib/grappa/accounts.ex
+++ b/lib/grappa/accounts.ex
@@ -83,6 +83,7 @@ defmodule Grappa.Accounts do
       Grappa.Accounts.Revocations,
       Grappa.Ecto.Like,
       Grappa.EncryptedBinary,
+      Grappa.IRC,
       Grappa.Repo,
       Grappa.Visitors.Visitor
     ],
@@ -92,9 +93,11 @@ defmodule Grappa.Accounts do
 
   alias Grappa.Accounts.{Passkey, RecoveryCodes, Revocations, Session, TOTP, TOTPRecoveryCode, User}
   alias Grappa.Ecto.Like
+  alias Grappa.IRC.Identifier
   alias Grappa.Repo
   alias Grappa.Visitors.Visitor
 
+  require Identifier
   require Logger
 
   @type subject :: {:user, Ecto.UUID.t()} | {:visitor, Ecto.UUID.t()}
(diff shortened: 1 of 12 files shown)


vjt:

Issue #1355, opened 2026-08-15
sqlite: the ZFS page_size move silently multiplied the WAL checkpoint threshold by 16, and journal_size_limit never truncates it


vjt:
## Measured

Production DB (`grappa_prod.db`), read-only PRAGMA read on the live file:

| PRAGMA | value |
|---|---|
| `page_size` | **65536** |
| `wal_autocheckpoint` | **1000** (SQLite default, never set in config) |
| `journal_size_limit` | **-1** (no limit, never set in config) |
| `journal_mode` | `wal` |

File sizes at the time of the measurement:

- `grappa_prod.db` — 1 517 092 864 B logical (~1.41 GiB), **399 M** on disk (ZFS compression)
- `grappa_prod.db-wal` — 176 094 192 B logical (~168 MiB), 54 M on disk
- WAL mtime moving, but the file had not been reset since 11:02 that morning

A self-hosted instance without the ZFS dataset reports a `-wal` of **5.8 M** under
comparable use (reported by a self-hoster on IRC, not measured by me).

## The defect

`wal_autocheckpoint` counts **pages, not bytes**. With SQLite's default
`page_size 4096` the default of 1000 pages means a passive checkpoint is
attempted every ~4 MiB. `docs/zfs-baseline-2026-07-31.md` deliberately moved
production to `page_size 65536` to match the ZFS `recordsize 64K` — a good
change on its own terms — but the checkpoint threshold was never re-tuned, so
the same `1000` now means **~64 MiB**. The change multiplied the checkpoint
interval by 16 as a side effect.

`journal_size_limit = -1` compounds it: after a checkpoint SQLite does not
truncate the WAL, it keeps and recycles the file at its high-water mark. So the
file never comes back down once a burst has grown it.

Neither PRAGMA appears anywhere in `config/runtime.exs`, which otherwise pins
`busy_timeout`, `journal_mode`, `cache_size`, `temp_store`, `synchronous` and
`foreign_keys` — the latter two pinned *explicitly* under the REV-B / C3
rationale that "defaults are right by accident". These two are the same class,
and here the default stopped being right the moment the page size changed.

## Impact

No data loss: `synchronous: :normal` under WAL is sound, and the WAL is a valid
part of the database. What it costs is:

- crash recovery has to replay a much larger WAL
- disk high-water mark that never recedes
- checkpoint work arrives in large infrequent batches instead of small frequent
  ones — the opposite of what the ZFS work was tuning for

## Suggested fix

[...]

Issue #506, opened 2026-07-28
boot: fresh-DB journal-mode switch races the migrate connection (benign log noise, fixed at the source)


vjt:
## Summary

The e2e integration stack emits a boot-time `Exqlite.Error: database is locked`
on the second pool connection (`db_conn_2`) during `Grappa.Repo` creation /
migration. It fired on **both** e2e runs tonight, disturbs test seeds, and
produces flaky reds. Sibling of #499 but a **different mechanism** (this one is
a boot/migration-time pool-connection race, not the runtime path #499 covers).

## Evidence (two independent runs, same night)

**Run A — full integration ship gate** (`/tmp/w2-integration-477.log:413`):

```
The database for Grappa.Repo has been created

03:30:13.190 [error] Exqlite.Connection (#PID<0.464.0> ("db_conn_2")) failed to connect: ** (Exqlite.Error) database is locked

03:30:14.001 [info] == Running 20260425000000 Grappa.Repo.Migrations.Init.change/0 forward
```

**Run B — isolated single-spec re-run** (`/tmp/w2-iso-cp15b6.log:158`):

```
03:53:03.080 [error] Exqlite.Connection (#PID<0.464.0> ("db_conn_2")) failed to connect: ** (Exqlite.Error) database is locked
```

Same PID label, same pool connection (`db_conn_2`), same phase: immediately
after "database has been created", before/around the `Init` migration.

## Why it matters

The lock lands during the window where the stack is being seeded, so a seed that
races the locked connection can silently under-populate — which then surfaces as
an unrelated, intermittent red downstream (e.g. a members-pane precondition
timing out). See the companion #495-recurrence report for a concrete instance
that this boot lock is a candidate root cause for.

## Not in scope

- Not #499 (different mechanism — that's runtime, this is boot/migration).
- No fix proposed here; this is a triage record so the mechanism is tracked
  rather than re-diagnosed from scratch next time.

vjt:
## Downstream consequence: a later API write 500s (not just seed disturbance)

Control-arm evidence from the #498 integration investigation. Isolated
single-variable run (`issue299-theme-cards.spec.ts` x3, chromium) with the
**#498 seeder REMOVED**, so this is exogenous — NOT a #498 regression:

- boot: `Exqlite.Connection (db_conn_1) failed to connect: ** (Exqlite.Error) database is locked`
[...]

PR #1369, opened 2026-08-16
sqlite(#1355): denominate the WAL checkpoint threshold in bytes

vjt:
Fixes the defect in #1355: `PRAGMA wal_autocheckpoint` counts **pages**, so the
ZFS baseline's `page_size` 4096 → 65536 move multiplied the effective checkpoint
threshold by 16 as a side effect, and `journal_size_limit = -1` meant the WAL
never came back down from its high-water mark.

## 🔴 This is a COLD deploy

The cure lives in `config/runtime.exs`. There is no hot path for it and no
attempt was made to invent one.

## What changed

- **`config/runtime.exs`** — one number, `wal_checkpoint_bytes = 16 MiB`, in
  BYTES, feeding both PRAGMAs. `journal_size_limit` takes bytes natively and is
  pinned to the same value.
- **`lib/grappa/repo.ex`** — `init(:supervisor, _)` reads the DB file's **live**
  `page_size` and divides, producing the page count the PRAGMA actually takes.
  It does this on the serial pre-pool connection that #506 already opens for the
  rollback→WAL switch, so no new connection and no new boot step. `:runtime`
  stays pure, as #506 requires.
- **Docs** — `docs/zfs-baseline-2026-07-31.md` gains a *What `page_size` also
  moves* section (with a pointer from the `page_size` table itself), so the
  coupling is written where the next person to change the page size will look;
  DESIGN_NOTES carries the decision record.

**A re-computed page count was rejected on purpose.** It would fix today's
number and leave the identical trap armed for the next `page_size` change,
which is the point of the issue rather than a detail of it. The invariant is:
never pin a page count; express the intent in bytes and derive.

## The rounding is load-bearing, not tidiness

The division rounds **UP**. `wal_autocheckpoint = 0` does not mean "checkpoint
constantly" — it **disables automatic checkpointing outright**. A floor division
with a large enough `page_size` and a small enough threshold would therefore
switch off the very mechanism this issue exists to repair, silently. That is the
same class of trap as the one being fixed, so it is guarded and tested rather
than assumed away.

## The number, and how it was arrived at

**16 MiB is a tuning choice, not a measurement.** I have no production DB in
this worktree; every figure in the issue is vjt-claude's, and I re-measured
none of them.

It is argued between two bounds:

- **Lower bound ~4 MiB.** That was the effective threshold here for the whole
  life of the deployment (SQLite's 1000-page default at `page_size 4096`), and
  nothing ever complained about it.
- **Why not simply restore it.** A 64 KiB page makes a single-row update dirty
  16× more WAL bytes than a 4 KiB one did. Pinning 4 MiB would therefore
  checkpoint far more often *in wall-clock terms* than this deployment has ever
  done — an over-correction in the opposite direction.
- **16 MiB** sits between them: it caps the WAL an order of magnitude below the
  168 MiB observed in the issue, keeps crash-recovery replay trivial, and is a
  whole number of pages at **both** 4096 and 65536, so the round-up never
  actually fires on the configured value.

`journal_size_limit` equal to the threshold is also deliberate: in steady state
the WAL is already at that size, so truncation is a no-op and there is no
truncate/regrow churn, while a burst-grown WAL is cut back at the next reset.

Deliberately **not** an env var — it is a tuning constant, not an operator knob,
and every env var carries the registry/compose/`.env.example` tax.

- **No before/after measurement.** The change is justified structurally, on the
  unit mismatch, not on a benchmark delta. Nobody should read a performance
  claim into it.
- **A lead I did not measure, flagged as a lead.** A passive autocheckpoint
  cannot reset the WAL while a reader still holds an older snapshot, and
  `journal_size_limit` only truncates when a reset actually happens. With a
  read-concurrent pool that could plausibly be part of why the observed WAL was
  much larger than the threshold alone predicts — but I did not measure it, I
  did not look at the live node, and I am attaching no numbers to it. If it
  holds up it is a separate issue, not this one.
- **Dev and test keep SQLite's default** (~4 MiB at their 4096 page size). The
  issue is prod-scoped and so is the pin.

## Gates

`scripts/check.sh` on the branch: **RC=0** — `8 doctests, 60 properties, 6322
tests, 0 failures`; Credo strict `found no issues` across 657 files; Dialyzer
`Total errors: 0`; Sobelow, doctor, deps.audit, wire-type drift gate and the
482-test bats suite all green. `mix format` produced no diff.
`scripts/design-notes-gate.sh` RC=0.

## The tests are not mirrors — measured

Nine new tests in `test/grappa/repo_wal_checkpoint_test.exs`, and two deliberate
mutants were run against them to prove they witness the defect rather than the
implementation:

| mutant | result |
|---|---|
| `div(bytes + page_size - 1, page_size)` → `div(bytes, page_size)` (drop the round-up) | **1 test, exactly 1 assertion**: `left: 0` — the value that disables checkpointing |
| the derivation → a bare `1000` (the defect itself) | **3 tests**, the load-bearing one reporting `left: 4096000, right: 16777216` — literally 4 MiB where 16 MiB was meant |

Both reverted; the tree is clean and the branch is the green source.

The decisive assertion is that the *same* config yields the *same BYTES* at both
4096 and 65536 — a bare page count produces the same number twice and therefore
two thresholds 16× apart, which is exactly the bug. A second test reads the real
prod block of `config/runtime.exs` through `Config.Reader` and fails if prod
ever pins `wal_auto_check_point` directly.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -148,6 +148,29 @@ if config_env() == :prod do
 
   config :grappa, :uploads_storage_root, uploads_storage_root
 
+  # #1355 — the WAL's steady-state byte envelope. ONE number, in BYTES,
+  # feeding both PRAGMAs below, because both express a byte-shaped intent:
+  #
+  #   * `journal_size_limit` takes bytes directly.
+  #   * `wal_autocheckpoint` takes PAGES, so `Grappa.Repo.init/2` divides this
+  #     by the DB file's LIVE `page_size` before the pool opens. NEVER pin a
+  #     page count here: `docs/zfs-baseline-2026-07-31.md` moved prod from
+  #     `page_size 4096` to `65536`, and the untouched 1000-page default
+  #     silently went from ~4 MiB to ~64 MiB.
+  #
+  # 16 MiB is a TUNING CHOICE, not a measurement, argued between two bounds:
+  # SQLite's own default was ~4 MiB here before the page-size change, but a
+  # 64 KiB page makes a one-row update dirty 16× more WAL bytes, so pinning
+  # 4 MiB would checkpoint far more often in wall-clock terms than this
+  # deployment ever did. 16 MiB keeps the checkpoint cadence in the same
+  # order as the pre-ZFS one while capping the WAL an order of magnitude
+  # below the 168 MiB observed in #1355 — and it is a whole number of pages
+  # at both 4096 and 65536, so no rounding either way.
+  #
+  # Deliberately NOT an env var: it is a tuning constant, not an operator
+  # knob, and every env var carries a registry/compose/.env.example tax.
+  wal_checkpoint_bytes = 16 * 1024 * 1024
+
   # NB: `:cic_dist_root` is derived ABOVE, hoisted out of this prod block
   # (all envs except :test) since #485 — see the comment there, which
   # folds in the #526 jail-CWD knowledge that used to live here.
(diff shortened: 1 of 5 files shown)


vjt:

Issue #1358, opened 2026-08-15
cic(iOS): keyboard closes when a long draft crosses back under the single-frame limit


vjt:
On iOS, deleting characters from a draft that is over the single-frame limit closes the keyboard as the draft crosses back under it. **The textarea loses the focus** — measured on the device by vjt (#grappa, 22:09).

## Why that is strange

Measured on `origin/main` (`0376738a`):

- **Nothing calls `blur()`.** The only occurrences of the word in `cicchetto/src/ComposeBox.tsx` are the comments at `:748` and `:789`, both explaining why earlier fixes avoided losing the focus.
- **`readOnly` is not on this path.** `:759` — `readOnly={isDraining(key()) || isQueueFull(key())}`: paced-send drain and full queue, not the frame limit. It was chosen over `disabled` precisely because `disabled` blurs (`:748`).
- **No `<Show>` on the seam path wraps the textarea.** The textarea is inside the `<form>` at `:713`; the countdown is a sibling `<p>` ABOVE the form (`:660`), the split warning a sibling `<p>` BELOW it (`:970`, `SplitWarningLine` at `:149`).

A focus loss with no `blur()` in the source means the focused node is **leaving the DOM**.

## Primary suspect: root-fragment reconciliation moves the form

`ComposeBox`'s root is a fragment whose children are, in order: countdown `<Show>` (`:660`), `<form>` (`:671`), uploads `<Show>` (`:893`), greyed `<Show>` (`:935`), feedback/split-warning `<Show>` (`:970`).

Crossing the limit downward flips **two** of those siblings in the same frame: the split warning is removed below, and the countdown appears above — `frameCountdown()` is non-null only for the last 10 bytes of the budget, which is exactly the band the draft re-enters. If the countdown's insertion is reconciled by re-inserting the following siblings rather than by `insertBefore` against a stable marker, the `<form>` is detached and re-attached, and a detached focused textarea loses focus with nobody calling `blur()`.

That is consistent with every observation: no blur in source, focus genuinely lost, and only on the transition that moves two seams at once.

The countdown `<Show>` is also `keyed` (`:660`), so its `<p>` is re-created on every value change, not only on appear/disappear.

## The measurement that settles it

Needs a DOM. Mount `ComposeBox`, focus the textarea, drive the draft across the frame limit in both directions, then assert:

1. the textarea is the SAME node object before and after, and
2. `document.activeElement` is still it.

[...]

PR #1371, opened 2026-08-16
cic(#1358): stop the frame seam from moving the focused composer

vjt:
## What was measured

vjt's report (#1358) is iOS: deleting characters from a draft that is over the
single-frame limit closes the keyboard as the draft crosses back under it. The
mechanism is not iOS. `ComposeBox`'s root is a fragment; the byte countdown is
a sibling ABOVE the `<form>`, the split warning a sibling BELOW it. Crossing
the limit downward flips both in the same flush, and Solid reconciled that by
**detaching the form and re-attaching it** after the new sibling. The focused
textarea rode along inside it, and a detached element loses the focus in every
engine with nobody calling `blur()` — which is why no `blur()` is findable in
the source.

Traced with a `MutationObserver` over the container:

```
REMOVED: FORM   (with the focused textarea inside)
ADDED:   P      (the countdown arrives above)
REMOVED: P      (the split warning leaves below)
ADDED:   FORM   (the SAME node object goes back)
```

### One transition out of eight, and it is the reported one

| transition | form detached | focus kept |
|---|---|---|
| countdown appears / changes value / disappears, alone | no | yes |
| split warning appears / disappears / changes text, alone | no | yes |
| **up** over the limit (countdown out above, warning in below) | no | yes |
| **down** under the limit (warning out below, countdown in above) | **yes** | **no** |

Asymmetric, and the asymmetry matches the report: a deleting thumb travels
downward.

### Two premises in the issue were wrong

- **"jsdom is broken here and cic tests run `--environment node`" is false for
  this repo.** `cicchetto/vitest.config.ts:14` sets `environment: "jsdom"`, and
  jsdom resets `document.activeElement` to `<body>` when the focused node
  leaves the document. Solid reconciles with plain `removeChild`/`insertBefore`
  in every DOM, so the movement and the focus loss are engine-independent.
  **No e2e, no browser and no device were needed to measure this.**
- **The issue's oracle (1) — "the textarea is the SAME node object before and
  after" — is a FALSE NEGATIVE.** Measured: across that transition Solid hands
  back the same `<textarea>` and the same `<form>`, so an identity assert
  passes with the keyboard already gone. The oracle is `document.activeElement`;
  the detach is reported beside it so a red says why.

## The change

`ComposeBoxFrameSeam.test.tsx` — its own file, because `ComposeBox.test.tsx`
mocks `lib/compose` wholesale, so the draft there is a constant and the seams
cannot be *crossed*, which is the entire subject. Here the compose store, the
frame budget and the ISUPPORT table are real; only `networkBySlug` is stubbed.
The invariant is stated over the **class** — no seam transition may unmount the
focused composer — so all eight rows are pinned, not just the reported gesture.
Each row asserts its pre-state before the gesture, so a row cannot pass having
crossed nothing. Committed red first (7 green, 1 red on the downward crossing).

`ComposeBox.tsx` — the countdown `Show` now sits inside an always-mounted
`div`, so the appear/disappear happens where the form is not a sibling.

### Alternatives measured and rejected

- **An always-mounted `<p>` whose text merely empties** (the issue's
  suggestion): also clean on all eight, but the defect is the identity of the
  node in the slot, not the number of slots — and this shape additionally
  changes an unrelated observable contract ("hidden means the node is absent"),
  turning two currently-green assertions in `ComposeBox.test.tsx` red. The slot
  leaves observable behaviour identical except for the bug.
- **A `fallback` on the existing `Show`**: measurably WORSE — it breaks the
  upward crossing too, because swapping the node in the slot is the very thing
  that relocates the form.
- **Re-focusing after the fact**: not attempted. It papers over a DOM move
  instead of preventing it and fights the UX-6-D `preventScroll` history
  (`composeAppend.ts:26`).

### The empty slot consumes no gap — checked, not assumed

An always-mounted child consumes a flex/grid `gap` even at zero height, which
would be a permanent phantom space above the box. Both `ComposeBox` mount sites
are inside `DropUploadZone`, so the fragment's only parent is
`.drop-upload-zone` (`themes/default.css:1394`): `display:flex;
flex-direction:column`, **no `gap`, no `row-gap`**. That file is the only
stylesheet in the codebase and contains **no `:empty` rule anywhere**. The slot
therefore carries no CSS rule and needs none.

## Gates

- `scripts/bun.sh run test` — **5499/5499 green**, 285 files, including the
  eight new rows. Zero churn on existing assertions.
- `scripts/bun.sh run check` (biome + `tsc --noEmit` over `src` and `e2e`) — RC 0.
- `scripts/design-notes-gate.sh` — 1 new entry, separator and marker present.
- No Elixir, no `VERSION`, no `mix.exs`, no migrations touched, so no server
  gate applies.

**Deploy class: cic bundle only** (`scripts/deploy-m42.sh --cic`) — verified,
not asserted: `git diff --name-only origin/main...HEAD` is two files under
`cicchetto/` and one under `docs/`. A server hot-reload would not ship this at
all.

- **That the slot is visually inert in a real engine.** The gap/`:empty` check
  above is static; no browser was driven (and none can be driven from this
  host).
- **The detach-to-keyboard chain on iOS.** What is established here is that the
  form is detached and the focus is lost. That this is what closes vjt's
  keyboard remains **vjt's device-verify**, still owed.

### DIFF
--- a/cicchetto/src/ComposeBox.tsx
+++ b/cicchetto/src/ComposeBox.tsx
@@ -657,17 +657,32 @@ const ComposeBox: Component<Props> = (props) => {
           visual: aria-hidden, because a live region that changes on every
           keystroke is noise, and the thing worth ANNOUNCING (the draft will
           split) is the polite seam line below. */}
-      <Show when={frameCountdown()} keyed>
-        {(label) => (
-          <p
-            class="compose-box-frame-countdown"
-            data-testid="compose-frame-countdown"
-            aria-hidden="true"
-          >
-            {label}
-          </p>
-        )}
-      </Show>
+      {/* #1358 — the slot is ALWAYS mounted; only its content comes and goes.
+          The countdown used to be a bare `Show` among the root fragment's
+          children, and when it arrived in the same flush the split warning
+          left — the draft shrinking back under the frame limit — Solid
+          reconciled the fragment by DETACHING the <form> and re-attaching it
+          after the new sibling. The focused textarea rode along, and a
+          detached element loses the focus with nobody calling blur(): on iOS
+          that closes the keyboard mid-edit. A stable parent keeps the
+          appear/disappear inside this slot, where the form is not a sibling.
+          Measured in ComposeBoxFrameSeam.test.tsx over all eight seam
+          transitions; only that one moved the form, and only downward.
+          The node identity of the textarea is NOT the oracle — Solid handed
+          back the same node object across the move, focus already gone. */}
+      <div class="compose-box-frame-countdown-slot">
+        <Show when={frameCountdown()} keyed>
+          {(label) => (
+            <p
+              class="compose-box-frame-countdown"
+              data-testid="compose-frame-countdown"
+              aria-hidden="true"
+            >
+              {label}
+            </p>
+          )}
+        </Show>
+      </div>
       <form
         class={`compose-box${greyed() ? " compose-box-greyed" : ""}`}
         onSubmit={(e) => {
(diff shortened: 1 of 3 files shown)


vjt:

Issue #1372, opened 2026-08-16
review: the unread-count aggregate lost its covering index, and nothing pins the covering property (P-S1, P-S4, P-S5)


vjt:
Bucket issue from the 2026-08-15 code review (`docs/reviews/codebase/2026-08-15-codebase-review.md`,
merged as `3a8eddec`). One theme: index coverage on `messages`, the highest-write-rate table.
Read the findings there — they carry the measurements and the file:line detail.

- **P-S1 (HIGH)** — `#532A`/`#576` added `exclude_own_authored/3` to `count_after_split/6`, and the new
  predicate reads two columns that appear in **no index on `messages`** (`sender` via
  `Identifier.nick_fold/1`, and `json_extract(meta,'$.new_nick')`). #393 had shipped its composites
  explicitly so that aggregate would be **covering** — its migration moduledoc records 80ms → 5-7ms,
  ~15×. The aggregate can no longer be satisfied from the index, so SQLite does one table row-fetch per
  post-cursor row. That is the shape #393 measured as the 2026-07-25 prod incident, and the worst
  window is the one that matters: a channel whose read cursor is far behind, which is what the cold-load
  `/me` seed and the per-channel join reply both route through (`WindowCounts.snapshot/7`).
  The DM arm has the same problem through the folded-COALESCE index.
  Review section: `#### P-S1` (`lib/grappa/scrollback.ex:634-667`, `:710-744`).

- **P-S4 (MEDIUM)** — four dead indexes on `messages`; write amplification on the busiest table.
- **P-S5 (MEDIUM)** — `user_settings.data` read-modify-write loses concurrent writes to different keys.

## Two constraints this work carries

1. **Measure before changing anything.** `EXPLAIN QUERY PLAN` against realistic data — the review says
   so, and the fix shape (extend the composites vs. restructure so the exclusion is not evaluated per
   row) is a decision that only a plan reading can make. A guessed index on the highest-write-rate
   table costs writes forever.
2. **Pin the covering property.** The review is explicit that *the absence of such a pin is what let
   this through*. ⚠️ A plan pin is harder than it looks: SQLite chooses the plan **per connection**, so
   a pool checkout that never served the `CREATE INDEX` plans a full scan and the pin flips run to run.
   The working shape — measured in #1355/#1353 — is **two assertions** (first that `sqlite_master`
   holds the index with the expected SQL, then the plan) plus `pool_size: 1` in the test setup with the
[...]

Issue #393, opened 2026-07-25
perf: badge-count fan-out + non-sargable rfc1459 fold saturate the SQLite pool (slow logon, /motd, send)


vjt:
## Symptom

Reported in production on 0.4.0: logon is slow, `/motd` is slow and intermittently returns nothing at all on repeat invocations, and sending messages became noticeably slower. All three degrade together.

## Measured root cause

Production logs (2026-07-25 08:43) show the SQLite pool saturating:

```
[warning] scrollback persist unavailable: SQLite pool saturated for the full
1500ms retry budget (1 attempts) — dropping row
** (DBConnection.ConnectionError) connection not available and request was
dropped from queue after 6187ms
[info] Sent 500 in 5023ms
```

Scrollback rows were dropped and at least one HTTP request returned 500. The interrupted query in the stack trace comes from:

```
(grappa 0.4.0) lib/grappa/push/badge_count.ex:176: Grappa.Push.BadgeCount.count_window/7
(grappa 0.4.0) lib/grappa/push/badge_count.ex:143: Grappa.Push.BadgeCount.accumulate/5
Function: #Function<... in GrappaWeb.ReadCursorController.create/2>
```

Two compounding factors:

**1. Per-POST fan-out.** `BadgeCount.count/1` folds over every window of the subject and calls `count_window/7` → `Scrollback.unread_content_tail/6` once per window. Measured on a copy of the production database: **average 8.5 windows per (subject, network), maximum 29**. One read-cursor POST therefore issues up to 29 scrollback queries.

**2. The DM read predicate is not sargable.** `Scrollback.where_dm_peer/2` applies the rfc1459 fold to the *column*:

```elixir
Identifier.nick_fold(m.dm_with) == ^folded_peer or
  (is_nil(m.dm_with) and Identifier.nick_fold(m.channel) == ^folded_peer)
```

`nick_fold/1` expands to `replace(replace(replace(replace(lower(?), '[', '{'), ']', '}'), '\', '|'), '~', '^')`. Wrapping the column in an expression means none of the eleven `messages` indexes can serve the comparison. `messages` currently holds 653k rows; the query planner falls back to `messages_network_id_index` and evaluates the fold row by row.

Timings taken directly against production, same query shape:

| variant | time |
|---|---|
| folded comparison (what ships) | 51 ms |
| plain equality, same shape | 8 ms |

[...]

Issue #372, opened 2026-07-21
cicchetto: incoming query-window routing is case-sensitive → duplicate split windows (outgoing 'debugserv' vs replies in 'DebugServ')


vjt:
## Summary

Incoming query-window routing in cicchetto is case-sensitive: when the outgoing query was opened with one casing and the reply arrives from a differently-cased nick, the reply lands in a **separate** query window instead of the one the user opened. Result: a "dead" outgoing window plus a second (often archived) window holding the replies, split purely by nick casing.

## Repro (reported on #grappa, testnet)

1. `/msg debugserv HELP` (lowercase) → opens a window keyed `debugserv` containing the sent command(s). It looks dead — no replies appear there.
2. The service replies as `DebugServ` (proper case). Those replies land in a **different** window, `DebugServ`, which ends up in *archived*.
3. `/msg DebugServ HELP` (matching case) works as expected — replies show in the same window.

So the two windows (`debugserv` = what I sent, `DebugServ` = the replies) coexist and are **not merged**.

Additional oddity: pressing X → "really delete" on **either** of the two windows deletes **both** — yet their message contents are not merged, they stay separate. So the delete path treats them as the same key while the message-routing / display path treats them as different keys. That inconsistency is itself a bug and a hint at where the mismatch lives.

## Why it happens (source)

The nick case-folding fix from 2026-05-17 (`canonicalQueryNick`, see `cicchetto/e2e/tests/nick-case-sensitivity.spec.ts`) is applied only at **outgoing / user-action** sites: `compose.ts` (`/msg`, `/query`), `MembersPane`, `NamesModal`, `ScrollbackPane`, `UserContextMenu`. The **incoming** path — an arriving PRIVMSG/NOTICE being routed to a query window by sender nick — does not canonicalise the sender against existing windows, so a differently-cased sender mints a phantom duplicate window. (`nickEquals()` — case-insensitive compare, already used for the members store in `lib/members.ts` — is not applied to incoming query-window keying.)

## Fix direction

- On the incoming message → query-window association, resolve the sender nick against existing windows case-insensitively (reuse `canonicalQueryNick` / `nickEquals`) so a reply lands in the window the user opened, regardless of casing.
- Reconcile the delete-coupling vs display-split inconsistency (delete matches case-insensitively, routing/display does not — pick one canonical key everywhere).
[...]

PR #1376, opened 2026-08-16
perf(#1372): restore the unread aggregate's covering index, and un-blind the pin

vjt:
Bucket B of the 2026-08-15 review: P-S1 (the unread aggregate lost its covering
index) and P-S4 (four dead indexes on the same table). P-S5 is deliberately not
here — it is a lost update on `user_settings.data`, another table, no index
dimension; the orchestrator is filing it separately.

## The finding that outlived the fix

P-S1 closes with *"the absence of such a pin is what let this through."* **The
pin existed, and it was green.** Three of them, asserting `USING COVERING
INDEX` — at `66759042:test/grappa/scrollback_test.exs:2013`, `:2173` and
`:2183` (line numbers as of the branch point, since this PR moves them).

They EXPLAINed `count_split_query/3` (`66759042:…:123`) — a **local rebuild**
of the aggregate, not the production query — and it passed `own_nick: nil` to
`channel_or_dm_where/3`. At nil, `exclude_own_authored/3` is the identity
clause (`lib/grappa/scrollback.ex:711`, unchanged here), so the predicate
#532 A and #576 added
never entered the pinned SQL. The pin asserted COVERING against a shape
production had stopped running and stayed green straight through the regression
it existed to catch.

Measured on one corpus, same channel, same cursor, 103,209 post-cursor rows:

| shape | plan | page fetches |
|---|---|---|
| `own_nick = nil` — what the pin asserted | **COVERING** INDEX | **1,676** |
| `own_nick = "vjt"` — what production runs | INDEX, no COVERING | **175,576** |

A pin on a branch production does not take is worse than no pin: it reports
safety. So it is not fixed by adding a second pin beside it — the plan is now
read off the SQL the production function emits, captured from
`[:grappa, :repo, :query]` telemetry, and the local rebuild is deleted. There
is no second copy left to drift.

The same defect class shows up twice more in the same file and both are cleaned
up here: two tests EXPLAINing the `(channel = ? OR dm_with = ?)` disjunction
#393 deleted from production, which were the only remaining readers of the four
indexes P-S4 drops.

## Method

No prod copy on this lane, so one was built: the real migrated DDL from
`runtime/grappa_dev.db` (13 indexes on `messages`), loaded with 650k synthetic
rows — 20 channels zipf-skewed, `#linux` heaviest at 103,209 rows on the busy
network; 192 peers plus the own nick at ~4% of `sender`; a realistic kind mix;
10% DM rows split inbound/outbound the way `Session.Server` persists them;
`meta.new_nick` on `nick_change` with 5% renaming *to* the own nick; ids
interleaved across channels so a channel's rows are contiguous in the index and
**scattered in the table**, which is where the whole cost lives.

**No `ANALYZE`** — verified absent from `lib/ priv/ scripts/ config/`, and no
`sqlite_stat*` in any runtime DB, so prod plans on default estimates and the
bench must too.

The primary metric is **page fetches**, not milliseconds: deterministic and
host-independent. The SQL is the real emitted statement, captured off
telemetry, not transcribed.

## P-S1 — the fix, and the shape that lost

Widen both covering families with `lower(sender)` and
`json_extract(meta, '$.new_nick')`:

- **Read:** 175,576 → 1,886 page fetches (**93x**), identical results
  (52112/46620 both sides). Page *misses* 23,427 → 1,885 is **14x**, which
  lands on the same ~15x #393 recorded from a completely independent
  measurement.
- **Write — the half that decides it**, on the highest-write-rate table in the
  codebase. 50k single-row INSERTs, 3 reps:

  | variant | insert | `messages` index bytes |
  |---|---|---|
  | main today | 0.50 / 0.50 / 0.48 s | 348 MB |
  | + widen (P-S1) | 0.48 / 0.49 / 0.48 s | 373 MB |
  | + widen + drop (P-S1 + P-S4) | **0.37 / 0.37 / 0.37 s** | **271 MB** |
  | drop alone (P-S4) | 0.35 / 0.36 / 0.36 s | 246 MB |

  Widening four existing entries adds no b-tree, so it costs nothing
  measurable. **The alternative — count own-authored rows as a second seek on a
  `sender`-leading index and subtract — costs +37% on INSERT** (0.67/0.67/0.68 s),
  because it adds four new b-trees. It is also not a subtraction: the predicate
  is a disjunction whose arms overlap (a case-only self-rename matches both), so
  it needs inclusion–exclusion across three seeks, each reproducing the GROUP BY
  bucketing, to replace one query. Dearer and more fragile.
- Widening with `lower(sender)` **alone** was measured and does **not** restore
  COVERING. Both expressions are load-bearing.

## P-S4 — measured dead, not just grepped

An 11-shape `EXPLAIN QUERY PLAN` sweep (channel fetch, DM fetch, both
`fetch_after`, both `count_after_split`, both content tails, `list_archive`
GROUP BY, `delete_for_dm`, `server_time` page) across main / widened /
widened+dropped:

- **No shape regresses.** Every plan is byte-identical across the three
  variants except the two `count_after_split` arms, which go `INDEX` →
  `COVERING INDEX`.
- **widened+dropped is identical to widened on all 11** — nothing was using the
  four dropped indexes.

The two `messages_archive_*_idx` are deliberately untouched: DESIGN_NOTES
17317-17332 records their #372 staleness and a deliberate KEEP.

## The pin, and how each assertion was bought

Two arms, because a plan assertion alone cannot say *which* expression went
missing, and "not COVERING" reads identically whether an expression drifted or
the planner merely passed a good index over (the #1353/#1355 shape). Arm 1 is
the DDL SQLite stored; arm 2 is the plan.

Every mutant restored afterwards, `git status` empty, baseline reprinted green
(`1 property, 181 tests, 0 failures`) at both ends:

| mutant | tally | reddens |
|---|---|---|
| **M1** index: JSON path `$.new_nick`→`$.newnick` | 181 tests, **5 failures** | 2 DDL arms + 3 plan arms |
| **M2** index: `fold("sender")`→`fold("id")` | 181 tests, **5 failures** | 2 DDL arms + 3 plan arms |
| **M3** *query*: JSON path drifts, index untouched | 181 tests, **3 failures** | **plan arms ONLY — DDL arms green** |
| **M4** *query*: fold `m.body` instead of `m.sender` | 181 tests, **5 failures** | 3 plan arms + 2 behavioural tests |

**Honest reading of that table.** M1 and M2 each kill two assertions, because
an index-side drift necessarily costs COVERING — the DDL arm implies the plan
arm on that class. The house rule says a mutant that kills two means the pair
needs rethinking, so **M3 is the one that answers it**: a query-side drift away
from a perfectly good index reddens the plan arms while the DDL arms stay
green. That is a real regression class the DDL arm cannot see, so the arms are
complementary, not redundant — and on the index-side class the DDL arm is what
names *which* expression died. M4 additionally shows the `#532 A`/`#576`
behavioural tests are not asleep.

**A trap worth naming:** M1 and M2 initially *survived*. `mix test` runs
`ecto.migrate`, which only applies **pending** migrations — so editing a
migration whose version is already in `schema_migrations` is a silent no-op and
an index-side mutant never takes effect. CI is unaffected (fresh DB per run);
locally the harness has to drop the DB between rounds. A harness without that
step would have reported these assertions bought when they were not.

## Deploy — COLD, and how long the window is

Two new migration files, Preflight Class 5. All four `CREATE INDEX` builds
share ONE transaction, so the transaction is what sizes the window, not the
per-index split.

**Measured at prod's actual row count.** The first pass of this used a 650k
corpus because that is what #393 recorded. That number is three weeks stale and
three times too small: vjt read prod on 2026-08-16 — **`messages` = 1,943,545
rows, DB 1.53 GB**. So the corpus was rebuilt at 1,943,545 rows (1.23 GB; prod
is larger because it carries the other tables and the archive indexes) and
re-measured, fresh copy per rep so no rep inherits the previous page cache:

| | rep 1 | rep 2 | rep 3 |
|---|---|---|---|
| **P-S1 `up/0`** — 4 drop + 4 create, 1 txn | 9.66 s | 9.55 s | 9.85 s |
| P-S4 `up/0` — 4 drop, 1 txn | 0.09 s | 0.08 s | 0.09 s |
| **both back to back — the real migrate step** | 10.14 s | 9.72 s | 9.72 s |

Per-CREATE at 1.9M: `user_channel` 2.16-2.31 s · `visitor_channel` 1.63-1.74 s ·
`user_dm` 2.28-2.32 s · `visitor_dm` 1.73-1.78 s. Internal consistency check:
2.99x the rows cost 3.24x the time (2.9 s → 9.7 s), the mild superlinearity a
sort-dominated build should show.

**Prod estimate — EXPLICITLY DERIVED, not observed.** The above is a macOS/APFS
laptop; prod is a FreeBSD bastille jail on ZFS. The scaling factor is #393's own
prod record for the same operation on the same table: 1.6-2.0 s per channel
index and 2.442/2.570 s per DM index at 654k rows, against 0.52-0.69 s locally
at 650k — a **3-4.5x host factor**, and conservative, because #393's indexes
were NARROWER than these (5 columns vs 7). Applying it to the measured 9.7 s:

> **expect roughly 30-45 seconds for the COLD migrate step on m42.**

This supersedes an earlier "order of 10 seconds" in this PR's history, which was
derived from the stale 654k figure. **Nobody has measured an index build on
prod's substrate at 1.9M rows** — the row count is observed, the local build is
observed, the host factor is borrowed from #393, and the product of the last two
is an extrapolation.

Not measured: the WAL commit cost. The transaction writes the four new indexes
into the WAL before committing, and #1355 has just changed the checkpoint
threshold. On a COLD deploy the BEAM is down, so at least concurrent write load
is genuinely absent rather than merely unmodelled.

The widen is expand-class (no schema-shape change, running old code unaffected);
the drop is contract-class but has no reader, which is what the sweep
establishes.

## What I did NOT establish

- **No prod copy.** Distributions are chosen, not observed. The ~4%
  own-`sender` share moves the absolute numbers; it does not move whether the
  plan is COVERING.
- **Wall-clock is not prod's** — warm macOS/APFS, no pool contention, no
  FreeBSD jail. Cold-cache timing is unmeasured (`purge` needs sudo); the
  23,427 page misses stand as the proxy for the I/O the incident was made of.
- **#1353's open question stays open.** Why a pool connection that did not
  serve the `CREATE INDEX` plans differently — schema cache or transaction
  visibility — is still unmeasured. `config/test.exs` is already `pool_size: 1`
  for the Sandbox-serialisation reason, which happens to satisfy the
  plan-stability need too; I have not re-derived the mechanism and am not
  claiming it.
- **`ReadCursor.own_authored_dynamic/2`** (`read_cursor.ex:483`) is the same
  predicate shape against the same table, flagged by the review as the same
  defect class and explicitly out of P-S1's scope. It plausibly benefits from
  these same widened indexes — **I did not measure it**, and nothing here
  claims it is fixed.
- Prod's real cursor-lag distribution. The cold `/me` seed is the worst case by
  construction; how far behind a typical cursor sits, I don't know.
- Three `users_folded_name_index` failures appeared in the full suite on the
  **shared** test DB while this branch was based on `66759042`. Established as
  contamination, not assumed: the index was in no migration on this branch, it
  came from `w1-1353-name-fold`, and the full suite under an isolated cache and
  DB (`GRAPPA_CACHE_ID`) was **6315 tests, 0 failures**. #1353 has since landed
  on main, so after the rebase the branch carries that migration legitimately
  and the three tests pass on any DB. Recorded because the *reasoning* is what
  matters here: a red naming something you never touched is a contamination
  suspect before it is a defect.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -43368,3 +43368,68 @@ when a reset actually happens. With `pool_size: 10` under continuous reads
 that is a real possibility, and it is the most likely reason prod's WAL was at
 168 MiB rather than the ~64 MiB the threshold alone predicts. The change lowers
 the pressure and bounds the recovery; it does not promise a ceiling.
+<!-- entry #1372 -->
+
+---
+
+## 2026-08-16 — #1372: the unread aggregate lost its covering index, and the pin that should have caught it was blind
+
+`Scrollback.count_after_split/6` is the query behind the cold-load `/me` seed
+and the per-channel join reply (`WindowCounts.snapshot/7`). #393 shipped two
+covering index families for it on 2026-07-25 — `kind` at the tail so the
+content-vs-event GROUP BY could be answered without touching the table —
+because the non-covering form was the prod incident of that morning: its
+migration moduledoc records 80ms → 5-7ms, ~15x.
+
+Four and six days later #532 A and #576 added `exclude_own_authored/3` to the
+same aggregate. The predicate reads `lower(sender)` and
+`json_extract(meta, '$.new_nick')`, neither of which was in any index on
+`messages`, so SQLite went back to one table row-fetch per post-cursor row.
+Neither entry mentions an index, a covering property or an EXPLAIN.
+
+**The finding that outlived the fix is the pin.** The 2026-08-15 review read
+this as "no covering pin exists". One did — three of them, asserting
+`USING COVERING INDEX` — and they were green throughout. They EXPLAINed a
+LOCAL REBUILD of the aggregate that passed `own_nick: nil`, and at nil
+`exclude_own_authored/3` is the identity clause, so the predicate that caused
+the regression never entered the pinned SQL. Measured on one 650k-row corpus,
+same channel, same cursor: the pinned shape reads 1,676 page fetches and says
+COVERING; the production shape reads 175,576 and does not. A pin on a branch
+production does not take is worse than no pin, because it reports safety. The
+cure is not a second pin beside it: the plan is now read off the SQL the
+production function actually emits, captured from `[:grappa, :repo, :query]`
+telemetry, so no second copy of the query exists to drift. Rule for any future
+plan pin: **EXPLAIN what production emitted, never what the test can rebuild.**
+
+**Why widening, and not seek-and-subtract.** Both shapes were measured on the
+same corpus before either was written. Appending the two expressions to the
+four existing indexes restores COVERING (175,576 → 1,886 page fetches, 93x,
+identical results) at **no measurable INSERT cost** — widening an entry adds no
+b-tree. Counting own-authored rows as a second seek and subtracting costs
+**+37% on INSERT**, because it adds four new b-trees to the highest-write-rate
+table in the codebase; and it is not a subtraction but inclusion-exclusion
+across three seeks, since a case-only self-rename matches both arms of the
+disjunction. Dearer and more fragile, for a read that is already served.
+Widening alone with `lower(sender)` was also measured: it does NOT restore
+COVERING. Both expressions are load-bearing.
+
+The four dead `dm_with` indexes (P-S4) are dropped in the same PR rather than
+deferred, because that is what makes the write side a net win: the pair is 25%
+faster on INSERT and 22% smaller in index bytes than main is today, where the
+widening alone is merely neutral. An 11-shape plan sweep confirms nothing else
+regresses, and that the four dropped indexes served no production read — the
+only thing referencing them was a pair of tests EXPLAINing the
+`(channel = ? OR dm_with = ?)` disjunction #393 deleted from production, the
+same defect class as the blind pin.
+
+**Not established.** There is no prod copy on this lane; the corpus is
+synthetic and its distributions — notably the ~4% own-`sender` share — are
+chosen, not observed. Page fetches are reported instead of milliseconds
+precisely because they are deterministic and host-independent; the wall-clock
+figures are warm macOS, not the m42 jail, and cold-cache timing is unmeasured.
+Local `CREATE INDEX` is 0.51-0.66s per index at 650k rows, but #393's own prod
+record of 1.6-2.6s per index at 654k rows is the better predictor for the COLD
+window. And the open question from #1353 stays open: *why* a pool connection
+that did not serve the `CREATE INDEX` plans differently — schema cache or
+transaction visibility — is still unmeasured, and `pool_size: 1` remains a
+remedy whose mechanism nobody has characterised.
--- a/priv/repo/migrations/20260816013504_add_own_authored_exprs_to_messages_covering_indexes.exs
+++ b/priv/repo/migrations/20260816013504_add_own_authored_exprs_to_messages_covering_indexes.exs
@@ -75,11 +75,28 @@ defmodule Grappa.Repo.Migrations.AddOwnAuthoredExprsToMessagesCoveringIndexes do
 
   ## Deploy
 
-  New migration file — Preflight Class 5 forces **COLD**. The four
-  `CREATE INDEX` builds took 2.6s in total over 650k rows locally (per-index
-  split in the PR body); they share one migration transaction, so schedule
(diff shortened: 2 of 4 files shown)


vjt:

Issue #1374, opened 2026-08-16
persistence: bring the write path under one busy-retry + immediate-transaction discipline (review bucket C)


vjt:
Bucket **C busy-retry** from the 2026-08-15 codebase review — four findings that share one root: writes that
reach `Repo` without the retry/transaction discipline the codebase already documents for itself. Full text with
file:line in `docs/reviews/codebase/2026-08-15-codebase-review.md` (P-S2, P-S3, P-S7, P-S8).

The findings, in the order they should be taken:

**P-S2 (HIGH) — the nick-rename migration chain runs in `Session.Server` with no `BusyRetry` and no transaction.**
`apply_effects/2` on `{:peer_nick_renamed, old, new}` runs `Scrollback.rename_dm_peer/4`, `ReadCursor.rename_dm_peer/4`,
`QueryWindows.rename/4` and `UserSettings.rename_muted_target/4` inline in the supervised process, each calling `Repo`
bare, each bound with a strict `{:ok, _} =` / `:ok =`. A transient `SQLITE_BUSY` raises and takes the session down
mid-migration; a crash between two of the steps strands the history under the new nick with a cursor keyed to the old
one. The trigger — a netsplit rejoin renaming many peers at once — is exactly the correlated write burst.

**P-S3 (HIGH) — seven write transactions use deferred `Repo.transaction/1` where `Repo.immediate_transaction/1` is
mandated** (`accounts.ex`, `accounts/webauthn.ex`, `accounts/totp.ex`). The sharpest is `TOTP.verify/3`: a read
followed by a conditional write inside one deferred transaction — the #524 shape — and the only one with no
`BusyRetry` at all, so a busy surfaces as a 500 at the 2FA login door instead of the 503 #518 promises. The other six
happen to write first today; that is statement ordering, not a property.

**P-S7 (MEDIUM) — `BusyRetry` nested inside an already-open `immediate_transaction`** (`visitors.ex` →
`user_settings.ex`), which `accounts.ex`'s own family contract names as the thing not to do. Plus a shape trap in the
error path: an atom reaches `Repo.rollback/1` through a binding named `changeset`.

**P-S8 (MEDIUM) — five writers missed by #523 B1 / #590 B2**: two session-driven credential persisters that fire on
every self-JOIN/PART/KICK and every away transition, and two web-reachable scrollback deletes that 500 today where
#523 B1's stated scope says they should degrade to 503.

Notes for whoever takes it:

[...]

Issue #524, opened 2026-07-28
db: Ecto deferred transactions hit an immediate SQLITE_BUSY that busy_timeout does not cover


vjt:
## Summary

A `Repo.transaction` whose first statement is a write (e.g.
`Grappa.Themes.do_copy/3`) is opened by ecto_sqlite3 as a **DEFERRED**
transaction: it begins as a reader and upgrades to a writer on the first write
statement. When another connection already holds the file-level write lock,
SQLite returns **SQLITE_BUSY immediately** on that read→write upgrade — the
configured `busy_timeout` (30s) does **not** apply to the upgrade inside a
deferred transaction, so the caller does not wait it out; it fails at once.

## Evidence

Same e2e failure that motivates the busy-retry issue (see linked). `POST
/themes/:id/copy` raised `Exqlite.Error: Database busy` after **724ms**, not
after the configured `busy_timeout: 30_000` (30s):

```
[chromium] › issue299-theme-cards.spec.ts:157 ... (724ms)
** (Exqlite.Error) Database busy
    INSERT INTO "themes" (...) RETURNING "id"
    lib/grappa/themes.ex:258: Grappa.Themes.do_copy/3
    lib/grappa/repo.ex:14: Grappa.Repo.transaction/2
```

A write that had genuinely waited out the busy_timeout would take ~30s and
*then* error. **724ms ≪ 30s** is the signature of an immediate BUSY on the
deferred→write upgrade, not an exhausted wait — i.e. `busy_timeout` never got a
chance to help.

Config (`config/dev.exs`, `config/runtime.exs`): `busy_timeout: 30_000`,
`journal_mode: :wal`, `pool_size: 5` (dev) / `10` (prod). Under WAL with
`pool_size > 1`, concurrent writers live on separate connections; a deferred
transaction upgrading to a write is the collision point that bypasses the
timeout.

## Proposed direction

Open write transactions as **`BEGIN IMMEDIATE`** (acquire the write lock up
front) so `busy_timeout` governs the wait and the transaction blocks-then-
proceeds instead of failing immediately. ecto_sqlite3 exposes a transaction
mode for this; whether to set it globally at the adapter level or per write
transaction is to be confirmed during implementation. Pairs with the busy-retry
issue below and is the same family as #518.

## Related
- #518 — DB-unavailable write path surfaces a 500 raise instead of a clean 503 (same family).
- #523 — DB writes outside Scrollback lack busy-retry (raise a 500 instead of retrying).

---
[...]

PR #1378, opened 2026-08-16
persistence(#1374): one busy-retry + immediate-transaction discipline for the write path

vjt:
Four findings from the 2026-08-15 codebase review (P-S2, P-S3, P-S7, P-S8),
one commit per finding plus their reds. They share one durable rule, and it is
the part worth keeping: **the retry goes OUTSIDE the transaction, the
transaction goes INSIDE it, and no retry loop ever opens inside an already-open
transaction.**

## What landed

**P-S2 — `Grappa.NickMigration` owns the nick-rename set.** The chain ran bare
against `Repo`, step by step, each bound `{:ok, _} =` or `:ok =`, inside the
supervised session. The new module is a TOP-LEVEL boundary rather than a
supervised child (the `Grappa.SpawnOrchestrator` shape), calls only public
context APIs, and lets `Session` span four contexts without acquiring the
`Grappa.Repo` dep its boundary deliberately lacks. The `query_windows_list`
broadcast stays in `Session.Server` after the `:ok` return — it is the #373
"rename fully applied" barrier and a retried attempt must not re-fire it.
Budget exhaustion is the #590 background-DROP posture.

**P-S3 — seven `Repo.transaction/1` call sites converted, and gated.** Plus a
retry at the 2FA door, and the two `@doc` paragraphs that quoted the dead
spelling.

**P-S7 — a non-retrying `!` seam for in-transaction callers.** `get_or_init!/1`
and `put_upload_ttl_seconds!/2`, bodies parameterised rather than duplicated.
The atom reaching `Repo.rollback/1` was bound to a variable called `changeset`,
which is what it is not; it is `reason` now.

**P-S8 — seven writers brought under a retry budget.** Five in `credentials.ex`,
two in `scrollback.ex`.

## Where the review was wrong, with the measurement

**The 2FA door does NOT answer 500.** The review says a transient busy surfaces
there as a 500 where #518 promises 503. Measured at HEAD, it already answered
**503**: the retry producing it sits DOWNSTREAM of `TOTP.verify/3`. The real
defect was that `totp_last_used_step` advanced anyway — 59561585 to 59561586 in
the red — so the user paid a redeemable code for a degraded response. The
boundary retry fixes that, but the status code was never the symptom.

**P-S2's killing site is not one the review names.** The red's crash is at
`server.ex:6161`, `UserSettings.rename_muted_target/4` — the FIRST step of the
arm. That call already carried a retry loop (23 attempts over the full 300ms
budget in the log); what killed the session is that its
`{:error, :db_unavailable}` terminal was bound `{:ok, _} =`. Same defect as the
sites the review does name, different cause: not a missing retry, an exhausted
one meeting a strict match. The bare-`Repo` steps downstream were never reached.
Because the mute had to move inside the transaction, `rename_muted_target!/4`
and `persist!/1` land in the P-S2 commit — so the mechanism P-S7 asks for is a
prerequisite of P-S2, and the P-S7 commit consumes the family instead of
inventing it.

**P-S8's count is wrong three ways at once.** The prose says five, four are
named, the file:line list points at six. Counted: seven.
`update_visitor_last_joined_channels/3` is not in the review at all — it is
session-driven exactly like its user-side twin, injected as
`last_joined_persister` from `session_plan.ex:237`, and retrying one subject
kind but not the other is the split into two patterns CLAUDE.md warns about.
`update_away/4` has no visitor twin; that was checked, not assumed.

**P-S7's nesting is narrower than implied.** It fires only on incognito
provisioning — `seed_incognito_upload_ttl/2` has a `false` clause — not on every
anonymous login.

**`totp.ex:48` was converted but #524 was never open there.** Its single
production caller already runs inside the outer transaction of
`confirm_totp_enrollment/5`, so the change is inert. Converted for the contract,
not for a bug.

## Why the transaction-mode gate is STATIC

It cannot be a runtime assertion. exqlite emits `BEGIN IMMEDIATE` only when
`transaction_status == :idle` (`deps/exqlite/lib/exqlite/connection.ex:297-316`);
nested, every mode collapses to `SAVEPOINT exqlite_savepoint`. Under the SQL
Sandbox every test already runs inside a transaction, so `immediate_transaction/1`
is always a savepoint there and no ExUnit oracle distinguishes it from
`transaction/1` — the "put the deferred spelling back" mutant is unkillable at
runtime. Undecidable at runtime, decidable at compile time, so the gate walks
the AST of `lib/**/*.ex`. It is an arch test rather than a Credo check because
the repo already has that shape (`error_tokens_drift_test.exs`), while a custom
check would want `requires:` in `.credo.exs` and a module outside the app tree.
Its blind spot is in its own moduledoc: a renaming alias escapes it, and a
genuinely read-only transaction must say so via a `deferred_transaction/1`
sibling, deliberately not written because it would be dead code today.

Bought by mutation: restoring the deferred spelling at one site reddens the gate
naming exactly `lib/grappa/accounts/totp.ex:56`.

## Why atomicity is UNKILLABLE here, and why the transaction stays anyway

This is a reason, not an omission.

The mutant that removes `immediate_transaction/1` from `NickMigration.migrate/1`
**survives** — measured, and deliberately left surviving rather than masked by an
assertion that passes for another reason. Two facts explain it.

First, the fault seam has a reach narrower than it looks. `BusyRetry`'s injected
fault fires at exactly one place: `maybe_inject_fault/0` at the top of a
`BusyRetry.run/1` attempt, before `op.()` — stated in the source and confirmed by
mutation, since removing the wrapper from `migrate/1` did not make the write fail,
it made the write LAND. Counted in the post-fix code: **zero** reachable
`BusyRetry.run` calls inside the transaction body across all five steps. So there
is exactly one check per migration and it fires before the transaction opens. The
irony is that the P-S7 fix is what removed the only thing that could have fired
mid-chain.

Second — and this is the part that closes the question — **no data-driven
intermediate failure exists to arrange either.** Every step is total on real data:

* the two scrollback renames are `update_all` returning `{:ok, n}`, with no
  constraint on the updated columns to violate;
* `ReadCursor.rename_dm_peer/4` handles a fold-collision by dropping the stale
  cursor (`read_cursor.ex:675`), and wraps the update in a `try`;
* `QueryWindows.rename/4` merges on collision (`query_windows.ex:296`) and even
  rescues the unique-index race, degrading to the same merge path (`:308`);
* the mute's `Repo.rollback` (`nick_migration.ex:183`) is reachable only on a
  `Settings` changeset error, and that changeset validates `data` presence, the
  subject XOR, two `assoc_constraint`s and `unique_constraint(:user_id)` — none
  of which a rename of an EXISTING row can violate.

`own_renamed/5` is the arm where the route should have worked, because there the
mute is LAST, after four stores. It fails for the same reason: the only step that
can roll back is the only step with nothing after it to roll back.

So the single failure the chain admits is the enclosing retry's budget running
out, which fires before step one. Nothing can fail in the middle, so no rollback
can be observed, so the mutant cannot die. **The transaction stays regardless**,
and the moduledoc now says why in prospective terms: CLAUDE.md requires every NEW
nick-keyed store to join this set, a new store is under no obligation to be
total, and on the day one arrives with a validation or a constraint a
half-applied identity becomes reachable. The test that will finally buy this
transaction is the one that adds a fallible step.

* **How long the P-S2 transaction holds on a large DM history.** It now wraps
  `update_all` over `messages` + `read_cursors` + `query_windows`. I have no
  honest way to measure that here, so it is declared rather than estimated.
* **That a real SQLITE_BUSY raises through the `!` variants to the caller's
  transaction.** The pool_size:1 Sandbox cannot produce one, and the injection
  seam reaches only code already inside a retry loop — precisely what the `!`
  variants lack. Argued from `Repo.update/1`'s behaviour, not measured.
* **That a raise escapes the P-S8 writers into the GenServer.** Same limit: an
  unretried writer has no fault point, so both P-S8 reds fail at HEAD by
  SUCCEEDING rather than by crashing. What they measure is the property the fix
  installs — each writer now runs under a retry budget and degrades on its own
  terms — and a mutant removing the wrapper flips each back.
* **The rest of `credentials.ex`'s write surface.** #523 B1 deferred
  `update_nick` on purpose and this does not widen past the seven. That exposure
  is declared here rather than quietly fixed.

## Test plan

- [x] `scripts/check.sh` green: 8 doctests, 60 properties, 6344 tests, 0
      failures; Dialyzer 0 errors; Credo strict clean; Sobelow no
      vulnerabilities; Doctor passed; bats green
- [x] every assertion bought by a named 1:1 mutation, each mutant verified
      applied with `grep -q` before its red was believed
- [x] one mutant deliberately left surviving, with the reason above
- [ ] integration suite — not run on this lane

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -170,7 +170,8 @@ Key invariants — break only with deliberate cause + DESIGN_NOTES entry:
   old→new (broadcasting mid-migration raced a follow-on
   `Scrollback.fetch`; #373 rename-order fix). A NEW nick-keyed store MUST
   be added to this migration set or a rename silently strands its
-  old-nick rows. Boundary limit: IRC delivers a NICK only to
+  old-nick rows — the set and its one retried transaction live in
+  `Grappa.NickMigration` (#1374), not in `Session.Server`. Boundary limit: IRC delivers a NICK only to
   channel-sharing peers, so a query with someone in no shared channel
   cannot follow. **Our OWN nick keys exactly one window — the SELF
   window (`/msg <ownnick>`, GH #948) — and it has its own set** on
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -43368,3 +43368,65 @@ when a reset actually happens. With `pool_size: 10` under continuous reads
 that is a real possibility, and it is the most likely reason prod's WAL was at
 168 MiB rather than the ~64 MiB the threshold alone predicts. The change lowers
 the pressure and bounds the recovery; it does not promise a ceiling.
+<!-- entry #1374 -->
+
+---
+
+## 2026-08-16 — #1374: one busy-retry discipline for the write path, and what the fault seam cannot see
+
+Four findings from the 2026-08-15 review (P-S2, P-S3, P-S7, P-S8) share one
+rule, and it is the durable part of this change: **the retry goes OUTSIDE the
+transaction, the transaction goes INSIDE it, and no retry loop ever opens
+inside an already-open transaction.** `Repo.BusyRetry.run(fn ->
+Repo.immediate_transaction(fn -> … end) end)` is the only correct nesting.
+Retrying a step is not retrying a transaction; a nested retry sleeps while
+holding the enclosing transaction's connection and extends the very contention
+it waits on.
+
+Two consequences fall out. Code reached only from inside an open transaction
+needs a NON-retrying seam that RAISES, because the raise is what aborts the
+transaction and hands the retry to the enclosing engine — that is the `!`
+family (`persist!/1`, `get_or_init!/1`, `put_upload_ttl_seconds!/2`,
+`rename_muted_target!/4`). And a set of writes that must land together needs
+one owner: `Grappa.NickMigration` is now the home of the nick-rename set, a
+TOP-LEVEL boundary rather than a supervised child, so `Session` can span four
+contexts without acquiring a `Grappa.Repo` dep it deliberately lacks. The
+`query_windows_list` broadcast stays in `Session.Server` after the `:ok`
+return: it is the #373 "rename fully applied" barrier and a retried attempt
+must not re-fire it.
+
+**The write-transaction spelling is now gated statically, because it cannot be
+gated at runtime.** exqlite emits `BEGIN IMMEDIATE` only when
+`transaction_status == :idle` (`connection.ex:297-316`); nested, every mode
+collapses to `SAVEPOINT exqlite_savepoint`. Under the SQL Sandbox every test
+already runs inside a transaction, so `immediate_transaction/1` is always a
+savepoint there and no ExUnit oracle can distinguish it from `transaction/1` —
+the "put the deferred spelling back" mutant is unkillable at runtime. The
+defect is undecidable at runtime and decidable at compile time, so
+`test/grappa/repo/transaction_mode_gate_test.exs` walks the AST of `lib/**/*.ex`
+instead. Its blind spot is declared in its own moduledoc: a renaming alias
+escapes it, and a genuinely read-only transaction must say so by calling a
+`deferred_transaction/1` sibling, deliberately not written because it would be
+dead code today.
+
+**The fault-injection seam has a reach, and it is narrower than it looks.**
+`BusyRetry`'s injected fault fires only at the top of a `BusyRetry.run`
+attempt. Established by mutation, not assumed: removing the wrapper from
+`NickMigration.migrate/1` did not make the write fail, it made the write
+LAND. Three things follow, and they bound what any future test here can claim.
+A code path with no retry loop has no fault point at all, so the seam cannot
+show a raise escaping an unretried writer — that claim stays argued from
+`Repo.update/1`'s behaviour. A saturating fault fires before the first step of
+a transaction, so removing `immediate_transaction` from a retried chain leaves
+every test green: atomicity is NOT measurable in-process, and the mutant that
+should have bought it survives. And a fault cannot be aimed mid-chain, because
+a first attempt that succeeds means there is no second attempt to arm.
+
+Two review claims did not survive measurement, recorded so nobody re-derives
+them. The 2FA door does NOT answer 500 on a sustained busy; it already answers
+503, from a retry DOWNSTREAM of `TOTP.verify/3`. Its real defect was that
+`totp_last_used_step` advanced anyway, so the user paid a redeemable code for a
+degraded response. And P-S8's count was wrong in three different ways at once —
+prose said five, four were named, the file:line list held six; seven are fixed,
+including `update_visitor_last_joined_channels/3`, which the review never
+mentions and which is session-driven exactly like its user-side twin.
--- a/lib/grappa/accounts.ex
+++ b/lib/grappa/accounts.ex
@@ -775,7 +775,7 @@ defmodule Grappa.Accounts do
     * **in-transaction** — the `!` variants
       (`revoke_other_sessions_for_user!/2` and the private
       `revoke_sessions_for_user!/1`). Reached only from inside a
-      `Repo.BusyRetry.run(fn -> Repo.transaction(…) end)`, where the
+      `Repo.BusyRetry.run(fn -> Repo.immediate_transaction(…) end)`, where the
       enclosing engine retries the WHOLE transaction. They deliberately do
       NOT retry: a nested retry would sleep while holding the open
       transaction's connection, extending the very contention it waits on,
@@ -944,7 +944,7 @@ defmodule Grappa.Accounts do
 
   In-transaction only — every caller (TOTP enrolment/disable, the passkey
   mode transaction) already runs inside a
-  `Repo.BusyRetry.run(fn -> Repo.transaction(…) end)`, so this RAISES on
+  `Repo.BusyRetry.run(fn -> Repo.immediate_transaction(…) end)`, so this RAISES on
   SQLITE_BUSY rather than retrying: the raise is what aborts the
   transaction and hands the retry to the enclosing engine. See the family
   contract on `revoke_session/1`. A caller with no such enclosure wants a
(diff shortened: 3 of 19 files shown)


vjt:
Held out of the union merge for now: this branch's `integration` run came back with two failures.

- `issue458-presence-page-yield` expected 3 `nick_change` scrollback rows and saw 2, and the call log shows the locator resolving to 2 elements across twelve polls over 10s — that is not a race about to settle. This branch is the nick-rename refactor, so it is inside the blast radius and gets treated as a real regression until measured otherwise.
- `issue498-badge-follows-live-nick` failed in `provisionSpecSubject` with `504 autojoin_timeout` on `#bofh` — a fixture-layer failure, unrelated to the assertions.

main was green on the same suite at `92df169a`, and a sibling PR on the same base passed, so the base is not the explanation. Next step is a measurement, not an argument: run that spec on this branch and on main and compare. No assert will be weakened and no timeout bumped to make it pass.

vjt:
## Diagnosis of the `integration` red — reproduced, attributed, and bought with a one-line mutant

Four runs of 5 repeats each, same host, identical command
(`scripts/integration.sh --project chromium --grep "#458|#498" --repeat-each 5`).

| side | tree | UTC | verdict | per-repeat span of the 3 `nick_change` rows | `SQLite pool saturated` | `scrollback row dropped` | `fault=busy_locked` |
|---|---|---|---|---|---|---|---|
| **A** | branch `6db45a1e` | 04:16 | **1 failed** / 9 | 5, 3, **34s**, 3, 4 | **4** | **2** | **2** |
| **B** | main `92df169a` | 07:14 | 10 passed | 4, 3, 4, 4, 4 | 0 | 0 | 0 |
| **A′** | branch + 1-line mutant | 07:30 | 10 passed | 4, 3, 3, 4, 4 | 0 | 0 | 0 |
| **A2** | branch `6db45a1e` | 07:45 | **1 failed** / 9 | 4, 3, **32s**, 2, 4 | **4** | **2** | **2** |

A2 exists because A had run three hours before B. Re-run **adjacent in time** to B and A′, the branch fails
again — same spec, same line 68, same position (`repeat2`), same signature counts. 2/2 red on the branch,
0/2 on the green sides. Log sizes are comparable (26.2k–29.0k lines), so the signature differential is not
a volume artefact.

### The mechanism, named by its fault string

```
fault=busy_locked … db write unavailable: SQLite pool saturated for the full 1500ms retry budget (1 attempts)
```

Someone holds the SQLite write lock for the entire retry budget. The mutant is one line in
`NickMigration.migrate/1`:

```elixir
- BusyRetry.run(fn -> Repo.immediate_transaction(fun) end)
+ BusyRetry.run(fn -> {:ok, fun.()} end)
```

With it, the branch reproduces main's profile exactly: spans 3–4s, zero saturation. **The defect is not the
transaction — it is opening one when there is nothing to migrate.** `peer_renamed/5` takes a `BEGIN
IMMEDIATE` (RESERVED write lock) on *every* peer rename, including the `:noop` arm where the peer has no
query window and no row moves. That is this spec's case. A NICK in `#bofh` fans out to 4 grappa sessions, so
one rename becomes 4 transactions racing for the single writer.

### Why CI sees 2 rows and the local repro sees 0 — same event, different severity

Not a lost row: **the session is stuck on the DB.** In the failing repeat the assertion window is
`04:17:07.7 → 04:17:17.7`; the three inserts land at `04:17:36` and `04:18:10`, 18s and 52s after it closed.
CI's "2 of 3" is the same stall with one row additionally dropped — `Scrollback.with_pool_retry/1` has the
documented #336 contract to discard the row rather than crash the session, logging
`scrollback row dropped: SQLite pool saturated — session continues`.

### Hypotheses refuted with the measurement

- **A leaked `hide_presence` pref.** The trace's boot `GET /me/settings/display-prefs` returns
  `presence_filter: {}` — unset, so presence is shown. Refuted.
- **ircd NICK throttling.** `IrcPeer.changeNick` awaits the upstream echo, so a throttled NICK would fail at
  `changeNick` with a NICK timeout, not at `toHaveCount`. Refuted.
- **The row was never produced.** `event_router.ex` is not in this diff at all, and `persist_effects`
  precede `:peer_nick_renamed` in the effect list. The rows are produced; they are late.

### `issue498-badge-follows-live-nick` — did NOT reproduce, 20/20

Five repeats on each of the four sides, all green, and zero `autojoin_timeout` in any container log. It is
also not #498's to own: `provisionSpecSubject` is the `_specSubject` fixture with `auto: true`, so the 504 on
`#bofh` comes from the **shared** per-spec provisioning — #498 merely held the ticket. Signature if it
returns: `subject_session.ex:46`, a 5s deadline after `:session_ready`.

### What this does not establish

The mutant is **diagnostic only** — it removes the atomicity the transaction exists for, and is not the fix.
The narrow shape (open the transaction only when there is work) is the likely fix and was not measured here.
`repeat2` twice is "reproducible in position", not proven deterministic; why the third repeat specifically
(WAL checkpoint?) is unbought.

vjt:
Closing by hand, superseded by content rather than by merge.

This branch's work now lives on the union branch `w2-1374-1375-union` (PR #1415), which
re-lands `settings(#1375)` (`83d90cb1`, cherry-picked verbatim so its message
and author date survive) and replays all fifteen commits from here on top of
it, unchanged.

Verified rather than asserted: the union tree is byte-identical to this
branch's head — `git diff` between them is empty. The union carries two
commits this branch does not, both DESIGN_NOTES:

  * one superseding #1375's "Accepted cost" section in place, since the
    cherry-pick puts that paragraph back on main and it declines the very probe
    landing beside it;
  * one dropping `persist!/1` from the `!` family roster (the rebase deleted it
    in favour of #1375's `write_data!/2`) and recording what the local
    reproduction attempt did NOT establish.

Why a union and not a rebase: main reverted #1375 in #1414 while this branch
was open, and this branch does not merely sit on top of #1375 — its rebase
resolution ADOPTED #1375's write path, deleting `persist!/1` so the module
would keep exactly one `Repo.update`. Merging this as-is would have re-landed
#1375 silently, inside a commit that says nothing about it. The union makes
that explicit and reviewable in one sequence: what #1375 did, why it was taken
out, what puts it back, and what changed in the cost it had accepted.

No code is lost and nothing is squashed.

Issue #1377, opened 2026-08-16
deploy/CI: the hot path never establishes MIX_ENV, and two gates assert shape where they should assert equality (review bucket I)


vjt:
Bucket **I infra-simplify** from the 2026-08-15 codebase review: D-S3 (HIGH), D-S4, D-S5. Full text with
file:line in `docs/reviews/codebase/2026-08-15-codebase-review.md`.

**D-S3 (HIGH) — `scripts/deploy.sh` seeds against the DEV database on a HOT deploy.** `substrate_build`
short-circuits on hot BEFORE the `.env` check and the `MIX_ENV=${MIX_ENV:-prod}; export MIX_ENV` line, so
the later `substrate_seed` runs `mix grappa.seed_themes` with no shell `MIX_ENV`; compose falls back to
`.env`, and `.env.example` ships `MIX_ENV=dev` uncommented. The result is asymmetric and silent: COLD
deploys seed prod, HOT deploys seed dev — and hot is the documented normal case. The die message names
three keys to fill and never mentions `MIX_ENV`, so an operator following it verbatim lands exactly here.
`infra/docker/deploy.sh` is immune because its installer writes `set_env MIX_ENV prod`; this script has no
installer. Fix shape: hoist the `.env` check and the `MIX_ENV` export to the top, before `deploy_main`, and
annotate the `.env.example` line — the drift test forces the key to be PRESENT, not to be correct.

**D-S4 — the POSIX gate covers 5 of ~20 `sh`-dialect files**, and misses two of the three shared libs the
operations doc calls POSIX, both declaring `# shellcheck shell=sh` on line 1. Fix shape: derive the set the
way `scripts/shellcheck.sh` already derives its own (shebang OR line-1 declaration) and feed that to
`dash -n`, instead of a hand-maintained list.

**D-S5 — the `oven/bun` digest is transcribed four times with no equality gate.** The existing bats asserts
each reference carries a well-formed `@sha256:`, never that the four agree; the invariant lives in a comment,
which is the shape #441 rejected. Two of the four bind the same host cache, so a bump touching three of them
surfaces only as a lockfile/cache disagreement in CI. Fix shape: extract every `oven/bun:1@sha256:` token
across tracked build files and assert one distinct value.

Notes for whoever takes it:

- The review is a document, not a measurement. Verify each claim before changing anything — in particular
  D-S3's chain (short-circuit → no export → compose falls back to `.env` → dev DB) should be demonstrated,
[...]

PR #1380, opened 2026-08-16
infra(#1377): fix the hot-path MIX_ENV asymmetry, derive the POSIX gate, refuse D-S5

vjt:
Bucket I (infra-simplify) of the 2026-08-15 codebase review: D-S3 (HIGH), D-S4, D-S5.
Two were real. D-S5 was already closed the day the review was written, and is refused
here with evidence rather than implemented.

## D-S3 — `scripts/deploy.sh` seeded the DEV database on a HOT deploy

The chain was demonstrated before anything was changed, leg by leg:

- `substrate_build` opens with `[ "$MODE" = cold ] || return 0`, and owned both the
  `.env` check and `MIX_ENV=${MIX_ENV:-prod}; export MIX_ENV` — so on hot neither ran.
- `deploy_common.sh` reaches `substrate_seed` on the hot path
  (`_deploy_hot` → `_deploy_seed`), a fresh `docker compose … run --rm grappa mix
  grappa.seed_themes`.
- With no shell `MIX_ENV`, compose interpolates from `.env`. Measured against the real
  files with the real compose.yaml:

  ```
  $ docker compose -f compose.yaml --env-file .env.example --profile prod config | grep -E 'MIX_ENV|DATABASE_PATH'
        DATABASE_PATH: /app/runtime/grappa_dev.db
        MIX_ENV: dev
  $ MIX_ENV=prod docker compose … config | grep -E 'MIX_ENV|DATABASE_PATH'
        DATABASE_PATH: /app/runtime/grappa_prod.db
        MIX_ENV: prod
  ```

So cold seeded prod, hot seeded dev, on the same box, silently — and hot is the
documented normal case.

**Fix.** Both preconditions move into `establish_deploy_env`, called from the top of
`substrate_pull`. Not from file scope: the library parses flags inside `deploy_main`,
so a check above that turns an unknown flag into a missing-`.env` error. The top of
the first hook is after the parse and before the first side effect, which is all the
placement has to satisfy.

**Red first.** The bats `docker` stub grew a second log carrying the `MIX_ENV` each
compose invocation inherited. Before the fix the new hot case reports
`hot seed ran with MIX_ENV=<unset>`; the cold twin was already green, which is the
asymmetry stated as a test.

**Operator-visible change:** a box with no `.env` now fails on hot exactly as it always
did on cold. Three sibling bats suites drove hot deploys against a fixture with no
`.env`; they now provision one, because their subject is the reload contract and the
worktree guard, not the absence of `.env`.

`.env.example` keeps `MIX_ENV=dev` (the registry drift test wants the key present, and
compose defaults to `dev` anyway, so commenting it out would de-trap nothing) and names
the residual hazard instead: a hand-driven `docker compose --profile prod …` inherits
whatever is written there, since only `deploy.sh` exports over it. Verified that the
declared-key set of `.env.example` is byte-identical in membership before and after
(33 keys, no delta).

## D-S4 — the POSIX gate covered 5 of 28

Counted, not estimated: 28 files under `bin/`, `infra/`, `scripts/` declare the sh
dialect on line 1; the `dash -n` step named 5. Missing: `infra/lib/beam_wait.sh` and
`infra/lib/cic_dist.sh` (two of the three libs `docs/OPERATIONS.md` calls strict POSIX,
same directory and same `# shellcheck shell=sh` line 1 as the one it did name), all
twelve `infra/freebsd/jail_*.sh` rails, `infra/packaging/version.sh`, `bin/start.sh`,
the packaging scriptlets, and the extensionless `infra/freebsd/rc.d/grappa`.

`scripts/posix-parse.sh` derives the set from each file's own line-1 declaration —
the same answer the file already gives shellcheck, so there is no second copy of the
dialect to drift. Same shape as `scripts/shellcheck.sh`: `--list`, an empty-set refusal,
and a hard exit when the tool is missing rather than a silent skip. All 28 parse clean
today, so the widening lands green.

**The gate is bought by mutation.** With `arr=(a b c)` appended to
`infra/lib/cic_dist.sh` — a file the old list did not name — it exits 1 with
`infra/lib/cic_dist.sh: 91: Syntax error: "(" unexpected`. The bats suite pins the same
thing in a sandbox, plus the membership rule, the exclusions, the empty-set refusal, and
that `ci.yml` runs the script and names no `dash -n` paths (that last assertion matches
five lines of the pre-change `ci.yml`, so it is not vacuous).

**Measured limit, stated in the script and the docs:** `dash -n` checks GRAMMAR.
`arr=(a b)` is a syntax error to it; `[[ 1 == 1 ]]` and `local x` are not — they parse
as ordinary commands and fail only when they run. shellcheck's `sh` dialect (SC3xxx)
covers that half, over the same files. Neither gate subsumes the other, and the review's
note that the `dash -n` list is "largely redundant" holds only for the half shellcheck
sees.

## D-S5 — refused

The finding says the four `oven/bun:1` transcriptions have no equality gate. They do:
`test/infra/base_image_digest_pin_test.bats:69` grew one in `1fb0bffd`,
`test(#1343): hold every copy of an image:tag to one digest (D-S5)`, landed 2026-08-15
— the same day the review was written. Verified by mutation, not by reading: a zeroed
digest in `scripts/bun.sh` fails the case with `DIVERGENT digests for the same
image:tag`.

The mutation did surface a smaller gap the finding did not raise. The failure printed
the family and then dumped every pinned token, and the four `oven/bun:1` tokens are
textually identical — so it proved a divergence and left the reader to grep for which of
four surfaces moved. It now prints the matching `path:line:content` refs for the
conflicting family; the same mutation on `.github/workflows/ci.yml:348` now names all
four, with the odd one out on its own line.

## Gates

- `scripts/bats.sh` (full set): 492 ok, rc=0
- `scripts/shellcheck.sh`: rc=0 (75 derived scripts)
- `scripts/posix-parse.sh`: rc=0 (28 derived scripts)
- `scripts/design-notes-gate.sh`: rc=0, `1 new entry heading(s), separator and marker present`
- Contribution `--numstat` pinned to a file across the rebase onto `origin/main`; both
  sides identical (additions unchanged, deletions zero on `docs/DESIGN_NOTES.md`).

- **Nothing ran on a real Docker substrate.** No container was started, no seed was run,
  no `grappa_prod.db` was written. The D-S3 chain rests on two separate measurements —
  `docker compose config` (client-side interpolation, no daemon) for the
  `.env` → `DATABASE_PATH` leg, and bats against a stubbed `docker` for the
  "the seed oneshot inherits `MIX_ENV=prod`" leg. I did not join them in one live run.
- **The jail and systemd substrates were not exercised.** The D-S4 widening asserts that
  28 files *parse* under dash; it says nothing about whether they *run* correctly there,
  and `dash -n` would not catch it if they did not (see the grammar-vs-runtime limit
  above).
- **No Elixir gate was run** — this branch has no COMPILE lane. The only Elixir test that
  reads a file I touched is `test/grappa/config/env_registry_drift_test.exs` (`.env.example`);
  I proved its `env_file_decls` key set is unchanged by re-implementing that regex over
  both revisions, which is an argument, not a run of the test.
- **The `.env.example` residual hazard is documented, not closed.** A hand-driven
  `docker compose --profile prod up` on a prod box still opens `grappa_dev.db` unless the
  operator edits that value. Closing it would mean changing compose's default, which
  breaks plain dev `docker compose up`; that trade was not mine to make here.
- **The bun-digest report change is only proven for a two-digest conflict** in one family.
  A three-way divergence would print all refs the same way, but I did not construct one.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -11,6 +11,14 @@ CONTAINER_GID=1000
 # Mix environment switch. `dev` for local hacking, `prod` for the
 # deploy.sh stack. Phoenix expects an atom without the leading colon
 # when read from env vars (just `prod`).
+#
+# NOTE: compose interpolates this into BOTH `MIX_ENV` and `DATABASE_PATH`
+# (`grappa_${MIX_ENV:-dev}.db`), so the value below picks which database
+# every container and oneshot opens. `scripts/deploy.sh` exports
+# `MIX_ENV=prod` for its own invocations and the shell wins over this
+# file (#1377 D-S3) — but a hand-driven `docker compose --profile prod
+# …` on a prod box inherits exactly what is written here. Set it to
+# `prod` on any box that is not a dev laptop.
 MIX_ENV=dev
 
 # Host port publish for the grappa container. #485 dropped the in-stack
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,24 +160,21 @@ jobs:
         # convention. Needs the fetch-depth: 0 above.
         run: scripts/design-notes-gate.sh
 
-      - name: POSIX-parse the shared deploy lib + its sh consumers (#503)
+      - name: POSIX-parse the sh-dialect scripts (#503, derived by #1377)
         # A DIFFERENT property from the lint above, with a different tool:
-        # the shared deploy lib (infra/lib/deploy_common.sh) must stay strict
-        # POSIX sh so dash can run it as the FreeBSD jail's /bin/sh. dash -n
-        # parses; shellcheck lints. A reintroduced bashism in the shared
-        # algorithm fails here instead of silently breaking the jail deploy —
+        # the shared deploy lib (infra/lib/deploy_common.sh) and its sh
+        # consumers must stay strict POSIX so dash can run them as the FreeBSD
+        # jail's /bin/sh. dash -n parses; shellcheck lints. A reintroduced
+        # bashism fails here instead of silently breaking the jail deploy —
         # the copy-paste-drift class #503 exists to kill.
         #
-        # Still a hand-written list, deliberately: #441 derived the SHELLCHECK
-        # set, and this one is not the same question — "which files must parse
-        # under dash" is a claim about where they run, and the honest
-        # derivation for it is a separate piece of work.
-        run: |
-          dash -n infra/lib/deploy_common.sh
-          dash -n infra/freebsd/deploy.sh
-          dash -n infra/docker/assert-abi-lockstep.sh
-          dash -n infra/docker/release-entrypoint.sh
-          dash -n infra/docker/get.sh
+        # The set is DERIVED, like the shellcheck one (#441). The hand list
+        # this replaced covered five of the twenty-eight files that declare
+        # the sh dialect, missing two of the three POSIX libs under
+        # infra/lib/ and all twelve jail rails; the derivation reads each
+        # file's own line-1 declaration instead of keeping a second copy of
+        # it. Covered by test/infra/posix_parse_gate_test.bats.
+        run: scripts/posix-parse.sh
 
       - name: Cloud provider-door drift-guard (#665)
         # Prove every provider door (infra/aws today, infra/terraform later)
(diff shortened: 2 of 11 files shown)


vjt:

Issue #1379, opened 2026-08-16
protocol: cicchetto never negotiates the version, and the published carrier topic for join/kick is wrong on all three surfaces (review bucket J)


vjt:
Bucket **J wire-pins**, the two HIGH findings, from the 2026-08-15 codebase review (X-S2, X-S3). Full text
with file:line in `docs/reviews/codebase/2026-08-15-codebase-review.md`.

**X-S2 — cicchetto never sends `client_proto` and never reads `/api/config`.** A grep for
`protocol_version|client_proto|min_protocol` across all of `cicchetto/` returns one hit, and it is a prose
comment. `socketEndpoint` builds the socket URL with no `?client_proto=`, so the server treats cic as
*current*; the join reply's `protocol_version` is received and ignored; nothing fetches `/api/config`, so
`min_protocol_version` and `push_content_encoding` are never observed. This matters more for cic than for a
generic client: it is a service-worker-cached PWA — precisely the client that can be arbitrarily stale
against a freshly deployed BEAM — and cic-only vs server-only deploys are normal here. The 426 refusal of
#447 exists to catch exactly that, and the reference client has opted out by omission.

Fix shape from the review: append `client_proto=<N>` from a cic-side constant; consume the join reply's
`protocol_version` (log/banner on mismatch); add a boot-time `GET /api/config` so `min_protocol_version` is
checked before the socket opens and `push_content_encoding` can gate the push UX.

**X-S3 — the documented carrier topic for `joined` / `join_failed` / `kicked` is wrong, in three places at
once.** `broadcast_window_state/2` broadcasts to the USER topic only; the per-channel topic sees those kinds
exclusively as the cold-subscribe snapshot pushed per-socket at join time. Three published statements say
otherwise: the CLAUDE.md window-state invariant (which lists all four kinds as per-channel, when only
`members_seeded` is), a comment inside the `:kicked` arm naming the per-channel topic two lines above a
user-topic call, and a client comment describing a "dual-broadcast" via a function that does not exist in
`lib/` at all. cic works anyway because both dispatchers feed the same idempotent setters — but a
third-party client that follows CLAUDE.md subscribes per-channel and never observes a live join or kick.

Fix shape: correct all three statements to the real model, and add the topic split to
`docs/CLIENT_PROTOCOL.md` §4 — it is a topic-selection fact a client author cannot derive.

Notes for whoever takes it:

- The review is a document, not a measurement. Establish both claims against the code first — X-S3 in
[...]

Issue #193, opened 2026-07-06
cicchetto: WS reconnect uses ws:// → hits :80 redirect → 301 → client stuck on splash after BEAM restart


vjt:
## Summary

After a backend (BEAM) restart, a subset of clients get stuck on the splash screen. Their Phoenix Channels WebSocket **reconnect** goes out over `ws://` (plaintext, port 80) instead of `wss://`. The port-80 vhost answers with `301 https://…`. A WebSocket handshake does **not** follow HTTP redirects, so the upgrade fails; the client retries in a backoff loop and never establishes the channel → splash never clears.

## Evidence

- The `:80 → 301 https` redirect vhost access log holds **34** `GET /socket/websocket…` → `301` entries (this location should never receive a WS handshake).
- They cluster tightly **right after each backend restart**: 2026-05-27 20:13, 07-03 01:07, 07-04 01:07, 07-06 01:06 UTC (the last = the COLD deploy at 03:05 CEST).
- **5 distinct real clients** affected (residential IPs, redacted), each retrying with exponential backoff (…:39, :40, :41, :43, :46, :52, :58, :04, :10).
- On those lines `remote_addr` = the grappa prod nginx upstream and `X-Forwarded-For` = the real client → the `ws://` reconnect **transits the grappa-side nginx**, it is not a direct browser hit.
- **Not fail2ban:** none of the affected IPs are banned — absent from pf, from all 9 jails (`http-400/404/ratelimit/url/recidive`), and from `fail2ban.log`. No 400/404/414 involved.

## Root cause

The socket endpoint is (re)constructed with a `ws://` scheme for these sessions instead of deriving `wss://` from the page origin. On a fresh device the first load can happen over `http://` before the HSTS pin exists, and/or a stale service-worker bundle pins a `ws://` endpoint; the Phoenix `Socket` then keeps that scheme across every reconnect. The BEAM restart is only the trigger — it forces the reconnect that exposes the wrong scheme.

## Fix (client — durable)

Force `wss://` unconditionally for the prod socket endpoint; never derive `ws://` on an https origin. Minimum: `scheme = location.protocol === "https:" ? "wss:" : "ws:"`. Ensure a stale service worker cannot pin an old `ws://` bundle across a deploy.

## Mitigation (server — no client change, cheap)

[...]

Issue #95, opened 2026-06-26
Phase 5: move bearer token off the WS query string


vjt:
The bearer currently rides `?token=...` on the WS upgrade URL. Phase 3 redacts it via `:filter_parameters` + nginx `access_log off`, but it's still visible to anyone who can see the URL pre-redaction. Move to `Sec-WebSocket-Protocol` or a post-connect `phx_join` payload -- needs phoenix.js + UserSocket protocol change.

Phase 5 hardening cluster.

vjt:
Coupled client+server fix **merged to main** in `a00aed3` (rides tonight's 03:00 CEST **cold** deploy — client + server swap atomically, which is exactly what this change requires). Leaving OPEN until verified live post-deploy.

**Approach:** Phoenix 1.8.8's native `Sec-WebSocket-Protocol` `auth_token` mechanism (blessed by both phoenix.js and Phoenix) — chosen over a `phx_join` refactor (connect/3 already establishes identity for socket-id / WSPresence / topic routing). Per vjt, the legacy query-string path is **retained as a server-side fallback** for one deploy cycle so a stale bundle mid-cold-deploy still authenticates (avoids a #193-class stuck client).

**Server**
- `endpoint.ex`: `auth_token: true` (TOP-LEVEL socket opt) + `websocket: [connect_info: [:auth_token]]`.
- `user_socket.ex` `connect/3`: `extract_token/2` prefers `connect_info.auth_token` (subprotocol), falls back to `params["token"]`.
- Auth-method observability (vjt ask): a Logger line (`auth_method:`) + `[:grappa, :ws, :connect]` telemetry tagging `:subprotocol` | `:query_string` — **method only, never the token** (the raw bearer IS the credential). `:auth_method` added to the Logger metadata allowlist.

**Client**
- `socket.ts`: `new Socket(ep, {authToken: token()})` — token via the subprotocol ONLY, deliberately NOT via `params` (phoenix.js appends `params()` to the URL query, which would re-leak the token and defeat the change). Rotation now rebuilds the socket (authToken is captured once at ctor).

[...]

PR #1381, opened 2026-08-16
#1379: cicchetto declares its protocol version; X-S3 already closed

vjt:
## Summary

Bucket J of the 2026-08-15 codebase review — the two HIGH cross-surface
findings. One is implemented; one is **refused with the measurement**, because
it had already been closed before the bucket was opened.

### X-S3 — refused: already landed by #1354, on all four surfaces in scope

The claim (the carrier topic for `joined` / `join_failed` / `kicked` published
wrong in CLAUDE.md, in `server.ex` and in `api.ts`, the last naming a
`broadcast_window_state_dual/3` that exists nowhere) was **true when written**
and stopped being true the same evening. The review landed as `b09577a1` at
17:47 on 2026-08-15; `31c452cf` (#1354) corrected every surface at 22:19,
including the `CLIENT_PROTOCOL.md` §4 addition the review asked for.

Re-measured on `origin/main` @ `92df169a`, not taken on trust:

| in scope per the brief | state on main |
|---|---|
| CLAUDE.md window-state paragraph | `CLAUDE.md:249` leads **"Every LIVE window-state transition broadcasts on the USER topic"** |
| the `server.ex` comment | `:joined` (`:5595`), `:join_failed` (`:5650`), `:kicked` (`:5785`) all name the USER topic |
| the `api.ts` comment | `api.ts:1194` — "broadcast on `Topic.user/1` ONLY"; no `_dual` |
| `CLIENT_PROTOCOL.md` §4 | carries **"Window state is a USER-topic event, not a per-channel one."** |

`grep -rn 'broadcast_window_state_dual' lib/ cicchetto/src/` → no hits (rc=1).

**Runtime path, as asked.** `broadcast_window_state/2` (`server.ex:6602`) does
exactly one thing: `Broadcaster.to_user(state, payload)` →
`Grappa.PubSub.broadcast_event(Topic.user(ctx.subject_label), payload)`
(`broadcaster.ex:66-68`). There is no per-channel call anywhere on that path.
The per-channel topic sees these kinds only as the cold-subscribe snapshot —
a per-socket `push/3` from `push_window_state_if_known/4`, not a broadcast.
Confirmed as described; **no edit was needed and none was made.**

### X-S2 — implemented

`grep -rn 'protocol_version|client_proto|min_protocol' cicchetto/src/` returned
exactly one hit and it was a prose comment (`socket.ts:742`). Confirmed.

#447 made declaring opt-in: a client that says nothing is treated as CURRENT.
cicchetto had never opted in, so the 426 refusal was **unreachable against the
one client it was written for** — a service-worker-cached PWA, on a project
where `--cic`-only and server-only deploys are both routine.

- `socketEndpoint` appends `?client_proto=` from an exported
  `CLIENT_PROTOCOL_VERSION`. Query param, not subprotocol: the version is
  public discovery data, so the #95/#202 keep-it-off-the-URL rule (about the
  *bearer*, a secret) does not reach it. In the endpoint rather than the
  Socket's `params` because `params` is the same query string by a longer
  route. phoenix.js's `Ajax.appendParams` switches to `&` when the URL already
  has a `?`, so the two compose.
- `joinUser` stops discarding the join reply's `protocol_version`. Diagnostic
  only — a difference is *legal* under the additive-only rule, so there is
  nothing to gate on — logged once per bundle load, since the `"ok"` hook
  re-fires on every phoenix auto-rejoin.
- **New cross-stack pin.** `protocol_test.exs` reads cic's constant out of
  `cicchetto/src` (mounted into the Elixir test container) and fails when
  `min_version/0` is raised past it. Arming the 426 against cic couples two
  numbers that were previously free to drift: from here, raising the floor
  without shipping a cic bundle 426s the reference client, and a browser
  cannot read a 426 body — the operator would see an endless silent
  reconnect, not an error. A missing constant **flunks** rather than returning
  nil, because Elixir orders every atom above every integer and
  `nil >= min_version()` is `true`: the comparison would have passed vacuously
  on the exact edit it guards.

**No runtime behaviour changes today, deliberately.** `version/0` and
`min_version/0` are both 1 and cic declares 1.

### No `protocol_version` bump is required

Both numbers stay at 1. Nothing here changes a field's meaning or withdraws a
frame; the client merely starts declaring a number it already satisfied.

## Two parts of the review's fix shape I did not implement

**The boot-time `GET /api/config` pre-check — declined, on the server's own
argument.** `ConfigController`'s moduledoc already rules: *"the WS handshake is
the real enforcement point for `min_protocol_version` … `/api/config` is only
the advisory pre-check."* cic was missing the enforcement, not the advice, and
the declaration supplies it. Measured, the pre-check today would fetch,
compare 1 against 1 and do nothing, **100% of the time** — a boot-critical
request on the CRT splash's critical path whose only outcome is a no-op. New
boot fetches are not free here: #75's `GET /me/theme` silently broke the
fake-token freeze specs, because any 401 clears the token. It becomes the
right answer on the commit that raises the floor — which is also when the
"what does cic show" UX call has to be made, and that is not a bugfix's to
make.

**`push_content_encoding` — true, and a different subject.** Nothing in
`service-worker.ts` or `lib/pushPayload.ts` reads it, and the consequence is
real. It is a push-decryption concern that rides in `/api/config` by
convenience, not part of the version seam. Named, not done.

## Gates (all local, this worktree, on a clean tree)

- `scripts/check.sh` → **RC=0**: 8 doctests, 60 properties, **6340 tests, 0
  failures**; Credo `5919 mods/funs, found no issues`; Dialyzer `Total errors:
  0`; Sobelow pass; bats `0` lines matching `^not ok`.
- `scripts/bun.sh run check` → **RC=0** (biome + `tsc --noEmit` + e2e tsconfig).
- `scripts/bun.sh run test` → **RC=0**, **285 files / 5508 tests, 0 failures**.

### Every new assertion bought by mutation — 10 mutants, 10 killed

| mutant | tests reddened |
|---|---|
| M1 `client_proto` dropped from the URL entirely | 9 (4 × #193 URL pins, the singleton ctor pin, 4 × #1379) |
| M2 URL literal drifts from the constant (const=2, URL says 1) | 3 — exactly the three that compare against the constant |
| M3 query dropped only on the `!l` fallback branch | **1** — the no-DOM test alone |
| M4 join reply left unread (pre-#1379 posture) | 2 — both warn-dependent tests |
| M5 once-per-load latch removed | **1** — the once test alone |
| M6 absent `protocol_version` no longer silent | **1** |
| M7 equality arm removed (same version warns) | **1** |
| M8 the check returns early, swallowing the reply | 2 — passthrough + the pre-existing visibility test |
| M9 `min_version` raised to 2 while cic declares 1 | **1** — the cross-stack floor pin alone |
| M10 constant reshaped so the source pin cannot read it | 2 — both cross-stack tests, via the `flunk` path |

M9 moves `version/0` **and** `min_version/0` together so the `min <= version`
invariant still holds and only the cross-stack pin can object.

- **No e2e ran and no browser was observed.** I did not hold the stack lane.
  The chain is pinned as far as the `Socket` constructor by unit test; the last
  hop — phoenix.js appending its own `vsn` to a URL that now already carries a
  query — is argued from `phoenix.mjs`'s `Ajax.appendParams`
  (`url.match(/\?/) ? "&" : "?"`), **not measured against a running server**.
  An e2e assertion that the real bundle's upgrade URL carries `client_proto`
  is writable (`page.on("websocket")` → `ws.url()`, the
  `issue1061-offline-does-nothing.spec.ts` pattern) and is *not* in this PR
  because I could not run it. Say the word and I will take the lane.
- **The 426 has never fired against cicchetto and cannot today**, both numbers
  being 1. The refusal is proven against synthetic below-floor clients only —
  `user_socket_test.exs:147,154,164` and
  `issue447-protocol-handshake.spec.ts:69`, both of which **already existed**;
  I added no new server-side refusal test because that gate was not missing.
- **shottino still declares nothing.** Out of scope for a finding about
  cicchetto; the absent-is-current guarantee still has holders, and I only
  corrected the e2e comment that named cicchetto among them.
- The `Sobelow` Low-confidence `Traversal.FileModule` finding in
  `lib/grappa/cic/bundle.ex:170` is pre-existing and below the gate threshold.

### DIFF
--- a/cicchetto/e2e/tests/issue447-protocol-handshake.spec.ts
+++ b/cicchetto/e2e/tests/issue447-protocol-handshake.spec.ts
@@ -101,11 +101,16 @@ test.describe("#447 protocol handshake + /api/config discovery", () => {
   test("WS handshake with NO client_proto → treated as current (403 auth-fail, NOT 426)", async ({
     request,
   }) => {
-    // The zero-behavior-change guarantee for cicchetto + shottino (which
-    // declare no version), proven at the handshake layer: an absent
-    // version is treated as current and falls through to auth (403), so it
-    // never trips the 426 upgrade path. If absent were mis-treated as
-    // too-old, this would be 426.
+    // The absent-is-current guarantee, proven at the handshake layer: an
+    // absent version falls through to auth (403) and never trips the 426
+    // upgrade path. If absent were mis-treated as too-old, this would be 426.
+    //
+    // #1379 narrowed who this covers. It was written for "cicchetto +
+    // shottino, which declare no version"; cicchetto now declares
+    // (`socket.ts` CLIENT_PROTOCOL_VERSION), so the un-declaring clients are
+    // shottino and any third party that skipped the negotiation. The
+    // guarantee is unchanged and still has holders — but it is no longer what
+    // keeps the reference client connecting, which is the point of #1379.
     const res = await request.get(`/socket/websocket?vsn=${PHX_VSN}`);
 
     expect(res.status()).toBe(403);
--- a/cicchetto/src/__tests__/socket.test.ts
+++ b/cicchetto/src/__tests__/socket.test.ts
@@ -121,8 +121,10 @@ describe("socket singleton", () => {
     // #95: authToken carries the bearer via the Sec-WebSocket-Protocol
     // subprotocol; the token is deliberately NOT passed via `params`
     // (phoenix appends params to the URL query — that would re-leak it).
+    // #1379: the URL also carries `?client_proto=` — the protocol version is
+    // public discovery data, unlike the bearer this test is about.
     expect(h.socketCtor).toHaveBeenCalledWith(
-      "ws://localhost:3000/socket",
+      "ws://localhost:3000/socket?client_proto=1",
       expect.objectContaining({ authToken: "tok-1" }),
     );
     const opts = h.socketCtor.mock.calls[0]?.[1] as {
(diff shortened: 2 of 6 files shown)


vjt:
**Held — this is a real regression, not the timeout it looked like.**

All three `integration` runs on this branch ended `cancelled` at exactly 45 minutes, which reads as the job timeout. It is not the whole story: the run does not merely take too long, it **fails nearly everything and therefore never finishes**.

Measured on run `31930058748` (head `a63bfb22`, rebased onto `0a650e82`):

- **92 passed, 288 failed**, cancelled at the cap while still running test 378.
- The first 19 tests pass (`_smoke`, `admin-network-crud`, `admin-server-crud`). Failures begin at test 20 (`archive-desktop-only.spec.ts:47`) and are then near-universal across every spec that drives the SPA.
- Each failure burns its own timeout, which is what pushes the job past `timeout-minutes: 45` — the duration is a symptom, not the cause.

Base is innocent: `integration` on `main` at `62285aaa` completed **green in 39 minutes**, and this branch was rebased onto `0a650e82` before this run.

The shape of the split — server-side and admin-REST specs green, SPA-driving specs red — points at the client handshake this PR changes. That is a hypothesis from the pass/fail boundary, **not a measurement**; it needs to be bought before it is believed.

No further re-runs from me: a re-run of a branch that fails 288 specs buys nothing. Next step is diagnosis by the author.

Issue #1387, opened 2026-08-16
auth: bring the share-token consume endpoint under the house rate-limit policy


vjt:
`POST /auth/share/consume` is an unauthenticated door by design, and it is currently the only credential-adjacent endpoint that sits outside the project's shared rate-limit policy. This aligns it with every other door rather than inventing a new posture for it.

## The house policy already exists

The credential doors share one default: a per-IP failure window of **10 failures over 15 minutes**, with a larger ceiling at 30. Those values live in `config/config.exs` and are not overridden in `runtime.exs` or `prod.exs`, so they are the production values.

## Ruling

1. **Adopt the house default unchanged** — 10 per IP per 15 minutes. A door that behaves like its siblings needs no separate explanation, and a freshly invented number would be the only thing left to justify.

2. **A `410` "already consumed" must NOT charge the window.** That is the shape of an honest double click, or of the PWA retrying, and charging it would lock out precisely the legitimate user. Only an invalid token charges — that is the shape worth limiting.

3. `:passkey_login_options`, which charges on success as well as failure, is out of scope here and stays as it is.

## Notes

- `:request_budget` is not the mechanism to reach for: it reads `current_subject`, and this door is unauthenticated by design.
- The sibling shape to follow is the existing per-IP failure window plus the login-throttle charge.
- Whether the window should also apply to a token that is well-formed but unknown is the same case as "invalid" and needs no separate rule.

vjt:
Closed by the pipeline change ruled by vjt on 2026-08-20: **`status:soon` is retired**. An issue is now closed when its work **merges**, and the **milestone** records which release carries it — the label duplicated what the milestone already said.

This issue was verified individually before closing, not swept: its work is on `origin/main` under commits that are not documentation-only. No behaviour changes here, only bookkeeping.

<sub>Applied by vjt-claude on vjt's instruction.</sub>

PR #1389, opened 2026-08-16
auth(#1387): the share-link redeem door joins the house failure window

vjt:
## What this does

`POST /auth/share/consume` is unauthenticated by design — the signed one-shot
link IS the credential — and it was the last credential-adjacent door standing
outside the per-source-address failure window the rest of them share. It now
runs that window, with the house numbers adopted unchanged: **ten failures per
address per fifteen minutes**, the same values its siblings carry. Inventing a
number here would have left this door the only one with something of its own
to justify.

The window is read before the link is, so a shut door answers the same 429
whatever the body held and costs the server no verification work — the
placement the sibling doors already use.

**Only a link this server cannot have issued loads the window.** Every other
refusal describes a link that WAS ours: already redeemed, run out, or
outliving the identity behind it. Those are the shapes an honest holder
produces on their own second click, and charging them would shut the door on
precisely the person it exists for.

The crossing failure carries the sibling doors' exactly-once operator signal,
so `:share_token_consume` joins the `login_throttled` door enum. The widening
is additive: a client that predates the value keeps the alert and drops only
the attribution it cannot name. That is the whole cic-side change — the door
renders as a label, so no per-door copy was needed.

`:passkey_login_options` is ruled out of scope and stays as it is.

## Tests

Written first. Two of the five were RED before the change; the other three
state what must NOT happen, so each was measured against its own mutant:

| Mutant (one line, at the site named) | Test killed |
| --- | --- |
| charge on a link that ran out | `a link that has run out never loads the window` |
| charge on a link already redeemed | `a spent link retried from the same address never loads the window` |
| charge on a link whose subject is gone | `a link whose subject is gone never loads the window` |

Each mutant killed exactly one test, and the right one — 1 failure of 33 in
all three runs.

## Gates

- `scripts/check.sh` — RC=0 (format, Credo, Dialyzer, Sobelow, doctor, bats;
  8 doctests, 60 properties, 6345 tests, 0 failures)
- `scripts/bun.sh run check` / `run build` / `run test` — RC=0 each
  (5506 cic tests). Run because the wire enum widened.
- `scripts/design-notes-gate.sh` — 1 new entry, separator + marker present
- STACK / integration not requested: no client-facing shape change beyond a
  generated enum member the client already tolerates, and the surface is a
  429 the FallbackController and cic already render for five other doors.

### DIFF
--- a/cicchetto/src/AdminEventsTab.tsx
+++ b/cicchetto/src/AdminEventsTab.tsx
@@ -219,7 +219,7 @@ function renderEvent(ev: WireAdminEvent): string {
       return `${ev.user_name} unbound from ${ev.network_slug}${actorSuffix(ev.actor_user_name)}`;
     case "login_throttled":
       // S6 — a credential door's brute-force gate tripped for this source
-      // IP. Seven windows across five doors share this event, so the door
+      // IP. Eight windows across six doors share this event, so the door
       // and the key that crossed are what make it actionable: the fine key
       // says one account is being hammered from that address, the ceiling
       // says the address is spraying across accounts. Both are additive —
--- a/cicchetto/src/lib/wireTypes.ts
+++ b/cicchetto/src/lib/wireTypes.ts
@@ -417,6 +417,7 @@ export const ADMIN_EVENTS_WIRE_LOGIN_THROTTLE_DOOR = [
   "totp_login",
   "passkey_recovery",
   "passkey_login_options",
+  "share_token_consume",
 ] as const;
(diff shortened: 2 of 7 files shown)


vjt:
Landed via the union PR #1434, which replayed this branch by its own sha — `8d866a69` is an ancestor of `main` at `1552312d`, so GitHub marked this MERGED on its own. Recorded here because the union, not this PR, is what CI gated: #1389 and #1417 were each green against a base that no longer existed and both touched `docs/DESIGN_NOTES.md`.

Issue #1375, opened 2026-08-16
user_settings: the data blob read-modify-write drops concurrent writes to different keys


vjt:
Split out of review bucket B by ruling on #1372 (different table, different defect, different fix).
Full text: `docs/reviews/codebase/2026-08-15-codebase-review.md`, finding P-S5.

All nine `UserSettings` setters follow `get_or_init` → `Map.put(settings.data, @key, value)` → changeset →
persist, with no transaction, no `optimistic_lock` and no conditional UPDATE. Two writers touching DIFFERENT
keys both read the same blob and both write a full replacement: last writer wins and the other key is silently
dropped.

Not hypothetical. `put_last_client_prefix64/2` is driven by `Vhosts.record_client_source/2` on EVERY client
connect, so a second tab connecting while the operator saves a settings-drawer toggle can eat the toggle;
`set_highlight_patterns/2` is reachable from two WS handlers that can interleave. The read-modify-write
discipline is documented for the CLIENT in DESIGN_NOTES; the same hazard on the server side is recorded nowhere.

Fix shape from the review: route all nine through one private `update_data(subject, fun)` doing
`get_or_init` + merge + `Repo.update` inside a single `Repo.immediate_transaction/1`, with the existing
`BusyRetry` moved OUTSIDE to wrap the whole transaction (the `Visitors.create_anon/4` precedent — and see
#1374 P-S7 for why a retry must never be nested inside an open transaction). One helper also deletes eight
copies of the merge idiom.

Notes for whoever takes it:

- The review is a document, not a measurement. Reproduce the lost update before fixing it — a test that
  interleaves two setters on different keys and fails for the stated reason — and buy the assert by mutation.
- If the interleaving turns out to be unreachable in practice, that is a result: say so with the evidence
  rather than shipping a fix for a race nobody can trigger.

vjt:
Closed by the pipeline change ruled by vjt on 2026-08-20: **`status:soon` is retired**. An issue is now closed when its work **merges**, and the **milestone** records which release carries it — the label duplicated what the milestone already said.

This issue was verified individually before closing, not swept: its work is on `origin/main` under commits that are not documentation-only. No behaviour changes here, only bookkeeping.

<sub>Applied by vjt-claude on vjt's instruction.</sub>

PR #1405, opened 2026-08-16
settings(#1375): hold the read and the write of `data` in one transaction

vjt:
Refs #1375.

## The defect

`user_settings.data` is one JSON column, so every setter writes it whole. Nine
setters plus `rename_muted_target/4` spelled the read-modify-write pair by hand:
two writers touching **different** keys both read the same snapshot and both
wrote a full replacement, so the second commit dropped the first's key — with
`{:ok, _}` on both sides and nothing in the logs.

Not theory. `put_last_client_prefix64/2` fires from
`Vhosts.record_client_source/2` on every client connect, in the socket's own
process and its own connection of the ten (`pool_size: 10`,
`config/runtime.exs:187`), while a settings-drawer PUT runs in another.

## The fix

Exactly one writer — `update_data/2` — holding the read and the write in a
single `Repo.immediate_transaction/1`, with the `BusyRetry` budget wrapping it
from **outside** (#1374's nesting). The init inside it is the non-retrying
`get_or_init!/1` for the same reason: a retry loop opened inside an open
transaction sleeps on the connection it holds. #523's degrade contract
(`{:error, :db_unavailable}` → 503) is unchanged, and the helper deletes eight
copies of the merge idiom.

`rename_muted_target/4` is in scope too — same blob, same hazard. It does not go
through `update_data/2` (a rename must never CREATE the row, and it reports
which of the two things it did) but shares the same `write_transaction/1` frame.

## Evidence

The interleave is **reproduced**, not argued. Ecto's per-query telemetry
(`[:grappa, :repo, :query]`) runs its handlers synchronously in the querying
process, so a handler that blocks on the writer's `user_settings` SELECT
suspends it exactly between its read and its write — no seam in production code
and no sleep-and-hope. Unfixed, the key written in that window reads back `nil`.

Mutants run, not predicted:

| mutant | result |
|---|---|
| drop the transaction (`BusyRetry.run(fn -> {:ok, fun.()} end)`) | **both tests red**, each on exactly one assertion (`get_last_client_prefix64` → `nil`) |
| `Repo.transaction/1` instead of `Repo.immediate_transaction/1` | **green — survives** |

The second one is the declared blind spot, and it is declared in the test's own
moduledoc: with one Sandbox connection (`config/test.exs`, `pool_size: 1`) the
second writer blocks on the *checkout* the open transaction holds, not on
SQLite's file-level write lock, and nested under the Sandbox every transaction
mode is a `SAVEPOINT`. These tests buy "the read and the write are ONE
transaction"; they do **not** buy the `:immediate` spelling, which rests on
`Repo.immediate_transaction/1`'s own #524 contract.

`scripts/check.sh` green end to end: 8 doctests, 60 properties, 6351 tests, 0
failures; dialyzer `Total errors: 0`; credo strict, sobelow, bats all pass.

## Overlaps with #1378 (PR of #1374, w2, held)

Landing order is w1 first, #1378 rebases after. Four of the five hunks #1378 has
in `lib/grappa/user_settings.ex` land on lines this PR rewrites; details in the
thread.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -43691,3 +43691,67 @@ lost for good. Do not build machinery on top of this.
 
 Repair of the rows already truncated in production is a separate,
 operator-owned decision — this change stops the loss, it does not undo it.
+<!-- entry #1375 -->
+
+---
+
+## 2026-08-16 — #1375: one write path for the settings blob, and what a one-connection harness can prove about a race
+
+`user_settings.data` is a single JSON column, so a setter that reads the blob,
+merges its key and writes it back necessarily writes the WHOLE column. Nine
+setters spelled that read-modify-write pair by hand, plus
+`rename_muted_target/4`, so two writers touching completely DIFFERENT keys both
+read the same snapshot, both write a full replacement, and the second commit
+drops the first's key — with `{:ok, _}` on both sides and nothing in the logs.
+
+The durable rule is the one the module now states in its moduledoc: **there is
+exactly ONE writer, `update_data/2`, and it holds the read and the write in a
+single `Repo.immediate_transaction/1`.** A new setter that spells its own pair
+reintroduces the defect for its own key. The nesting is #1374's — retry
+OUTSIDE, transaction INSIDE — which is why the row init inside it is the
+non-retrying `get_or_init!/1`: a retry loop opened inside an open transaction
+sleeps on the connection it holds.
+
+### The interleave was reproduced, not argued
+
+Ecto's per-query telemetry (`[:grappa, :repo, :query]`, the event
+`Grappa.DbLatency` already folds) runs its handlers SYNCHRONOUSLY in the
+process that issued the query. A handler that blocks on the writer's
+`user_settings` SELECT therefore suspends it exactly between its read and its
+write, with no seam in production code and no sleep-and-hope — the second
+writer is signalled from inside that window.
+`test/grappa/user_settings_concurrency_test.exs` drives the pair the issue
+names as reachable in production (`put_last_client_prefix64/2` fires from
+`Vhosts.record_client_source/2` on every client connect, in the socket's own
+process and its own connection of the ten, while a settings-drawer PUT runs in
+another). Unfixed, the key written in the window comes back `nil`.
+
+### What the harness substitutes, and what it therefore cannot buy
+
+In production the serialisation comes from SQLite's file-level write lock: the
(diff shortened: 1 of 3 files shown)


vjt:
## Overlap with #1378, measured

Both branches rewrite `lib/grappa/user_settings.ex`. #1378's change there is 118
added/removed lines across five hunks; **four of them land on lines this PR
rewrites**, and the fifth is a function this PR deletes. So the overlap is
effectively the whole of #1378's `user_settings.ex` hunk — it is a re-do, not a
merge.

| #1378 hunk | state after this PR |
|---|---|
| `get_or_init/1` → `do_get_or_init(subject, run)` + public `get_or_init!/1` | **already done here**, byte-identical in intent, with `get_or_init!/1` `defp`. Rebase = make it `def` and move the `@doc` onto it. |
| `rename_muted_target/4` → `do_rename_muted_target(…, persist_fun)`, `rekey_muted_target/4` gains the param | **conflicts.** Here `rename_muted_target/4` carries a `write_transaction/1` frame and `rekey_muted_target/3` returns bare `:renamed \| :noop` (the `{:ok, _}` wrapper is now the transaction's). |
| `put_upload_ttl_seconds/2` → `do_put_upload_ttl_seconds(…, init_fun, persist_fun)` | **conflicts.** Here it is a one-liner over `update_data/2`. |
| `persist!/1` added beside `persist/1` | **`persist/1` is gone.** Its replacement is `write_data!/2`, which takes `(settings, data)` rather than a changeset and rolls back on error. |

### What the two `!` callers need after the rebase

This is the part worth deciding before redoing it, so it is written down rather
than rediscovered.

`Grappa.NickMigration` and `Visitors.create_anon/4` call into this module from
inside an *already open* `BusyRetry.run(fn -> immediate_transaction(…) end)`.
Before this PR the only thing their `!` twins had to skip was the retry, so an
injected `persist_fun` was enough. After it, the plain functions carry **both**
the retry and the transaction, in `write_transaction/1` — so the `!` twins must
skip the frame, not the persist. Concretely: the seam moves up one level from
`persist_fun` to the frame, i.e. `update_data/2` gains a private twin that runs
the same body bare, and `rename_muted_target!/4` calls `rekey_muted_target/3`
without `write_transaction/1`. Same three call sites, one seam instead of two,
and `write_data!/2` is already the non-retrying writer `persist!/1` was going to
be.

Nothing here blocks #1378; it changes what its `user_settings.ex` hunk should
look like.

Issue #1404, opened 2026-08-16
hardening: close the thirteen medium-severity items from the 2026-08-15 security review


vjt:
Tracking issue for the **medium-severity items** of the 2026-08-15 full-project security review, opened on the same terms as #1353 (which covered the high-severity ones and is on `main`).

**Details are deliberately out of band.** The review output stays in a local, gitignored file and is not reproduced here, in the commits, or in the PRs. Fixing these is necessarily public; describing them is not, and self-hosters need time to update before the shape of a defect is legible from the diff.

Scope: **13 items.** Counts for the whole review, which are all that belongs in public: 0 critical, 3 high (closed — #1353, on `main` and deployed), **13 medium**, 4 low-medium, 17 low, 8 informational.

By nature — stated at the level of the control, not the defect — the thirteen fall into four groups:

1. **Ceilings that are absent on paths that accept remote input.** Several accumulators and caches grow with input and have no cap; several answer paths have no budget. The correct behaviour is a bound, and a bound that is asserted by a test rather than by a comment.
2. **Controls whose documented justification names a mechanism that does not cover that path.** In more than one place a moduledoc leans on a quota, a budget or a limit that exists elsewhere. Each is two fixes: put the control where the prose already promised it is, and correct the prose. Where the two disagree, the prose is the one that has been misleading readers.
3. **Supply-chain and build-substrate pinning.** Third-party build inputs are pinned by mutable references. The correct behaviour is pinning by digest, with a documented update path so pinning does not decay into staleness.
4. **Premises in defence-in-depth layers that no longer hold.** One design note asserts a property the implementation does not provide; one policy permits a channel the same policy exists to close. Neither is a standalone break, and neither should be described in public in more detail than that.

Working rules for this issue, unchanged from #1353:

- Commit and PR messages state the guard that now exists and the behaviour that is now correct — never the consequence of its absence, no reproduction, and **no `file:line` that would serve as a map**.
- Each fix needs a test that fails without the guard.
- Several of the thirteen carry a **NOT established** note out of band (unmeasured throughput, unverified chains, environment-dependent reachability). Read it before assuming the declared severity: at least one item's severity is bounded by a configuration the default does not select.
- When in doubt about how much a message reveals, ask rather than publish.

Sequencing note: two of the thirteen sit on paths that also appear in the architecture review's buckets. Where a hardening fix and a structural fix touch the same door, land the hardening first and on its own — the structural change should not be the carrier for a security guard.

[...]

PR #1412, opened 2026-08-16
#1404: three guards on the session process

vjt:
First slice of #1404. Three items, all on `Session.Server`'s inbound path.

**These land first and on their own.** Two of them sit on doors the
2026-08-15 architecture review already has buckets for, and the issue's
sequencing note asks that the hardening not travel inside the structural
change:

- the outbound-reply arm the review's `E outbound-door` bucket proposes to
  replace with a single `send_outbound/2`;
- the state map the `A session-decomposition` bucket proposes to regroup —
  the redaction test is written against VALUES, not key names, precisely so
  that regrouping breaks it loudly rather than quietly stepping around it.

The third is not door-sharing; it is the trigger the second one answers,
and splitting them would ship half a chain.

## What is guarded now

- **A valueless IRCv3 message-tag carries `""`**, the value the spec gives
  it, and the tag type narrows to `String.t()`. The consumer that splits on
  `nil` versus binary is now total by construction rather than by a
  catch-all clause.
- **The CTCP VERSION/PING auto-reply has a ceiling**, and one token buys
  both of its costs — the upstream frame and the scrollback row — because
  a row claiming an answer that never went out is worse than neither. The
  arithmetic is pure and lives in the session's own state; the reasoning
  for not reusing the node-global bucket is in the commit and in
  DESIGN_NOTES.
- **`format_status/1` keeps upstream credentials out of a status dump.**
  The house `@derive Inspect` pattern is unavailable here (bare map, not a
  struct), so this is the callback that owns the same question.

## Gates

- `scripts/check.sh` green on the branch.
- Every guard has a mutant that was RUN, not predicted — six in all. Two
  of them corrected me: one exposed a test measuring the wrong thing, and
  one showed a divergence assertion was never being reached.
- Integration CI has not been run from here; that gate is outstanding
  before any merge.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -158,6 +158,23 @@ config :grappa, :send_throttle,
   oper_capacity: 50,
   oper_refill_per_sec: 5.0
 
+# #1404 — the ceiling on answers a session emits on a STRANGER's command
+# (the CTCP VERSION / PING auto-replies). Distinct from `:send_throttle`
+# above in WHO sets the pace: that one meters the operator's own sends and
+# lives in a node-global `RateLimit.TokenBucket`; this one meters a rate
+# chosen by whoever sends the query, so the arithmetic is pure and the
+# bucket lives in the session's own state — a remote-paced path must not
+# be able to enqueue serialized calls into a node-global process.
+#
+# Same pair of numbers as the ORDINARY send drip on purpose: an answer
+# grappa emits on a stranger's command may not consume more of the
+# upstream's flood allowance than the operator's own client is allowed to.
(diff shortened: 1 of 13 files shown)


vjt:
Superseded by the union PR #1422, which landed on main as `50c00362` (23 commits, b8b09a1f..50c00362). This slice's commits were cherry-picked into that branch, so their SHAs were rewritten and GitHub cannot detect the merge — closing by content instead.

The sec-medium slice was gated on its own, before the union, so a red on the union would not have been attributable to it; the union then gated once as a whole and went 5/5 green at `50c00362`.

The work is on main. Nothing here is lost.

Issue #1413, opened 2026-08-16
cicchetto: the 500ms long-press shows nothing while it runs — the hold reads as a dead touch


vjt:
## What was reported

A cicchetto user on iOS, on IRC: the long-press that opens the message menu feels
like a dead touch. Not "500ms is too long" — *"un delay ignoto all'utente senza
feedback"*: nothing on screen acknowledges the press until the menu is already up.

Their first proposal was to fire the menu on a timer from touchstart instead of
on release. That is already what happens (see below), so the number is not the
defect. The absence of a signal during it is.

## Measurement (read on `origin/main` @ `83d90cb1`; static, no browser)

- `cicchetto/src/lib/messageGestures.ts:131-140` — the hold arms a
  `setTimeout(..., LONG_PRESS_MS)` on **touchstart** and calls `onLongPress`
  when it fires. Nothing is drawn while the timer runs; the handler's only other
  effect is to disarm the swipe.
- `cicchetto/src/lib/keepKeyboard.ts:128` — `LONG_PRESS_MS = 500`, chosen to
  match the iOS long-press convention, feel accepted 2026-07-04.
- `cicchetto/src/lib/keepKeyboard.ts:178` — the **same constant** decides
  tap-vs-hold for the keyboard gate, imported rather than redeclared so one
  press cannot be a hold for one handler and a tap for the other. **Shortening
  the threshold therefore also shortens the keyboard gate** — it is not a free
  knob.
- `grep -n 'vibrate|haptic|pressed|:active'` over `messageGestures.ts` and
  `ScrollbackPane.tsx`: **no hits**. No pressed state on the row, no haptic, no
  progress cue.

Contrast with the sibling gesture: the swipe is self-describing — the row slides
under the finger, and #1156 refused to arm that slide where nothing could be
quoted, on the grounds that *the slide is the promise*. The hold makes the same
promise and shows nothing.

## Direction proposed

Signal the hold rather than shorten it:

1. A pressed state on the row from touchstart (or a progress cue that runs the
   500ms), cancelled on `touchend` and on movement past
   `HOLD_MOVE_TOLERANCE_PX` — the same escape the timer already honours.
2. Optionally `navigator.vibrate` when the timer fires, where the platform has
   it.

Both leave `LONG_PRESS_MS` and the keyboard gate untouched.

## Not established

- **Not device-verified.** This is a static read of `origin/main`; no browser,
  no phone. Whether iOS Safari paints anything of its own during those 500ms
[...]

Issue #1156, opened 2026-08-09
cicchetto: presence rows arm the reply swipe and then do nothing


vjt:
Reported by vjt on `#grappa`, 2026-08-09: presence rows (join / part / quit) can be
swiped like any message, but nothing happens — the row slides, snaps back, and the
compose box stays empty.

## Measured on `origin/main` (`3383abf6`), read end to end

The gesture arms on **every** row, by construction:

- `messageGestures.ts` resolves the row with `MESSAGE_ROW_SELECTOR = ".scrollback-line"`
  (`const` near the top of the file), and `ScrollbackPane.tsx:958` renders **every** row —
  presence included — with `class="scrollback-line"` plus `data-msg-id` and `data-kind`.
- So `onStart` arms, `onMove` claims the rightward drag, the row follows the finger up to
  `SWIPE_MAX_SLIDE_PX = 72` with the `scrollback-line-swiping` class on it, and `onEnd`
  calls `params.onReply(row)`.
- `ScrollbackPane.tsx:1841` resolves the message (`messageForRow`, by `data-msg-id`) and
  hands it to `replyToMessage`.
- `replyQuote` (`lib/replyQuote.ts`) returns `null` on the first line: `if
  (!isContentKind(msg.kind)) return null`. `CONTENT_KINDS` is `privmsg` / `notice` /
  `action` only (`lib/api.ts:641-650`), and the eight presence kinds are the closed set at
  `ScrollbackPane.tsx:913` (`join`, `part`, `quit`, `nick_change`, `mode`, `topic`, `kick`,
  `server_event`).
- `replyToMessage` then returns without touching compose.

This is not an oversight — the current comment states the posture out loud: *"A no-op for a
row with nothing to quote — the gesture still slid and snapped back, which is the honest
feedback for 'armed, but this row has no reply'."* The report says that posture does not
survive contact with a real finger: a 72px slide **is** a promise, and nothing arrives.

## The other door already handles it, and differently

The long-press menu — the *same* verb, the *same* issue (#1067) — does the honest thing:
`MessageContextMenu.tsx` renders **Reply** with `enabled: replyQuote(target.msg) !== null`,
so on a join the item is visible-but-disabled, with the comment *"Disabled-but-visible on a
row with nothing to quote (a join, a part)"*. **Copy** stays enabled ("even a presence row
is worth copying").

So the two doors of one feature disagree: the menu says *this row has no reply* before the
[...]

Issue #1067, opened 2026-08-08
Swipe left→right to reply; long-press → menu (Copy / Select…) instead of programmatically select-all'ing the row


vjt:
## Pivot — Reply is a swipe, not a menu item (2026-08-08)

Direct call from vjt on `#grappa`, after Hypnotize argued a swipe is more intuitive than a
long-press for this:

> `<vjt>` ok vjt-claude pivot usiamo swipe
> `<vjt>` sx->dx si

So **Reply** leaves the long-press menu and becomes a **swipe left→right on the message
row**, Telegram/WhatsApp style: the row slides, snaps back, and the compose box is filled
with `<nick> quoted message<< ` with the caret at the end. Exactly the same fill as before —
only the gesture changes.

What does **not** change: the long-press menu still ships, and still carries **Copy** and
**Select…**. The two symptoms this issue diagnoses (whole-row select-all with the keyboard
up, unmovable selection endpoints) are unaffected by the gesture choice — they are
`keepKeyboard.ts` `handleTouchEnd` and the never-re-enabled `-webkit-touch-callout`, both
still exactly as described below. Reply also stays in the menu as a second path to the same
action; it costs nothing and it is the discoverable one.

**Not settled — out of scope until vjt says otherwise:** the right→left swipe. Hypnotize
proposed `dx -> sx` = open a query with the sender, then floated `dx -> sx` = `!addquote`
instead; vjt closed the thread with "vabe vediamo come viene". Nothing is implemented for
that direction, and picking one here would be inventing a decision that was not made.

**New open point:** the swipe has to coexist with horizontal scrolling inside the
scrollback and with whatever the browser does with a horizontal drag near the screen edge
(iOS back-swipe). The gesture needs a distance/angle threshold that does not fire on a
diagonal scroll, and it must not arm on the inline controls already excluded via
`SELECTABLE_TEXT_EXCLUDE`.

**Acceptance, amended:** a left→right swipe on a message fills the compose box as above,
keyboard open and closed; a vertical scroll never triggers it; the long-press menu keeps
Copy and Select… with the behaviour specified below.

---

On mobile, a long-press on a message should open a small context menu — **Copy**,
**Reply**, **Select…** — with the pressed message zooming slightly (~1.1x) as the
press-feedback. Same behaviour with the keyboard open and closed.

* **Copy** writes the message to the pasteboard.
* **Reply** puts the quoted message into the compose box as `<nick> quoted message<< `,
[...]

PR #1417, opened 2026-08-16
cic(#1413): the long-press shows nothing while it runs

vjt:
## Summary

The long-press message menu armed its 500ms timer on `touchstart` and drew
**nothing** until the menu was already up, so on iOS the press read as a dead
touch. This adds the missing signal.

The number was never the knob. Firing the menu from a timer on `touchstart` is
already what happens, and has since #1067 — which is what the reporter asked
for. The absence of a signal during it is the defect.

Verified still live on `origin/main` at `4ed33d2c` (the issue measured
`83d90cb1`, which had since been replaced by the #1375 revert), and no open PR
touched the gesture files.

### What it does

A cue on the row, driven off the **same state machine as the timer**:

- painted where the timer arms, on `touchstart`;
- unpainted through `cancelHold`, which is already the one exit for **every**
  way the timer can die — release, drift past `HOLD_MOVE_TOLERANCE_PX`,
  `touchcancel`, dispose, and now the fire itself.

A second observer of the touch stream would have been free to disagree with the
one that actually opens the menu; a row still lit under a scrolling finger
promises a menu that is no longer coming.

It is tracked apart from the swipe's arming state on purpose: #1156 leaves a
presence row unarmed for the swipe while the hold still arms on it (Copy and
Select… are useful on a join), so keying the paint to the swipe's state would
have left exactly those rows silent. A mutant pins that.

The cue **ramps** rather than snapping on, because the report is of an unknown
*delay*: a cue that appears instantly says "touched", one that fills says "wait,
and for this long" — the same information the sibling swipe conveys for free by
sliding under the finger, which is the ground #1156 stood on when it refused to
arm a slide that could not be honoured. The duration reaches CSS as a
`--hold-ms` custom property written from `LONG_PRESS_MS`, so the stylesheet
keeps no second copy of 500 to drift from.

### What it deliberately does not do

- **`LONG_PRESS_MS` is untouched.** `lib/keepKeyboard.ts` imports the same
  constant to tell a tap from a hold for the keyboard gate, deliberately rather
  than redeclaring it, so that one press cannot be a hold for one handler and a
  tap for the other. Shortening it to make the menu feel quicker would silently
  shorten that gate too.
- **No `navigator.vibrate`.** iOS Safari does not implement it, so it would have
  done nothing for the person who reported this. A haptic can only ever be an
  extra on platforms that have one — never the fix.

### Why the CSS sits where it sits

The tint is a `background-image` **layer**, declared *after* the mention and
highlight rules. Those set the `background` **shorthand**, which resets
`background-image`, so a tint declared above them would have been wiped off
exactly the rows that already carry a colour — and a `background-color` would
have replaced those colours instead of composing over them. The placement is
load-bearing and commented as such.

## The selection path, explicitly

**This change does not touch it.** No `preventDefault`, no `user-select`, no
`touch-action`, no `position`, no `filter` — nothing that moves a box or joins
the selection machinery iOS begins over copyable text during a long press.
`keepKeyboard.ts` exists because that path is delicate, and the diff stays out
of it: what is added is a paint-only background layer plus a class toggle in the
existing binder.

That is an argument from the properties changed, not a measurement. See below.

1. **Whether iOS Safari already paints something of its own during those
   500ms** — a touch-callout, a selection affordance. That cannot be read from
   the source, and jsdom cannot tell it either. So the claim made here is
   narrow and deliberate: **we** drew nothing before this change and draw
   something after it. Not that the user saw nothing.
2. **Whether the paint disturbs the selection iOS starts over copyable text.**
   The reasoning above says it should not, because no property involved in
   selection or layout is touched. It has not been observed on a device.
3. **The feel.** Whether the ramp reads as an acknowledgement in time, on a real
   phone, in a real hand.

## Device verification is owed, and Playwright does not cover it

The e2e buys the **mechanics**: the paint is present while the hold runs and
gone by every exit. It does **not** buy the feel — Playwright's webkit is not
iOS Safari and does not reproduce real iOS touch physics.

Stated plainly so it is not read as verified: **this ships without device
verification, and the device verification remains due** — it is declared, not
parked.

## Test plan

- [x] `bun run test` — 5516 cic unit tests green (286 files), including 10 new
      jsdom tests over the hold's state machine.
- [x] **Mutation-tested.** Seven mutants; every one of the 10 new tests dies for
      at least one, and three die for exactly one: no `--hold-ms` var, the paint
      keyed to the swipe state (which silences #1156 presence rows), and a row
      painted whose hold never armed. That last mutant had to be written after
      its test survived all the others — a test that never dies is decorative.
- [x] `bun run check` — biome + tsc over `src` and `e2e`, clean; no diagnostic
      on the new rule or the new spec.
- [x] `scripts/design-notes-gate.sh` — checks the entry this branch adds
      (re-run after the rebase, with the entry committed, so it is a real pass
      and not a "nothing to check").
- [ ] **e2e, on CI** — `cicchetto/e2e/tests/issue1413-hold-press-feedback.spec.ts`.
      Three exits (menu fires, early release, drift past the tolerance). The
      oracle is `getComputedStyle(row).backgroundImage`, sampled while the touch
      is live, **not** the class: a class assertion would have gone green with
      the stylesheet never matching at all. The spec never writes the threshold
      down — it reads the `--hold-ms` the binder hands to CSS and derives its own
      wait from it.
- [ ] **On-device**, per the section above.

### DIFF
--- a/cicchetto/e2e/tests/issue1413-hold-press-feedback.spec.ts
+++ b/cicchetto/e2e/tests/issue1413-hold-press-feedback.spec.ts
@@ -1,0 +1,210 @@
+// #1413 — the long-press used to run its 500ms in total silence: the timer
+// armed on touchstart and nothing was drawn until the menu was already up, so
+// the press read as a dead touch. The sibling gesture never had that problem —
+// the swipe slides the row under the finger, and #1156 refused to arm that
+// slide where nothing could be quoted, on the grounds that the slide IS the
+// promise. The hold makes the same promise, and now shows the same evidence.
+//
+// The oracle here is the PAINT, not the class: `getComputedStyle` on the row
+// during the hold. A class assertion would survive the stylesheet never
+// matching — a typo in the selector, or the rule sitting above the mention and
+// highlight rules whose `background` shorthand resets `background-image`. The
+// tint is a background LAYER, so "painted" reads as a `background-image` that
+// is not `none`.
+//
+// Harness + limits, same as #1067's sibling spec:
+//   * chromium, untagged. Chromium's TouchEvent/Touch constructors are
+//     reliable (webkit's are not — Playwright webkit ≠ real iOS), so the
+//     gesture is synthesized in-page on the ROW and reaches the production
+//     listener by bubbling to `.scrollback`, exactly as a real finger does.
+//   * This buys the MECHANICS: the cue appears while the timer runs and is
+//     gone by every way the timer can die. It does NOT buy the FEEL — whether
+//     the ramp reads as an acknowledgement on a real phone, and whether iOS
+//     Safari paints anything of its own during those 500ms, is vjt's on-device
+//     dogfood and is not settled by this file.
+import type { Page } from "@playwright/test";
+import { composeSend, loginAs, selectChannel } from "../fixtures/cicchettoPage";
+import { AUTOJOIN_CHANNELS, NETWORK_SLUG } from "../fixtures/seedData";
+import { expect, specNick, specUser, test } from "../fixtures/test";
+
+test.use({ viewport: { width: 390, height: 844 }, hasTouch: true });
+test.setTimeout(90_000);
+
+const CHANNEL = AUTOJOIN_CHANNELS[0];
+
+// Outside both 20px edge zones (#1041 sidebar / #308 members), so the hold is
+// the only gesture in play.
+const AT = { x: 200, y: 400 };
+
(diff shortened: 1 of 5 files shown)


vjt:
Landed via the union PR #1434. Verified by content at `main` = `1552312d`: for every file this branch touches, the number of lines the branch has and `main` lacks is **0** (checked per file, three-dot merge-base diff).

Closed rather than merged because the union replayed these three commits onto current `main` under a single CI gate, together with #1389. Both branches touched `docs/DESIGN_NOTES.md`, so landing one would have left the other conflicting and ungated.

Issue #1430, opened 2026-08-16
POST /messages answers 201 row OR 202 ack, and cic casts both to ScrollbackMessage


vjt:
## Status of this report: NOT observed

🔴 **This is a structural reading of three files. No trace was run, no log
was read, no reproduction was attempted.** The chain below is what the code
says; whether it fires in the field, and how often, is unmeasured. Please do
not re-state it as an observed bug — turning a read structure into an assumed
magnitude is exactly the error this preamble exists to prevent.

Found while reading the 24 REST cast sites for #1400 (bucket K rest-boundary),
one of which is this endpoint.

## What was read

`POST /networks/:network_id/channels/:channel_id/messages` has **two success
shapes**, discriminated by HTTP status. `GrappaWeb.MessagesController`'s
`render_send_result/2` (`lib/grappa_web/controllers/messages_controller.ex:418-428`):

- `%Scrollback.Message{}` → **201** + `Grappa.Scrollback.Wire.to_json/1`
- `:no_persist` → **202** + `%{ok: true}`

The `:no_persist` arm is not an error path. It is the ordinary outcome for a
`*Serv` target on `send_privmsg` (`lib/grappa/session.ex:628`), for `/notice`
to a service, and for a no-persist CTCP
(`lib/grappa/session/server.ex:4238,4254`).

Client side, `sendMessage` (`cicchetto/src/lib/api.ts:2425-2447`) branches on
`!res.ok` and then casts:

```ts
if (!res.ok) throw await readError(res);
return (await res.json()) as ScrollbackMessage;
```

**`res.ok` is true for 202.** So `{ok: true}` is cast to `ScrollbackMessage`.
The discriminant here is the status, and the client does not read it.

## What the code then does with it

`cicchetto/src/lib/scrollback.ts` awaits that value and reads `row.id`
(`:1178`):

```ts
const current = getReadCursor(slug, name);
if (hasRenderedRow && (current === null || row.id > current)) {
  void setReadCursor(t, slug, name, row.id);
}
```

On the 202 arm `row.id` is `undefined`. `undefined > current` is `false`, so
with an existing cursor the branch declines. With `current === null` the first
disjunct is already true, so — given a pane that holds a rendered row — the
guard passes and `setReadCursor` is called with `undefined` as the message id.

## Why it is worth its own issue

This is the defect #1400 describes in the abstract ("surfaces as `undefined` in
a renderer rather than as a dropped-and-logged payload at the boundary"), but
[...]

Issue #1400, opened 2026-08-16
arch: 83 as-casts on res.json() at the REST edge while the WS edge narrows from unknown (bucket K rest-boundary)


vjt:
Bucket **K rest-boundary** of the 2026-08-15 architecture review (`docs/reviews/architecture/2026-08-15-architecture-review.md`). One HIGH: one wire, two disciplines.

## What was measured

**Type leverage A3 (HIGH).** **83** `(await res.json()) as X` casts outside tests, against a WS edge that narrows everything from `unknown`. That includes the generic at `api.ts:1608` — an unconstrained type parameter asserted from parsed JSON.

## Why it matters

cic deploys independently of the server, so a response one vintage behind is **normal**, not exceptional. Under a cast it lands as a type-checked value with a missing field, and surfaces as `undefined` in a renderer rather than as a dropped-and-logged payload at the boundary. The WS edge already treats the same wire correctly, so the two doors have different drift-detection guarantees for the same domain data.

## Proposed direction

Route REST through the same `validate(S_X, raw)` the WS edge uses. Where a response envelope has no Wire module, that absence is the reason to create one rather than a reason to cast.

## The real blocker (corrected 2026-08-16)

The earlier version of this body said the work depends on bucket D (#1393) "making the generated schemas consumable, sequence it after". **That is wrong, and it was measured wrong.** `cicchetto/src/lib/wireSchema.ts` is a generated file carrying **162 runtime schema literals**, and `wireValidate.ts` already interprets them — the machinery this issue needs is present today. #1393's defect is **adoption** (3 of 162 consumed), not consumability. The first tranche therefore does not wait on it. **Correction (17:20):** the *rest* does — of the 83 sites, 24 are validated in PR #1431 and **56 remain gated on #1393**, because their envelopes have no generated schema yet. My earlier "nothing here waits on #1393" was too broad; it holds for the tranche that had schemas, not for the remainder.

The blocker is elsewhere and was not planned for: the REST envelope types are **hand-written mirrors in `api.ts`** (112 local type declarations there, against a handful of `wireTypes` imports), so for most target envelopes there is no generated schema to validate against — one has to exist before `validate/2` can be pointed at it. That makes the first step **emitting the missing envelope schemas**, not rewiring call sites.

## Ruling on missing required fields (vjt, 2026-08-16)

[...]

Issue #1393, opened 2026-08-16
arch: 162 generated wire schemas, 3 consumed — the codegen migration stopped at the admin channel (bucket D wire-codegen)


vjt:
Bucket **D wire-codegen** of the 2026-08-15 architecture review (`docs/reviews/architecture/2026-08-15-architecture-review.md`). Four HIGH findings — the same arrested migration measured by four independent concern agents — plus three gating MEDIUMs. This is the largest bucket in the review.

## What was measured

**Duplication A1 (HIGH).** `wireSchema.ts` is 1,571 lines / **162** generated `S_*` consts, of which **3** are imported, while ~1,150 hand-transcribed narrower lines still stand (`userTopic.ts:174-1021` — 848 lines — plus five narrowers in `wireNarrow.ts`). #429 emitted the schema for the entire Wire surface from the same typespec walk, under one `--check` gate, precisely so the hand narrowers could be deleted; the migration took the admin channel (3 shapes) and stopped. #429 also settles the standing objection: exactly three of ~30 arms needed a hand-written tolerance, and all three are wrappers *around* `validate(S_…)`.

**Type leverage A1 (HIGH).** 165 generated types, **73** with no consumer — including all three auto-emitted discriminated unions (`WireSessionEvent`, 37 arms; `WireScrollbackEvent`; `WireNetworksEvent`). The hot path is still a hand-transcribed 42-arm union plus a 42-arm hand narrower, while the generated 37-arm union sits unimported next to it. Sampling the arms shows the divergence is comment volume, not shape.

**Abstraction A4 (HIGH).** Membership in the codegen walk is decided by FILENAME (`gen_wire_types.ex:91`, `lib/grappa/**/*wire.ex`, widened once by hand for `error_tokens.ex`). `Grappa.Push.Payload` (`push/payload.ex:76-82`) is a client-consumed wire contract in every respect — including an `optional(:badge)` the emitter handles — but the file is called `payload.ex`, so the client hand-maintains a mirror plus a hand-written narrower. DESIGN_NOTES 2026-07-04 records this as "`push` has no `wire.ex` (hand-written type) — `gen_wire_types --check` green": **the gate reporting green is the symptom.**

**Extension A1 (HIGH).** The same finding measured as change-cost: every new wire event pays the hand-narrower tax in a file the codegen cannot see, and #429 proves that tax is paid with silent omissions.

## Gating MEDIUMs in this bucket

[...]

PR #1431, opened 2026-08-16
cic: validate the 24 one-to-one REST doors instead of casting (#1400)

vjt:
Bucket **K rest-boundary** of the 2026-08-15 architecture review, first slice.

The WS edge narrows every payload from `unknown`; the REST edge asserted
`(await res.json()) as X` at **83** sites. Same wire, two drift-detection
guarantees. Under a cast a response one vintage behind — normal, since cic
deploys independently of its server — lands as a type-checked value with a
missing field and surfaces as `undefined` in a renderer rather than as a
rejection at the boundary.

## The cut: 24 / 3 / 56 = 83

- **24 converted here** — the sites whose target is one-to-one with a
  generated schema. **12 shapes**, 24 call sites, 20 controller actions behind
  them.
- **3 left out** — the bare arrays (`as ScrollbackMessage[]` ×2,
  `as ChannelEntry[]` ×1). The ELEMENT has a schema; the top-level array does
  not. Composing that array node by hand in the client would be **a second
  SSOT of the schema** — the very defect this issue removes, moved one file
  over. The right place for it is server-side.
- **56 left out** — no generated envelope exists for them at all, because REST
  envelopes are inline literals in controllers rather than `*.Wire` modules.
  **Gated on #1393**: what unblocks them is giving those envelopes a Wire
  module so the codegen can emit their schemas. Named, not forgotten.

## The failure mode: FAIL LOUD, in one place

Per vjt's ruling (the single authoritative version lives in the body of
#1400): a **missing required field** makes the boundary reject the payload and
the surface fail **visibly**. It does not fall through to `undefined` and
degrade silently. **Extra fields stay harmless** — the contract is additive
(#447), and `validate` already ignores undeclared keys rather than rejecting.

That mode lives in exactly one function, `narrowRest`. The twelve narrowers
pick a schema and a label and decide nothing else; thirteen copies of a throw
would be thirteen chances to soften one of them into a silent default. When a
tolerance is genuinely needed it goes where the house already puts it — a
named wrapper around the validate call, as `old_nick` does in
`narrowSessionLogEntry`. #1393 measured that exactly three of ~30 arms ever
needed one.

`wireNarrow.ts` remains the only module that names a generated `S_*` constant.

## `POST /messages` is a union discriminated by the STATUS

This door does not have one shape, and the client was not reading the
discriminant. `MessagesController.render_send_result/2` answers:

- **201** + the persisted row (`Grappa.Scrollback.Wire.to_json/1`)
- **202** + `{ok: true}` — an ack for a send the server deliberately did not
  persist: a `*Serv` target, `/notice` to a service, a no-persist CTCP

**`res.ok` is not the discriminant — the status is.** 202 is `res.ok`, so the
old cast covered both arms and handed the caller an object with no `id`.

It now reads the status: 201 narrows strictly, 202 returns `null` — "accepted,
no row", which is not a failure. The 202 body is neither narrowed nor cast;
writing a schema for two keys the server never typespec'd would be the second
SSOT again. The `scrollback.ts` consumer skips the read-cursor advance on
`null` rather than running its gate against an absent id.

Filed separately as **#1430**. This **mitigates it at this one door and does
not close it**: it does not audit the other REST doors for the same
status-discriminated shape, and it does not establish whether the `undefined`
cursor write ever reached the server. #1430 stays open, deliberately with no
`Closes` keyword.

## What the conversion changes about the value

A cast handed the caller the parsed body **verbatim**. `validate` hands back an
object **RECONSTRUCTED** from the declared fields only (`walkObject`,
`wireValidate.ts:223` — "Undeclared keys are dropped, never rejected"). A key
the server sends that its own typespec does not declare is now dropped here
rather than travelling on unnoticed.

Whether any such key exists was measured before a line was written, because
silent data loss would be worse than the defect being fixed. **Zero hits**, and
established by READING, not by grep:

- **The 20 controller actions behind the 12 shapes, across 11 server files,
  read one by one.** Every one renders a `*.Wire` / `*.AdminWire` result and
  none attaches sibling keys to it — the "half-Wire" pattern #1393 records as
  A7 does not occur on any of them.
- **The 12 Wire builders too**, because a `Map.put` grep sees neither a
  pass-through nor a conditional omission. Ten are single unconditional map
  literals whose keys are exactly the `@type`'s. The two pass-throughs are the
  real candidates and both come back negative: `ChannelDirectory.Wire` receives
  entries its context builds with three explicit keys and adds the declared
  `featured`; `FeaturedChannels.Wire` wraps a `select` that projects exactly
  the two declared keys.

A corollary worth keeping: no builder omits a declared key conditionally, so
strict validation cannot reject ordinary traffic for a shape that varies.

## Where the evidence stops

🔴 The reading above covers the **success** arms. That the other 19 render
paths have a single shape on **every error arm** is a **deduction about the
routing** — error arms exit through `FallbackController`, hence `!res.ok`,
hence cic throws before reaching a narrower — **not a reading of
`FallbackController`**. Anyone extending this work should see where the
measurement ends and the inference begins.

Equally: the reconstruction is shape-preserving **as measured today**. That is
a measurement, not a guarantee.

## The nine local aliases are kept

`ScrollbackMessage = ScrollbackWireT` and its eight siblings are **pure aliases
to the generated types** — they carry no drift channel, they *are* the codegen
output. Replacing them with direct imports would touch dozens of files and
close nothing. (An earlier plan for this work said to delete them; it was
written before the aliases were distinguished from the hand-written mirrors.)

## Tests

13 new cases. They were written **after** the implementation, so a green run
proves little on its own — each was falsified by injecting a mutant on the
commit and reverting it:

- **`narrowRest` returns the value instead of throwing** (the fail-loud policy
  removed, everything else identical) → **6 assertions red**, exactly the six
  that name the throw. The cases covering reconstruction, additive tolerance,
  valid pass-through and 202→`null` stayed green, which is correct: they do not
  speak about the throw.
- **`if (res.status === 202) return null;` deleted** → **1 assertion red**, the
  one that names it.

The second mutant surfaced a guarantee rather than just a test result: with the
status branch gone, the 202 body reaches `narrowMessageResponse` and **throws**.
**#1430 is therefore no longer reintroducible silently** — the worst case is
loud.

## Gates

cic-only, from the worktree root:

| gate | result |
|---|---|
| `scripts/bun.sh run check` | RC=0 — 59 warnings, **identical to the pre-change baseline** |
| `scripts/bun.sh x tsc --noEmit` | RC=0 (run separately: a biome failure short-circuits tsc) |
| `scripts/bun.sh run test` | RC=0 — 287 files, 5533 tests (baseline 5520) |

No server-side gate was run: this branch touches six `cicchetto/` files and
nothing else, but that scoping is a reading of the diff, not a verdict from the
integration suite.

### DIFF
--- a/cicchetto/e2e/tests/issue282-vhost-reconnect-button.spec.ts
+++ b/cicchetto/e2e/tests/issue282-vhost-reconnect-button.spec.ts
@@ -50,6 +50,29 @@ const VHOST_VIEW = {
   selection: [],
 };
 
+// The `PATCH /networks/:slug` 200 body — a mirror of
+// `Grappa.Networks.Wire.credential_to_json/1`, which the client validates
+// against `S_NetworksWireCredentialJson` before returning. `connection_state`
+// is overwritten per request with the state that was asked for, so the stub
+// answers like the server (the response IS the updated credential). If the
+// wire grows a required field this fixture goes stale and this spec reddens —
+// truthfully, and with the narrower naming the shape.
+const CREDENTIAL_JSON = {
+  network: ANCHOR,
+  nick: "e2e-p282",
+  ident: null,
+  realname: null,
+  sasl_user: null,
+  auth_method: "none",
+  auth_command_template: null,
+  autojoin_channels: [],
+  connection_state: "parked",
+  connection_state_reason: null,
+  connection_state_changed_at: null,
+  inserted_at: "2026-08-16T00:00:00Z",
+  updated_at: "2026-08-16T00:00:00Z",
+};
+
 async function waitForNetworkState(
   token: string,
   slug: string,
@@ -102,11 +125,24 @@ async function bootToVhostPage(
 
   // Record + short-circuit the connection_state PATCH (the reconnect verb's
   // wire trace). Non-PATCH requests to this exact path fall through.
+  //
+  // The fulfilled body is a FULL `credential_json` echoing the requested
+  // state, not `{}`. Since #1400 `patchNetwork` runs the response through
+  // `narrowCredentialResponse`, which THROWS `WireShapeError` on a shape the
+  // bundle cannot read — and `bounce()` awaits the park before issuing the
+  // reconnect, so one unreadable response silently costs the second PATCH.
+  // `{}` is a body the server never sends: `Wire.credential_to_json/1` always
+  // renders all thirteen fields.
   await page.route(`**/networks/${ANCHOR}`, (route) => {
     if (route.request().method() === "PATCH") {
       const body = route.request().postDataJSON() as { connection_state?: string } | null;
-      if (typeof body?.connection_state === "string") patches.push(body.connection_state);
-      return route.fulfill({ status: 200, contentType: "application/json", body: "{}" });
+      const state = body?.connection_state;
+      if (typeof state === "string") patches.push(state);
+      return route.fulfill({
+        status: 200,
+        contentType: "application/json",
+        body: JSON.stringify({ ...CREDENTIAL_JSON, connection_state: state ?? "parked" }),
+      });
     }
     return route.continue();
   });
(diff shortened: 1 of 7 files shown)


vjt:

Issue #1429, opened 2026-08-16
e2e: a green integration run destroys its own evidence — ~449 of 626 runs are blind by construction


vjt:
A green `integration` run writes no container logs at all — not in CI, not
locally. The dump is gated on a non-zero exit inside the EXIT trap, so roughly
**91% of runs leave no server-side evidence behind by construction**. This is
not a CI policy and not bad luck: it is one `if` in the script, and it is
currently **pinned by a test that asserts it**.

I hit this while censusing a 30.1 s stall regime across CI artifacts (#1420,
#1421). The census can never acquire an honest denominator until this changes.

## The fact, in source

`scripts/integration.sh`, `cleanup()` (**:88-95**), installed at **:104**:

```bash
cleanup() {
    local rc=$?

    # Failure-only. A green run's logs answer no question, and capturing
    # them every time is how an artifact store becomes a landfill.
    if [ "$rc" -ne 0 ]; then
        capture_container_logs
    fi
```

`capture_container_logs()` is **:58-86** and writes one file per compose
service into `cicchetto/e2e/container-logs/`.

Verified in person, not read off: a full local `scripts/integration.sh`
baseline came back **748 passed, 28.2 min, rc=0**, and
`cicchetto/e2e/container-logs/` **did not exist** afterwards.

On CI the three artifact uploads in `.github/workflows/integration.yml` are all
`if: failure()` (**:179**, **:190**, **:203**; `retention-days: 14`), but that is
downstream of the same decision — the files the third step would upload were
never produced.

## A test asserts the blindness

`test/scripts/integration_log_capture_test.bats:121-132`:

```bash
@test "a green run captures nothing" {
    ...
    refute grep -q 'docker compose logs' "$LOG"
    [ ! -d "$LOGS_DIR" ]
```

So the gap is not an oversight anyone can quietly close: it is a stated,
gated property. That is the right place to change it, and the reason this is
filed as a defect rather than a patch.

## Counts (already measured; carried, not re-derived)

- **626** `integration` runs since 2026-08-06; **~449 green** ⇒ blind.
- **59** runs carry a `container-logs` artifact ≈ **9%** of runs.
- Of those 59, **7** show the stall signature. The only honest phrasing is
  **"7 of 59 runs that produced logs"** — never "7 of 626", because the other
  567 were never asked.
- Weight of one capture: **~275 MB uncompressed** per run (`grappa-test.log`
  alone ~263 MB), **~20 MB compressed** in the artifact.

[...]

Issue #1421, opened 2026-08-16
persistence: BusyRetry's 1500 ms budget is unreachable — SQLite's 30 s busy_timeout sits inside the first attempt


vjt:
## Measured

`BusyRetry` carries a retry budget of **1500 ms** (`lib/grappa/repo/busy_retry.ex:47`),
while SQLite's `busy_timeout` is **30 000 ms** (`config/dev.exs:9`,
`config/runtime.exs:199`).

The SQLite timeout therefore sits **inside the first attempt**. Observed in a CI
log, verbatim:

```
[warning] db write unavailable: SQLite pool saturated for the full
          1500ms retry budget (1 attempts)
[debug]   QUERY ERROR source="messages" db=30066.6ms
```

`db=30066.6ms` followed by `(1 attempts)` against a 1500 ms budget is the whole
statement: the first attempt alone spent 30 s inside SQLite, so the budget was
already blown when it returned.

## Consequence

**Against real lock contention the retry engine can never make more than one
attempt.** The linear backoff and the budget are unreachable by construction —
whatever the numbers are set to, the caller waits one `busy_timeout` and then
gives up.

This is coherent with the code rather than an implementation slip: nothing is
wired wrong. But the budget and its comment describe a retry behaviour that does
not occur, which is worse than no comment — the numbers read as tunable and are
not.

## Direction

The two timeouts have to be chosen together: either `busy_timeout` drops below
the retry budget so the engine gets its attempts, or the budget stops claiming
attempts it cannot make. Which way round is a design call — the observable that
should decide it is what a caller is meant to do while a legitimate writer holds
the lock.

## Not established

Not profiled, and no claim about which of the two numbers is the wrong one. This
was found while reading logs for another defect and is reported rather than
chased.

vjt:
## Closing the cure: options with measured cost

Recorded state before this: `docs/DESIGN_NOTES.md` (#1420-instrument) already
carries this defect under **"Named, not cured"**. What follows is the cure
decision, priced.

**A + B are implemented** (PR to follow): documentation, plus the terminal log
line reporting the wait it OBSERVED. **C / D / E / H are NOT taken** — they
change retry behaviour under contention, which is #1420's contested axis.

---

### 1. The mechanism, stated exactly

The budget is a deadline consulted **between** attempts (`busy_retry.ex`, the
`System.monotonic_time(:millisecond) < deadline` arm). It cannot preempt an
attempt already running: the engine does not own the wait, SQLite does. So
[...]

PR #1433, opened 2026-08-16
e2e(#1429): give a green integration run a measurement instead of silence

vjt:
Refs #1429.

A green `integration` run wrote no container logs at all, so ~449 of 626 runs
since 2026-08-06 left no server-side evidence — **blind by construction**, not by
bad luck. The blindness was also *pinned*: `integration_log_capture_test.bats`
carried a test named "a green run captures nothing".

That mattered because the open question about the 30.1 s stall regime
(#1420, #1421) is whether it also hits the runs that **pass** — and those were
exactly the runs whose evidence was discarded.

## What changes

The EXIT trap now streams every service's log through a tracked scanner on
**every** run and keeps the measurement. The raw bytes are materialised on a
red, and on a green only when the measurement says something.

* `scripts/log-gap-scan.awk` — one streaming pass per service: largest silence
  and when it started, every silence at or over the threshold, and counts of the
  damage signatures (`db=3xxxx`, `idle=3xxxx`, dropped scrollback row, saturated
  pool). A few hundred bytes instead of a few hundred megabytes.
* `scripts/integration.sh` — `tee` gives one code path for both exits: the
  stream always reaches the scanner, the disk only when the bytes are kept.
* `.github/workflows/integration.yml` — one added `if: always()` step uploading
  `gap-scan.tsv` alone. The three existing `if: failure()` steps are untouched.

**This is evidence collection, not a gate.** No branch of the trap touches the
exit code, and a test asserts a green run stays green.

## Retention: attribution and retention are different jobs

Retention fires on **a gap on an eligible service, OR a non-zero `dropped`**,
and the census records which:

```
RUN	RETENTION	kept=yes	by_service_gap=0	by_dropped=1
```

Attribution stays single-criterion. Retention answers a different question —
what evidence do I want to still have tomorrow — and there a gap is only a
*proxy* for a mechanism while a dropped row *is* the damage. Retaining on the
gap alone would make **damage without a stall** permanently unobservable, which
is the class nobody can currently prove exists. The field is what turns
*retained* into *countable*: one grep over the uploaded artifacts answers
whether that class happens, and how often.

**Eligible is derived from `docker compose ps`, and the discriminant is the
pair `(State, ExitCode)`.** A container that exited cleanly is a one-shot that
finished; its silence is the interval between invocations. Measured:
`cicchetto-build-test` exceeded the threshold on **both** runs observed
(111.3 s and 12.9 s), so retaining on any gap would retain on essentially every
run — the ~9 GB landfill this change exists to avoid, restored. A container that
exited non-zero crashed, and its silence does mean something.

The pair is load-bearing, and it was sampled on a live stack rather than
assumed:

```
hub|running|0          <- a RUNNING container also reports 0
cert-init|exited|0     <- a completed one-shot
```

Keying on the exit code alone would have made every service ineligible:
retention would have stopped firing while the census kept printing a verdict it
was no longer computing. There is a test for exactly that mutant.

## Reading a gap: the direction of causation is not given

Documented in the scanner's own header and in `docs/OPERATIONS.md`, because it
is the part a future reader is most likely to get wrong.

| census shows | licenses |
|---|---|
| gap **+ a damage signature** | a stall — the environment is implicated |
| **bare gap**, damage counters zero | **nothing** — traffic drought and a stalled process look identical |

A spec hung on a locator stops driving the stack, and a server with no traffic
logs nothing, so the silence can be the **effect** of a failure rather than its
cause. Not hypothetical: the first real red this instrument met showed a 16.7 s
silence on `grappa-test` spanning exactly the failing spec's lifetime
(provisioned 14:40:14, silence 14:40:15.008 → 14:40:31.720, failed 14:40:32) —
and that spec then passed 3/3 in isolation with every damage counter at zero.
Read as an environment stall, it would have exonerated a spec on no evidence.

## Cost, measured

| | |
|---|---|
| whole capture window, 13 services | **11.23 s** |
| of which the awk pass over `grappa-test` (248 MB / 1,092,461 lines) | **8.99 / 8.94 s** |
| ⇒ extraction + disk write add | **~2.3 s** on top of the scan |
| total captured on a red | **253 MB** |
| `gap-scan.tsv` | **1,511 bytes** |
| share of a 30.3 min suite | **~0.6%** |

The three stages run in a pipe, so the cost is the `max`, not the sum: the
pipeline is **awk-bound**. Extraction is not the expense — scanning is, and it
is already paid on every red today. If the 11 s ever becomes a problem, the
lever is the scanner, not the `tee`.

## Verification

* **bats 512/512, RC=0** on the final tree; TDD throughout, each red read by
  name and committed before its green. One exception is stated in its own commit
  message (`a0b793ed`): that red was reconstructed *after* the implementation
  was written, and the commit says so rather than implying the original order.
* **Mutation, one mutant at a time**, each killed by a distinct assertion:

  | mutant | killed by |
  |---|---|
  | green always keeps the bytes | raw-absence assertion |
  | green never keeps the bytes | the measured-stall case |
  | census withheld from stdout | the job-log copy assertion |
  | service label hardcoded | the derived-set assertion |
  | threshold `>=` to `>` | scanner boundary case |
  | midnight rollover removed | scanner rollover case |
  | `dropped` not counted | scanner damage case |
  | eligibility on `ExitCode` alone | the running-service case |
  | eligibility on `State` alone | the crashed-service case, and only that one |
  | no eligibility filter | the clean-green cases |

* The scanner is pinned clause by clause (`log_gap_scan_test.bats`): threshold
  boundary from both sides, arithmetic, midnight rollover — without which a
  stall spanning midnight reads as minus 86390 s and discards the very bytes
  that prove it — one case per damage signature, the `db=3xxxx` neighbours at
  2999.9 and 4000.1 that must not score, an empty stream reading as quiet rather
  than missing, and the refusal to default a missing threshold. Every expected
  value was read off the scanner before being asserted.

## Not established

* **That green runs stall.** This unblocks the question; it does not answer it.
  No claim about frequency, or about the phenomenon crossing the green/red line.
* **That the census is sufficient.** It captures the signatures the census of
  #1420 needed. A future investigation wanting different evidence would still
  find a clean green unable to answer it — this raises the floor.
* **The extraction cost on the green path specifically.** The run measured went
  red, so the number above is the red path, which bounds green from above.
* **A known local wart, reported rather than fixed here:** `container-logs/` is
  a single fixed directory that every run overwrites. Before this change only
  reds wrote there; now greens that trip do too, so consecutive local runs
  clobber each other's evidence more readily. It cost me the raw log of the red
  described above. Harmless on CI (fresh runner, one artifact per run), real
  locally, and left for a separate change rather than widened into this one.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -207,3 +207,16 @@ jobs:
           path: cicchetto/e2e/container-logs/
           retention-days: 14
           if-no-files-found: ignore
+
+      - name: Upload log gap census
+        # #1429 — the measurement taken from the same streams, on EVERY
+        # run. The artifact above stays failure-only because it is ~20 MB
+        # compressed; this one is a few KB, and it is what stops the ~91%
+        # of runs that pass from being unaskable about a stack stall.
+        if: always()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: log-gap-census
+          path: cicchetto/e2e/container-logs/gap-scan.tsv
+          retention-days: 14
+          if-no-files-found: ignore
--- a/docs/OPERATIONS.md
+++ b/docs/OPERATIONS.md
@@ -912,15 +912,69 @@ extraction on the rare interesting green and nothing at all on a clean
 one — the alternative, teeing 253 MB to disk on every green and
 unlinking it afterwards, pays on the common case to save on the rare.
 
+**🔴 READING THE CENSUS: a gap does not tell you which way the
+causation runs.** A silence in a container's log is not evidence that
+the container was stuck. A Playwright spec hung on a locator stops
+driving the stack, and a server with no traffic logs nothing — so the
+silence can be the EFFECT of the failure rather than its cause. This is
+not hypothetical: the first real red this instrument met showed a
+16.7 s silence on `grappa-test` spanning exactly the failing spec's
+lifetime (provisioned 14:40:14, silence 14:40:15.008 → 14:40:31.720,
+failed 14:40:32) — and that spec then passed 3/3 in isolation, with
+every damage counter at zero. Read it as an environment stall and you
+would have exonerated a spec on no evidence.
+
+So:
+
+| what the census shows | what it licenses |
+|---|---|
+| gap **+ a damage signature** (`db30` / `idle30` / `dropped` / `saturated`) | a stall — the environment is implicated, the spec is not |
+| **bare gap**, every damage counter zero | **nothing.** Traffic drought and a stalled process look identical from here |
+
+The counters therefore live on the same line as the gap, deliberately:
+the gap alone is not a verdict, and putting them in a separate report
+invites reading it as one.
+
 **ATTRIBUTION and RETENTION are different jobs, and only the first is
 single-criterion.** Blaming a red uses ONE rule — a silence at or over
-`GAP_THRESHOLD` — and that rule stands alone. Deciding what evidence to
+`GAP_THRESHOLD` **carrying a damage signature** — and that rule stands
+alone. Deciding what evidence to
 still have tomorrow answers a different question, and there a silence
 is only a PROXY for the mechanism while a dropped row IS the damage. So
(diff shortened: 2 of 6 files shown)


vjt:

Issue #428, opened 2026-07-26
Codegen the api.ts type-mirrors (97 hand-copied Wire types drift silently)


vjt:
Deferred seam surfaced by #410 (the STRUCT-mirror + error-token epics it explicitly did not take on). Anchors verified on `main` at `cc7ac706`, not estimated.

## The problem

`cicchetto/src/lib/api.ts` hand-copies **97** `export type` declarations mirroring server Wire shapes (`grep -cE '^export type ' api.ts` = 97), including the client twin of the error-token union from #411/D6b.

Only `wireTypes.ts` is codegen-pinned (drift-gated against `wire.ex`). The other 97 are hand-mirrors with **no gate at all**, so they drift silently: the server changes a Wire shape, the TypeScript keeps compiling against the stale mirror, and nothing fails until runtime.

## Fix

- Extend `mix grappa.gen_wire_types` to emit **structs and unions** too (as generated re-exports), not just the currently-covered subset.
- **Widen the source glob.** The current `lib/grappa/**/wire.ex` pattern misses `admin_wire.ex` — **10 modules are uncovered**. A glob that silently skips modules is worse than no glob, because it looks like coverage.

## Why it matters

This is the same class as the #364 wireTypes gate: the exhaustive generated mirror is what makes `tsc` a real cross-language gate. Every hand-mirror is a hole in that gate.

Unlabelled — vjt or the ircbot decides whether this enters the queue.

vjt:
Shipped to production in **v0.6.0** (`0.6.0-6ba1235a`) — https://github.com/vjt/grappa-irc/releases/tag/v0.6.0

PR #1435, opened 2026-08-16
wire(#1393): give the six identical admin index envelopes a Wire home

vjt:
Slice 1 of #1393 (bucket D, wire-codegen). It stands on its own and does
not depend on any later slice.

## What was actually missing

The admin index **rows** have been codegen-visible since #428 widened the
glob to `lib/grappa/**/*wire.ex` — which also catches `admin_wire.ex`, and
every admin controller already builds its rows with an
`AdminWire.*_to_admin_json/…`. The part that never had a home is the
collection **envelope**: `json(conn, %{users: rows})` is an inline map
literal in the controller, so the shape a client actually receives has no
`@type`, no generated alias and no runtime schema, and cic casts it to a
hand-written mirror instead of validating it.

`Grappa.Networks.FeaturedChannels.Wire.index_payload/1` is the one REST
envelope that already lived in a Wire module, and therefore the one that
was already generated. It is the template these six follow: an
`@type index_payload` next to the existing row type, a builder next to the
existing row builder, and one line of controller.

## The six

`Accounts`, `LiveIntrospection`, `Networks.Credentials`,
`Networks.FeaturedChannels`, `Networks.Servers`, `Visitors` — all
`AdminWire`, all `%{key: [t()]}`, all identical in shape. That identity is
the whole value of the slice, which is why the seventh is not here.

`Grappa.Networks.AdminWire` is deliberately excluded. Its controller
composes `circuit_state` and `live_counts` onto the row, so the envelope's
element is wider than `t()`, and its moduledoc already records why that
composition cannot move into the wire: a `Networks → Admission` reference
would close a boundary cycle against the existing `Admission → Networks`
edge. That is a design question, not a mechanical swap, and it travels
separately.

## Order of the commits

1. `test(#1393)` — the red, pinning all six against the codegen.
2. `wire(#1393)` — the six Wire modules and the six controller lines.
3. `codegen(#1393)` — `mix grappa.gen_wire_types`, no hand editing.
4. `style(#1393)` ×2 — formatter and Credo strict on the new test.

The red was written before a build was available and its commit message
says so. It has since been read: at `9d3a9711` the tests compile and fail
on the assertion (`Grappa.Accounts.AdminWire emits no
AccountsAdminWireIndexPayload`), 2 of 2 — the red that was claimed, not a
compilation error wearing its clothes.

## A note on the regenerated diff

`wireSchema.ts` reads +40/-10 rather than +30/-0. The emitter also moved
`S_NetworksCredentialsAdminWireSessionAction` and
`S_NetworksCredentialsAdminWireSpawnError` to sit after
`S_NetworksCredentialsAdminWireT` and the new index payload. Both blocks
are byte-identical to their previous text, the exported-symbol set loses
nothing and gains exactly six (162 → 168), and running the same task on
the base commit with a clean tree leaves the tree clean — so `main` is
drift-free and the reordering belongs to this change.

## What comes later, and why the gate is last

#1393 blocks 56 of the 83 REST casts in #1400. The remaining envelopes
ship as further slices (`themes`/`notify`/`circuit`/`vhosts`, the auth
ceremony, `uploads`, `settings`, the `%{ok: true}` acks, the one-liners,
and the error shapes). **The structural gate lands last**, once everything
it would flag is already converted — so no exclusion list ever exists in
the tree.

## Verification

`scripts/check.sh` green end to end: compile with Boundary, `mix format
--check-formatted`, Credo strict, Sobelow, `mix deps.audit`, 6399 tests /
60 properties / 8 doctests with 0 failures, doctor, Dialyzer `Total
errors: 0`, the `grappa.gen_wire_types --check` drift gate reporting both
`wireTypes.ts` and `wireSchema.ts` in sync, and the bats suite (499 tests,
no failures).

### DIFF
--- a/cicchetto/src/lib/wireSchema.ts
+++ b/cicchetto/src/lib/wireSchema.ts
@@ -46,6 +46,11 @@ export const S_AccountsAdminWireT = {
   },
 } as const;
 
+// Grappa.Accounts.AdminWire.index_payload/0
+export const S_AccountsAdminWireIndexPayload = {
+  o: { users: { a: S_AccountsAdminWireT } },
+} as const;
+
 // Grappa.Accounts.Wire.client_token_json/0
 export const S_AccountsWireClientTokenJson = {
   o: {
@@ -558,6 +563,11 @@ export const S_LiveIntrospectionAdminWireT = {
   },
 } as const;
 
+// Grappa.LiveIntrospection.AdminWire.index_payload/0
+export const S_LiveIntrospectionAdminWireIndexPayload = {
+  o: { sessions: { a: S_LiveIntrospectionAdminWireT } },
+} as const;
+
 // Grappa.Networks.Network.services_flavor/0
 export const S_NetworksNetworkServicesFlavor = {
   e: [...NETWORKS_NETWORK_SERVICES_FLAVOR],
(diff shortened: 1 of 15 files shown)


vjt:

Issue #1391, opened 2026-08-16
arch: five reply-bundle accumulators are bare map() across 19 producer sites and 43 readers (bucket B accumulator-types)


vjt:
Bucket **B accumulator-types** of the 2026-08-15 architecture review (`docs/reviews/architecture/2026-08-15-architecture-review.md`). Two HIGH findings, one from the abstraction concern and one from the type-leverage concern, measuring the same untyped contract from both ends.

## What was measured

**Abstraction A2 (HIGH).** The five reply-bundle accumulators are typed `map()` across the whole path. `EventRouter` accumulates into `state.*_pending` (**55** references), emits `{:whois_bundle, target, accum :: map(), reply_to}`, and `Session.Wire` unpacks it with **43 `Map.get(accum, :key)` calls** — `whois_bundle/3` alone reads 30 keys. The OUTPUT type is fully declared and codegen-mirrored; the shared contract between writer and reader is not.

**Type leverage A2 (HIGH).** Twelve of the 83 `Session.Server` state keys are `map()`, `nil | map()` or `%{_ => map()}`. For WHOIS alone: **19 producer sites** writing atom-keyed literals in `event_router.ex` against **43 `Map.get` reads** in `Session.Wire`, joined by nothing but the atom spelling. Dialyzer runs with `:underspecs` and catches none of it, because `map()` is inhabited by every map.

## Why it matters

A key typo yields `nil`, which the wire renders as `null`, which the generated TS says is legal — so the client cannot distinguish "the server does not know" from "the peer is not idle". `Map.get(accum, :is_operator, false)` cannot distinguish "upstream said no" from "we misspelled the key". This is the same failure #429 identified one layer up, still open one layer down.

## Proposed direction (revised 2026-08-16 — the first version did not buy the claim)

**Structs, not declared map types.** The original direction here was to declare `whois_accum` and friends as map typespecs next to their `*_payload` twins, on the theory that a writer-side typo would then become a Dialyzer failure. **That theory was wrong, and it was measured wrong by a mutant pair on the same commit with the same compiler:** a struct-field typo is a *compile* error, while a bare-map key typo compiles clean — a declared `map()` shape does not make `Map.get(accum, :nikc)` or a misspelled producer key fail anything. Typing the map leaves all 43 reads exactly as unsafe as they are today.

[...]

Issue #1407, opened 2026-08-16
review: two admin boundaries behave the opposite of their own prose — silent key swallow and a missing catch-all (W-S3, W-S7, bucket E)


vjt:
Bucket **E boundary-shape** of the 2026-08-15 codebase review (`docs/reviews/codebase/2026-08-15-codebase-review.md`) — the two gating MEDIUMs. W-S1, the HIGH of this bucket, is already covered by #1338; these two were left without an issue. Both are the same shape: a boundary that behaves the opposite of what the module's own prose says.

## W-S3 (MEDIUM) — admin `PUT /settings` silently accepts unknown body keys, alone among the admin controllers

**Files:** `lib/grappa_web/controllers/admin/settings_controller.ex:206-209`, `:304-309`

An unrecognised key under `upload` or `addressing` is logged at warning and the action returns **200 with the full settings view** — indistinguishable from a successful write. The moduledoc promises that any invalid value collapses to 422 `invalid_setting` with the offending dotted key; that covers bad VALUES, not bad KEYS. An operator who types `image_per_file_cap` instead of `image_per_file_cap_bytes` watches a save succeed and change nothing.

Every sibling admin controller does the opposite — `Admin.NetworksController.create_attrs/1` / `settings_attrs/1`, `Admin.ServersController.server_attrs/2`, `Admin.CredentialsController.update_attrs/1` / `create_attrs/1`, `Admin.FeaturedChannelsController.channel_attrs/1`, `Admin.UsersController.*_attrs/1`, `Admin.VhostsController.vhost_attrs/1` all compute `extra = Map.keys(params) -- allowed` and return `{:error, :bad_request}`. CLAUDE.md's "no silent-swallow at boundaries" and "same problem, same solution" both point the same way.

## W-S7 (MEDIUM) — `AdminChannel.handle_info/2` has no catch-all, against the module's own stated posture

**Files:** `lib/grappa_web/channels/admin_channel.ex:77-97` (clauses), `:99-105` (the `handle_in/3` catch-all and its rationale)

Phoenix defines no default `handle_info/2`; `Phoenix.Channel.Server` dispatches whenever the channel exports it. `AdminChannel` does, so any message matching none of its three clauses (`:after_join`, `:overview_tick`, `%Broadcast{topic: "grappa:session_log", event: "event"}`) raises `FunctionClauseError` and kills the operator console's pid.

[...]

PR #1437, opened 2026-08-16
session(#1391): make the five reply accumulators structs, and put the check in the pattern

vjt:
## What

The five in-flight reply accumulators — WHOIS, WHOWAS, LUSERS, LINKS and the
type-A LIST-MODE — were bare maps in `Session.Server` state, read in
`Session.Wire` through fifty-five `Map.get/2` calls. `Map.get/2` has no
key-membership check, so a misspelt field read `nil` and the payload shipped
a null the client cannot distinguish from a genuinely absent one.

They are structs now, with the three nested `Entry` structs under WHOWAS,
LINKS and LIST-MODE, and every read is a field access. A map `@type` was
considered and rejected: it documents the key set without enforcing it, and
all fifty-five reads would have stayed exactly as unsafe.

## What the mutants bought

One mutant at a time, injected exactly once, restore verified between each.

| layer | misspelt read | caught by |
| --- | --- | --- |
| accumulator | `accum.total_userz`, `accum.msk` | **compile** — `unknown key`, `typing violation`, exact line |
| entry, before this PR's guard | `e.msk`, `last_entry.usr` | runtime `KeyError` only |
| entry, after the guard | same two | **compile**, like the accumulator |

The split is the finding. Elixir 1.19 infers from VALUES, not from `@type`
specs, and does not carry an element type out of a list — so `[head | _]` or
`fn e ->` leaves the binding `dynamic()` and every read through it
unchecked, even though `entries` is declared `[Entry.t()]`. Naming the
struct in the PATTERN restores the check. **A struct declaration is not a
guard; the guard is the struct appearing in a pattern on the path the value
travels.**

That stricter clause then caught something no mutant could: a test feeding
`banlist_bundle/4` entries that were bare maps. It could never have been
flagged, because a map and the struct built from it project to the same
payload — the assertion passed while the declared element type was fiction.

On the write side `struct!/2` is the only door, which is also why there is
no separate type for a fold's delta: a second declaration of the key set is
a second thing to drift.

## What this does NOT claim

`struct!/2` added **no** coverage the WHOIS tests did not already have —
eight sampled fields were all asserted at accumulator level, so the writer
mutant would have died regardless. What it buys is that the guard no longer
depends on that coverage. The reader guard is the net gain; the writer guard
is insurance.

Out of scope on purpose: `who_pending` and `names_pending` stay bare maps
and keep their bracket reads. One rewrite of the WHO drain slipped in with
the typing work and was reverted — on a bare map `Map.get/3` answers a
default where field access raises, and that is a behaviour change riding a
types commit.

## Shape notes

* The payload stays a plain map. `banlist_entry/0` and `links_entry/0` are
  what the TS codegen reads and what Jason encodes, so the builders project
  struct → map; the tests assert structs on the input side and bare maps on
  the payload side. Two shapes at the two ends of one function.
* `Entry` modules are nested in their accumulator's file:
  `Deploy.Preflight.collect_state_blocks/1` collects every `defstruct` in a
  file, so an entry inherits its parent's cold-deploy check with no second
  `@state_helpers` line. Separate files would have cost three files and six
  registry lines for identical coverage.
* `WhoisAccum.channels` / `extra_lines` keep a `nil` default — on the wire
  `null` and `[]` are different values.

## Gates

`scripts/check.sh` green on the rebased branch: format, Credo
(`6011 mods/funs, found no issues`), `8 doctests, 60 properties, 6397 tests,
0 failures`, Dialyzer `Total errors: 0`, Sobelow, `deps.audit`, bats
`499 ok`. `design-notes-gate.sh`: `1 new entry heading(s), separator and
marker present`.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -44539,3 +44539,76 @@ and cannot be read in jsdom either. What is established is that WE draw
 nothing before this change and something after it. Whether the reporter
 now sees an acknowledgement in time is a device verification, still
 owed — and stated as owed rather than parked.
+<!-- entry #1391 -->
+
+---
+
+## 2026-08-16 — #1391: the reply accumulators are structs, and the check lives in the pattern
+
+The five in-flight reply accumulators — WHOIS, WHOWAS, LUSERS, LINKS and the
+type-A LIST-MODE — were bare maps carried in `Session.Server` state and read
+in `Session.Wire` through fifty-five `Map.get/2` calls. `Map.get/2` has no
+key-membership check, so a misspelt field read `nil` and the payload shipped
+a null the client could not tell from a genuinely absent one. A map `@type`
+would not have helped: it documents the key set without enforcing it, and
+all fifty-five reads would have stayed exactly as unsafe.
+
+They are structs now, and the reads are field accesses. The gain is measured
+rather than asserted, with one mutant at a time, injected exactly once and
+the restore verified:
+
+  * READER — misspell a field in `wire.ex` and it is a COMPILE error:
+    `unknown key`, `typing violation`, the exact line. Elixir 1.19's
+    set-theoretic inference, not Dialyzer, and `--warnings-as-errors` makes
+    it a red build.
+  * WRITER — misspell a key in a producer and it compiles clean, then dies
+    at runtime on `struct!/2` with a `KeyError`. `struct!/2` is deliberately
+    the only write door, which is also why there is NO separate type for a
+    fold's delta: a second declaration of the key set is a second thing to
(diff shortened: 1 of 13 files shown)


vjt:
Superseded by the union PR #1440, which replayed this branch's commits onto current main and gated them together with #1439 (#1407). Verified by content, not by ancestry: with mb = merge-base(origin/main, b63ab6b2), every file this branch touched has zero lines present here and absent from main (0 orphan lines). The DESIGN_NOTES entry marker `entry #1391` is present exactly once on main, with the boundary shape read on the file itself (text / marker / blank / --- / blank / heading). Merged as 38e75561.

Issue #1409, opened 2026-08-16
review: PORT is honoured by nothing on Docker, the oneshot overlay is bypassed six times, and db.sh dies silently from a worktree (D-S6, D-S7, D-S11, bucket I)


vjt:
Bucket **I infra-simplify** of the 2026-08-15 codebase review (`docs/reviews/codebase/2026-08-15-codebase-review.md`) — the three gating MEDIUMs. D-S2 and D-S3, the HIGHs, are in #1384 and #1377; these three had no issue.

**All line references below were re-measured on `origin/main` `9808cb51`.** The review's own citations were taken against the tree of 2026-08-15 and most of the deploy-side ones no longer resolve: the deploy path has since moved out of `scripts/deploy.sh` (now a 69-line wrapper, `:53`/`:67`) into `infra/lib/deploy_docker.sh` + `infra/lib/deploy_common.sh`. Nothing here was executed by me — no docker, no compose; the one executed result is attributed inline.

## D-S6 (MEDIUM) — `PORT` is documented and forwarded but honoured by no Docker-side consumer

`.env.example:114` presents `PORT=4000` as an ordinary knob with no caveat, `compose.yaml:98` forwards it as `PORT: ${PORT:-4000}` specifically so the override is not a silent no-op (#369 X1), and the app really does read it (`config/runtime.exs:318`). But every Docker-side consumer hardcodes 4000:

| Site | what breaks at `PORT=4001` |
|---|---|
| `compose.yaml:33` — `"${GRAPPA_PUBLISH:-127.0.0.1:4000}:4000"` | container side of the publish points at a dead port |
| `compose.yaml:101` — compose healthcheck | healthy container reported unhealthy ⇒ `depends_on: service_healthy` never satisfied |
| `Dockerfile:50` — image healthcheck | same |
| `scripts/healthcheck.sh:13` | probe fails |
| `infra/lib/deploy_docker.sh:172` — `POST /admin/reload` | every hot deploy fails |
| `scripts/deploy-cic.sh:59` — `POST /admin/cic-bundle-changed` | every cic deploy fails |

[...]

Issue #1384, opened 2026-08-16
deploy(docker): two divergent deploy implementations for one substrate, and the HOT path never sets MIX_ENV


vjt:
Surfaced by the 2026-08-15 architecture review (D-S2, D-S3) and re-confirmed alive on `origin/main` by a liveness sweep on 2026-08-16.

## D-S2 — one substrate, two implementations

Two complete `substrate_*` hook sets exist for the Docker substrate:

- `scripts/deploy.sh:52-181` — 14 `substrate_*` functions
- `infra/docker/deploy.sh:627-720` — 14 `substrate_*` functions

All seven drift points between them were re-confirmed:

1. `.env` precondition
2. `MIX_ENV` floor
3. pull-failure message
4. `assert_box_ownership`
5. `migrate_publish_env`
6. `--no-pull`
7. empty-array expansion

Two implementations of the same substrate means an operator's outcome depends on which entry point they happened to run, and a fix applied to one silently leaves the other wrong.

## D-S3 — CLOSED by #1377, not by this issue

`00554be3 fix(#1377): establish .env + MIX_ENV on both deploy paths` landed in the #1382
union (main `62285aaa`) and closes D-S3: `establish_deploy_env` is called from the top of
`substrate_pull`, so both paths of `scripts/deploy.sh` carry the preconditions, and the new
cases in `test/infra/deploy_docker_test.bats` pin it.

The liveness sweep that re-confirmed D-S3 alive checked `origin/main` and was right about
that — the fix was in flight in an open PR at the time. Verifying against `origin/main`
alone is not enough; open PRs have to be excluded too.

What #1377 could NOT reach is the twin: it touches only `scripts/deploy.sh`, where "both
paths" means hot and cold, not the two implementations. That remainder is D-S2 below, and
it is where this issue now lives.

<details>
<summary>Original D-S3 text, kept for the record</summary>

## D-S3 — the HOT path never sets `MIX_ENV` (subsumed by D-S2)

`scripts/deploy.sh:97` reads `[ "$MODE" = cold ] || return 0`, and that early return precedes the `MIX_ENV=${MIX_ENV:-prod}` assignment at `:103-104` inside `substrate_build`. No other line in the 195-line script sets it — the only other occurrence is the preflight's throwaway `-e MIX_ENV=dev` at `:89`. `.env.example:14` ships a bare `MIX_ENV=dev`.

[...]

PR #1444, opened 2026-08-16
infra(#1409): three infra findings from the 2026-08-15 review, plus the cache-id refusal

vjt:
Bucket I of the 2026-08-15 codebase review — issue #1409. All three
findings survive; none survives in the shape the review wrote it.

## What is in here

**D-S7 — `compose.oneshot.yaml`.** The overlay claimed a oneshot without
it "inherits `container_name: grappa` and the host port-publishes, and
both collide with the long-lived copy". Six `compose run --rm` on the
deploy path never layered it and never collided, so both statements
could not be true. Measured with `docker inspect` on the ephemeral
container (`--service-ports` as the positive control): the name is
`/grappa-grappa-run-<hash>`, `PortBindings` is `{}`, `RestartPolicy` is
already `no` — three no-ops. Only `healthcheck: disable: true` is
load-bearing, and it is inherited in full. The file reduces to that one
key and the prose goes. Scope of the measurement, stated in
DESIGN_NOTES: Docker Desktop on macOS, one Compose version, one host,
2026-08-16 — not verified on the Linux CI runs on.

**D-S7, fourth piece — `GRAPPA_CACHE_ID`.** The review's other branch was
to route the six oneshots through `in_oneshot` so they gain
`CACHE_ENV`/`CACHE_VOLUMES`. Declined on a measured asymmetry: `compose
up` exposes no `-v` and `run` does, so the oneshots could take the per-id
binds and the long-running container they migrate FOR could not — a
deploy that migrates inside `.caches/<id>` and then boots from the shared
`_build`/`deps`, with nothing in the output to say which half it is on.
Half an isolation is worse than none. So the substrate says what it
cannot do: ONE guard, in the hook set both Docker doors have shared since
#1384, refusing with the variable named. Unset — every deploy that runs
today — the path is byte for byte the previous one.

**D-S11 — `scripts/db.sh` died mute.** `die` is an `exit 1`, and an
`exit` inside a command substitution kills the subshell, so the
`|| echo dev` fallback was dead the day it was written: the script exited
1 with zero bytes on both streams. The review's advised repair (quote
`$MODE_ARG`) would have added a second bug — an empty quoted scalar
reaches sqlite3 as a filename — so the mode is an array now.

**D-S6 — `PORT` on Docker.** Honoured by exactly one Docker-side
consumer; everything else hardcodes 4000. Marked unsupported in the two
places an operator reads, with `GRAPPA_PUBLISH` named as the knob that
works. The drift gate points its grep at `infra/linux/install.sh` first —
a substrate that DOES honour `${PORT}` — so a broken probe says so
instead of passing vacuously.

## Not observed

The cache-id defect was **deduced by reading the two compose surfaces,
not reproduced**. No half-isolated deploy has been reported or is known
to have been run.

## Testing

- `scripts/bats.sh` (full set): 532 ok / 0 not ok, rc 0 — before the
  rebase and again on `origin/main` after it. Baseline was 528; the four
  new cases account for the difference exactly.
- `scripts/design-notes-gate.sh`: rc 0, one entry heading, separator and
  marker present, re-checked after the rebase.
- Red first for all three findings, and for the guard: with
  `GRAPPA_CACHE_ID` set both doors exited 0 before it existed. The unset
  twin is the control — green before AND after, which is what makes the
  red attributable to the variable rather than to a broken fixture.
- Mutants: the message stripped of the variable name kills exactly the
  two message assertions; a guard narrowed to `scripts/deploy.sh` kills
  exactly the standalone-door case; `[ -n ]` → `[ -z ]` kills 37,
  including the control. A fourth mutant — the guard moved below the
  `.env` check — SURVIVED the first three cases, which proved the
  "holds on an uninstalled box too" claim was prose; the case that pins
  it was added, and that mutant now dies to exactly one assertion.
- `scripts/shellcheck.sh` was not run locally (it is a docker run, and
  the lane is allocated elsewhere). CI runs it.

## Note for reviewers

`bats` runs on the host and inherits the caller's environment, so an
ambient `GRAPPA_CACHE_ID` would have turned every deploy case in the two
Docker suites red once the guard landed. Both `setup()`s unset it, the
way `cache_id_isolation_test.bats` already did.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -111,6 +111,18 @@ VAPID_PRIVATE_KEY=
 POOL_SIZE=10
 
 # Phoenix HTTP listener port. Defaults to 4000.
+#
+# 🔴 NOT SUPPORTED ON DOCKER (#1409 D-S6). Compose passes this into the app,
+# but every other Docker-side consumer hardcodes 4000: the container side of
+# the publish, the compose healthcheck, the Dockerfile EXPOSE + HEALTHCHECK,
+# scripts/healthcheck.sh, and the /admin/reload + /admin/cic-bundle-changed
+# POSTs. Set it to anything else and the container is healthy while reported
+# unhealthy, `depends_on: service_healthy` never resolves, and the deploy
+# fails at the reload POST. To move the port on Docker, change
+# GRAPPA_PUBLISH (the HOST side) and leave this alone.
+#
+# It IS honoured on the Docker-free substrates — infra/linux/install.sh and
+# the release image both thread ${PORT} end to end.
 PORT=4000
 
 # UX-6-B1 (2026-05-20): embedded image uploader storage directory.
--- a/compose.oneshot.yaml
+++ b/compose.oneshot.yaml
@@ -2,14 +2,20 @@
 # `in_oneshot()` for ephemeral mix tasks (test, credo, dialyzer, format,
 # deps.get, ecto.migrate) that have to coexist with a running container.
 #
-# Without it a oneshot inherits `container_name: grappa` and the host
-# port-publishes, and both collide with the long-lived copy.
+# ONE key, because one is all that is needed (#1409 D-S7). This file used to
+# also `!reset` `container_name` and `ports` against a stated collision with
+# the long-lived copy. Measured 2026-08-16 with `grappa` up and holding
+# 192.168.53.12:4000: `compose run` names its container
+# `grappa-grappa-run-<hash>` and never `grappa`, creates it with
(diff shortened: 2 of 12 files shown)


vjt:
Landed via the union PR #1451 (merged as f59b0c7e). This branch's commits were replayed by cherry-pick onto a fresh branch off origin/main together with #1449 and #1450, gated once as a single unit, and fast-forwarded into main. Verified by content: every line this branch adds is present on main. Closing here because the replay rewrote the SHAs, so GitHub cannot mark it merged on its own.

Issue #1402, opened 2026-08-16
arch: the window-state set is restated in six hand-maintained places, none derived, none gated (bucket M closed-sets)


vjt:
Bucket **M closed-sets** of the 2026-08-15 architecture review (`docs/reviews/architecture/2026-08-15-architecture-review.md`). One HIGH plus four gating MEDIUMs — every one of them a closed set restated by hand in places nothing compares.

## What was measured

**Extension A3 (HIGH).** The window-state set is restated in **six** hand-maintained places: the server type, CLAUDE.md's invariant, `windowState.ts`, `pseudoChannels.ts`, and byte-identical `NOT_JOINED_STATES` sets duplicated between `ComposeBox.tsx:112` and `Sidebar.tsx:93`. None derived, none gated — because the type never appears in a `*.Wire` module, so codegen never sees it. The last addition touched **20 files**.

## Gating MEDIUMs in this bucket

- **Type leverage A8 / Duplication A5** — the theme token vocabulary is declared five times (server `@color_keys` / `@font_families` / `@size_modes` plus four TS declarations), with only one pair pinned, and pinned to each other rather than to the server.
- **Duplication A8** — the upload MIME taxonomy has three server copies with a CI pin and a fourth cross-language copy pinned only by a comment.
- **Extension A8** — the cross-language drift-gate pattern exists and works, but covers 4 mirrors and **misses `CONTENT_KIND_PROJECTION`**, the one pair CLAUDE.md and both moduledocs single out as THE single source; it is held together by prose.
- **Type leverage A8 (third leg)** — `me_json.ex` re-spells `WindowCounts.t()` longhand: one shape, three declarations, zero references between them.

## Why it matters

The project already built the mechanism that fixes this class — the `as const` closed-set emitter and the cross-language drift gate — and applied it to four mirrors. The sets left outside are not the obscure ones: window state drives join/compose behaviour, and `CONTENT_KIND_PROJECTION` is the pair the documentation itself nominates as canonical.

## Proposed direction

Publish the window-state set on `Session.Wire` so codegen emits the const, and hoist the two duplicated `NOT_JOINED_STATES` sets into `windowState.ts`. Extend the existing drift gate to `CONTENT_KIND_PROJECTION`, the theme vocabulary and the fourth MIME copy — the harness exists, this is coverage, not design.

## Not established

Nothing was executed and no gate was run. "The last addition touched 20 files" is taken from the review's measurement of a prior change, not re-derived here.

[...]

Issue #1303, opened 2026-08-13
STATUSMSG: a multi-sigil target (@%#chan) misroutes — one lands in $server, the other silently persists to a phantom +#chan


vjt:
Third report from the same operator session, and the one with a second failure mode hiding behind the reported one. Related: #218 (inbound routing), #1247 (level recorded), #1301 (outbound validator), #1302 (badge vocabulary).

## Symptom as reported

A NOTICE whose target carries **more than one** STATUSMSG sigil lands as a private notice in the `$server` tab instead of the channel window.

## Cause

`strip_statusmsg_prefix/2` (`event_router.ex:430-435`) peels exactly **one** sigil, and only if what remains starts with a channel prefix:

```elixir
defp strip_statusmsg_prefix(<<sigil::binary-size(1), rest::binary>> = target, statusmsg) do
  if sigil in statusmsg and channel_target?(rest), do: {rest, sigil}, else: {target, nil}
end
```

with `channel_target?/1` (`event_router.ex:3096`) testing membership in `#&!+`.

Two distinct outcomes follow, depending on which sigil sits second:

**A — the reported one.** `@%#chan` (networks advertising `STATUSMSG=@%+`): `@` peels, the remainder `%#chan` fails `channel_target?/1` because `%` is not a channel prefix, so the guard declines and the target stays `@%#chan`. Not a channel → the non-channel arm → a query/`$server` row. Same for `~&#chan`-style owner/admin stacks on UnrealIRCd-shaped networks.

**B — worse, and not reported because it is silent.** `@+#chan`: `@` peels, the remainder `+#chan` **passes** `channel_target?/1` — because `+` is a channel prefix in its own right (the `+` collision the docstring already calls out). So the function returns `{"+#chan", "@"}`, and the row is canonicalised and persisted against `+#chan`, a channel nobody is in, rather than `#chan`. The recorded level is `@`, and the `+` is lost. This does not land in `$server`; it lands in a window that should not exist, which is why it reads as "the message vanished" rather than "the message went to the wrong tab".

Case B also means the `#1247` level is wrong whenever the outer sigil is not the only one — the badge would claim ops-only for a message actually delivered at the intersection of both levels.

## Shape of a fix (not a ruling)

[...]

Issue #1302, opened 2026-08-13
cic: the STATUSMSG badge names only @ and + — a halfop level renders as a bare sigil, and richer PREFIX networks fare worse


vjt:
Follow-up to #1247, raised by two operators on a self-hosted 1.1.0 install.

## Symptom

The STATUSMSG badge is worded for two levels and raw for the rest. On a network advertising a halfop level, a `%#chan` notice renders a bare `%` next to `@#chan`'s `ops-only` and `+#chan`'s `voice-only` — three rows, two vocabularies. The operators' words: the labels read as noise, and `@` already says "ops only" to anyone reading IRC.

## Where

`ScrollbackPane.tsx:641`:

```ts
const STATUSMSG_LABEL: Record<string, string> = { "@": "ops-only", "+": "voice-only" };
```

with `STATUSMSG_LABEL[level] ?? level` at `:652`. This is deliberate — the comment above it says only the two levels #1247 named get a word, and an unnamed one falls back to its sigil so "unknown level" and "everyone" cannot look alike. This issue is not "the fallback was overlooked"; it is a request to revisit the choice now that a third level is visible in practice.

## Why the current shape does not scale

The sigil set is per-network and open-ended. bahamut advertises `@+`, others `@%+`, and UnrealIRCd-style networks add owner/admin (`~`, `&`) via `PREFIX=(qaohv)~&@%+`. A two-entry hardcoded map cannot name them, so every network richer than bahamut shows a mixed vocabulary.

The repo already solved this exact problem one file over. `channelModes.ts:216-236` (`editorSigils/1`) derives its sigil set from the network's own `isupport.prefix` **specifically to avoid hardcoding `@`/`%`**, with the rationale written in the docstring. The client has that map: `isupport.ts:41,87` carries `prefix: Record<string, string>` (default `{o: "@", h: "%", v: "+"}`). The badge is the one place that still hardcodes.

## Ruling (vjt, 2026-08-16) — sigils in the row, words in the tooltip and the accessible name

- **The row shows the sigils, always.** `STATUSMSG_LABEL` stops being the source of truth: no level gets a word in the visible row, so there is one vocabulary on every network and nothing to maintain per sigil. This is what both operators asked for.
[...]

Issue #1301, opened 2026-08-13
cic: /notice to a STATUSMSG target (@#chan) is rejected as malformed at the POST validator


vjt:
## Symptom

`/notice @#chan message` is refused by the client as malformed. Reported by an operator on a self-hosted 1.1.0 install; `@#chan` is the form operators use most.

## Where it is refused

Not in the client parser. `slashCommands.ts:526-535` accepts any non-empty target and emits `{kind: "notice", target, body}` — deliberately, since the channel-target refusal that `/msg` carries (#12/#343) does not apply to a NOTICE.

The refusal is server-side, at the POST boundary:

- `messages_controller.ex:294` → `validate_post_target_name(notice_target)`
- `validation.ex:100` → `validate_target_name/1`
- `validation.ex:~76` → `Identifier.valid_channel?/1 or Identifier.valid_nick?/1`
- `identifier.ex:83` → `@channel_regex ~r/^[#&+!][^\s,\x07]{1,49}$/`

A STATUSMSG sigil is not a channel prefix, so `@#chan` matches neither validator and the request becomes `{:error, :bad_request}` → 400.

## Measured

Both regexes transcribed verbatim from `origin/main`'s `identifier.ex:69,83` and matched against the candidate targets. This exercises the regexes, not the controller path — no Elixir toolchain on the machine I measured from:

| target | `valid_channel?` | `valid_nick?` | outcome |
|---|---|---|---|
| `#chan` | true | false | ok |
| `@#chan` | **false** | false | **400 bad_request** |
| `%#chan` | **false** | false | **400 bad_request** |
| `+#chan` | true | false | ok — **but for the wrong reason** |
| `&#chan` | true | false | ok — **but for the wrong reason** |

The `+` and `&` rows are the interesting part: they pass because `+` and `&` are RFC channel *prefixes* in `@channel_regex`, not because anything recognised them as STATUSMSG sigils. So the voice-level notice works by accident, while the ops-level one — the common case — is refused. Any network whose ISUPPORT `STATUSMSG=` includes `%` (halfop) is refused too.

## Why this survived #218

[...]

PR #1449, opened 2026-08-16
STATUSMSG: peel the whole run, and let a wire recipient carry a sigil

vjt:
Fixes the two server-side halves of the STATUSMSG group: #1303 (inbound
misroute) and #1301 (outbound refusal). #1302, the cic badge, is
deliberately a separate change.

## #1303 — a multi-sigil target misrouted, one of the two silently

`strip_statusmsg_prefix/2` peeled exactly one sigil and accepted the
result only if what remained started a channel. Two outcomes followed,
depending on which sigil sat second:

- `@%#chan` — `@` peels, `%#chan` is not a channel, the guard declines
  the peel whole, the row lands in `$server`. Loud.
- `@+#chan` — `@` peels and `+#chan` **passes**, because `+` is an RFC
  channel prefix in its own right. The row is persisted and broadcast
  against a channel nobody is in. Silent: it reads as "the message
  vanished", not "the message went to the wrong tab".

The peel now takes the longest leading run of advertised sigils whose
remainder still starts a channel, rolling back whole when no split
works. **Plain greedy was rejected because it regresses two targets
that work today**: on bahamut's `@+`, `@+chan` and `++chan` are
ops-level and voice-level notices to the modeless channel `+chan`, and
a peel with no backtracking strips both bytes, finds `chan` is no
channel, rolls back, and sends them to `$server`. Both are covered by
regression guards that were green before this change.

`meta.statusmsg` now carries the whole peeled run (`"@+"`, not `"@"`).
A STATUSMSG target reaches the union of the named levels, so recording
only the outermost would badge as ops-only a line voiced members also
read. **Consumers must stop assuming length 1** — the field was already
`String.t()`, so no schema change.

## #1301 — `/notice @#chan` was refused as malformed

The POST boundary tested the raw target against the channel regex and
the nick regex, and a membership sigil is neither. `@` and `%` were
rejected; `+#chan` and `&#chan` passed **by accident**, because `+` and
`&` are channel prefixes — so the voice-level notice worked while the
ops-level one, the form operators use most, did not.

This adds a separate `validate_wire_recipient_name/2` rather than
widening `validate_post_target_name/1`, and the measurement is where
the row is keyed: `handle_notice_send/4` keys the echo to the SOURCE
window and carries the recipient in `meta.notice_target`, so a sigil on
a notice recipient never becomes a window key. The plain arm is the
opposite — there the URL channel is both target and key, so admitting
`@#chan` would manufacture a phantom `@#chan` window, the outbound twin
of the defect above. A test pins that the plain arm still refuses it.

The target reaches the wire **verbatim**: peeling to validate is not
permission to rewrite. What `@%#chan` means is the ircd's ruling, and a
refusal on our own authority would hide the ircd's error behind ours.

`Session.statusmsg/2` is the sibling of `casemapping/2` (#537) — the
stateless POST boundary has no `state.isupport`. It degrades to the
bahamut default rather than the empty set, because an empty set would
refuse every sigil on a parked network, which is the reported failure.

## One sigil list

`Identifier.peel_statusmsg/2` is now the single answer to "what is a
STATUSMSG sigil on this network", shared by ingress and egress, with
the set still read from the per-network 005. `EventRouter`'s
`channel_target?/1` delegates to `Identifier.channel_sigil?/1` instead
of keeping a second copy of `#&!+`.

## Gates

- `scripts/check.sh` RC=0 (rc read from the log file): 8 doctests, 61
  properties, 6445 tests, 0 failures; Dialyzer `Total errors: 0`; bats
  0 `not ok`; format, Credo, Sobelow clean.
- Red first, by displacement: with `lib/` checked out at the base and
  the tests committed, the router and controller suites fail by name —
  `@%#chan` to `$server`, `@+#chan` to `+#italia`, the run recorded as
  `"@"`, and `400 bad_request` on the notice. The controls
  (`@+chan`, `++chan`, the per-network run, the URL-arm refusal) were
  green before the fix.
- Seven mutants, none surviving. Three kill exactly one assertion each:
  widening the shared validator onto the plain arm, rewriting the
  target before the wire, and degrading `statusmsg/2` to the empty set.
  A shortest-first peel kills only the `@+#chan` family and leaves
  `@%#chan` green, so the two cases are discriminated rather than
  covered by one blanket assertion.

## Not in this change

The member-prefix sigils are a different, adjacent list, and about ten
sites still spell `@`/`%`/`+` by hand for ranking, glyphs and tier
labels — that is the #1402 class. The badge that renders
`meta.statusmsg` still names two levels and still assumes one
character: #1302.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -45042,3 +45042,101 @@ setup-beam call sites read back from the parse tree, which proves the
 files are well-formed and wired, not that the runner resolves the
 versions. `scripts/shellcheck.sh` was not run and did not need to be —
 no shell script changed.
+<!-- entry #1303 -->
+
+---
+
+## 2026-08-16 — #1303 + #1301: STATUSMSG is a run, the 005 is the only sigil list, and a recipient is not a window key
+
+Three operator reports on a self-hosted 1.1.0 landed in the same place: the
+STATUSMSG prefix. `@#chan` was refused on send, `@%#chan` misrouted on
+receive, and `@+#chan` misrouted silently. They are one defect class — every
+place that has to decide "is this a membership sigil?" had its own answer,
+and only one of them read the network's 005.
+
+### The peel is the longest run whose remainder is still a channel
+
+`@%#chan` peeled the `@`, found `%#chan` was not a channel, declined the peel
+whole and landed in `$server`. `@+#chan` peeled the `@` and accepted `+#chan`
+as the underlying channel — `+` is an RFC channel prefix in its own right —
+so the row was persisted and broadcast against a channel nobody is in. The
+second is the one that matters more, precisely because nobody reported it:
+it reads as "the message vanished", not "the message went to the wrong tab".
+
+The fix peels every leading sigil the network advertises, and rolls the peel
+back **whole** when what remains is not a channel. Plain greedy was tried on
+paper and rejected, because the `+` collision turns it into a regression:
+on bahamut's `@+` it strips both bytes of `@+chan` — an ops-level notice to
+the modeless channel `+chan` — finds `chan` is no channel, rolls back, and
+sends the row to `$server`. That target routes correctly today under the
+single-sigil peel, so plain greedy would have traded two fixed shapes for two
+broken ones. Backtracking one sigil at a time until the remainder is a
+channel resolves `@+chan`, `++chan`, `@+#chan` and `@%#chan` with no special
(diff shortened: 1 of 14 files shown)


vjt:
Landed via the union PR #1451 (merged as f59b0c7e). This branch's commits were replayed by cherry-pick onto a fresh branch off origin/main together with #1450 and #1444, gated once as a single unit, and fast-forwarded into main. Verified by content: every line this branch adds is present on main. Closing here because the replay rewrote the SHAs, so GitHub cannot mark it merged on its own.

Issue #1406, opened 2026-08-16
review: 16 of the 37 generated union arms are transcribed by hand with no drift pin (X-S1 + X-S4/X-S8/X-S9/X-S10, bucket J remainder)


vjt:
Completes bucket **J wire-pins** of the 2026-08-15 codebase review (`docs/reviews/codebase/2026-08-15-codebase-review.md`). #1379 took X-S2 and X-S3; the HIGH and the four MEDIUMs below were left without an issue.

## X-S1 (HIGH) — 16 of the 37 generated union arms carry no drift pin at all

**Files:** server `lib/grappa/session/wire.ex:145`; generated `cicchetto/src/lib/wireTypes.ts:1327`; hand-roll `cicchetto/src/lib/api.ts:753`, `:1084`; exemption `cicchetto/src/lib/wireTypesAssert.ts:145-149`

`wireTypesAssert.ts:145-149` exempts `WireUserEvent`, `WireChannelEvent`, `WireAdminEvent`, `MeResponse` and `Network` from the `Equal<>` pins, on the promise that their per-arm payloads are pinned below. Two things in that sentence are wrong as written.

**The exemption list names five unions; only four are hand-rolled.** `api.ts:1381` reads `export type WireAdminEvent = AdminEventsWireEvent;` — a pure alias to the generated type, so it cannot drift by construction (the same argument #410 used for the leaf enums). It does not belong in the exemption list.

**The gap is not missing arms — cic declares all 37.** The generated `WireSessionEvent` union (`wireTypes.ts:1327`) has 37 arms and cic has a copy of each; what varies is what guards the copy:

| class | count | arms |
|---|---|---|
| full pin (`Extract`, discriminator + body) | **12** | `away_confirmed`, `connection_progress`, `joined`, `join_failed`, `kicked`, `lusers_bundle`, `presence_changed`, `presence_error`, `presence_snapshot`, `window_invited`, `window_invite_declined`, `window_pending` |
| body only (`Equal<X, Omit<Payload,"kind">>`) | **5** | `whois_bundle`, `whowas_bundle`, `names_reply`, `who_reply`, `links_bundle` |
| element only (a nested type is pinned, the envelope is not) | **4** | `topic_changed`, `channel_modes_changed`, `members_seeded`, `mentions_bundle` |
[...]

Issue #410, opened 2026-07-26
wire codegen: finish the half-adopted migration (Wire kinds, unimported generated unions, api.ts hand-mirrors, invisible narrowers, AdminWire outside the glob)


vjt:
Spun out of #369 (architecture review 2026-07-20) — **theme 1, the largest cluster: ~10 findings across 4 concerns.** #369 is closed as an exhausted umbrella; this carries the remaining work.

`gen_wire_types` is sound and CI-gated, but the migration stalled halfway — which is worse than not having started, because two patterns now coexist and whichever is closest gets copied.

## Still open

- ~8-10 Wire modules still type `kind: String.t()`, defeating union emission at the source.
- Generated unions exist but **nothing imports them**.
- `api.ts` hand-maintains ~90 mirror types with opt-in asserts.
- Runtime narrowers (~1,400 lines) are codegen-invisible.
- The 10-module `AdminWire` surface escapes the glob **by filename**.
- Scrollback `meta` crosses the wire as `Record<string, unknown>`.

## Already landed — do not redo

#369 batch 3 converted closed-set wire *kinds* to atom unions (`archive_wire_entry.kind`, `ChannelDirectory` status, `AdminWire.subject_kind`), deleted the drifted dead `Networks.connection_state_changed_event` `@type`, and typed `Themes.Wire.payload`. Wire bytes unchanged (Jason round-trip tests).

## Constraints

Total consistency or nothing (CLAUDE.md) — a partial conversion is what created this issue. If it must be split, split on a **complete seam** (all server `@type`s, then the client), never 'some modules now, some later'.

⚠️ **Six #369 findings failed hand-verification.** Treat its text as a lead list; verify at the source before building.

vjt:
**Two claims in this issue's body are wrong — correcting them at the source (I wrote them, derived from the #369 review rather than verified against the code).**

1. *"Generated unions exist but nothing imports them"* — **false.** Several generated types are imported today (Themes / SessionLog / ServerSettings, plus `ERROR_TOKENS_REST`/`CHANNEL` since #411). What is true is narrower: a few generated enum consts still have hand-written duplicates alongside them.
[...]

PR #1450, opened 2026-08-16
wire(#1406): pin the closed sets that never reached the codegen (X-S8 + X-S9)

vjt:
Third slice of #1406, covering **X-S8** and the pin-able part of **X-S9**. X-S1
landed in #1443; X-S4 is #1448. X-S10 follows in its own PR.

## The defect

`grappa.gen_wire_types` walks `lib/grappa/**/*wire.ex` only. Two closed sets cic
depends on live outside that glob — the window states
(`session/window_state.ex`) and the theme token vocabularies
(`themes/token_model.ex`) — so nothing generated referenced them and cic held a
hand transcription of each, with a "mirror of …" comment and no gate. CLAUDE.md
says adding a window state is a server change cic just mirrors: true as prose,
unenforced as code.

## The mechanism

`Grappa.Session.Wire` re-exports `WindowState.window_state/0`;
`Grappa.Themes.Wire` re-exports the font-family and background-size
vocabularies. No module moves and the glob is untouched — a `@type` at the wire
boundary is enough for the walk to see it.

A re-export rather than a field annotation because **no payload carries the
window-state union**: each event pins its own literal (`state: :joined`,
`state: :failed`, …) and `:parked` never rides a `state` field at all. The set
exists as a set nowhere on the wire, which is why a payload-walking codegen
could not see it.

**The theme specs are DERIVED, not restated.** `TokenModel`'s source of truth is
`@font_families` / `@size_modes`, the string lists `sanitize_font/1` and
`sanitize_size/1` guard with. An atom-union `@type` written next to them would
be a second closed set to keep in step — the defect this issue series exists to
close — so the specs are spliced out of the attributes at compile time with
`unquote/1`. One list, two consumers.

cic then **aliases** the generated types rather than pinning a transcription
with `Equal<>`: the #410 / `WireAdminEvent` posture, since a derived type cannot
drift and the assert would be an identity — the finding X-S1 already measured on
the five `Omit<…, "kind">` pins.

## Evidence: the mutation campaign

Each mutant was applied one at a time with a one-shot injector, the artifact was
checked to confirm the mutation actually arrived, and the tree was restored
after each.

| mutant | layer | outcome |
|---|---|---|
| `:parked` out of the window-state SSOT | ExUnit | **killed** — `the window-state SSOT is emitted as a const…`, one assertion, `left` missing `"parked"` |
| the derived font spec replaced by a hand twin list | ExUnit | **killed** — `the derived font-family spec emits EXACTLY the sanitizer's allowlist`, one assertion |
| `iosevka` out of `@font_families` | tsc | **killed** — `ThemeEditor.tsx(44,3) TS2322` |
| `repeat` out of `@size_modes` | tsc | **killed** — `customTheme.ts(107,16) TS2367`, the comparison losing its overlap |
| `iosevka` out of `@font_families` | ExUnit | **survives, by design** — both sides derive from the attribute, so the test is a construction proof, not a drift detector. Its detector is the twin-list mutant above |
| `:kicked` out of the window-state SSOT | tsc | **survives** — see below |

## What this does NOT buy, stated plainly

**The `WindowState` alias detects nothing at the TypeScript layer.** Dropping
`:kicked` from the server SSOT and regenerating (the generated array verifiably
lost the member) left `bun run check` green: `"parked"` has no writer at all
(`setParted` deletes the key) and the five states cic does write go in through a
computed key — `{...prev, [key]: "kicked"}` — which TypeScript does not check
against `Record<ChannelKey, WindowState>`. For `window_state` the alias buys the
derivation and the codegen test does the detecting. The theme vocabularies are
the opposite: they fail loudly, as the table shows.

**Two of the four theme vocabularies stay unpinned, both declared in the code
that carries them.**

`ThemeColorKey` keeps its hand mirror: the nick slots are the pattern template
literal `nick_${number}`, which TS degrades to an index signature — a payload
missing `nick_5` type-checks today while the server sanitiser requires all 27
keys. Narrowing it is a behaviour change with consumer fallout across the theme
editor and its tests, not a re-export.

`BuiltinBackground` keeps its hand mirror too, and that one is not a choice:
**the codegen cannot render it.** The re-export was written and run.
`BuiltinBackgrounds.t/0` carries `variant: variant()`, a same-module
`user_type`, and the external-type path renders one type at a time with no
sibling registry — so the types half fell back to `camelize/1` and emitted
`variant: Variant` (`TS2304`, a name nothing declares) while the schema half
computed `THEMES_BUILTIN_BACKGROUNDS_VARIANT` and imported it (`TS2724`, a const
neither half emits). The generator exited 0; the breakage surfaced on the client
compiler. `gen_wire_types.ex` already rules out teaching the external path to
resolve refs, and restating the four fields server-side would have rebuilt the
twin this slice deletes — so the emitter's fail-loud travels with the other
codegen work, and a test here pins the ABSENCE meanwhile.

## Gates

`scripts/check.sh` RC=0 (format, credo, dialyzer `Total errors: 0`, sobelow,
520 bats, and the codegen drift check reporting both `wireTypes.ts` and
`wireSchema.ts` "is in sync") · `scripts/test.sh` RC=0, 6413 tests, 0 failures ·
`scripts/bun.sh run check` RC=0 · `scripts/design-notes-gate.sh origin/main`
RC=0. Rebased onto `651aa289`, contribution numstat identical either side of the
rebase and the entry boundary read on the file.

The first `check.sh` was red, by name: Credo consistency at `themes/wire.ex:46`,
because dropping the `BuiltinBackgrounds` alias left two single directives under
one prefix. Fixed in its own commit and the whole gate re-run, not resumed.

### DIFF
--- a/cicchetto/src/lib/themesApi.ts
+++ b/cicchetto/src/lib/themesApi.ts
@@ -17,23 +17,22 @@
 import { buildHeaders, readError } from "./api";
 import { getOrCreateClientId } from "./clientId";
 import { narrowThemeResponse } from "./wireNarrow";
-import type { ThemesWireT } from "./wireTypes";
+import type { ThemesWireBackgroundSize, ThemesWireFontFamily, ThemesWireT } from "./wireTypes";
 
-// The closed font-family allow-list — mirror of
-// `Grappa.Themes.TokenModel.font_families/0`. `mono-default` maps to the
-// existing `--font-mono` stack (no font file); the rest each get a
-// self-hosted `@font-face` (sub-task 8, deferred).
-export type ThemeFontFamily =
-  | "mono-default"
-  | "jetbrains-mono"
-  | "fira-code"
-  | "iosevka"
-  | "hack"
-  | "cascadia-code"
-  | "source-code-pro"
-  | "ibm-plex-mono";
+// The closed font-family allow-list — #1406 X-S9, DERIVED from
+// `Grappa.Themes.TokenModel.font_families/0` (the same list `sanitize_font/1`
+// guards with) instead of mirrored by hand. `mono-default` maps to the existing
+// `--font-mono` stack (no font file); the rest each get a self-hosted
+// `@font-face` (sub-task 8, deferred).
+export type ThemeFontFamily = ThemesWireFontFamily;
 
-// The 27 color keys — mirror of `Grappa.Themes.TokenModel.color_keys/0`.
+// The 27 color keys — still a HAND mirror of
+// `Grappa.Themes.TokenModel.color_keys/0`, and the one vocabulary #1406 X-S9
+// leaves unpinned: `nick_${number}` is a pattern template literal that TS
+// degrades to an index signature, so narrowing it to the server's exact 16
+// nick slots is a behaviour change (a payload missing `nick_5` type-checks
+// today and would stop) with real consumer fallout, not a re-export. Declared
+// debt, tracked on #1406, deliberately out of this slice.
 // Each value is a strict `#rrggbb` string (validated server-side; cic
 // treats them as opaque CSS color literals via `customTheme.ts`).
 export type ThemeColorKey =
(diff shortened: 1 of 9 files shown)


vjt:
Landed via the union PR #1451 (merged as f59b0c7e). This branch's commits were replayed by cherry-pick onto a fresh branch off origin/main together with #1449 and #1444, gated once as a single unit, and fast-forwarded into main. Verified by content: one single line differs, and in main's favour — this branch still carried the loose generated form 'export type ScrollbackMetaT = Record<string, unknown>', while main now carries the narrower derived one, 'Partial<Record<ScrollbackMetaTKey, unknown>>', keyed off the closed SCROLLBACK_META_TKEY set. Nothing of this branch is missing. Closing here because the replay rewrote the SHAs.

Issue #1392, opened 2026-08-16
arch: Grappa.Subject's promotion stopped at three contexts of five — three spellings of one subject predicate (bucket C subject-SSOT)


vjt:
Bucket **C subject-SSOT** of the 2026-08-15 architecture review (`docs/reviews/architecture/2026-08-15-architecture-review.md`). One HIGH — mechanical, no behaviour change, and with the measurement below the risk is **duplication, not divergence**.

## What was measured

Counted at `origin/main` `b50fc6a6`; the review's own figures (24 / 13) do not reproduce and are corrected here.

`Grappa.Subject`'s moduledoc declares it the SSOT for the `{:user, uuid} | {:visitor, uuid}` discriminator and states that `subject_where/2` was "promoted" from the per-context privates. The promotion never finished:

- `Subject.subject_where/2` has **19** call sites across **7** modules — `notify.ex` (6), `query_windows.ex` (4), `channel_directory.ex` (3), `push.ex` (2), `themes.ex` (2), `user_settings.ex` (1), `vhosts.ex` (1). The review named only three of the seven.
- `Scrollback` keeps a private `subject_where/2` (`scrollback.ex:1135`) with **12** call sites, and does not reference `Grappa.Subject` at all.
- `ReadCursor` keeps a third spelling, `subject_filter/2` (`read_cursor.ex:823`, 8 call sites), plus a private `subject_attrs/1` duplicating `put_subject_id/2`.
- **7** modules re-declare the type literal, not four: `accounts.ex:103`, `accounts/revocations.ex:79`, `rate_limit/request_budget.ex:63`, `read_cursor.ex:102`, `scrollback.ex:337`, `session.ex:120` — plus `account_deletion.ex:74`, which is a **different** type (`{:user, User.t()} | {:visitor, Visitor.t()}`, the rich-struct shape) and must not be folded into `Subject.t()`.

**The three spellings are exact synonyms on the valid domain.** All three emit the same `where` on `user_id` / `visitor_id` under the same `is_binary/1` guard; read side by side they are the same predicate. They differ only *outside* the domain, and both branches raise: `Scrollback` has an explicit catch-all raising `ArgumentError` with the inspected subject (`:1148`), while `Subject` and `ReadCursor` fall through to `FunctionClauseError`.

## Why it matters

[...]

PR #1457, opened 2026-08-17
arch(#1392): fold the two private subject predicates into Grappa.Subject

vjt:
Bucket C subject-SSOT of the 2026-08-15 architecture review. Issue #1392.

`Grappa.Subject` has claimed to be the single source of truth for the
`{:user, uuid} | {:visitor, uuid}` discriminator since the V1 visitor-parity
promotion, and its docstring named `Grappa.Scrollback` and `Grappa.ReadCursor`
as the contexts it was promoted out of. Neither referenced `Grappa.Subject` at
all. This finishes the promotion.

## What changed

- `Scrollback.subject_where/2` (private, 12 call sites) and
  `ReadCursor.subject_filter/2` (private, 7) deleted; both now call
  `Subject.subject_where/2`.
- `ReadCursor.subject_attrs/1` (1 call site) folded into
  `Subject.put_subject_id/2`.
- Scrollback's `ArgumentError` fall-through promoted INTO `Grappa.Subject`,
  because the fold could not be a deletion — see below.
- Three of the four verbatim `@type` clones collapse to `Subject.t()`
  (`scrollback.ex`, `read_cursor.ex`, `session.ex`).
- Two false docstring sentences corrected.

## This is a behaviour change, and the issue's "no behaviour change" is retired

The predicate never diverged: all three spellings emitted the same equality
against the same FK column under the same `is_binary/1` guards, and all three
bound POSITIONALLY at index 0 (`[m]` in one, `[row]` in the other two). That is
what makes the 19 replacements a substitution rather than a rewrite — no call
site's query shape can change the result.

The DIAGNOSTIC diverged. Scrollback carried an explicit fall-through raising
`ArgumentError` with the inspected subject (B5.4 L-pers-2), precisely because a
`FunctionClauseError` hides both the offending value and the function name. The
other two had none. Deleting the privates would have stripped that from 12 call
sites, so the clause travels WITH the function — which moves 19 existing
`Subject.subject_where/2` call sites and the 7 former `subject_filter/2` ones
from `FunctionClauseError` to `ArgumentError`.

Nothing in `test/` asserted `FunctionClauseError` on these functions, and the
two pins that already existed in `scrollback_test.exs` pass **untouched** —
measured, not predicted. That is the evidence the fold is transparent at the
Scrollback boundary.

`put_subject_id/2` deliberately keeps its bare `FunctionClauseError`: giving it
the same clause is a second behaviour change across its own call sites that
nothing here asks for.

## Two figures in the issue do not reproduce at this base

- `subject_filter/2` has **7** call sites, not 8. All twelve occurrences of the
  token classified: 7 calls, 2 clauses, 1 `@spec`, 2 in prose.
- "Seven modules re-declare the type literal" merges three different types.
  Only **four** are verbatim clones. `revocations.ex` and `request_budget.ex`
  spell the id as `String.t()`, strictly WIDER than `Ecto.UUID.t()`;
  substituting `Subject.t()` there narrows rather than deduplicates.
  `themes.ex` is the rich-struct shape the issue excluded `account_deletion.ex`
  for, but did not name.

## Deliberately left out

- `accounts.ex:103`, the fourth clone. `Grappa.Accounts` is a DEP of
  `Grappa.Subject`, so referencing `Subject.t()` from inside it would close
  `Subject -> Accounts -> Subject` **if** Boundary tracks remote type
  references. That is not established — the repo has no discriminating case —
  and the declaration is worth two spec sites.
- The two wider `String.t()` spellings and the rich-struct clones, per above.
  `request_budget.ex`'s own typedoc justifies staying local to avoid a
  `Grappa.Session` dep, which is stale (the type lives in `Grappa.Subject`);
  noted in DESIGN_NOTES, not acted on.

## The write-side asymmetry is real and unreachable

`subject_attrs/1` had no guard, so `{:user, nil}` returned `%{user_id: nil}`
where every sibling raises. Traced: both paths into `upsert_cursor/5` call
`get/3` first, and both public entries also pass `message_belongs?/4`, all
subject-guarded — a malformed subject raises before `subject_attrs/1` is
reached. So the fold is deduplication, not a fix, and no boundary test was
fabricated for a path that does not exist.

## Gates — run, with numbers

Measured at `b10bb29d`, rebased onto `60657249`.

- `WRITABLE_CIC=1 scripts/check.sh` — **RC=0**: 8 doctests, 61 properties,
  **6461 tests, 0 failures**; Dialyzer `Total errors: 0`; Credo `found no
  issues`; `deps.audit` `No vulnerabilities found`; Sobelow clean; bats
  **540 ok / 0 not ok**. Includes the `grappa.gen_wire_types --check` drift
  gate.
- `scripts/design-notes-gate.sh origin/main` — **RC=0**, `1 new entry
  heading(s), separator and marker present`. (A first run passed VACUOUSLY —
  `nothing to check` — because the entry was written but not yet committed;
  the gate compares committed branch state. The second run is the one that
  counts.)
- Rebase verification: `git diff --numstat` pinned to a file before and
  re-compared after — additions unchanged AND deletions unchanged. Boundary
  shape read on the file for BOTH this entry and its neighbour `#1302`, since
  `merge=union` can damage the neighbour rather than you.

### TDD, bought with mutants

The five new tests were committed failing first: all five got
`FunctionClauseError` where `ArgumentError` was expected, naming
`Grappa.Subject.subject_where/2` (3) and `Grappa.ReadCursor.subject_filter/2`
(2).

- **M1** — `"unknown subject: #{inspect(other)}"` -> a static string: kills
  **only** the 5 new assertions. The two `scrollback_test.exs` pins survive,
  because they match the prefix while these match the inspected value.
- **M2** — catch-all clause deleted: kills **7**, the 5 new plus both existing
  Scrollback pins. This is the "the promotion is load-bearing" mutant.
- A third mutant (restoring `ReadCursor`'s private) was not run: that is
  exactly the pre-implementation state, and the initial red already measured
  it.

## Gates NOT run, with the reason

- **cic (`bun run check`, `tsc`)** — this branch adds **zero** `cicchetto/`
  lines (`git diff --name-only origin/main...HEAD | grep -c '^cicchetto/'` ->
  0), and the wire drift gate inside `check.sh` is green, so the wire has not
  moved. Running it would measure `main`, not this branch.
- **`scripts/integration.sh` and e2e** — not run. This is a core-context
  refactor with no wire, controller, or client surface; the full suites are the
  ship gate and belong to whoever merges.
- **`mix boundary.find_violations` standalone** — not run separately. Boundary
  is a compiler (`compilers: [:boundary]` in `mix.exs`) with
  `check: [in: true, out: true]`, so the two new deps were enforced by the
  clean compile inside `check.sh`. The no-cycle argument was also derived by
  hand: the transitive closure of `Grappa.Subject` is 10 modules and contains
  neither `Scrollback` nor `ReadCursor`.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -45535,3 +45535,122 @@ jsdom. `role="img"` stays — ARIA does prohibit the name there, so no AT
 owes us the reading — but that is the spec, not a measurement, and the
 tests now pin the spec-legal SHAPE with a byRole query instead of
 pretending the name assertion proves it.
+<!-- entry #1392 -->
+
+---
+
+## 2026-08-17 — #1392: three spellings of one subject predicate, and the claims measurement retired
+
+Bucket C of the 2026-08-15 architecture review. `Grappa.Subject`'s moduledoc
+has declared it the single source of truth for the
+`{:user, uuid} | {:visitor, uuid}` discriminator since the V1 visitor-parity
+promotion, and named `Grappa.Scrollback` and `Grappa.ReadCursor` as the
+contexts it was promoted OUT of. Neither of them referenced `Grappa.Subject`
+at all. The promotion reached nine modules and stopped one short of the two it
+was named after.
+
+Counted at `f59b0c7e`: `Subject.subject_where/2` served 19 call sites across
+seven modules, `Scrollback.subject_where/2` (private) 12, and
+`ReadCursor.subject_filter/2` (private) 7. A fourth spelling sat on the write
+side — `ReadCursor.subject_attrs/1`, one call site, duplicating
+`Subject.put_subject_id/2`.
+
+### The predicate did not diverge; the diagnostic did
+
+All three built the same clause: the same `is_binary/1` guards on both
+branches, the same equality against the same FK column, and the same
+**positional** binding — `[m]` in one, `[row]` in the other two, both index 0.
+That last detail is what makes the fold a substitution rather than a rewrite:
+no call site's query shape can change the result, because the binding is
+resolved by position and the position is identical. No value in the valid
+domain makes any two of them build a different query.
+
+They differed OFF the domain, and there the difference is real.
+`Scrollback` carried an explicit fall-through raising `ArgumentError` with the
+inspected subject, added by B5.4 L-pers-2 precisely because a
+`FunctionClauseError` hides both the offending value and the function name.
+`Subject` and `ReadCursor` had no such clause.
+
+So the fold could not be a deletion. Dropping the two privates and letting
+`Subject.subject_where/2` answer would have removed that diagnostic from 12
+call sites; the clause was promoted **with** the function instead. That is a
+behaviour change and the commit says so rather than claiming none: 19 existing
+`Subject.subject_where/2` call sites and 7 former `subject_filter/2` ones move
+from `FunctionClauseError` to `ArgumentError`. Nothing in `test/` asserted
+`FunctionClauseError` on these functions, and the two pins that already
+existed (`scrollback_test.exs`, `~r/unknown subject:/`) stay green untouched —
+which is the evidence the fold is transparent at the Scrollback boundary.
+
+### What the issue text claimed and the code did not
+
+Two figures in the issue body do not reproduce, and the code wins:
+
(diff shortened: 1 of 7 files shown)


vjt:

Issue #1080, opened 2026-08-08
e2e: issue168-scroll-authority.spec.ts:169 settles at 1417 (expected <= 50), the same number twice, independent of the #1078 cascade


vjt:
Observed on the dev host, 2026-08-08, across six full `scripts/integration.sh` runs
(659–660 tests, 1 worker, chromium, deterministic order).

`cicchetto/e2e/tests/issue168-scroll-authority.spec.ts:169` fails its settled-offset assertion with
**`Expected <= 50, Received 1417`** — red in 2 of 6 runs, and **the number is identical both times**.

## It is independent of the #1078 cascade — proved by removal

| run | tree | cascade poisoner | issue168 |
|---|---|---|---|
| (first) | head | @ #91 | green @ #114 |
| A1 | base | @ #380 | green @ #113 |
| A2 | base | @ #380 | green @ #113 |
| B | head | @ #91 | **RED @ #114** |
| C | head | @ #380 | green @ #113 |
| D | head | **absent (spec deleted)** | **RED @ #113** |

* Red **with** the poisoner upstream (B) **and with it gone** (D) ⇒ not a victim of that cascade;
  #1078's model explains the `issue625` / `p0b` pair and does not extend to this spec.
* Green twice in the very same head tree at the same position ⇒ not deterministic on fixed content.
* Both reds report `1417` byte-for-byte ⇒ the intermittency is in *whether* it trips, not in *what it
  measures*.

## Named suspect — HYPOTHESIS, unmeasured

Same as the sibling residual: `cicchetto/src/ScrollbackPane.tsx:1851`, the #778 ResizeObserver arm
that re-pins to the tail on a container height change. A late mount landing inside the assertion
window fits the shape. **Not measured — do not act on it without a test.**

## What would separate it from host noise (not done)

`--repeat-each` **inside a full-suite context** (isolation never reproduces it), plus instrumentation
of the ResizeObserver arm rather than inference from the failure text.

vjt:
Third local observation, different worker, different branch (2026-08-08, same host): red once inside a full-suite run, green in isolation on the same tree. Reported as **not reproduced**, explicitly not called a flake. Consistent with the pattern above — it trips under suite context and clears alone, which is exactly the environment isolation cannot reproduce.

vjt:
Third independent sighting, same number — and it came with a clean displacement, from #1089/PR
#1092 work that has nothing to do with this spec.

[...]

PR #1458, opened 2026-08-17
e2e(#1336): make the cursor spec's gestures prove they landed

vjt:
Slice of #1336 (*the e2e suite's own instruments are not deterministic*),
sub-shape 1: **a gate satisfied by the pre-existing state is not a gate.**

`cursor-forward-only.spec.ts` drove nine wheel gestures through two
hand-rolled `mouse.move` + `mouse.wheel` helpers, each caller then sleeping a
fixed `SETTLE_WAIT_MS` (500 ms of product debounce + 500 ms of slack) and
reading the cursor ONCE, with no assertion that retries. `page.mouse.wheel()`
resolves before the scroll is applied — ~250 ms, measured on #1080 — so half
that slack was already owed to a dispatch nobody waited for.

## What was measured, and against which base

Everything below ran on this branch, chromium, `--grep "forward-only
contract"`, 6 tests, and was **re-verified after the rebase onto
`f3c37191`**. Every run's #1429 log census was clean on all 12 services
(`db30=0`, `idle30=0`, `dropped=0`, `saturated=0`); the only non-zero signal
is a build gap on the `cicchetto-build-test` oneshot, which is a build and
not a service stall.

**Before touching anything — delete one gesture at a time:**

| mutant | kills | what survived |
|---|---|---|
| `scrollByPx` deleted outright | **2 of 6** | only the two STRICT inequalities notice |
| `scrollToBottom` deleted outright | **0 of 6** | nobody was watching it |

The three survivors of the first mutant are an equality assertion, a
no-retreat assertion, and the audit-"strengthened" disjunction
`inVisibleBand || stayedAtOrBelowPrior` — all three accept "nothing moved".
And *"scroll back to bottom advances cursor to tail"* passes with the
scroll-back-to-bottom deleted, because the cursor is forward-only and the
preceding scroll-up cannot retreat it off the tail it already occupies: the
contract under test is what makes its own verification vacuous.

**The attribution — one displacement (remove the hover, so the wheel is
dispatched away from the pane), run on both sides of the change:**

| | reds | what the red says |
|---|---|---|
| before | **2 of 6** | `expect(701).toBeLessThan(701)` — accuses the product |
| after | **6 of 6** | `never moved the pane (scrollTop stayed 4029)` — accuses the gesture, by name |

**Gates:** converted spec **6/6 green in 29.1 s** (baseline was 32.4 s — a
condition-wait that resolves on arrival costs nothing when the gesture
lands). `scripts/bun.sh run check` **RC=0, 59 warnings + 6 infos**, re-run
after the rebase and identical to main's baseline.
`scripts/design-notes-gate.sh` **RC=0**, also re-run after the rebase.

## What this does NOT do

- It does **not** repair those four assertions. The disjunction still accepts
  "nothing moved", and `scrollToBottom`'s effect is still asserted by no one.
  One of the two holes is closed — being green about a gesture that never
  happened — and the other is left open, named rather than papered over.
- A prediction of mine died here: I expected the converted `scrollToBottom`
  to reject as "already clamped" and thereby prove itself a no-op. It did
  not reject. The pane does move; the gesture is **real and merely
  unobserved**, which is a narrower claim than the one I set out to make.
- No claim that this fixes a CI red. The natural failure was never
  reproduced, and the full `scripts/integration.sh` has not been run on this
  branch — the scoped `--grep` above is an iso-rerun, not a ship gate.
- The five other spec files still driving a bare `mouse.wheel` are untouched.
  One of them, `issue230-wheel-underfill-loadmore`, is not named in any
  report filed under the epic.

### DIFF
--- a/cicchetto/e2e/fixtures/cicchettoPage.ts
+++ b/cicchetto/e2e/fixtures/cicchettoPage.ts
@@ -1386,7 +1386,7 @@ export async function inlineNickColorVar(locator: Locator): Promise<string> {
   return match[0];
 }
 
-// #1336 — page the scrollback UP with a wheel gesture that has demonstrably
+// #1336 — move the scrollback with a wheel gesture that has demonstrably
 // LANDED before the caller moves on.
 //
 // `page.mouse.wheel()` resolves before the scroll is applied (measured: the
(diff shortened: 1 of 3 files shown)


vjt:

PR #1459, opened 2026-08-17
persistence(#1420): measure the write-lock holder, not only its victims

vjt:
## Why

Six CI runs stalled for exactly 30.1 s each. The census could name the
victims every time and the cause not once — and that is not bad luck, it is a
property of the instruments.

**Every DB signal grappa had was completion-driven, so all of them measured
the victim by construction:**

- `[:grappa, :repo, :query]` fires when a query FINISHES. A process that opens
  `BEGIN IMMEDIATE` and then sits still emits nothing at all while it sits;
  the only rows reaching the log are the 30-second `begin`s and `SELECT`s of
  everybody queued behind it.
- `BusyRetry`'s `:on_contention` hook lives inside a `rescue`, so it counts
  whoever caught the exception.
- The #1429 log census greps for the silence after the fact.

One blind spot, three times: **the holder is the only process in the system
that is not finishing anything, and finishes were all we measured.**

## What this is

`Grappa.Repo.LockWatch` reads at the seam instead of at the completion.
`Repo.immediate_transaction/1` is the sole producer of `BEGIN IMMEDIATE`, and
the ordering is **read in the dependencies, not assumed**: DBConnection enters
`run_transaction/5` only on `{:ok, _}` from `begin/3`
(`db_connection.ex:1103-1107`) and Exqlite emits `BEGIN IMMEDIATE TRANSACTION`
for `mode: :immediate` on an idle connection (`connection.ex:307`). So the
transaction fun's first statement is the exact acquisition edge: before it a
WAITER, after it the HOLDER.

A stall is reported only when a holder is past the threshold **with a queue
behind it** — a slow uncontended write is not a stall. The holder's live
`current_stacktrace` is sampled: a `Logger` frame means blocked on logging, a
`DBConnection` frame means a pool-topology deadlock, anything else is a third
answer nobody predicted. That distinction is the datum #1420 records as
missing ("separating the two needs a running stack").

Transaction semantics are untouched: `observe/1` is `try/after` around
`transaction/2`, passing the return value and any raise or `rollback/1` throw
through unchanged. The seam self-disables when its ETS table is absent, so a
watchdog restart can never abort a caller's write.

## What this is NOT

**Not the cure.** Narrowing `BEGIN IMMEDIATE` trades away what #1374/#1375
were written to buy. This turns that decision from a guess into a reading and
leaves the decision where it belongs.

## Cost

**Seam, resolvable:** 355.8 ns/call armed vs 95.5 ns/call disarmed, block
ranges non-overlapping — **260.3 ns per write transaction**. The disarmed
**micro-cost is ~95 ns**: the `persistent_term` read, the `ets.whereis`, and
the unconditional depth bookkeeping that stops the nesting counter leaking
across a watchdog restart.

**End-to-end: NOT RESOLVABLE BY THIS METHOD.** Observed delta **2.124 %**
against a **within-regime spread of 3.8 %** — the spread is larger than the
signal and the blocks overlap. The clause was fixed before the numbers were
seen and is not adjusted after: 2.124 % is **not a measurement**, and it is
above the 1 % threshold declared in advance. Composing the two resolvable
quantities (260.3 ns on a 47.15 us transaction) gives ~0.55 %, under
threshold — but that is **DERIVED, not measured end-to-end**, and it is
host-specific.

## Switch

Default ON (`config/config.exs`); OFF only under `:test`, where the Sandbox's
`pool_size: 1` makes every write transaction a holder and the output would be
noise. CI is covered — the integration stack runs `MIX_ENV: dev`
(`cicchetto/e2e/compose.yaml`, service `grappa-test`).

**Armed is not observed:** `integration.sh` discards container logs on a GREEN
run (#1429), so a stall in a passing run is written and then thrown away. The
#1420 question about green runs stays open, and this change cannot close it.

## The red that buys it

Real `SQLITE_BUSY` contention, not injected: a private `TmpRepo`
(`pool_size: 2`) with one process parked inside a write transaction and a
second genuinely blocked in its `BEGIN IMMEDIATE`. Six mutants, all confirmed
as ARRIVED (`Compiling 2 files` in each log), all killed:

| mutant | predicted | measured |
|---|---|---|
| roles swapped in `detect/2` | test 1 | test 1 |
| `acquired` made a no-op | all 4 | all 4 |
| `acquired` called outside the transaction body | test 1 + barriers | tests 1, 2 |
| emit without checking for a queue | test 3 | test 3 |
| stack sampled from `self()` | test 1 | test 1 |
| threshold comparison dropped | test 2 | test 2 |

The third is the one that matters: it validates that **the edge order IS the
measurement**. Only the per-test attribution was recorded, not which
assertion fired.

## Gates

`scripts/check.sh` **RC=0** (read from a file, not from a wrapper's exit
code): Credo `found no issues`, Dialyzer `Total errors: 0`, Sobelow, doctor
passed, **8 doctests / 61 properties / 6462 tests / 0 failures**,
`deps.audit` `No vulnerabilities found`, bats **540 ok / 0 not ok**.

Four rounds were needed because `check.sh` short-circuits; every red was
mine, cross-checked against the diff (Credo, an architecture test refusing an
unregistered long-lived GenServer, and two Dialyzer errors where my own
`scan/1` and `detect/2` specs contradicted each other).

## What I do NOT claim

- `scripts/integration.sh` and e2e: **never run** on this branch.
- The end-to-end cost: **not measured** (see above).
- `check.sh` went green on base `60657249`; `origin/main` has since moved to
  `e1892e80`, so **this branch is not fast-forwardable** and the gate predates
  the current head.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -21,6 +21,7 @@ Top-level supervision tree:
 ```
 Grappa.Application
 ├── Grappa.Vault                       (Cloak — encrypts at-rest creds; before Repo)
+├── Grappa.Repo.LockWatch              (#1420 write-lock holder/waiter observer; before Repo — owns the ETS table the BEGIN IMMEDIATE seam writes to)
 ├── Grappa.Repo                        (Ecto + sqlite)
 ├── Phoenix.PubSub                     (name: Grappa.PubSub)
 ├── Registry                           (name: Grappa.SessionRegistry)
@@ -52,6 +53,9 @@ Grappa.Application
 
 Child order is load-bearing — see `lib/grappa/application.ex` for the
 why-comment per child. Vault before Repo (Cloak schema callbacks);
+LockWatch before Repo (#1420: it owns the ETS table
+`Repo.immediate_transaction/1` writes on every write transaction, and the
+seam self-disables while the table is absent);
 Backoff/WSPresence/NetworkCircuit before SessionSupervisor (ETS
 tables read directly from `Session.Server`'s start path); Bootstrap
 LAST (depends on Registry + SessionSupervisor existing). `Grappa.SpawnOrchestrator`
--- a/config/config.exs
+++ b/config/config.exs
@@ -121,6 +121,26 @@ config :grappa, :session, connection_stable_ms: 60_000
 #   * backoff_ms — base per-attempt linear backoff (× attempt).
 #   * backoff_cap_ms — ceiling per sleep so late attempts don't stretch a
 #     single wait past a fifth of a second.
+# #1420 — the write-lock holder/waiter observer (`Grappa.Repo.LockWatch`).
+# Every DB signal that predates it is completion-driven and therefore measures
+# the VICTIM: Ecto's per-query event fires when a query FINISHES, so a process
+# holding `RESERVED` while it sits idle emits nothing, and only the 30.1s
+# `busy_timeout` of everybody queued behind it reaches the log. This observer
+# takes the reading at the `BEGIN IMMEDIATE` seam instead, which separates the
+# HOLDER from the WAITERS, and samples the holder's live stack when it stalls.
+#
+#   * enabled — arms both the watchdog timer and the write-path seam. OFF
+#     costs one :persistent_term read per write transaction and nothing more.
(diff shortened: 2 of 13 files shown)


vjt:

PR #1460, opened 2026-08-17
cic(#1402): fold the duplicated window-state sets, and measure what that does not buy

vjt:
First slice of bucket M / A3. It deletes the window-state copies the finding
names, and it does NOT deliver the detection everyone — the issue, the code
comment, and my own recon — assumed those copies were costing us. That second
half is measured, not conceded, and is the reason to read this body before the
diff.

## What the slice removes

| site on `2617474b` | before | after |
|---|---|---|
| `ComposeBox.tsx:112` | `const NOT_JOINED_STATES = new Set(["failed", "kicked", "parked"])` | imports the one definition |
| `Sidebar.tsx:93` | byte-identical to the above | imports the one definition |
| `pseudoChannels.ts:63` | `state: "pending" \| "failed" \| "kicked" \| "parked"` | `Exclude<WindowState, "invited" \| "joined">` |
| `pseudoChannels.ts:86` | `state: state as PseudoRow["state"]` | `state` |

The two `Set`s were `Set<string>`, which accepts any string: two copies was the
A3 finding, but the missing TYPE is what made them worth touching. The cast
removal is not cosmetic either — the two `continue` guards above the push
already narrow the union, so the cast was not carrying the assignment, only
silencing the compiler.

## The detection half is NOT bought, and that is deliberate

A seventh state (`"locked"`) was injected into the emitted
`SESSION_WINDOW_STATE_WINDOW_STATE` const, and `tsc --noEmit` run on both
sides:

| mutant | pre-fix | post-fix |
|---|---|---|
| **7th state ADDED** | rc=0, 0 errors | **rc=0, 0 errors** — unchanged, not bought |
| **`"parked"` REMOVED** | rc=0, 0 errors | **rc=1, `TS2769` at `windowStateSets.ts:26`** — bought, by name |
| control (deliberate required field on `PseudoRow`) | — | rc=1, `TS2741` — instrument proved able to fail |

So the slice detects a state REMOVED or RENAMED upstream and is blind to one
ADDED. The reason is structural: `Exclude<T, …>` is defined relative to `T` and
grows with it, and a `Set` need not be exhaustive. Both are "cannot drift"
constructs; neither is a "must be handled" construct.

**Blind is the documented behaviour, not a gap left open by laziness.**
CLAUDE.md's window-state invariant closes with "New states automatically
inherit synthetic-row + greyed-class treatment as long as they land in
`windowStateByChannel`". An exhaustiveness assertion against the emitted const
would catch the addition and would also change what that invariant promises,
which is a ruling rather than a slice. It is deliberately absent here.

For the same reason the sentence at `windowState.ts:37-39` — "a seventh state
arrives here … and tsc reports every site that does not handle it", written by
#1406 X-S8 — is knowingly left standing even though this PR measures it false.
Correcting it presumes the ruling it depends on.

## Where the constant lives, and why not where it belongs

`NOT_JOINED_STATES` is in a new `lib/windowStateSets.ts`, not beside
`WindowState` in `windowState.ts`. The natural home was tried first and failed
24 tests: `windowState.ts` builds a Solid module root at import time via
`selection.ts`, and Sidebar, ComposeBox and Shell each replace the whole module
with `vi.mock`. Restating the set in three mocks rebuilds the duplication in
test land; `importOriginal` drags the real module's import-time side effects
through those suites' other partial mocks and dies on an unrelated
`../lib/networks` stub. The new module imports `WindowState` as a type, which
is erased, so it has no runtime edge to the store and no suite needs to mock
it. No test file is touched by this PR.

## Gates

| gate | result |
|---|---|
| `scripts/bun.sh run test` | **rc=0 — 289 files, 5579 tests, 0 failures** (unchanged count: pure refactor, existing tests are the characterization) |
| `scripts/bun.sh run check` | **rc=0 — 59 warnings + 6 infos**, identical to the declared baseline (biome green, so tsc ran) |
| `scripts/design-notes-gate.sh` | **rc=0** — marker `<!-- entry #1402 -->` verified absent on the base first, unique after |

Branched from `origin/main` = `2617474b`, re-fetched before pushing and
unmoved, so every gate above ran on the live base. Contribution: 5 files,
100 additions, 8 deletions.

## Not run, and not claimed

- **No new test.** The behaviour is unchanged by construction — a cast and a
  duplicate definition are both erased or inert at runtime — and the 289 files
  already passing are the characterization. A test that pinned the set's
  membership would assert the copy rather than the derivation.
- **No e2e, no `scripts/integration.sh`, no Elixir gate.** Cic-only branch, no
  `lib/grappa/**`, no lane taken.
- **The other three restatements in the A3 list are untouched**: `CLAUDE.md`'s
  prose, the server type (which is the SSOT), and `windowState.ts` itself
  (already derived by #1406). The four gating MEDIUMs of bucket M are untouched.
- **`NETWORK_GREYED_STATES` is left alone deliberately.** It is a second
  byte-identical `Set` duplicated across the same two files, which the issue
  does not name — but it is keyed on `net.connection_state`, a DIFFERENT closed
  set with its own SSOT, so folding it here would have been the boundary
  violation and not the fix.
- **No claim that a seventh state is handled correctly at runtime.** What is
  measured is that nothing reports it. Whether inheriting the greyed treatment
  is right for an unknown future state is the invariant's claim, not this PR's.

Refs #1402.

### DIFF
--- a/cicchetto/src/ComposeBox.tsx
+++ b/cicchetto/src/ComposeBox.tsx
@@ -47,6 +47,7 @@ import {
   uploadState,
 } from "./lib/uploadOrchestrator";
 import { windowStateByChannel } from "./lib/windowState";
+import { NOT_JOINED_STATES } from "./lib/windowStateSets";
 
 // Sticky-bottom compose surface. Reads + writes compose.ts state;
 // dispatches submit on Enter; arrow keys walk per-channel history.
@@ -109,7 +110,6 @@ export type Props = {
 // it is not on screen while composing normally.
 const COUNTDOWN_FROM = 10;
 
-const NOT_JOINED_STATES = new Set(["failed", "kicked", "parked"]);
 const NETWORK_GREYED_STATES = new Set(["parked", "failed"]);
 
 // #356 — how long a green success/notice stays up before self-clearing.
--- a/cicchetto/src/Sidebar.tsx
+++ b/cicchetto/src/Sidebar.tsx
@@ -28,6 +28,7 @@ import {
   SERVER_WINDOW_NAME,
 } from "./lib/windowKinds";
 import { windowStateByChannel } from "./lib/windowState";
+import { NOT_JOINED_STATES } from "./lib/windowStateSets";
 import NickText from "./NickText";
 import WindowBadges from "./WindowBadges";
 
@@ -87,10 +88,6 @@ import WindowBadges from "./WindowBadges";
 //   per-window `:parked` events from `Session.Server.terminate/2`; cic
 //   derives the cascade from the network-level state.
 
-// #902 — `invited` left this set with the pseudo-row itself: an unanswered
-// invite is announced by the top banner now, so no sidebar row carries that
-// state to grey out.
-const NOT_JOINED_STATES = new Set(["failed", "kicked", "parked"]);
 const NETWORK_GREYED_STATES = new Set(["parked", "failed"]);
 
 // #96 — a row's state, spoken. Every non-live sidebar row is rendered muted +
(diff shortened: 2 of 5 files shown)


vjt:

PR #1462, opened 2026-08-17
cic(#1406 X-S1): assert the pin population is complete, not just correct

vjt:
## What

X-S1 reads *"16 of the 37 generated union arms are transcribed by hand with no
drift pin"*. That describes a tree that no longer exists — the earlier #1406
slice already replaced the per-arm pins for the whole `WireSessionEvent`
population with a walk. What survived it is not a missing pin, it is a missing
**census**.

Measured at `0a5c32b0`, over the two hand-rolled unions in `api.ts`:

| set | count |
|---|---|
| kinds cic declares by hand (`WireUserEvent` 42 + `WireChannelEvent` 11, 4 dual-declared) | **49** |
| reached by the Session walk | 37 |
| cross-module, carrying a hand-written `_Assert_*` | 11 |
| **cross-module, carrying nothing at all** | **1** — `web_session_severed` |

`web_session_severed` has been declared by cic since #630 and widened since
#1338 X-S14, and was never pinned. It was found by subtracting sets, not by
reading the file — which is the finding: each of the eleven asserts was
correct, and none of them had anything to say about a twelfth.

## How

The eleven asserts become eleven entries in a `CrossModuleArm` registry, and
the existing walks quantify over `Session ∪ registry`. Naming the counterparts
as DATA is what makes the census expressible at all: `walked ∪ registered` is
only a set if "registered" is a type-level set, and an exported `_Assert_*`
alias is not one.

Three new asserts:

- **`_Assert_NoUnpinnedHandArm`** — the completeness check: the only assert
  whose subject is the other asserts rather than a shape. Read for exactly
  what it buys and no more — see Scope below.
- **`_Assert_NoStrayPin`** — a registry key cic no longer declares, or one the
  Session walk already covers (which would resolve to two candidates and fail
  obscurely instead of naming the overlap).
- **`_Assert_NoUnresolvedPin`** — an entry whose generated type no longer
  carries the kind it is filed under, i.e. a server-side discriminator rename.
  Subsumes `_Assert_BundleHashKind`, whose only content was that same check.

Two constraints were discovered rather than chosen. `bundle_hash` is a
CONVERSION, not a widening (`version: string | null` against `version?:
string`; `undefined` is not in `string | null`), so it is declared
discriminator-only rather than forced into a bound it would correctly overrun.
And the widening registry was scoped to `WireSessionEvent["kind"]`, so the one
arm whose widening was argued in prose was the one arm that could not be
declared in types — widening that scope is what lets
`web_session_severed.code` be both exempted and *bounded*.

Counterpart resolution is `Extract` over both sources (one distributive
conditional the walks can index) and the registry stores the TYPE, because two
generated types carry `kind: "web_session_severed"` — the admin-topic one with
five fields, the user-topic one with one.

## Evidence

Baseline `tsc --noEmit` clean on `0a5c32b0`. The completeness assert was
watched fail before it was allowed to pass:

```
src/lib/wireTypesAssert.ts(388,3): error TS2344: Type '{ web_session_severed:
  "cic-declared kinds no assert in this file reaches"; }' does not satisfy the
  constraint 'true'.
```

Five mutants, five killed, each naming the intended assert:

| mutant | assert killed |
|---|---|
| registry entry removed | `_Assert_NoUnpinnedHandArm` |
| registry key shadows a Session arm (`joined`) | `_Assert_NoStrayPin` |
| exemption points at the wrong field | `_Assert_NoWideningOverrunUser` |
| a cic field retyped (`archive_purged.target`) | `_Assert_NoUserArmDrift` |
| widening exemption removed | `_Assert_NoUserArmDrift` |

The fourth is the proof the walk really replaced the eleven deleted pins; the
fifth is the proof the widened arm is compared rather than skipped. Two mutants
also produced collateral errors that are not assert kills and are reported as
such: the first an unused-import error caused by the mutation itself, the
fourth two genuine `userTopic.ts` consumer errors.

Gates: `bun run check` RC=0 (biome + tsc src + tsc e2e), `bun run test` RC=0
(289 test files), `scripts/design-notes-gate.sh` RC=0.

## Scope

cic + docs only — no `lib/` file, so no server behaviour and no runtime code
path moves. This does **not** remove the hand-maintained line: a new
cross-module arm still needs a registry entry, and the change is that
forgetting one is now a `tsc` error naming the arm instead of silence.
Deriving the entry needs a topic axis in the codegen, which stays open.

### DIFF
--- a/cicchetto/src/lib/wireTypesAssert.ts
+++ b/cicchetto/src/lib/wireTypesAssert.ts
@@ -230,8 +230,12 @@ type CrossModuleArm = {
 //     topics, which pins the two copies to each other as a side effect.
 //   * `_Assert_NoUnpinnedHandArm` — a kind cic declares that NOTHING here
 //     checks: absent from the Session union and absent from the registry.
-//     This is the one that closes X-S1, because it is the only assert whose
-//     subject is the COMPLETENESS of the others rather than a shape.
+//     It is the only assert whose subject is the COMPLETENESS of the others
+//     rather than a shape. Read it for exactly what it buys and no more: it
+//     does NOT remove the hand-maintained line — a new cross-module arm
+//     still needs its registry entry — it makes FORGETTING one a `tsc`
+//     error that NAMES the arm instead of silence. Deriving the entry
+//     needs a topic axis in the codegen, which stays open.
 //   * `_Assert_NoStrayPin` / `_Assert_NoUnresolvedPin` — the registry's own
 //     two rot modes. A key cic no longer declares (or one the Session walk
 //     already covers, which would silently shadow it) makes its entry
--- a/cicchetto/src/lib/wireTypesAssert.ts
+++ b/cicchetto/src/lib/wireTypesAssert.ts
@@ -36,10 +36,11 @@
 //   * Add an assert for every api.ts type that has a wireTypes.ts
 //     counterpart. When server-side adds a new Wire module + type,
 //     the codegen emits it; if a cic consumer needs the new shape,
-//     add the assert + the api.ts mirror. EXCEPT for arms of
-//     `WireSessionEvent`: #1406 walks that population instead, so a new
-//     Session arm needs no line here and gets none of the silence that
-//     a forgotten line used to buy.
+//     add the assert + the api.ts mirror. EXCEPT for UNION ARMS, which
+//     are walked rather than listed (#1406): a `WireSessionEvent` arm
+//     needs nothing at all, and a cross-module arm needs one line in the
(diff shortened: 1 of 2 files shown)


vjt:
Landed on main as `e321fba5` (fast-forward from `0a5c32b0`, three commits, numstat identical on both sides: 147/87 in `wireTypesAssert.ts` + 94/0 in `DESIGN_NOTES.md`). #1406 stays open — X-S1 is the only slice this carries; X-S4/X-S8/X-S9/X-S10 remain.

Issue #1390, opened 2026-08-16
arch: Session.Server is 7,230 lines with two effect vocabularies and a 32-field contract nothing enforces (bucket A session-decomposition)


vjt:
Bucket **A session-decomposition** of the 2026-08-15 architecture review (`docs/reviews/architecture/2026-08-15-architecture-review.md`). Three HIGH findings that are one shape seen from three angles, plus two gating MEDIUMs.

## What was measured

**Abstraction A1 (HIGH).** `EventRouter`'s declared input contract (`lib/grappa/session/event_router.ex:101-107`) names 4 required keys plus `optional(any()) => any()`, while the module reads or writes **32** distinct fields of `Session.Server`'s state (`server.ex:561-956`). The catch-all makes Dialyzer accept all of them. The sibling module solves the same problem the opposite way: `NumericRouter` takes a narrow projected view, 4 required keys built by an explicit constructor (`numeric_router.ex:171-176`, `:853`).

**Responsibility A1 (HIGH).** `Session.Server` is 7,230 lines, 83 state fields, 108 callback clauses, 219 private functions, 34 `apply_effects/2` clauses. The three early extractions (`AuthFSM`, `EventRouter`, `NumericRouter`) moved behaviour out and share one interpreter. Every extraction after them moved only data or a pure predicate and left a bespoke interpreter inline: `GhostRecovery.step/2` and `RecoverIdentity.step/2` both return `{verdict, next, lines}` — a second, private effect vocabulary `apply_effects/2` cannot see — and the same shape recurs for away, presence and perform/identify. Each owns its own timer field: **eight** timer refs on one state map, cancelled by hand at each terminal (`server.ex:4585` documents a release wired to one of two terminals).

**Extension A4 (HIGH).** `Session.Server` went 4,701 → 7,230 lines and its test 7,844 → 12,491 in 26 days; `handle_call` 52 → 66, `handle_info` 29 → 40, state 87 keys.

Two premises of the 2026-07-20 review were **retracted, not restated**: the `:sys.get_state` pin tax is gone (98 → 7 occurrences) and hub concentration is flat (server hubs 20.8% → 20.0% of `lib`, client 20.8% → 17.3%). The hubs grow with the codebase, not faster than it. What is real is absolute width.

## Gating MEDIUMs in this bucket

[...]

PR #1464, opened 2026-08-17
refactor(#1390): fold the ten injected session callbacks into one Deps struct

vjt:
First slice of #1390. `Grappa.Session.Server` carried ten injected
callbacks as ten top-level state keys. They become one `deps` field
holding a `Grappa.Session.Deps` struct, in a new
`lib/grappa/session/deps.ex`. Session state goes from **85 to 76** keys.

`start_link/1` does not change: the ten opt keys are byte-identical, and
that is what keeps eight other files out of this slice.

The ten: `visitor_committer`, `visitor_password_rotator`,
`visitor_nick_persister`, `credential_failer`, `credential_committer`,
`registration_committer`, `last_joined_persister`, `recover_source`,
`away_persister`, `query_window_open?`.

## Three things the diff does not explain about itself

1. **`EventRouter` matches a PLAIN map, not `%Deps{}`.** `Deps` already
   names `EventRouter.query_window_open?()` as a type, so matching the
   struct would close a compile/export cycle. A struct is a map, so the
   plain match still works.

2. **The absent branch holds `false`, NOT `%Deps{}`.** The struct's
   default for `query_window_open?` is `&QueryWindows.open?/3` — a real
   DB read. Defaulting to the struct would cost `EventRouter` its "no
   Repo" purity in every test that omits `deps`.

3. **Two of the twelve call sites were `Map.get(state, key, default)`**,
   deliberately tolerant of a process that predates the field (hot
   reload). Bundling had hardened them, and the #581 test caught it.
   They are now `Map.get(state, :deps, %Deps{})` — an exact equivalent,
   because the struct's defaults ARE what those two were passing.
   The #581 test was **re-aimed, not weakened**: it simulated staleness
   with `Map.delete(state, :recover_source)`, which after the bundle is
   a **no-op** — it would have passed while exercising nothing. It now
   deletes `:deps`; the assertion is unchanged.

## Three reds, in order

1. **TDD red 1** — the bundle pin, before `Deps` existed.
2. **TDD red 2** — the attributable red: the ten loose keys still present
   at the top level of state.
3. **Dialyzer red, and it was hidden.** `ci.check` chains its steps
   through `mix cmd`, which **halts on the first non-zero**, and
   `credo --strict` sits ahead of `dialyzer`. A Credo red in the test
   file therefore meant Dialyzer never ran at all — so its silence was
   the absence of a measure, not a pass. With Credo fixed, Dialyzer
   produced nine `unknown_type` errors.

## The slice is 7 files, not 6

The original count was six, read off the edited set. It was wrong for a
structural reason worth stating: **moving an `@type` out of a module
breaks every module that names it remotely, and the diff never opens
those.** Nine of the ten types moved from `Session.Server` to
`Session.Deps`, but `Grappa.Session`'s `start_opts/0` still spelled them
`Server.visitor_committer()` and friends — a seventh file,
`lib/grappa/session.ex`, that nothing in the six-file diff touched.

Nine and not ten: `query_window_open?` needs no rename there, because
that map never carried the key — it does not on `main` either. That
asymmetry with `Server.start_opts/0` is pre-existing and deliberately
left alone.

## Gate

`scripts/check.sh` → **RC=0, full chain**:

- `mix compile --warnings-as-errors`, `mix format --check-formatted`
- Credo `--strict`: `6098 mods/funs, found no issues`
- suite: `8 doctests, 61 properties, 6479 tests, 0 failures`
- Dialyzer: `Total errors: 0`
- Sobelow and the bats set: through

## What is NOT measured

- **The style commit was never gated on its own.** The gate ran once
  over the **union** of both fixes; the two were then split into two
  logical commits.
- Behaviour invariance beyond the suite: no mutants on this slice. It is
  behaviour-free by construction and the two pins are structural, so
  mutants were not required — but they were not run.
- **85 → 76** is arithmetic on the typedef. It was not re-read off live
  state after the change.
- Risk A1 (dropping the catch-all makes a typo go red) is **not bought**
  and does not touch this slice. It needs a mutant before slices 3-4 of
  #1390 get sized.
- SEC-28 is out of scope, on a **boundary** argument rather than a cost
  one: its call sites live in `Push` and `WindowCounts`, not `Session`.

## Base

Branched at `0a5c32b0`; `main` has since gained three #1406 commits. They
overlap only in `docs/DESIGN_NOTES.md`, and `git merge-tree` reports the
merge as clean.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -46365,3 +46365,79 @@ error naming the arm instead of silence. Deriving the entry needs the topic
 axis in the codegen, which stays open. Nor does it touch a narrower: no
 runtime behaviour moves, and the 1.150-odd lines of hand narrowing that the
 rest of #1406 is about are untouched here.
+<!-- entry #1390 -->
+
+---
+
+## 2026-08-17 — #1390 slice 1: the ten injected callbacks become one `Deps` struct
+
+Bucket A of the 2026-08-15 architecture review wants `Session.Server`
+decomposed. This is the first slice, and it was picked first because it is the
+only one whose value does not rest on a theory about types: it is
+behaviour-free, the ten fields are contiguous in the typedef, and it takes a
+closure off the `EventRouter` state contract that had no business being read as
+if it were state.
+
+The ten are dependencies, not state. Nothing in a session's lifetime writes
+them — `init/1` reads them out of the resolved plan and every later touch is a
+call. They now live in `Grappa.Session.Deps`, and `Session.Server`'s top-level
+state goes from **85 keys to 76**.
+
+### What did not change, deliberately
+
+`start_link/1`'s keyword API. All ten option keys are spelled exactly as
+before, so no producer of a session plan moved: `Networks.SessionPlan`,
+`Visitors.SessionPlan`, `Bootstrap` and the rest write into `opts`, and
+`Deps.from_opts/1` is the only thing that had to learn the new shape. That is
+what keeps this slice at six files instead of fifteen.
+
+### Why `EventRouter` matches a plain map and not `%Deps{}`
+
+`Deps` already names `EventRouter.query_window_open?()` for one of its field
+types. Pattern-matching `%Deps{}` inside `EventRouter` would put a compile
+dependency back the other way and close a cycle, so the read there is a plain
+`%{query_window_open?: fun}` map pattern — a struct is a map, so the production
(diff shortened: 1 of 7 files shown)


vjt:
Landed on main as `a60082eb` (fast-forward from `d3f40354`, five commits, numstat identical on both sides across all seven files). #1390 stays open — this is slice 1 of the decomposition.

Issue #1137, opened 2026-08-09
e2e: the per-spec CHANNEL slice — long-lived users still accrue ~2.9 rows per test from the shared #bofh


vjt:
Follow-up to #1078. The per-spec SUBJECT namespace (PR #1135) closed
every surface keyed by `user_id`. This is the part it provably did not
close, and the measurement that says so was registered as a prediction
*before* it was taken (commit `9f7b0e24` on `1078-spec-namespace`).

## The residue, measured

Census of the e2e sqlite after a 20-test batch (ten scroll/marker specs,
chromium, 1 worker, 2026-08-09) run on the per-spec-subject mechanism:

```
m9b-test    #bofh    59
m9b-victim  #bofh    60
vjt         #bofh   261      (200 seeded + 61 events)
```

**~+2.9 rows per test, unchanged from before the subject namespace.**

The mechanism is straightforward: each per-test subject JOINs and QUITs
the SHARED `#bofh`, and the long-lived users sitting in that channel
record those events in their OWN scrollback. A subject-level namespace
cannot reach them by construction — `delete_for_channel({:user, s...},
…)` is scoped to the subject that is being torn down, not to the other
occupants of the channel it was in.

For contrast, on the same run: vjt's `$server` stayed at **14** rows
(the boot value) instead of the 294 the old reset would have produced,
and there were **0** orphan messages — the subject-keyed half is shut.

Also open on the same axis, though not measured here: bahamut-side
channel state. `#bofh` never empties (m9b-test and m9b-victim are
permanent occupants), so a topic or mode a spec sets on it persists for
the rest of the run. A per-spec channel is destroyed by the ircd when
its only occupant quits, which removes that class too.

## Why this cannot be a fixture

#1078 suggested "a fixture deriving the name from `testInfo.file` rather
than 134 hand edits". That is not available, and the number is not 134.

**222 spec files capture the channel into a MODULE-SCOPE const:**

```
$ grep -hE "^(const|let|var) [^=]*=.*AUTOJOIN_CHANNELS" tests/*.spec.ts \
    | sort | uniq -c | sort -rn | head -3
 180 const CHANNEL = AUTOJOIN_CHANNELS[0];
  30 const SEED_CHANNEL = AUTOJOIN_CHANNELS[0];
   3 const CHANNEL = AUTOJOIN_CHANNELS[0]; // #bofh — vjt's seeded autojoin channel
```

A module body is evaluated ONCE per worker at import; a Playwright
[...]

Issue #1079, opened 2026-08-08
e2e: scroll-on-window-switch.spec.ts:410 settles at 337 (expected <= 50), the same number twice, on both trees


vjt:
Observed on the dev host, 2026-08-08, across five full `scripts/integration.sh` runs
(660 tests, 1 worker, chromium, deterministic order).

`cicchetto/e2e/tests/scroll-on-window-switch.spec.ts:410` fails its settled-offset assertion at
`spec:512`: **`Expected <= 50, Received 337`** — red in 2 of 5 runs, and **the number is identical
both times**.

## Why this is not filed as a flake

* It trips on **both trees** in the comparison (a feature branch head @ `49804a8e`-side and the
  baseline @ `40c5576b`), so it is not attributable to either diff.
* It is **not** a victim of the run-position cascade filed as #1078: it appears in run A1, where that
  cascade's poisoner sits at #380 and its victim is a different spec.
* **When it trips, it trips on the same wrong offset, byte-for-byte.** The intermittency is in
  *whether* it fires, not in *what it measures*. Jitter does not repeat a number exactly.

## Named suspect — HYPOTHESIS, unmeasured, do not act on it without a test

`cicchetto/src/ScrollbackPane.tsx:1851` — the ResizeObserver arm added by #778, which re-pins to the
tail on a container HEIGHT change (`if (moved && followMode()) applyActivation("tail-only", false)`).
Its own comment describes this shape: a row mounting AFTER the tail write. Under full-suite load that
mount could land inside the assertion window; alone it does not. **Nothing about this was measured.**

## What would separate it from host noise (not done)

Run the victim under `--repeat-each` **inside a full-suite context** — isolation is precisely the
environment that never reproduces — and correlate the trip against the ResizeObserver arm by
**instrumenting it**, not by inference.

Same assertion family and same "identical number every time it trips" signature as the residual filed
alongside this one; they may be two faces of one latent defect, which is itself unproven.

vjt:
Consolidated into #1336 — *the e2e suite's own instruments are not deterministic* — on vjt's call (#grappa, 2026-08-15), following the #653 precedent: one phenomenon, one investigation, instead of nine parallel spec-level patches.

Nothing here is dismissed. The measurement, artifacts and run ids in this issue are the starting point for that work; the epic links back rather than restating them.

*Closed by vjt-claude from vjt's GitHub account, on his instruction.*

PR #1465, opened 2026-08-17
e2e(#1336 S2): settle the pane before measuring it, not just move it off the tail

vjt:
## What #1079 is, measured

#1079 reports `scroll-on-window-switch` settling **337 px** from the bottom
against an expected 50 — the same number twice, on two trees. The shared cause
with #1080 was established a slice ago (a scrollTop DECREASE after a send
disarms `followMode`, the pane freezes), but that spec has no wheel, so its
**trigger** stayed inference and the previous slice deliberately refused to
cure an inference.

An in-page scroll recorder, four iterations at CPU throttle 1x/6x/12x/20x —
identical shape in all four, only the timings stretch:

| t (1x) | scrollTop | what it is |
|---|---|---|
| 0 ms | 354 | the `$server` pane we switch away from |
| 45 ms | **7** | the rows recreation resetting to the top |
| 59 ms | 1055 | the marker jump, in flight |
| 78 ms | **1078** | the marker, at rest |

Geometry there: `scrollHeight` 1627, `clientHeight` 212, so `maxScroll` 1415.
Two things follow by arithmetic on those measured numbers:

* `1415 − 1078 = **337**` — the reported failure is the pane sitting exactly
  on the marker. Reproduces independently what the previous slice derived.
* `1415 − 7 = **1408**` — and the barrier guarding the send is
  `distance-to-bottom > 50`.

**So the barrier is satisfied by the reset just as readily as by the marker.**
It cannot tell "the switch landed on the marker" from "the pane is at the top
with the marker jump still to come": it says *not at the bottom* while meaning
*the activation has finished*. That is the #1080 shape — a gate satisfied by a
state that is not the one it names — except the state here is **transient**
rather than pre-existing, which is why reading the spec never surfaced it.

## The fix

`waitForScrollRest` — the gesture core with the wheel taken off: hold until
two **adjacent** samples agree, reject if the pane never stops. Unlike
distance, rest cannot be true of a pane mid-jump. Both waits now share one
sampling loop; `requireMove` is the only thing they disagree about, because a
gesture resting back on its baseline was never delivered while a barrier is
satisfied by a pane that was already still.

Applied at all three sites in the spec where an activation is followed by
reads of a position — including the app-startup test, which has no send and so
no freeze to suffer, because its marker-offset assertions are measurements of
a position just the same.

## The instrument had the disease it was built to cure

The first version put the `pollMs` sleep at the **end** of the loop body, so
the first comparison — baseline against first sample — was separated by one
round trip instead of a poll. A pane read twice in the same millisecond reads
equal whatever it is doing, so the barrier resolved *instantly* against a pane
an injected `setInterval` was moving 1 px every 20 ms.

The unit fakes could not catch it — they answer from a script, so "no gap" and
"a poll apart" are the same sequence. The regression test answers from a
**clock** instead, and it is the only one of the eleven cases the
sleep-position mutant kills.

## Evidence

Two-sided displacement (`setInterval` writing the pane every 20 ms), same test:

| | what the red says |
|---|---|
| barrier removed | `expect(received).toBeLessThanOrEqual(expected)` — accuses the **product** |
| barrier present | `waitForScrollRest: the pane never came to rest within 10000ms (last 576)` — accuses the barrier's subject, **by name** |

`576` is itself attributable: `1078 − 10000/20 = 578`.

An earlier attempt at that displacement was silently **inert** — injected at
the first of three call sites while the run was scoped to the third — and both
sides passed identically. Recorded because it is the same inert injection the
gesture helper's "never moved" reject exists for.

**Mutants: five declared, five killed, none survived.**

| mutant | killed |
|---|---|
| gap moved back to the end of the body | **1** — only the clock case, which is the proof it is the only test that can see it |
| rest compared against baseline, not previous | 5 (incl. 3 `scrollByGesture` — the sampler really is shared) |
| adjacency dropped | 7 |
| move requirement ignored | 1 |
| never-rest resolves instead of rejecting | 2 |

Gates: `bun run test` 289 files / 5589 tests RC=0 — the five pre-existing
`scrollByGesture` cases stayed green **through** the re-expression on the
shared sampler, which is the regression oracle for the refactor.
`bun run check` RC=0. Target spec on the stack: 12 passed (4 tests ×
`--repeat-each 3`). `design-notes-gate.sh` RC=0.

* **The natural failure is still not reproduced.** In eight recorded samples
  the barrier exited at the marker (1078) every time, never at the reset. The
  vacuity is proven — the predicate demonstrably admits a state that is not
  the one it names, and that state demonstrably occurs 45 ms into every switch
  — but its **exploitation** is unobserved. This does not claim to be the
  mechanism behind #1079's 337.
* **CPU throttling cannot arbitrate it, measured.** The margin between the
  marker landing and the barrier clearing *grows* with the rate (105 ms at 1x,
  167 ms at 20x) because the barrier polls through the page too. A late
  network-timed rows recreation (the "307 race" the spec's own comments name)
  remains the uninstrumented candidate.
* **The post-send decrease at throttled rates is not the freeze** — it is
  `1431 → 1415`, the tail-follow overshooting and clamping to `maxScroll`.
* **The five specs with a raw `mouse.wheel`** (`cp14-b2`, `issue230`,
  `issue535`, `issue887`, `login-advanced-scroll-reachability`) are untouched
  and unmeasured — candidates by shape only.

## Full-suite note, stated rather than buried

A local full `scripts/integration.sh` on this tree came back **755 passed / 3
failed** (`freshness-on-activation`, `issue1229-unread-retention-bound`,
`issue269-visitor-reconnect-toggle`). All three pass in isolation locally
(3/3, 16.3 s), none references any instrument this PR touches, and CI's own
integration was green at this branch's original base. Those are different
hosts, so they do **not** attribute: I have not established the three are
pre-existing, only that they are not reproducible in isolation here and have
no overlap with the changed surface.

### Update — the three reds, measured rather than argued

The full-suite note above is now settled by measurement. Three complete
`scripts/integration.sh` runs on the **same host**, alternating sides:

| run | side | reds |
|---|---|---|
| 1 | this branch (`bad57ae1`) | 3 — `freshness-on-activation:81`, `issue1229:134`, `issue269:54` |
| 2 | clean `main` (`d3f40354`) | 4 — `error-banners:121`, `issue498:208`, `issue947:71`, `nick-follow-query:49` |
| 3 | this branch, rebased (`91c9faad`) | 1 — `issue1041-left-edge-sidebar:167` |

**Eight distinct names across three runs, pairwise overlap zero.** The run that
matters is the third: same side, same regime, **none of the original three
reproduced**. A deterministic breakage caused by this change would have come
back.

Containment is closed by construction, not by argument. Of the 53 exports in
`cicchettoPage.ts` exactly **two** reach the changed code (`pageScrollbackBy`,
plus the new `pageScrollbackRest`). Walking the transitive import closure of
all eight red specs (7–10 modules each) finds **no caller** of any changed
function; the same walker does find the callers in the three specs that do use
them (`scroll-on-window-switch`, `cursor-forward-only`, `issue168-scroll-authority`),
so the negative result discriminates. Those three consumers were green in run 3.

**What this does not establish.** Not that the three reds are pre-existing —
they were never reproduced anywhere, so they are simply **unattributed**. Three
runs exclude a *deterministic* fault in this change; they have essentially no
power against a small probabilistic shift. The mechanism behind the host's
run-to-run randomisation is not identified (fixture leak #1137 and load are
unmeasured candidates), and `issue1041` is unattributed in its own right. A red
count of 1 here against 4 on `main` is **not** evidence this side is better.

Also in this branch: `91c9faad` reformats the clock case the way biome wants it.
That job was red because `check` is `biome && tsc && tsc`, so a formatting error
short-circuited the type checks — meaning tsc had never actually run on the file.
Both `tsc` invocations were therefore re-run **standalone** (rc 0 each), plus
`check` (rc 0) and `test` (289 files / 5589 tests, rc 0).

### DIFF
--- a/cicchetto/e2e/fixtures/cicchettoPage.ts
+++ b/cicchetto/e2e/fixtures/cicchettoPage.ts
@@ -62,7 +62,7 @@
 
 import { expect, type Locator, type Page } from "@playwright/test";
 import type { SeededUser } from "./grappaApi";
-import { type ScrollGestureResult, scrollByGesture } from "./scrollGesture";
+import { type ScrollGestureResult, scrollByGesture, waitForScrollRest } from "./scrollGesture";
 
 const SHELL_READY_TIMEOUT_MS = 10_000;
 const MOBILE_BREAKPOINT_PX = 768;
@@ -1425,4 +1425,27 @@ export async function pageScrollbackUp(
   timeoutMs: number,
 ): Promise<ScrollGestureResult> {
   return await pageScrollbackBy(page, -pixels, timeoutMs);
+}
+
+// #1336 S2 — the same wait with no gesture in front of it: hold until the
+// pane STOPS being written to, and report where it stopped.
+//
+// For a spec that parks on the unread marker through a PROGRAMMATIC
+// activation rather than a wheel. Recorded in-page, such a switch writes
+// `scrollTop` three times — the rows recreation resetting to the top, the
+// marker jump in flight, the marker — and a "distance from the bottom is
+// large" barrier is satisfied by the FIRST of the three, which is the pane on
+// its way somewhere rather than the pane where the test means it to be.
+//
+// Sampling every 50 ms, matching `pageScrollbackBy`: two agreeing samples mean
+// the pane held still for 50 ms.
+export async function pageScrollbackRest(page: Page, timeoutMs: number): Promise<number> {
+  const pane = page.locator('[data-testid="scrollback"]');
+  return await waitForScrollRest(
+    { scrollTop: () => pane.evaluate((el) => el.scrollTop) },
+    {
+      timeoutMs,
+      pollMs: 50,
+    },
+  );
 }
--- a/cicchetto/e2e/fixtures/scrollGesture.test.ts
+++ b/cicchetto/e2e/fixtures/scrollGesture.test.ts
@@ -12,7 +12,7 @@
 // inside the next step.
 
 import { describe, expect, it } from "vitest";
-import { type ScrollPane, scrollByGesture } from "./scrollGesture";
+import { type ScrollPane, scrollByGesture, waitForScrollRest } from "./scrollGesture";
 
 type Recorded = { calls: string[]; pane: ScrollPane };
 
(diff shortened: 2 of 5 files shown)


vjt:

Closing by content, not by ancestry: the union replayed each PR's own commits onto a branch cut from current `main`, so the SHAs were rewritten and GitHub cannot mark this PR merged by itself. Verified landed: taking this branch's merge-base (`0330053c`), diffing the files it touches against `main` leaves **zero** lines that this branch has and `main` lacks.

Why the union rather than three merges: all three PRs were green against the same stale base and none had ever been gated *with the others*. #1471 is the bench that **measures** the guard #1472 changes — merged separately, both would have been green and neither would have asked whether the bench still told the truth after the change. One branch, one gate, one merge.

PR #1467, opened 2026-08-17
refactor(#1390): move the channel-directory ETL out of the session GenServer

vjt:
Second slice of #1390. The channel directory is a domain of its own —
`Grappa.ChannelDirectory` has been its own context all along — yet its
whole ETL lived inside the session GenServer: three compile-env constants,
four state fields, a watchdog, and seven private functions parsing
RPL_LIST rows, batching them, throttling progress pings and flushing the
tail.

The four state keys collapse into one `directory` holding a
`Session.DirectoryIngest`. `run == nil` IS the in-flight guard, exactly as
`directory_refresh == nil` was.

## Why this cluster and not the three the issue suggests

The issue's proposed direction lists identity/recovery, away and
auth-handshake as the continuation. Measured footprints decided otherwise:

| cluster | fields | outside `server.ex` |
|---|---|---|
| away | 3 | already has its own `away_state.ex` |
| identity/recovery | 6 | reaches `session.ex`, `deps.ex`, `grappa_channel.ex` |
| auth-handshake | 4 | `pending_auth` alone: **86** occurrences, 2 modules |
| **directory** | **4** | **comments only**, in `numeric_router.ex` + `wire.ex` |

Those three regroup session-domain fields inside the module that already
owns them: they shrink the state map without moving anything across a
boundary. Only the directory does, and the receiving context already
exists.

## The red that buys it

`directory_ingest_test.exs` runs `async: true` on plain `ExUnit.Case` — no
DataCase, no Repo, no `Session.Server`, no fake ircd — 14 cases in 0.09s.
Its sibling `directory_test.exs` needs all four: 240 lines, `async: false`,
25 references to `start_server`/`IRCServer`, because before this the parse,
batch and throttle decisions were reachable only by booting a GenServer.

That is a falsifiable boundary rather than a metric. If a later change
re-entangles the decisions with the process, the file stops being writable
that way.

## The structural pins are not vacuous, and that was measured

The two `server_test.exs` pins were written AFTER the implementation, so
they had never been seen red. Two surgical mutants fixed that, and they are
disjoint — **one mutant, one assertion**:

- a plain map in place of the struct kills only *"a live session carries the
  directory ingest in one struct, idle"*;
- one re-added legacy key kills only *"the four directory names are gone
  from the top level of state"*.

## A pin was re-pointed, not deleted

`directory_test.exs:105` was `assert :sys.get_state(pid).directory_refresh
== nil` — a `:sys.get_state` pin on the NAME of a private field, not an
assertion of behaviour. Its behavioural twin three lines above (the
`assert_receive` of the `directory_failed` broadcast with `reason:
"timeout"`) passed throughout, so the observable never moved.

It now reads `.directory.run`, in its own commit (`60215800`) so a reviewer
can tell a moved pin from a softened assert. Deleting it on the grounds that
the behavioural test covers the same case was the tempting move; that
reasoning is how pins disappear one at a time.

This also retracts a commitment made when the slice was proposed — that
having to touch `directory_test.exs` would BE the behaviour change. The
dichotomy had no room for a structural pin, which is neither an assert nor a
behaviour.

## Two behaviours preserved on purpose

Both were named as risks before any code was written, and both look like
untidiness:

- the `directory_complete` total is re-read from the DB snapshot with
  `ttl_ms: 0`, **not** taken from the accumulator's running `count`. The two
  diverge the moment `ChannelDirectory.ingest/3` dedupes or upserts, so
  collapsing them would change the number cic renders.
- the watchdog calls `abort/1`, which **drops** the buffered rows and never
  finalises. A truncated refresh leaves the prior snapshot intact and loses
  everything buffered since the last batch. Flushing on timeout would change
  the DB on every stalled refresh.

`flush_directory_buffer/2` carried a `:batch | :final` discriminator in its
`@spec` that BOTH clauses ignored. It is not carried into the new module —
dead in, dead out — and it is recorded rather than left to be rediscovered
as an omission.

## Struct, not a declared map type

Measured rather than assumed: #1391's mutant pair established that a
struct-field typo is a *compile* error while a bare-map key typo compiles
clean, and that declaring a map's shape buys neither. The tracker was an
anonymous map with `buffer: [map()]`.

## Scope: 8 commits, 7 files — one file over the declared count

The declared perimeter was **5 files (+1 conditional)**. It came in at
**7**, and both extras are stated as an overrun rather than as foresight:

- `numeric_router.ex` — the conditional became real: its comment described
  the dispatch head by a field that no longer exists, so it read as false
  rather than dated. `wire.ex` was left alone on purpose: it names
  `directory_progress_throttle_ms`, and that spelling survives as the
  start-opt key, so it is still true.
- `directory_test.exs` — the one line of the re-pointed pin.

A sixth stale comment was missed by the first sweep, because that grep
searched `directory_refresh:` **with** the colon and the comment wrote the
identifier bare. Fixed in its own commit rather than amended away.

## Gate

`scripts/check.sh` → **RC=0, full chain**:

- `compile --warnings-as-errors`, `format --check-formatted`
- Credo `--strict`: `6114 mods/funs, found no issues`
- suite: `8 doctests, 61 properties, 6495 tests, 0 failures`
- Dialyzer: `Total errors: 0`
- Sobelow and the bats set: through

Two self-inflicted gate reds on the way, both in code this slice
introduced: 17 Credo issues (variable rebinding in the new test, and
`_name` where the local strategy is a bare `_`), and one
`contract_supertype`. The latter was a correct design signal — it pushed the
construction out of a private `server.ex` helper and into
`DirectoryIngest.from_opts/1`, the exact shape of its slice-1 sibling
`Deps.from_opts/1`.

## What is NOT measured

- **Behaviour-freedom is not proven, only evidenced.** The suite is green
  and `directory_test.exs` needed one line, which is a pin — that is
  evidence, not an equivalence proof.
- **The key count is not the win here**: 4 → 1, against slice 1's 10 → 1.
  The claim is the boundary and the testability, not the arithmetic.
- Whether `ChannelDirectory.ingest/3` actually dedupes was not checked. The
  first preserved risk says `count` and `.total` **can** diverge, not that
  they do today.
- No mutants beyond the two that pinned the structural assertions.
- **No rebase was performed and no union-driver check is claimed.**
  `origin/main` is `a60082eb`, which is this branch's merge-base — slice 1
  landed fast-forward, so a rebase would be a no-op and the `merge=union`
  driver would never run. What was measured is the file state: both
  `<!-- entry #1390 -->` and `<!-- entry #1390 slice 2 -->` appear exactly
  once by `grep -Fxc`, no duplicate markers anywhere, and the entry adds 85
  lines with zero deletions.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -46441,3 +46441,88 @@ _Not measured here: whether the slice is behaviour-free beyond the two test
 files it touches. The 847 tests in `server_test.exs` and `event_router_test.exs`
 are green; the rest of the suite and the type gates are a separate run, and
 this entry does not claim their verdict._
+<!-- entry #1390 slice 2 -->
+
+---
+
+## 2026-08-17 — #1390 slice 2: the channel-directory ETL leaves the session
+
+Slice 1 bundled ten injected callbacks into `Session.Deps` — a co-mutation
+group, valuable by key count. This slice is chosen on a different axis, and
+the axis is the point: of the four clusters the issue's proposed direction
+lists, only this one crosses a DOMAIN boundary. Away, identity/recovery and
+auth-handshake all regroup session-domain fields inside the module that
+already owns them; they shrink the state map without moving anything. The
+channel directory is a foreign domain — `Grappa.ChannelDirectory` has been
+its own context all along — running a full ETL inside the hottest process in
+the tree.
+
+The measured footprint outside `server.ex` decided it. `pending_auth` alone
+appears 86 times across `event_router.ex` and `ns_interceptor.ex`, so the
+auth cluster is not the cheap bundle it reads as; `away_state` already has
+its own module. The four directory keys are named outside `server.ex` only
+in comments, in `numeric_router.ex` and `wire.ex` — nothing reads them.
+
+### What moved, and what the shape buys
+
+Four state keys collapse into one `directory` holding a
+`Session.DirectoryIngest`: the three config tunables plus the in-flight
+tracker, with `run == nil` carrying exactly the guard `directory_refresh ==
+nil` used to. Unlike its `*Accum` siblings, which are pure data drained by
+`EventRouter`, this struct also owns the row parse, the batch boundary and
+the throttle window — because that is what the extraction is FOR.
+
+A struct rather than a declared map type, and that choice is measured rather
+than reasoned: #1391's mutant pair established that a struct-field typo is a
+compile error while a bare-map key typo compiles clean, and that declaring a
+map's shape buys neither. The tracker was an anonymous map with
+`buffer: [map()]`.
+
+### The red that buys it
+
+`directory_ingest_test.exs` runs `async: true` on plain `ExUnit.Case`: no
+`DataCase`, no Repo, no `Session.Server`, no fake ircd, 14 cases in 0.09s.
+Its sibling `directory_test.exs` needs all four — 240 lines, `async: false`,
+25 references to `start_server`/`IRCServer` — because before this the parse,
+batch and throttle decisions were reachable only by booting a GenServer.
+
+That is a falsifiable boundary rather than a metric: if a later change
+re-entangles the decisions with the process, the file stops being writable
+that way. The two `server_test.exs` pins were each killed by exactly one
+surgical mutant (a plain map in place of the struct kills the first; one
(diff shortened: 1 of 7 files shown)


vjt:
Landed on main as `0330053c` (fast-forward from `a60082eb`, eight commits, numstat identical on both sides across all seven files). #1390 stays open — slice 2 of the decomposition.

Issue #1255, opened 2026-08-12
ISUPPORT: only CHANMODES and PREFIX reach the client — CHANTYPES, CASEMAPPING, MAXLIST and the length limits are dropped or stay server-side


vjt:
## Problem

`Grappa.Session.ISupport` parses **six** 005 tokens, and the client is told about **two** of them.

Parsed today (`lib/grappa/session/isupport.ex`, `merge_token/2` at :275-:313):

| token | where it lands | reaches cic? |
|---|---|---|
| `CHANMODES=` | `chanmodes.{a,b,c,d}` | yes |
| `PREFIX=` | `prefix` | yes |
| `STATUSMSG=` | `statusmsg` (#218) | **no** |
| `CASEMAPPING=` | `casemapping` (#537) | **no** |
| `MONITOR=` / `WATCH=` | `monitor` / `watch` (#247) | **no** |

`merge_token(_, acc), do: acc` (:315) drops everything else on the floor: `CHANTYPES`, `NICKLEN`, `CHANNELLEN`, `TOPICLEN`, `MAXLIST`, `MODES`, `TARGMAX`, `NETWORK`, `EXCEPTS`, `INVEX`, `ELIST`.

The wire payload (`Grappa.Session.Wire.isupport_changed/3`, `wire.ex:886`) carries `chanmodes_a..d`, `prefix`, `frame_budget_base` — nothing more.

## Consequences already visible in the tree

1. **`CHANTYPES` is open-coded.** `cicchetto/src/lib/compose.ts:80-88` says it plainly: the correct source is ISUPPORT, nothing in the stack parses it, so the RFC 2812 sigil class `[#&+!]` is a literal repeated across `slashCommands`, `linkify`, `pushPayload`, `ScrollbackPane`. (The `TODO(#30)` comment there points at a **closed, unrelated** issue — tab-completion for channel names. Bad reference, retarget it here.)
2. **`CASEMAPPING` is hardcoded `ascii` on the client.** `cicchetto/src/lib/nickEquals.ts` mirrors `Grappa.IRC.Identifier.canonical_nick/1` as a compile-time ASCII fold (#525). The server models three mappings (`:ascii | :rfc1459 | :rfc1459_strict`) and negotiates the right one per network — the client cannot, because the value is never sent. Against an rfc1459 network the two folds disagree and `foo[1]` / `foo{1}` become the same identity on one side and distinct on the other.
[...]

Issue #1251, opened 2026-08-12
Channel list modes: only +b is queryable — every other type-A CHANMODES list is unreachable


vjt:
Reported on IRC by a grappa user. Measured on `origin/main` `44358ad1`.

## What works today

Only the **ban** list is reachable. The `/banlist` path is complete end to end:

* `cicchetto/src/lib/slashCommands.ts:609` — `/banlist`, plus `/mode #chan +b` with no mask (`isBanlistQuery`, `:663`/`:674`)
* `cicchetto/src/lib/socket.ts:615` — pushes `"banlist"` on the user channel
* `lib/grappa_web/channels/grappa_channel.ex:760` — the `"banlist"` handler
* `lib/grappa/session/event_router.ex:2059` / `:2072` — 367 `RPL_BANLIST` appends `%{mask, setter, set_ts}`, 368 `RPL_ENDOFBANLIST` emits `{:banlist_bundle, ...}` and clears the accumulator
* `cicchetto/src/BanlistModal.tsx` — the modal that renders the bundle (#386)

## What is missing

**Every other type-A (list) channel mode the network advertises is unreachable.** There is no
command to query it, no numeric handler to accumulate it, and no surface to render it. Verified
by grepping the numeric-router forms on `origin/main`: `:numeric, 346`, `347`, `348`, `349`,
`728`, `729` — **no matches anywhere in `lib/`**. Only 367/368 exist.

Concretely, per network:

* **Azzurra (bahamut)** advertises `CHANMODES=bz,k,l,BcdijmMnOprRsStuU` and `MAXLIST=b:100,z:100`.
  `z` is the **restrict list** (`azzurra/bahamut` `src/channel.c:1478`), listed with
  `728 RPL_RESTRICTLIST` / `729 RPL_ENDOFRESTRICTLIST` (`src/s_err.c:812-813`). Note this
  ircd has **no** `+e` and no `+I` — the type-A set really is `bz`.
* **Other ircds** (solanum & co.) advertise `e` and `I`, listed with
  `348`/`349 RPL_EXCEPTLIST` and `346`/`347 RPL_INVITELIST`.

So the gap is not "add +e and +I": it is that the list-query path is **hardcoded to `b`** while
grappa is multi-network and the authoritative set is per-network data we already parse —
`Grappa.Session.ISupport` decodes `CHANMODES` into `%{a:, b:, c:, d:}` (`isupport.ex:53`,
[...]

Issue #1108, opened 2026-08-09
Warn before sending when the draft exceeds one IRC frame (yellow seam + byte countdown)


vjt:
Tell the operator, before sending, that the draft no longer fits in one IRC message.

Requested on IRC, in two parts:

> se il composebox eccede il singolo messaggio dobbiamo mostrare sotto di esso nel seam che ora mostra messaggi verdi / rossi un nuovo messaggio giallo "your message will send as X separate messages" E MAGARI mettere un piccolo contatore sopra il compose box che appare quando ti mancano 10 char con -10 -9 -8 etc e stai per esaurire i caratteri che entrano in un solo msg

and, on the copy:

> magari no senza (xx) chars nel seam, distrae troppo

So: **no character count in the seam line** — just "your message will send as X separate messages".

## Two surfaces

**1. Seam line, yellow.** The seam already has two states and they are siblings with identical geometry so the line never reflows when one swaps for the other (`default.css`): `.compose-box-error` in `var(--mode-op)` (red) and `.compose-box-notice` in `var(--mode-voiced)` (green, #356). This adds a third in the same shape; `var(--mode-halfop)` (`#d7af5f` light / `#7f5f00` dark) is the theme's existing amber and the obvious token. Follow #356's markup posture — `role="status"`, polite. Decide precedence when an error and the split warning are both live: an error is more urgent, and the geometry is shared, so presumably the error wins the single line.

**2. Countdown above the compose box.** Appears only in the last 10 characters of the current frame's budget: `-10 -9 -8 …`. New element, no precedent to reuse.

## The hard part: cic does not know the number

Length splitting is **server-side on purpose**. `lib/grappa/irc/line_split.ex` owns it, and its moduledoc is explicit that framing belongs to grappa, not cic — cic POSTs an arbitrary-length string and grappa fans out the fragments. The client-side splitter, `cicchetto/src/lib/messageLines.ts`, deliberately splits on newlines only (user intent), and says so: "the server still owns 512-byte length splitting for any single long line".

The per-frame budget is `linelen - relay_frame_overhead(target)`, and that overhead is not a constant the client can guess:

- `linelen` comes from `005 RPL_ISUPPORT LINELEN=<N>`, or RFC 2812's 512 default — per network.
[...]

PR #1471, opened 2026-08-17
test(#1393): census the user-topic wire boundary arm by arm

vjt:
## What this is

A **measurement** of the user-topic wire boundary, not a migration. `cicchetto/src/__tests__/wireUserBoundary.test.ts` points the existing `wireAdminBoundary` mutate-every-field harness at `narrowUserEvent`, so that bucket D of the architecture review (#1393) can be sized from a list instead of by eye.

**No arm is converted here.** `userTopic.ts` is untouched; this is one test file, zero production change.

## Why

`userTopic.ts` hand-narrows 42 arms and uses zero generated schemas, while a schema for every one of those arms is already emitted next to it. #1393 proposes continuing the migration arm by arm — but a swap that changes no verdict buys line count, and a swap that closes a real permissiveness hole buys safety. Nothing said which arms were which, so every slice was sized by guess.

## The census

Reconciled both ways rather than asserted — the walk goes over schemas indexed by `kind` literal, and is checked against the hand `switch`:

| | |
|---|---|
| arms in the hand `switch` | **42** |
| arms censused | **42** |
| in switch, not censused | none |
| censused, not in switch | none |
| census runs | 43 (one arm against two candidate schemas) |
| schemas rejecting their own sample | none |

Matching by `kind` literal rather than a camelised name is load-bearing: three arms (`connection_state_changed`, `query_windows_list`, `server_settings_changed`) have schemas under differently-named Wire modules, and the name heuristic reports them as missing. `web_session_severed` has two candidates (`S_AdminEventsWire…`, `S_RateLimitWire…`) and is censused against both.

### Both directions, because they cost opposite things

A disagreement is measured twice, and one direction alone cannot settle whether an arm is worth migrating:

* **hand ACCEPTS what the schema REJECTS** — a permissiveness hole, the only place a migration buys safety rather than line count;
* **schema ACCEPTS what the hand REJECTS** — strictness the migration would silently LOSE, since the typespec is the looser of the two and swapping in the generated check drops the hand-written guard.

Measured on `origin/main`: **8 arms diverge in the first direction, ZERO in the second.** The parity figure `armsAtParity: 34` is derived in the snapshot, over distinct arms rather than reports (an ambiguous `kind` literal is censused once per candidate and would otherwise overcount the boundary).

⚠️ **An earlier revision of this PR measured only the first direction and reported the 34 as "full parity".** That was a stronger claim than the measurement supported: with one direction, an asymmetry reads as a parity. The second direction is now measured, and it came back empty — so the parity holds, but it holds as a result rather than as an assumption.

### 34 arms at parity — a proved dead end

Neither direction disagrees. Migrating these buys line count, not safety:

`archive_changed`, `archive_purged`, `auto_away_debounce_changed`, `away_confirmed`, `channels_changed`, `connection_progress`, `connection_state_changed`, `directory_complete`, `directory_failed`, `directory_progress`, `invite_ack`, `join_failed`, `joined`, `kicked`, `mentions_bundle`, `names_reply`, `notify_list`, `own_nick_changed`, `peer_away`, `presence_changed`, `presence_error`, `presence_snapshot`, `query_windows_list`, `recover_progress`, `recover_result`, `server_reply`, `session_identity_changed`, `supported_umodes_changed`, `umode_changed`, `web_session_severed`, `who_reply`, `whowas_bundle`, `window_invite_declined`, `window_pending`.

### 8 arms where the hand narrower ACCEPTS what the schema REJECTS

| arm | fields | divergence (field/mutation) |
|---|---|---|
| `isupport_changed` | 15 | `list_modes_queryable`, `prefix_order`, `chantypes`, `casemapping`, `maxlist` (drop/null/wrong-type); `nicklen`, `channellen`, `topiclen` (drop/wrong-type); `frame_budget_base` (drop/null/wrong-type) |
| `lusers_bundle` | 13 | twelve numeric fields, drop + wrong-type each |
| `whois_bundle` | 30 | `source` (drop/null/wrong-type), `extra_lines` (drop) |
| `banlist_bundle` | 4 | `mode` (drop/null/wrong-type) |
| `window_invited` | 4 | `inviter` (drop/null/wrong-type) |
| `server_settings_changed` | 2 | `http_host_aliases` (drop/null/wrong-type) |
| `bundle_hash` | 2 | `version` (null/wrong-type) |
| `links_bundle` | 3 | `mask` (drop) |

## The verdict: every one of the 8 is divergent BY POLICY

A divergence has two readings — a deliberate tolerance to preserve, or unintended permissiveness to close — and `wireValidate.ts:18-31` states which side of that line the generated interpreter deliberately does not cross:

> A narrower is not just types — it is the defence at the boundary, and part of that defence is a judgement call the typespec cannot express: a DEFAULT that rescues an event instead of dropping it; a TOLERANCE toward a payload minted by a peer of a different vintage (cic deploys independently of the server) […] Those stay hand-written at the call site, named as policy.

Read arm by arm against the source, **all eight already carry that ruling in writing, with the issue number that made it**:

| arm | written rationale | issue |
|---|---|---|
| `isupport_changed` | five named fallbacks — `list_modes_queryable`→`["b"]`, `prefix_order`→`[]`, `chantypes`→the RFC 2812 sigil class, `casemapping`→`ascii`, `maxlist`→`{}` — plus `frame_budget_base`→`null` | #1251, #1302, #1255, #1108 |
| `lusers_bundle` | "per-field null-coercion of a non-number is RETAINED deliberately": a display-only card, one garbled optional count must not blow away the eleven good ones | codebase review S44, 2026-07-08 |
| `whois_bundle` | `source`: "Tolerant, NOT a drop condition (old servers omit it)"; `extra_lines`: optional, `undefined` tolerated | #606, #221 |
| `server_settings_changed` | `http_host_aliases`: "a malformed / absent value degrades to []" — dropping instead would strand the byte caps too | #324 |
| `bundle_hash` | `version`: "Normalise absent / malformed → null" | #292 |
| `links_bundle` | `mask`: absent → `null`, and a wrong type is **explicitly rejected** by a dedicated guard | #513a |
| `window_invited` | `inviter`: absent → the `"*"` sentinel, "so the banner has one nameless shape, not two" | #902 |
| `banlist_bundle` | `mode`: absent → `"b"`, "additive-tolerant like `links_bundle`'s `mask`" | #1251 |

So the count that matters is not 8 divergences but **zero arms worth migrating**: the 34 buy line count, and each of the 8 would reverse a decision that is written down, numbered, and load-bearing for cross-vintage deploys. `frame_budget_base` (#1108) is not the exception in that list — it is the typical case.

**Whether #1393 closes on this is not decided here** — the issue stays open, and the ruling belongs to its owner. This PR supplies the measurement it would rest on.

**One follow-up is arbitrated and ships separately.** `banlist_bundle.mode` is the only arm whose rationale falsifies itself: it claims parity with `links_bundle`'s `mask`, but `links_bundle` writes a guard rejecting a wrong-typed value and `banlist_bundle` coerces one to `"b"` — rendering a bundle as the ban list when it is not. That is a behaviour change and gets its own PR; the #1251 absent-tolerance survives, named.

## A correction to the issue's unused-type count, by method

#1393 states "165 generated types, 73 with no consumer". Re-measured at `a60082eb`: **187 generated**, **87 named outside** the generated files, but **153 reached structurally** inside a consumed aggregate — leaving **34 genuinely inert**. A name-based count reports 100, i.e. **66 false positives**.

This bears directly on the issue's fourth proposed direction: a gate failing on "a generated type with no importer", implemented the obvious way, would be wrong about two thirds of what it flags. It has to walk reachability, not names.

## Why the numbers are trustworthy

- **A positive control per direction.** A deliberately weakened narrower (the `state` check removed) is caught as a hole, reporting `["state"]`. A deliberately strengthened one (refusing a `whowas_bundle.user` null the typespec allows) is caught as a loss, reporting `["user"]`. An arm agreeing in one direction says nothing about the other, so a zero in the second column needs its own mutant or it means "blind", not "measured".
- **The second column was probed inside the census itself**, not only in its control: making `hand` reject `whowas_bundle.user === null` surfaces `whowas_bundle → user/null` and drops `armsAtParity` from 34 to 33. One mutant, exactly those two assertions, reverted after the reading.
- **Oracle checked from outside**: every arm asserts the schema accepts its own sample. None failed — had one failed, that arm's verdicts would be noise rather than measurement.
- **Two-way reconciliation** against the 42 `case` labels transcribed from the switch, a reference outside the schema walk.
- `kind` is excluded from the matrix on purpose: mutating the discriminator exercises dispatch, not field validation. Both sides reject it by different mechanisms (switch fallthrough vs literal mismatch) — an identical verdict carrying no information.

- **Not that the 34 are correct.** They are *the same as each other in both directions*, not *right*. If the schema is wrong and the hand arm is wrong the same way, this reads as parity. `wireAdminBoundary`'s recorded SHAPES are the anchor against exactly that, and they are **not** replicated here.
- **Not that the seven preserved tolerances are the right calls** — only that they are deliberate and written down. Whether each premise still holds (a BEAM of that vintage still in the field) is unmeasured.
- **Mutations are three** (drop / null / wrong-type) and **first-level fields only**: no nested mutation, no empty arrays, no out-of-enum strings. A deeper hole would not show.
- **Field-by-field agreement between the 42 schemas and the SERVER contract is unmeasured.** This compares two client-side boundaries against each other, not the client against the server.
- **No claim that a wrong-typed `banlist_bundle.mode` was ever observed in production.** The follow-up rests on an inconsistency between two sibling arms, measured in the source, not on an incident.

## Gates

- `bun run check` rc 0 — biome plus **both** `tsc` invocations. Worth stating: a biome failure short-circuits that chain, so a green recorded before the last edit is not a green, and the type checks it "passed" never ran.
- `bun run test` rc 0 — 290 files / 5587 tests.

#1393 stays open: this measures the bucket, it does not migrate it.

### DIFF
--- a/cicchetto/src/__tests__/wireUserBoundary.test.ts
+++ b/cicchetto/src/__tests__/wireUserBoundary.test.ts
@@ -13,9 +13,18 @@ import { validate } from "../lib/wireValidate";
 // list. For each arm it synthesises a valid payload from the GENERATED
 // schema, mutates it field by field, and records what the hand narrower and
 // the schema each do with it. Arms where the two agree are a proved dead end
-// — nothing to gain by swapping them. Arms where the hand narrower ACCEPTS
-// what the schema rejects are permissiveness holes, and they are the only
-// place a migration buys safety rather than line count.
+// — nothing to gain by swapping them.
+//
+// A disagreement is measured in BOTH directions, because they cost opposite
+// things and one alone cannot settle whether an arm is worth migrating:
+//
+//   * the hand narrower ACCEPTS what the schema rejects — a permissiveness
+//     hole, the only place a migration buys safety rather than line count;
+//   * the schema ACCEPTS what the hand narrower rejects — strictness the
+//     migration would silently LOSE, since the typespec is the looser of the
+//     two and swapping in the generated check drops the hand-written guard.
+//
+// Measuring one direction only would report an asymmetry as a parity.
 //
 // Arms are matched to schemas by their `kind` LITERAL, not by a camelised
 // name: the name heuristic missed three arms whose schema lives under a
@@ -97,6 +106,7 @@ type ArmReport = {
   schema: string;
   fields: number;
   handAcceptsSchemaRejects: string;
+  schemaAcceptsHandRejects: string;
   schemaRejectsValid: boolean;
 };
 
@@ -106,15 +116,17 @@ function censusArm(kind: string, schemaName: string, node: WireNode): ArmReport 
   const fields = Object.keys(valid).filter((f) => f !== "kind");
 
   const holes: string[] = [];
+  const losses: string[] = [];
   for (const f of fields) {
     for (const [label, mutated] of [
       ["drop", without(valid, f)],
       ["null", { ...valid, [f]: null }],
       ["wrong-type", { ...valid, [f]: wrongType(valid[f]) }],
     ] as const) {
-      if (verdict(hand, mutated) === "accept" && verdict(generated, mutated) === "reject") {
-        holes.push(`${f}/${label}`);
-      }
+      const byHand = verdict(hand, mutated);
+      const bySchema = verdict(generated, mutated);
+      if (byHand === "accept" && bySchema === "reject") holes.push(`${f}/${label}`);
+      if (bySchema === "accept" && byHand === "reject") losses.push(`${f}/${label}`);
     }
   }
 
@@ -123,6 +135,7 @@ function censusArm(kind: string, schemaName: string, node: WireNode): ArmReport 
     schema: schemaName,
     fields: fields.length,
     handAcceptsSchemaRejects: holes.length === 0 ? "-" : holes.join(", "),
+    schemaAcceptsHandRejects: losses.length === 0 ? "-" : losses.join(", "),
     // The oracle's own sanity check: a schema that rejects its OWN sample
     // means the sampler and the schema disagree, and every verdict on that
     // arm is noise rather than a measurement.
@@ -160,6 +173,17 @@ const weakened: Narrower = (raw) => {
   if (typeof raw !== "object" || raw === null) return null;
   const r = raw as Record<string, unknown>;
   return typeof r.network === "string" ? { kind: "away_confirmed", ...r } : null;
+};
+
+// The control for the OTHER direction, which needs its own mutant: an arm
+// agreeing in one direction says nothing about the other. `whowas_bundle`'s
+// `user` is `string | null` in the typespec, so the schema accepts a null the
+// hand narrower here refuses ON PURPOSE — and `user/null` has to show up as a
+// loss.
+const strengthened: Narrower = (raw) => {
+  if (typeof raw !== "object" || raw === null) return null;
+  const r = raw as Record<string, unknown>;
+  return typeof r.user === "string" ? { kind: "whowas_bundle", ...r } : null;
 };
 
 // The 42 `case` labels of `narrowUserEvent`, transcribed from the switch so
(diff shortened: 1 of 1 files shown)


vjt:

Closing by content, not by ancestry: the union replayed each PR's own commits onto a branch cut from current `main`, so the SHAs were rewritten and GitHub cannot mark this PR merged by itself.

The content check leaves **exactly one** differing line, and it is the one the union changed on purpose:

```
-            "handAcceptsSchemaRejects": "mode/drop, mode/null, mode/wrong-type",   <- this branch (pre-guard)
+            "handAcceptsSchemaRejects": "mode/drop",                               <- main (post-guard)
```

That is the bench recording the narrower tolerance #1472 introduced. It was **not** hand-edited to match: the bench was run against the new guard and the red was read first — one line, one failing test of four, no unexpected divergence. `armsAtParity` stays 34, verified in the output as a context line rather than deduced.

This PR is therefore landed *and* reconciled: the union is the only arrangement in which the bench and the guard it measures were ever gated together.

Issue #1468, opened 2026-08-17
recover: two terminal clauses leave a step `:running` forever, against the rule #623 pt3 set for itself


vjt:
Found while surveying `Session.Server`'s recover-progress code for the #1390 decomposition. Measured by
a worker on `a60082eb`, not deduced — the reasoning is reproduced below so it can be checked.

## The rule, and where it is broken

#623 pt3 set the rule that **no progress step may be left `:running` when the recovery reaches a terminal
state**. Its sibling leg encodes it in a test, textually: *"reconciled to `:failed` (NOT left running)"*.

**Two of the five terminal clauses violate it.** On leg (a) — nick taken → `RECOVER` → the re-NICK is
never observed → deadline — the terminal emits only `nick:failed`. The in-code comment justifies this
with *"identify never started"*, and that is **false**: `:awaiting_nick` is only reachable by passing
through `:awaiting_r`, which turns `identify:running` on at entry.

## Why it is user-visible

Verified client-side, not inferred:

- `recoverProgress.ts` upserts by step name and does not touch rows nobody updates, so a row left
  `:running` stays `:running`.
- `RecoverModal.tsx:125` keeps the `is-running` treatment regardless of the outcome.

Net effect: **the modal closes reporting a failed recovery while the identify step spins forever.**

## Test coverage today

None encodes the broken path: the leg (a) test asserts only `nick:failed`. The rule exists, a test for
the rule exists on the sibling leg, and the clause next to it does not honour it.

## Suggested red (not prescribed)

An `async: true` property that drives the FSM over every reachable transition, accumulates the emitted
steps under the client's upsert rule, and asserts the #623 pt3 invariant on every terminal — rather than
one example per leg, which is how the gap survived.

## Not established

- Whether the other three terminal clauses are correct for the same reason or by accident.
- How often leg (a) is reached in practice — no production frequency has been measured.

vjt:
Closed by the pipeline change ruled by vjt on 2026-08-20: **`status:soon` is retired**. An issue is now closed when its work **merges**, and the **milestone** records which release carries it — the label duplicated what the milestone already said.

This issue was verified individually before closing, not swept: its work is on `origin/main` under commits that are not documentation-only. No behaviour changes here, only bookkeeping.

<sub>Applied by vjt-claude on vjt's instruction.</sub>

PR #1479, opened 2026-08-17
recover(#1468): reconcile every started step at a terminal, and prove it

vjt:
## The defect

#623 pt3 set a rule — **no progress step may be left `:running` when the
recovery reaches a terminal state** — and two of the five terminal clauses did
not honour it. A deadline out of `:awaiting_verb_settle` or `:awaiting_nick`
reported a single step, while every path into those phases had gone through
`:idle -> :awaiting_r`, which sets `identify: :running`.

It is user-visible, not theoretical: `recoverProgress.ts` upserts one row per
step name and never touches a row nobody updates, and `RecoverModal` keeps the
`is-running` treatment regardless of the outcome. The modal closed reporting a
failed recovery with the identify step spinning forever.

**The comment that justified leg (a) claimed the identify step "never
started".** That is false — `:awaiting_nick` is reachable only through
`:awaiting_r` — and it is **deleted**, in the same commit as the clause it was
defending, rather than softened into something vaguer.

## The shape of the fix, and why not a property

Curing the two clauses is the small half. The interesting question is why the
gap survived a rule, a test for the rule on the sibling leg, and a review: each
leg was pinned by its **own example**, so a clause nobody wrote an example for
answered to nothing.

The issue suggested a StreamData property. **I used an exhaustive walk
instead** — a breadth-first search to fixpoint over the reachable
`{phase, verb, rows}` space, asserting the invariant at every terminal:

- the space is **finite and small** (seven phases, one status per step name),
  so the search is a **proof** over every reachable path, retry loops included,
  where a generator is a sample whose seed can get lucky;
- it cannot flake, and it cannot pass by luck;
- `rows` mirrors the **client's** rule (upsert by step name, last write wins),
  because the question being asked is what the modal shows when the modal
  closes;
- it carries its own vacuity floors — a minimum state count, a minimum terminal
  count, and the assertion that a terminal is reached out of **each** of the
  four phases that can fail. A walk that explored nothing would otherwise
  satisfy the invariant by having nothing to check.

**The walk immediately paid for itself.** With the cure reverted it reported a
path no example covered — `[:start, {:nick_error, 437}, :settle,
{:nick_error, 433}, :timeout]`, the RELEASE leg with a bounded retry — leaving
**two** rows stranded (`nick` and `identify`), where the issue described only
the RECOVER legs and only `identify`.

## The other three clauses: two by the rule, one vacuously

The issue left this unestablished. Established here:

- `:awaiting_r` and `:awaiting_final_r` are correct **by the rule** — the steps
  that can still be `:running` at those terminals are exactly the ones they
  reconcile.
- The catch-all `terminal_steps(_, _, reason)` is correct **only vacuously**:
  no reachable path reaches it, since `:failed` is only ever entered from the
  four named phases. It reconciles `nick` alone and would strand rows the day a
  new phase can fail. **Named as a second, latent defect and left as found** —
  a cure nothing can exercise is a claim, not a fix — and the walk is what
  makes leaving it safe: the terminal becomes reachable and checked the moment
  such a phase exists.

One inaccuracy is knowingly left in place: on the bounded-retry path the settle
marked the reclaim verb `:ok` and the terminal marks it `:failed` again. It
predates this issue and strands nothing, so curing it would widen the change
past what the red buys.

## Evidence

- **Red before green, by reverting the cure**: 3 failures — both cured examples
  and the walk. Restored: **12 tests, 0 failures**.
- **`scripts/check.sh` RC=0**, read from a file. The first run was **RED**, on
  two Credo findings of mine in the new walk (a single-item `++` append and a
  `_reason` discard where the file's inferred strategy is a bare `_`); both
  fixed in their own commit and the gate re-run **from the top**, since
  `check.sh` halts at the first red phase and a downstream green from the
  earlier run would not have existed.
- That clean run covers Credo, Sobelow, doctor, **Dialyzer (`Total errors: 0`,
  `done (passed successfully)`)**, ExUnit **6519 tests + 8 doctests + 61
  properties, 0 failures**, and **540 bats**.
- 3 files: the projection, its test, and the decision-log entry.

## What this PR does not claim

- **No production frequency was measured.** How often leg (a) is reached in
  practice is unknown, here as in the issue.
- **The catch-all's unreachability is a reading of today's `step/2`**, not a
  theorem about future clauses — which is precisely why the walk, and not the
  reading, is what guards it.
- **Nothing was exercised against a live session.** This projection is pure and
  the walk drives the real FSM in-process; that the server broadcasts these
  triples unchanged is covered by the existing `server_test.exs` integration
  describe, not by anything added here.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -46877,3 +46877,61 @@ because reaching it means fabricating a state the FSM cannot emit. And the
 claim that no OTHER consumer names the moved functions was scoped to `lib` and
 `test` — a comment in `cicchetto/src/RecoverModal.tsx` named one, and was
 found only after the move._
+<!-- entry #1468 -->
+
+---
+
+## 2026-08-17 — #1468: the terminal rule is now proven, not exemplified
+
+#623 pt3 set a rule — no progress step may be left `:running` when the recovery
+reaches a terminal state — and two of the five terminal clauses did not honour
+it. A deadline out of `:awaiting_verb_settle` or `:awaiting_nick` reported one
+step, while every path into those phases had passed through
+`:idle -> :awaiting_r`, which sets `identify: :running`. The client upserts one
+row per step name and never touches a row nobody updates, so the modal closed
+reporting a failed recovery with the identify step spinning forever.
+
+The in-code comment that justified leg (a) claimed the identify step "never
+started". That was false — `:awaiting_nick` is reachable only through
+`:awaiting_r` — and it is deleted rather than softened, in the commit that
+cures the clause it was defending.
+
+### What the red buys, and what it deliberately does not
+
+The cure is two clauses. The interesting part is why the gap survived a rule, a
+test for the rule on the sibling leg, and a review: **each leg was pinned by its
+own example, so a clause nobody wrote an example for answered to nothing.**
+
+The replacement is an **exhaustive walk**, not a StreamData property — a
+breadth-first search to fixpoint over the reachable `{phase, verb, rows}` space,
+asserting the invariant at every terminal. The space is finite and small (seven
+phases, one status per step name), so the search is a PROOF over every reachable
+path, retry loops included, where a generator would be a sample with a seed that
+can get lucky. It carries its own vacuity floors: a minimum state count, a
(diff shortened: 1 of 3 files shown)


vjt:

Issue #1419, opened 2026-08-16
cicchetto: modalChrome's whole-shape test lives at 71-102% of its 5s budget and reds the gate on a loaded runner


vjt:
## Measured

`cicchetto/src/__tests__/modalChrome.test.ts` →
`leaves no rule outside the bases carrying a base's whole shape`

Two consecutive attempts of the **same CI run**, same commit:

| attempt | result | test duration | budget |
|---|---|---|---|
| 1 | **failed** — `Test timed out in 5000ms` | 5086 ms | 5000 ms |
| 2 | passed | **3565 ms** | 5000 ms |

The whole file is 29 tests in 3868 ms, so this single test is **~92% of the
file's runtime**, and it passes with roughly 1.4 s of headroom. A ~40% slowdown
under runner load flips it.

## Why this matters beyond one red

The test body is two assertions. The time is spent inside the `classesResolvingTo`
helper, which resolves classes against the stylesheet — so the cost is in the
instrument, not in what is being asserted.

When it fires, the job reports `Test timed out in 5000ms` and the gate goes red,
which reads as a code failure on whatever branch happened to be running. It cost
one such investigation on an unrelated PR.

## Not the fix

**Raising the timeout.** That hides the margin rather than removing it, and the
margin drifts back as the stylesheet grows — which is the direction it grows in.
The assertion itself is correct and must not be weakened: it is keyed on the whole
resolved shape deliberately (see the `#740` comment above it), and an exclusion
list was explicitly rejected there.

## Direction

Make the helper's cost proportionate — resolve the stylesheet once for the file
rather than per call — so the test's duration reflects the two comparisons it
performs. The durable check is that the file's runtime stops being dominated by
one test.

## Not established

Where the time goes inside the helper was not profiled; the 92% figure is from the
reported per-test and per-file durations, not from a profile.

vjt:
Datapoint from main, run 31967220081 (head `38e75561`, the #1440 union merge):

```
FAIL src/__tests__/modalChrome.test.ts > #407 — the modal chrome extraction changes no pixel > leaves no rule outside the bases carrying a base's whole shape
Error: Test timed out in 5000ms.
 ❯ <anonymous> src/__tests__/modalChrome.test.ts:270:30
Test Files  1 failed | 286 passed (287)
Duration  163.40s
```

[...]

PR #1485, opened 2026-08-17
test(#1419): memoise the stylesheet parse the clone census re-runs

vjt:
## What

Memoises the stylesheet parse inside `cicchetto/src/__tests__/modalChrome.test.ts`.
One file, one commit, +16/-0.

## Why now

The test *"leaves no rule outside the bases carrying a base's whole shape"*
has hit the 5000 ms vitest default **nine times**, twice in one day on
branches that cannot touch modal chrome (#1471 wire bench, #1478 packaging).
Most recent CI sighting: **5201 ms — a 201 ms, 4% overshoot** in a run that
was otherwise 289/290 files and 5586/5587 tests green. It taxes every PR.

## The measurement that chose the cure

Two cures were available and they are not interchangeable, so the cost was
measured before either was written. `topLevelRules()` was instrumented with a
call counter and a `performance.now()` accumulator for the duration of that
one test:

| | |
|---|---|
| stylesheet | **434 793 bytes** (`src/themes/default.css`) |
| bare class selectors, `N` | **712** |
| `topLevelRules()` calls in the test | **1427** |
| time inside those calls | **1117.2 ms** |
| test wall time | **1184.8 ms** |
| **share spent re-parsing** | **94.3 %** |

1427 is exactly `2 × (N + 1)` plus the one probe call: `classesResolvingTo`
parses once to build the name list, then `resolve([name])` parses again for
each of the 712 names, and the test calls it twice. So the re-parsing does
dominate, and the cure that **removes** the cost was the available one.

After: that test runs in **49 ms** (24×), and the file's tests drop from
**1750 ms to 372 ms**.

## Why not a bigger timeout

This body has no temporal oracle — it asserts *which CSS rules carry a base's
shape*. The 5 s was vitest's default, never an assertion, so a raised budget
would have hidden 1117 ms rather than removed it. Distinct from #1336, where
the deadline WAS the measurement.

## Why the memo is safe

`themeCss` is a `readFileSync` evaluated once at import time, so the parse is
a pure function of a constant and the cache cannot go stale. No caller mutates
the returned array — `resolve` and the census only read it.

## Positive control

The test must still blush for the right reason. Injecting a clone of
`.modal-chrome-button` under a fresh class name — the exact #740 failure mode
this assertion guards — into the stylesheet:

```
× leaves no rule outside the bases carrying a base's whole shape
  AssertionError: expected [ 'modal-chrome-button', …(1) ] to deeply equal [ 'modal-chrome-button' ]
  +   "zzz-clone-mutant"
  Tests  1 failed | 28 passed (29)
```

**One mutant, exactly one reddened assertion**, naming the offender.

## Gates

- `scripts/bun.sh run check` — **RC=0** (`biome && tsc && tsc`; the chain is
  `&&`, so RC=0 means all three ran). The 59 biome warnings are pre-existing
  CSS-specificity notices, zero errors.
- `scripts/bun.sh run test` — **RC=0, 290/290 files, 5598/5598 tests**, run on
  this branch rebased onto `97b0ee3a`.

## Not claimed

The 5201 ms was measured on CI and the 1184.8 ms on a developer host. I make
**no claim** about why they differ, having measured neither host's load.

### DIFF
--- a/cicchetto/src/__tests__/modalChrome.test.ts
+++ b/cicchetto/src/__tests__/modalChrome.test.ts
@@ -74,8 +74,24 @@ type Rule = { selectors: string[]; body: string };
  * rules nested inside one, which resolve under a condition this file does not
  * model. The "bare class selectors only" assertion below is what makes that
  * safe: it fails if a site is ever reached from inside a media query.
+ *
+ * Memoised, because the clone census below resolves EVERY bare class in the
+ * sheet and each `resolve` re-parses: 1427 parses of a 435 KB stylesheet for
+ * one assertion pair, measured at 1117 ms of that test's 1185 ms — 94%, and
+ * the 5 s default is what #1419 kept tripping on. `themeCss` is read once at
+ * import, so the parse is a pure function of a constant and the cache cannot
+ * go stale. The budget is left at the default deliberately: nothing here
+ * asserts a duration, so a timeout bump would have hidden the cost instead of
+ * removing it.
  */
+let parsedRules: Rule[] | undefined;
+
 function topLevelRules(): Rule[] {
+  parsedRules ??= parseTopLevelRules();
+  return parsedRules;
+}
+
+function parseTopLevelRules(): Rule[] {
   const stripped = themeCss.replace(/\/\*[\s\S]*?\*\//g, "");
   const rules: Rule[] = [];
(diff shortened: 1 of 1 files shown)


vjt:

Issue #1395, opened 2026-08-16
arch: the user login ladder is reachable only from an HTTP action while the visitor twin is a context (bucket F login-policy)


vjt:
Bucket **F login-policy** of the 2026-08-15 architecture review (`docs/reviews/architecture/2026-08-15-architecture-review.md`). One HIGH. **Placement finding only — no weakness is asserted in the ladder itself.**

## What was measured

**Responsibility A3 (HIGH).** One job, two homes, chosen by subject kind:

- Visitor login policy is a context module with a documented state machine — `visitors/login.ex`, 827 lines.
- User login policy is a `defp` chain inside `auth_controller.ex` — **888 lines, 67 functions** — including the second-factor ladder, challenge minting, throttles and 21 `visitor_error_response/1` clauses, composing `Accounts`, `TOTP`, `WebAuthn`, `Phoenix.Token` and `LoginThrottle` itself.
- `Grappa.Accounts` has primitives only and no login orchestration.

## Why it matters

The user ladder is unreachable from anywhere but an HTTP action, untestable without a `conn`, and every new failure mode edits a controller. The visitor half of the same polymorphism already has the shape this one lacks.

## Proposed direction

Lift it into `Grappa.Accounts.Login` as the sibling of `Grappa.Visitors.Login`, taking a plain input map and returning a tagged outcome. The error clauses move to `FallbackController`, where the rest of the token vocabulary already lives.

## Not established

The review states explicitly that this is a placement finding: it asserts **no** defect in the authentication logic itself. Nothing was executed. Note also that the security review of the same date is tracked out of band — do not read this issue as a statement about the security posture of the login path either way.

vjt:
## Read-only recon: the proposed direction is refuted on three points of four

Measured against `origin/main` = `5bb8c68f`. Nothing was executed — no compile, no test run,
no gate. Every number below is `git show origin/main:<path>` piped to a counter, and each is
re-derivable from the paths and line numbers given. The four files involved
(`auth_controller.ex`, `visitors/login.ex`, `accounts.ex`, `fallback_controller.ex`) are
byte-identical between the base this was first measured on (`1bdc624a`) and `5bb8c68f`.

**The finding is live**, and one part of it is real and worth fixing (see "What survives").
But the *proposed direction* — lift into `Grappa.Accounts.Login` as the sibling of
`Grappa.Visitors.Login`, error clauses to `FallbackController` — does not survive
measurement, and the sibling it is modelled on is what refutes it.

[...]

PR #1486, opened 2026-08-17
accounts(#1395): give the user login decision a non-HTTP entry point

vjt:
## What

Gives the user password door a decision function reachable without a
`Plug.Conn`: `Grappa.Accounts.Login.authenticate/1`.

```elixir
@spec authenticate(%{name: String.t(), password: String.t()}) ::
        {:ok, User.t()}
        | {:second_factor, :passkey | :totp, User.t()}
        | {:error, :invalid_credentials | :passwordless}
```

6 files, +366/-29. Base `main`.

## Scope — the measured residue only

The ladder was a `defp` chain inside `GrappaWeb.AuthController`, so every
test over it needed an HTTP connection. What moved is the decision: a name,
a password, and the account's own second-factor configuration. What is
genuinely request-shaped stayed at the edge.

**Deliberately NOT moved**, each for a reason that is a boundary and not a
preference — the moduledoc argues all four:

- **The throttles.** The twin `Grappa.Visitors.Login` holds no rate limiting
  at all, and `GrappaWeb.LoginThrottle`'s own moduledoc argues that an
  IP-keyed window is request-edge policy. `Grappa.Accounts` depends on
  neither `Grappa.RateLimit` nor `Grappa.AdminEvents`, exactly as
  `Grappa.Visitors` does not.
- **The second-factor challenge.** Minting it needs `Phoenix.Token` against
  `GrappaWeb.Endpoint` and `GrappaWeb.PasskeyOrigin`. `GrappaWeb` already
  depends on `Grappa.Accounts`, so reaching back would close a cycle
  `Boundary` rejects outright.
- **The client-token door**, already a context function returning a tagged
  outcome, and an effect rather than a decision.
- **The 22 `visitor_error_response/4` clauses**, which belong to the visitor
  half of the controller. 20 are identity no-ops, one inverts a written
  decision, and the catch-all is load-bearing: without it `:no_server`
  answers a raw 500 instead of the JSON envelope.

## `:passwordless` is not a spelling of `:invalid_credentials`

Both leave as the same uniform 401, so the distinction is invisible on the
wire — but only one of them is a guess. A wrong password charges the mode-1
window; an account in `:passwordless` passkey mode presented the RIGHT
password and is refused because this door is shut for it, so charging would
spend a user's own window on a configuration choice they made. Before this
slice the distinction was implicit in the call order (the throttle was
charged upstream of the ladder); one function now answers both, so it had to
become explicit.

**Nothing asserted it.** Measured before writing the test: `passkey_mode:
:passwordless` never appears in `auth_controller_test.exs`, which only ever
sets `:second_factor`. The nearby *"a passwordless login never advances the
counter"* is a false friend — it lives in the VISITOR throttle describe and
means "with no password supplied", not `passkey_mode`.

So the new test: ten right-password attempts against a `:passwordless`
account, all 401; the eleventh still 401. `@mode1_max_failures` is 10, so a
429 on the eleventh is precisely the signal that one of the ten charged.

### Positive control

Collapsing the two tags — adding `charge_mode1_failure/1` to the
`{:error, :passwordless}` arm, making it behave identically to the
`:invalid_credentials` arm:

```
1) test ... a passwordless account's RIGHT password is refused without charging
   ** (RuntimeError) expected response with status 401, got: 429, with body:
   "{\"error\":\"too_many_attempts\"}"
   auth_controller_test.exs:1495

75 tests, 1 failure
```

**One mutant, exactly one reddened assertion** — the eleventh attempt, the
only observable the two tags differ on.

## Two traps this branch walked around

- In `AuthController`, `Login` is ALREADY aliased to `Grappa.Visitors.Login`
  — the other half of the same polymorphism. A bare `Login.authenticate(...)`
  would have pointed at the wrong module **silently**. The call site is
  spelled `Accounts.Login.authenticate(...)`.
- `@mode1_bucket` and its siblings are declared BELOW `password_login/4`, and
  a module attribute is in scope only after its declaration. The charge
  therefore lives in `charge_mode1_failure/1`, next to the constants.

## Gates

Run in this worktree, rebased onto `97b0ee3a` first so the green attests the
tree that is being pushed.

- `scripts/check.sh` — **RC=0**, read from a file rather than `$?`, and
  phase by phase, because this script aborts at the first red and a green
  downstream of an abort does not exist:

  | phase | result |
  |---|---|
  | Credo strict | 684 source files, no issues |
  | Sobelow | 15 findings, **all Low confidence**, none in a file this branch touches (uploads, reaper, background_image, isupport, hot_reload, hardened_cmd, repo) — the gate is Medium-or-above |
  | Doctor | passed |
  | ExUnit | **8 doctests, 61 properties, 6529 tests, 0 failures** |
  | Dialyzer | **Total errors: 0**, Skipped: 0 |
  | bats | **553 ok / 0 not ok** |

- `scripts/design-notes-gate.sh` — **RC=0**, one new entry heading, separator
  and marker present. Contribution numstat pinned to a file before the rebase
  and re-diffed after: **identical**, `docs/DESIGN_NOTES.md` **62/0**
  (deletions zero).

## Not claimed

- The first commit was written in a previous session and had never been
  compiled. `scripts/check.sh` duly halted on it at Credo with 9 issues in
  the new conn-free test file, all of them inert rebinds or unused-variable
  naming; the third commit clears them and explains why the rebinds were
  inert rather than load-bearing.
- The file count is **6**, against the 5 the scope ruling anticipated. The
  sixth is `auth_controller_test.exs`, added deliberately for the
  `:passwordless` pin above; 6 was the declared cap and this is at it.
- `login_test.exs` carries **10** conn-free tests, not the 8 the ruling
  estimated. Counted, not estimated.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -46985,3 +46985,65 @@ the CONFIGURATION — that the package is thin, that its version comes from
 the header, that its output directory is outside every upload glob — which
 is what can be held still without a Linux toolchain. Whether nfpm accepts
 these two files is proven by the first CI run, not here._
+<!-- entry #1395 -->
+
+---
+
+## 2026-08-17 — #1395: the user login DECISION moves; the throttle, the challenge and the render do not
+
+The user ladder was a `defp` chain inside `GrappaWeb.AuthController`, reachable
+only from an HTTP action: all 64 tests over that controller need a `Plug.Conn`,
+against 33 conn-free tests for the visitor twin. `Grappa.Accounts.Login.authenticate/1`
+now answers the decision — name plus password in, and the door the account still
+has to walk through out — with no `conn` anywhere in it.
+
+**What the issue proposed and this did NOT do.** #1395 asked for a full sibling
+of `Grappa.Visitors.Login`, with the second-factor ladder, the challenge minting
+and the throttles all lifted, and the 22 `visitor_error_response/4` clauses moved
+to `FallbackController`. Measurement refuted three of those four, and the ruling
+kept only the residue:
+
+  * **The throttles stay at the edge.** `Grappa.Visitors.Login` — the twin held
+    up as the model — contains no rate limiting at all: the visitor window lives
+    in the controller, above the call into the context. `GrappaWeb.LoginThrottle`
+    argues the placement in its own moduledoc, and `Grappa.Accounts` depends on
+    neither `Grappa.RateLimit` nor `Grappa.AdminEvents`, exactly as
+    `Grappa.Visitors` does not. Lifting the window would be the trade a shipped
+    module refuses, for a context standing in the identical position.
+
+  * **The challenge cannot be lifted.** Minting it needs `Phoenix.Token.sign/3`
+    against `GrappaWeb.Endpoint`, and the passkey options need
+    `GrappaWeb.PasskeyOrigin`. `GrappaWeb` already depends on `Grappa.Accounts`,
+    so carrying either into the context closes a cycle `Boundary` rejects.
+
+  * **The 22 error clauses stay put**, and are visitor-half code that has nothing
+    to do with this door. Twenty are pure identity pass-throughs whose relocation
+    would change no behaviour while deleting a documented audit index (L-web-1);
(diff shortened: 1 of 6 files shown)


vjt:

PR #1487, opened 2026-08-17
push(#1404): hand the detached work to the supervisor that exists for it

vjt:
Part of #1404 — the structural half of one of the thirteen. The measurement half of the same item was settled separately and needed no code; this is what remained.

## What changes

Three places hand work off the session hot path so the session never blocks on it: the window-counts snapshot and the two push-notification dispatches. All three started that work with a bare spawn, so the worker belonged to nobody — outside the supervision tree, invisible to the operator, and its crash a disappearance rather than a report. Every other detached spawn in the tree already goes through `Grappa.TaskSupervisor`, and S37 wrote the rule down; these three predate it.

They now follow it, and a test asserts that each of them does.

## What this buys — and what it does not

**Supervision and visibility. Not a concurrency ceiling.**

The number of these workers that can run at once is exactly what it was before. A ceiling on detached work is a child-spec setting (`max_children`) and the tree configures none — measured: the string appears nowhere in `lib/` or `config/`. Routing a spawn through a supervisor changes the spawn verb, not the child spec, so it cannot add a bound that is not configured.

This is stated narrowly on purpose. The moduledocs that described these workers as unlinked are corrected, and they are corrected to say what is true rather than something better: a docstring implying a bound here would be the same shape of defect this issue exists to remove, relocated from the code into the prose where it is harder to notice.

## The one part that is not a mechanical verb swap

**The start result is discarded, not matched.** A bare spawn could not fail, so a strict match around it was free. A supervisor can refuse. Matching strictly would convert a skipped best-effort optimisation into a crash on the session hot path the day a ceiling is introduced — the opposite of what a ceiling would be for. Two older call sites already carry that strict match; this change declines to add two more, and disarming those two is a separate decision with its own blast radius.

## How the guard catches the worker

A worker that has already finished is indistinguishable from one that was never supervised — the supervisor's child list is empty either way. So each test parks its worker and asks while it is parked, and both parks are real infrastructure rather than a stand-in for the subject under test: a process registered where the session would be, which never answers, holds the snapshot worker inside its live-count call; OTP's own `:sys.suspend/1` on the presence singleton holds the two notification workers where they consult it.

The assertion is the worker pid's membership in the supervisor's child set, not the size of that set. An earlier version compared counts and failed against code that was already correct, because another test's worker was still dying while it measured — a count cannot tell "one arrived" from "one left".

## Evidence

- **Red read before the fix, and the first red was the wrong red.** Two of the three assertions initially failed because the preference setter validates the whole map and rejected a partial one — the same colour as the real failure with the opposite diagnosis. Corrected, then re-read: three red for the property itself.
- **Three mutants, one per site**, injected by line number with an assertion on the expected shape and a verified clean restore. Each kills **exactly one** assertion: the snapshot site kills only the window-counts test, the first notification site only the message test, the second only the presence test. Never two, never zero.
- **Green on two seeds**, including the seed that exposed the count-based flake.
- **`scripts/check.sh` RC=0** — 8 doctests, 61 properties, 6521 tests, 0 failures; Dialyzer `Total errors: 0`; Sobelow passed; bats 0 not-ok. The first run of the gate was RED on two Credo strict findings, both this branch's own; they were fixed in their own commit and the gate was re-run **from the start**.

## Not established

- **Nothing here measures concurrency.** How many of these workers can be alive at once, and whether that wants a bound, is untouched and unanswered. Both the test and the decision-log entry say so.
- **The two older strict matches are still armed.** They are named, not fixed.
- **The gates ran on the previous `main`.** The branch has not been rebased since `main` moved; the intersection with what `main` brought is `docs/DESIGN_NOTES.md` alone, which is append-only and marker-protected, but the green above was measured on the earlier base.
- One prose site outside this change still describes these workers as unsupervised. It is in a file this slice deliberately did not open, and it is left for a follow-up rather than silently widened here.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -46985,3 +46985,53 @@ the CONFIGURATION — that the package is thin, that its version comes from
 the header, that its output directory is outside every upload glob — which
 is what can be held still without a Linux toolchain. Whether nfpm accepts
 these two files is proven by the first CI run, not here._
+<!-- entry #1404e -->
+
+---
+
+## 2026-08-17 — #1404: the detached hand-offs become supervised work, and only that
+
+Three places hand work off the session hot path so the session never blocks
+on it: the window-counts snapshot and the two push-notification dispatches.
+All three started that work with a bare spawn, so the worker belonged to
+nobody — outside the supervision tree, invisible to the operator, and its
+crash a disappearance rather than a report. Every other detached spawn in the
+tree already goes through `Grappa.TaskSupervisor`, and S37 wrote the rule
+down; these three predate it. They now follow it, and a test asserts each of
+them does.
+
+**What this buys, stated narrowly on purpose: supervision and visibility. Not
+a ceiling.** The concurrency of that work is exactly what it was. A ceiling on
+detached work is a child-spec setting (`max_children`) and the tree configures
+none, so routing the spawn through a supervisor cannot add one — routing
+changes the spawn verb, not the child spec. The moduledocs that used to say
+"unlinked task" are corrected, and they are corrected to say what is true
+rather than to say something better: a docstring that implied a bound here
+would be the very shape of defect this work exists to remove, moved out of the
+code and into the prose where it is harder to notice.
+
+**Why the start result is discarded rather than matched.** This is the one way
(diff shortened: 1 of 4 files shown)


vjt:

Issue #1410, opened 2026-08-16
review: init/1 does three DB reads inside the blocking start, against two documented contracts (L-S2, bucket D)


vjt:
Bucket **D session-lifecycle** of the 2026-08-15 codebase review (`docs/reviews/codebase/2026-08-15-codebase-review.md`), gating MEDIUM **L-S2** — the one finding of that bucket left without an issue. L-S1, L-S6 and L-S7 are in #1349.

## L-S2 (MEDIUM) — `Session.Server.init/1` performs DB reads, contradicting two documented contracts

**Files:** `lib/grappa/session/server.ex:1083-1093`; contracts at `lib/grappa/session.ex:41-44` and `server.ex:1057-1062`; chain at `networks/session_plan.ex:185-198` → `resolve_attempt/2` at `:77-91`

`init/1` invokes the injected `refresh_plan` closure, which runs `Credentials.get_credential_by_ids/2`, `Repo.preload(network: :servers)` and `Accounts.get_user!/1` — **three SQLite round-trips inside `init/1`**, i.e. inside the caller's blocking `DynamicSupervisor.start_child/2`.

Two live documents are now false:

- `Grappa.Session`'s moduledoc: "the Server's `init/1` is therefore a pure data consumer (no `Repo`, no `Networks`, no `Accounts`, no `Visitors` reads)" — this is the **A2 cycle-inversion contract**, i.e. the sentence the dependency architecture leans on.
- `server.ex:1057`: "`init/1` is intentionally non-blocking". The upstream-socket half of that still holds; the DB half does not.

Consequence beyond the drift: Bootstrap's spawn loop is serialised on N × 3 queries.

## Why this one is worth its own issue rather than a sweep

It is the point where a documentation defect and an architecture defect are the same defect. The architecture review's bucket I (dependency-inversion, #1398) reasons about the `Networks → Session` inversion partly from this contract; if the contract is false, the reasoning that cites it needs re-checking rather than restating. Fix the code to match the contract, or amend both documents and tell #1398 that its premise moved — but not neither.

## Not established

Nothing was executed: no query log, no timing of the spawn loop. "Serialised on N × 3 queries" is derived from the call chain, **not measured** — measure it before quoting a boot-time cost, the same way #1372's corpus lesson applies.

vjt:
## Recon on `e1892e80` — read-only

Everything below was read with `git show origin/main:<path>` at `origin/main` =
[...]

Issue #1398, opened 2026-08-16
arch: the acyclic Boundary graph hides 5 dirty_xref cycles, and every escape hatch grew since July (bucket I dependency-inversion)


vjt:
**Status:** the finding was half-dead before this issue was rewritten, and the extraction it
proposed is the wrong shape. What remains is small and specified below. **No milestone
decision is needed any more — the earlier "move to 1.4 or hold 1.3" question is moot.**

Bucket **I dependency-inversion** of the 2026-08-15 architecture review
(`docs/reviews/architecture/2026-08-15-architecture-review.md`).

This body carries the current state. Earlier correction comments have been folded in and
removed, per the one-version-in-the-body rule.

**Measurement base: `origin/main` = `2ed5fb51`.** Everything below was measured on that
tree, not inferred from the review.

---

## 1. Half the finding is already dead

The finding rests on five `dirty_xrefs` waivers that manufactured acyclicity, and on the
31 elementary cycles over 12 boundaries they hid.

**There is no active `dirty_xrefs` in the tree.** `git grep -nE 'dirty_xrefs' -- lib`
returns three hits, all of them comments or docstrings speaking of them in the past
(`networks/network.ex:27`, `networks/server.ex:18`, `visitors/visitor.ex:57`).

They were closed by the schema-owns-itself work (#1399): `Grappa.Networks.Network`
(`deps: [Grappa.IRC]`), `Grappa.Networks.Server` (`deps: []`) and
`Grappa.Visitors.Visitor` (`deps: []`) are now boundaries of their own.
`networks/network.ex:27` states it in the tree: "With the schema owning itself, the waivers
become either a declared edge to a leaf or nothing at all."

**=> The waivers, and with them the 31 cycles, are gone. Not waived — gone.**

## 2. `Grappa.Session.Control` was never the point, and must not be extracted

The cycles never ran through it. Measured lifecycle surface of `Grappa.Session`:
`start_session/3`, `stop_session/2,3`, `whereis/2`, `start_opts/*`. Their only
application-level dependency is `Session.Server` (`Server.registry_key/2`,
`Server.auto_away_debounce_for/1`, the `{Server, full_opts}` child spec) — everything else
is OTP (`DynamicSupervisor`, `Registry`, `Process`, `Logger`).

`Server` is already an export of the `Session` boundary. An extracted `Control` therefore
**cannot be a leaf**, because it needs `Server`. That is exactly why the review's version
[...]

PR #1488, opened 2026-08-17
docs(#1410): say that init/1 reads the DB, because it does

vjt:
Documentary slice for #1410 (L-S2, bucket D of the 2026-08-15 codebase review).
Three files, no behaviour change, no new tests.

## What was measured

The finding is alive on current `main`. `Session.Server.init/1` invokes the
injected `refresh_plan` closure before building any state, and there are **two**
injectors, not the one the issue counts: `Networks.SessionPlan` (user side) and
`Visitors.SessionPlan` (visitor side). Both walk the DB, on every spawn and every
`:transient` restart. The two sentences that said otherwise are both still in the
tree, and both are false — the A2 paragraph in `Grappa.Session` entirely, the
`init/1` comment in half.

## Why this slice corrects the prose instead of the code

The obvious remedy — move the re-resolve into `handle_continue` — is wrong, for a
boundary reason rather than a stylistic one. `init/1` returning `:ignore` is the
only **synchronous** "subject no longer viable" signal the spawn door has:
`SpawnOrchestrator` maps it to `{:ok, :ignored}` and `Bootstrap` counts it.
Deferred, the caller would observe `{:ok, pid}` followed by a silent death, which
is the swallow CLAUDE.md forbids at boundaries. That path is already pinned end to
end by an existing spec in `spawn_orchestrator_test.exs`, so the refactor would go
red rather than ship quietly — and so **no test is added here**, because the claim
being defended is already bought.

The reads are load-bearing for a second, independent reason: the #93 endpoint ring
resolves at the failure count, and that counter is authoritative only at this
instant (the bump is a synchronous call from the dying session's `terminate/2`).

So the code is right and the text was wrong. The text moves.

## What the corrected text now says

- `Grappa.Session` keeps the invariant A2 actually buys — no `Credential` /
  `Network` / `Server` / `Visitor` struct refs cross the boundary, which is what
  `Boundary` enforces — and drops the over-claim about runtime reads that the
  dependency mechanism never bought. Both injectors are named.
- The `init/1` comment keeps the upstream-socket half of "non-blocking" and
  retires the DB half.
- Both passages name what is **not** fixed: the spawn-door re-resolve stays
  redundant (`DynamicSupervisor` caches the child spec, so an "already fresh" flag
  would persist into every respawn — the opposite of what it must mean), and
  `Boundary` stays structurally blind to a closure that carries no module
  reference.

## Not established

Nothing was executed. No query log, no timing of the spawn loop, no `EXPLAIN`. The
issue's "N x 3 queries" stays a derivation from the call chain, and the corrected
text deliberately quotes **no** boot-time figure — it states that the cost exists
and is unmeasured. This slice closes the false statement, not the redundancy and
not the enforcement gap; those are named in the decision-log entry and left open.

#1398 (bucket I) reasons partly from the sentence corrected here, and is being
told its premise moved.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -47121,3 +47121,88 @@ three axes (timeout `1_000` ×11 vs `5_000` ×2, prefix `"USER"` ×12 vs `"USER 
 decision: two files have already found one second insufficient. Consolidating it
 is choosing a barrier value, not removing a duplicate, and it must be bought with
 its own reason.
+<!-- entry #1410 -->
+
+---
+
+## 2026-08-17 — #1410: `init/1` does read the DB; the two sentences that denied it now say so
+
+Two live documents claimed `Grappa.Session.Server.init/1` performs no DB reads.
+Both were false, in different ways and to different degrees, and this entry
+records which half of each was wrong, why the obvious code fix is the wrong fix,
+and what the correction deliberately leaves open.
+
+### What the code actually does
+
+`init/1` invokes the injected `refresh_plan` closure before building any state.
+There are **two** injectors, not one: `Grappa.Networks.SessionPlan` (user side)
+and `Grappa.Visitors.SessionPlan` (visitor side). Both walk the DB — the user
+side reads the credential, preloads `network: :servers` and fetches the user;
+the visitor side reads the visitor row, reloads the network, then resolves.
+So the reads happen on every spawn and every `:transient` restart, for both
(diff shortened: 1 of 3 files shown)


vjt:
Superseded by content: replayed into the two-way union PR #1492, merged to main as 5dbe6f68 (fast-forward). The replay rewrote the SHAs, so GitHub cannot close this automatically — verified landed by content (merge-base diff against main leaves zero lines this branch has and main lacks).

PR #1490, opened 2026-08-17
arch(#1398): pin the cycle count the acyclicity proof cannot see

vjt:
A ratchet for the half of the Boundary graph nothing watches. Two files, no
production code, no new behaviour.

## What this pins

The declared graph is acyclic and the compiler already enforces that half —
`mix.exs` puts `:boundary` in `compilers` with `check: [in: true, out: true]`.
The half nothing watches is `dirty_xrefs`: a waiver is a real dependency the
checker was told to ignore, so counting the five current waivers as the edges
they are yields **31 elementary cycles over 12 boundaries** while the build stays
green.

That population went from 3 waivers to 5 between the 2026-07-20 and 2026-08-15
reviews with nothing to notice it. Both times a human found it by reading
annotations. This gate makes the next one argue for itself in the diff that
introduces it.

It is explicitly **not** a fix for #1398. The direction that would zero the set is
recorded in the decision-log entry and deliberately left unspent.

## Why it reads the compiled attribute, not the source

The gate reads the `Boundary` module attribute the compiler persists — the same
annotation `GrappaWeb.BoundaryTest` and `Grappa.Visitors.VisitorBoundaryTest`
read, and what CLAUDE.md's architecture-test rule points at ("use `Boundary`
annotations — not string-matching").

That choice is load-bearing rather than stylistic. Two independently written
source parsers agreed that there are 31 cycles but **disagreed with each other
about how many `use Boundary` blocks exist at all** (99 against 107). Agreeing on
the output while disagreeing on the input is precisely the failure mode reading
the compiler's own record removes. Three measurements now agree on 31, and this
one has no parser of its own.

## Red first, and actually made to fail

The gate was committed green, then a mutant was applied to production source: a
sixth waiver on `Grappa.Mentions`, plus a real reference so the waiver is not
merely unused.

- With the mutant: **red — "cycle count rose to 37, above the pinned budget of 31"**,
  and all **6** new cycles listed in the failure message carry the injected
  `Mentions -> Networks` arc. One mutant, one assertion killed, the delta
  attributable.
- Mutant reverted, `git status --porcelain` empty, re-run: green.

`check.sh` was run twice. The first run **aborted at Credo** on a named discard
(`_mode`), which means nothing downstream of that phase had run — that is fixed in
the second commit and the suite was re-run from the start: Credo clean, 6522
tests / 0 failures, Sobelow 0 errors, Dialyzer passed, bats green.

## The blind spot, stated in the test body

The gate counts declared edges plus waivers and nothing else. The runtime
mechanisms this codebase actually leans on for cycle avoidance — the injected
closures, behaviours resolved from `:persistent_term`, the supervisor-level
`held_source_fn` — carry no module reference, so no compile-time mechanism can see
them and **none of them move this number**. A closure that opens a cycle passes
this gate silently. That limit is written into the test rather than left for a
reader to infer, because an undocumented blind spot gets read as coverage.

## Not established

The 31 is a count over declared plus waived edges on one commit; the true runtime
graph is denser and was not enumerated. Nothing here was tested against production
or a live session.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -47121,3 +47121,68 @@ three axes (timeout `1_000` ×11 vs `5_000` ×2, prefix `"USER"` ×12 vs `"USER 
 decision: two files have already found one second insufficient. Consolidating it
 is choosing a barrier value, not removing a duplicate, and it must be bought with
 its own reason.
+<!-- entry #1398 -->
+
+---
+
+## 2026-08-17 — #1398: pin the cycle count the acyclicity proof cannot see
+
+The declared Boundary graph is acyclic and the compiler enforces that: `mix.exs`
+puts `:boundary` in `compilers` with `check: [in: true, out: true]`, so a declared
+cycle never reaches CI. What nothing watches is the other half.
+
+### Why a waiver is an invisible edge
+
+A `dirty_xrefs` entry is a real dependency the checker has been told to ignore.
+Counting the five current waivers as the edges they are turns the acyclic graph
+into **31 elementary cycles over 12 boundaries**. That is not a new defect — it is
+the same one the 2026-08-15 architecture review found, and the one the 2026-07-20
+review found before it with three waivers instead of five. The population grew and
+nothing noticed, twice, until a human read the annotations by hand.
+
+This entry adds no fix. It adds a **ratchet**: a test that counts the cycles closed
+by declared deps plus waivers and fails when the number rises above a pinned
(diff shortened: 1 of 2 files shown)


vjt:
Superseded by content: replayed into the two-way union PR #1492, merged to main as 5dbe6f68 (fast-forward). The replay rewrote the SHAs, so GitHub cannot close this automatically — verified landed by content (merge-base diff against main leaves zero lines this branch has and main lacks).

Issue #1399, opened 2026-08-16
arch: 'a boundary needs only a schema' has three incompatible resolutions and #415's is the rarest (bucket J boundary-schemas)


vjt:
Bucket **J boundary-schemas** of the 2026-08-15 architecture review (`docs/reviews/architecture/2026-08-15-architecture-review.md`). One HIGH — an unfinished #415, where the partial application left the tree worse than before it started.

## What was measured

**Dependency A2 (HIGH).** One problem — "a boundary needs only a schema" — now has **three mutually incompatible resolutions**. #415 chose the right one (promote `Visitors.Visitor` to a `top_level?: true` leaf, retiring twelve waivers) and applied it to one of three populations:

- `Networks.Network` kept **5** `dirty_xrefs`.
- `Accounts.User` / `Session` are named as a real dep by **24** boundaries, of which **13 reference no `Accounts` function at all**.
- `Visitors.Visitor` is the promoted leaf.

`Grappa.Subject` is the sharpest illustration: a real dep on `Grappa.Accounts` for `User` and a real dep on the promoted `Visitors.Visitor` leaf — asymmetric treatment of the two halves of the polymorphism it models. (See also bucket C, which is the same module from the SSOT angle.)

## Why it matters

The result is worse than before #415: three patterns where there were two, and the correct one is the least represented. A reader cannot infer the rule from the tree, because the tree states it three ways.

## Proposed direction

Finish it in one mechanical commit: promote `Networks.Network` and `Accounts.User`/`Session` the same way, narrow the 13 struct-only deps, and **record the rule in CLAUDE.md next to the Boundary paragraph** so the next boundary that needs only a schema does not invent a fourth resolution.

## Related MEDIUM

**Dependency A7** — 11 `top_level?: true` boundaries live inside another boundary's namespace. Several are correct carve-outs, but nothing distinguishes a deliberate one from a context internal, so the module name is unreliable in 11 places. Worth settling in the same pass, since it is the same "what does this annotation mean" question.

## Not established

`mix boundary.find_violations` was not run; the populations were derived from the annotations and a whole-tree grep. Verify the 13 struct-only deps compile clean after narrowing rather than assuming the grep was exhaustive.

vjt:
## Recon: most of this issue's measurements have expired

Measured today on `origin/main` `5779c755`, and re-measured on the
review's own base `2ef9eeb0` (the commit that landed
`docs/reviews/architecture/2026-08-15-architecture-review.md`) so the
[...]

PR #1493, opened 2026-08-17
arch(#1399): drop seven Accounts deps that never carried an edge

vjt:
Slice A of bucket J (#1399): the seven boundaries that declared `Grappa.Accounts`
without ever holding a reference the checker could see now stop declaring it.

`Grappa.ChannelDirectory`, `Grappa.Notify`, `Grappa.Push`, `Grappa.QueryWindows`,
`Grappa.ReadCursor`, `Grappa.Scrollback`, `Grappa.UserSettings`.

**Declared ceiling: 9 files. Actual: 8** — the seven boundary files plus the
`docs/DESIGN_NOTES.md` entry. No test was added; see "Why no test" below.

## This contradicts the issue text, deliberately

#1399 asks to **narrow** the struct-only `Grappa.Accounts` deps. For these seven
there is nothing to narrow, and the proof is in the tree:

- Each reaches `Accounts.User` exactly twice, in one schema file, in the same two
  shapes — a `belongs_to :user, User` association argument and a
  `user: User.t()` typespec. Both are module names that become atoms in
  metadata, not remote calls, so no xref edge exists for Boundary to police.
- `Grappa.Uploads` holds that identical pair — `uploads/upload.ex:48` aliases
  `Grappa.Accounts.User`, `:75` declares `belongs_to :user, User` — with **no**
  `Grappa.Accounts` dep and no waiver, and compiles green today.
  `Grappa.TestSupport.SubjectSession` is a second instance, via a typespec.

So the review's "three incompatible resolutions" undercounts. A fourth —
*declare nothing* — was already live and already correct, and seven of the
thirteen belong to it. Narrowing them would encode a dependency the compiler
cannot see, on a module that is not a boundary today, and would have to be
undone the day `Accounts.User` is promoted. Deleting states what the compiler
can check. Where the prose and the measured tree disagree, the tree wins.

## The green here is worth something, because the red was armed first

Boundary reports cross-boundary references as **advisory warnings**; only
`mix compile --warnings-as-errors` — the first step of `ci.check`, for exactly
this reason (`mix.exs`) — makes them fail a build. A green compile after seven
deletions therefore proves nothing unless the checker is known to bite.

| step | command | result |
|---|---|---|
| baseline | `mix compile --force` at `d8b58e2d` | rc=0, 326 files |
| **mutant** | drop the `Grappa.Accounts` dep from `Grappa.Subject`, which holds a real `%User{}` reference, then `mix compile --force --warnings-as-errors` | **rc=1**, exactly **one** `forbidden reference to Grappa.Accounts.User`, at `lib/grappa/subject.ex:134`, `(references from Grappa.Subject to Grappa.Accounts are not allowed)` |
| restore | revert the mutant | `git status --porcelain` empty |
| **slice** | seven deps deleted, `mix compile --force --warnings-as-errors` | **rc=0**, `forbidden reference` count **0** |

One mutant, one assertion, the predicted line and the predicted target — nothing
else moved. Had any of the seven carried an edge the source measurement missed,
that is where it would have surfaced. The slice was designed to be falsified by
the compiler and was not.

## A stale number settled while the lane was held

The #1398 cycle-budget gate records an open question: two hand-written parsers
disagreed on how many `use Boundary` declarations the tree contains, **99 against
107**. The compiled attribute settles it at **99** — counted from the `Boundary`
module attribute over `:application.get_key(:grappa, :modules)`, the same source
`GrappaWeb.BoundaryTest` reads. The source measurement agrees exactly, and the
arithmetic explains itself: `lib` holds 111 textual occurrences of
`use Boundary`, 12 of them prose in moduledocs and comments, leaving 99.

**107 corresponds to no population that could be reconstructed here** (99
compiled, 99 line-anchored in `lib`, 104 adding `test/support`, 111 textual). No
explanation for it is offered rather than invented. The gate's comment still
carries the 99-vs-107 wording and lives in PR #1492, which this branch does not
touch — updating it there is a separate, one-line follow-up.

## Why no test

The compiler is the gate. `mix compile --warnings-as-errors` already fails the
build on a boundary violation, and CI runs it as the first step of `ci.check`, so
a test asserting "these seven do not declare `Grappa.Accounts`" would restate the
diff rather than check behaviour. The mutant above is the evidence that the
existing gate is live.

## What this deliberately leaves open

- **The other six of the thirteen are untouched.** Four hold real edges
  (`Subject`, `Themes`, `Admission`, `Vhosts`) and need `Accounts.User` and
  `Accounts.Session` promoted first. That change is larger than the review
  states: promoting them removes both from `Grappa.Accounts`' `exports:`, so
  every boundary holding a real edge must declare the new dep — measured at 19
  boundaries, not 11.
- **The `Networks.Network` half is not an annotation change** but a
  schema-cluster move, and is tracked separately.
- **A7 — the 11 nested `top_level?: true` boundaries — is not addressed.**
  Without a convention distinguishing a deliberate carve-out from a context
  internal there is no oracle to satisfy.
- **The rule is not written next to the Boundary paragraph in CLAUDE.md**, which
  is what #1399 asks for last. That file is not edited here; the edit is vjt's.
  Until it lands, a future boundary that needs only a schema can still invent a
  fifth resolution, so this issue is not closed by this PR.

## Not established

- Nothing here is exercised against production or a live session; these are
  compile-time annotations and no runtime behaviour changes.
- The seven deletions are proven by the compiler on this branch at this base.
  The source-level reference measurement that selected them is a regex over
  source, so a reference built by macro or `Module.concat/1` would be invisible
  to it — the compiler, not the parser, is what makes the claim safe.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -47291,3 +47291,80 @@ _Not established: nothing here measures how many of these workers can be alive
 at once, or what happens at any particular number, and the file says so. The
 question of whether that concurrency wants a bound is open and is not answered
 by this change._
+<!-- entry #1399 -->
+
+---
+
+## 2026-08-17 — #1399: seven boundaries stop declaring a dependency they never had
+
+Bucket J of the 2026-08-15 architecture review says one problem — "a boundary
+needs only a schema" — has three incompatible resolutions, and asks for the
+thirteen struct-only `Grappa.Accounts` deps to be *narrowed*. Measuring the tree
+first changed the answer for seven of them: there is nothing to narrow, because
+there was never an edge.
+
+`Grappa.ChannelDirectory`, `Grappa.Notify`, `Grappa.Push`, `Grappa.QueryWindows`,
+`Grappa.ReadCursor`, `Grappa.Scrollback` and `Grappa.UserSettings` each declared
+`Grappa.Accounts` in `deps:`. Each reaches `Accounts.User` exactly twice, in one
+schema file, in the same two shapes: a `belongs_to :user, User` association
+argument and a `user: User.t()` typespec. Both are module names that end up as
+atoms in metadata rather than as remote calls, so the xref-based checker has no
+reference to police. The dep bought nothing.
+
+**The tree already answered this, and nobody had noticed.** `Grappa.Uploads`
+holds the identical pair — `uploads/upload.ex` aliases `Grappa.Accounts.User`
+and declares `belongs_to :user, User` — with **no** `Grappa.Accounts` dep and no
+waiver, and it has been compiling green for as long as it has existed.
+`Grappa.TestSupport.SubjectSession` is a second instance, via a typespec. So the
+review's "three resolutions" undercounts: a fourth, *declare nothing*, was
+already live and already correct. Seven of the thirteen belong to it.
+
+**Why deleting rather than narrowing.** Narrowing `Grappa.Accounts` to
+`Grappa.Accounts.User` would encode a dependency the compiler cannot see, on a
+module that is not a boundary today, to satisfy a rule none of these seven
+actually violate — and it would have to be undone the day `Accounts.User` is
+promoted. Deleting states the truth that the compiler can check: these
+boundaries do not depend on `Grappa.Accounts`. Where an issue's prose and the
+measured tree disagree, the tree wins.
+
+**The deletion is falsifiable, and the falsifier was armed first.** Boundary
+reports cross-boundary references as advisory warnings; only
+`mix compile --warnings-as-errors` — the first step of the `ci.check` alias, for
+exactly this reason — turns them into a failed build. So a green compile after
+seven deletions proves nothing unless the checker is known to bite. Before the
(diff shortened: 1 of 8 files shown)


vjt:
Superseded by content, via #1496 -> #1511, which merged to `main` as 556bdd64. This one is a step removed from the other two: its commit was already folded into #1496 before #1511 existed, so #1511 deliberately did **not** cherry-pick it a second time — applying the same removal twice would have double-counted against the cycle-budget pin that #1496 had already recomputed with this change included.

Verified by content on the merged tree, not by reference:

* the seven `lib/` modules — `ChannelDirectory`, `Notify`, `Push`, `QueryWindows`, `ReadCursor`, `Scrollback`, `UserSettings` — each declared `Grappa.Accounts` inside its `deps:` block at f339b271 and none of them does now. 7 of 7.
* the DESIGN_NOTES entry: this branch's `<!-- entry #1399 -->` block is **byte-identical** to the one now on `main`, 77 lines, compared with `cmp`.

That second check is worth spelling out because I had previously declined to claim it — the earlier evidence was a same-title commit and a matching code diff, not a line-by-line comparison of the entry. It has now been run, so the claim is measured rather than inferred.

Issue #1403, opened 2026-08-16
arch: one new IRC verb costs 23 files across five mechanical plumbing layers (bucket N verb-plumbing)


vjt:
Bucket **N verb-plumbing** of the 2026-08-15 architecture review (`docs/reviews/architecture/2026-08-15-architecture-review.md`). One HIGH — the change-cost of the product's most routine extension.

## What was measured

**Extension A5 (HIGH).** One new IRC verb costs **23 files** across five mechanical server layers (controller → facade → `handle_call` → `Client` builder → meta), plus four client layers, plus 9 test files. This was measured as the union of the #1225 commits, not estimated. Every layer grew since the previous measurement. The parallel WS door adds **39** hand-written `do_handle_in` arms.

## Why it matters

Adding a verb is the single most predictable kind of work this project does, and it is currently the most expensive per unit of behaviour. The cost is entirely mechanical plumbing — none of the 23 files encodes a decision.

## Proposed direction

This is the one place a declarative per-verb table earns its keep: name, arity, target shape, echo policy, meta keys — consumed by the facade, the dispatcher and the builder, collapsing three server layers into one row. `slashCommands.ts` already has table shape, so codegen could make the client half free.

## Related MEDIUM

**Extension A7** — the Logger metadata allowlist is 306 lines / 106 atoms / **47% of `config/config.exs`**, and a one-directional sync test forces every scrollback meta key into it whether or not anything logs it; four entries carry apologetic comments saying exactly that. Same shape: a mechanical list that grows with every addition and that nothing derives.

## Not established

Nothing was executed. The 23-file figure is the union of one verb's commits (#1225) — a second verb might differ, and the table design has not been prototyped against the arms that carry policy rather than plumbing. Identify those arms before committing to a table, or the exceptions will live outside it and the count will not drop.

vjt:
Closed by the pipeline change ruled by vjt on 2026-08-20: **`status:soon` is retired**. An issue is now closed when its work **merges**, and the **milestone** records which release carries it — the label duplicated what the milestone already said.

This issue was verified individually before closing, not swept: its work is on `origin/main` under commits that are not documentation-only. No behaviour changes here, only bookkeeping.

<sub>Applied by vjt-claude on vjt's instruction.</sub>

PR #1496, opened 2026-08-17
arch(#1398): give the four Networks schemas their own boundaries, and retire every waiver

vjt:
The four `Grappa.Networks` schemas become their own boundaries, which retires
every `dirty_xrefs` waiver in the tree and takes the 31 declared cycles with
them. **This branch also contains #1399 slice A** (previously PR #1493), folded
in rather than shipped separately — see "Why the two are one branch" below.
#1493 should be closed by content at merge; the replay rewrote its SHA, so
GitHub cannot close it.

## What changes

`Network`, `Server`, `Credential` and `FeaturedChannel` each carry
`use Boundary, top_level?: true` with a deps list measured from their real
outbound edges — `Grappa.IRC` for three of them, nothing at all for `Server`.
A consumer that needs the struct now declares an edge to a leaf, and a leaf
closes nothing.

Five commits, one per schema plus the gate, each of which **compiles on its
own**:

| step | commit | files | `compile --force --warnings-as-errors` |
|---|---|---|---|
| 1 | `Networks.Server` | 2 | rc=0 |
| 2 | `Networks.FeaturedChannel` | 2 | rc=0 |
| 3 | `Networks.Credential` | 9 | rc=0 |
| 4 | `Networks.Network` + waivers retired | 12 | rc=0 |
| 5 | the cycle budget | 2 | rc=0 |

That was not free. The first pass of the same table was rc=1 on steps 3, 4 and
5, and the compiler named two consumers the source measurement had missed —
`Grappa.Operator` (`%Networks.Network{}`, qualified through the parent alias)
and `Grappa.TestSupport.SubjectReset` (`Credential.changeset/2`, reached
through an `alias Grappa.{…, Networks, Networks.Credential}` block whose keys
collide). Both fixes are folded into the step that introduced the breakage, and
the table above is the re-run.

## The premise this was planned under was wrong

Two earlier reconnaissance passes, one of them mine, concluded that four
separate leaves were impossible — the three sibling schemas name `Network`
back, so four leaves looked like four two-cycles, and the only expressible
form looked like one boundary holding all four schemas. Since no namespace
contains exactly those four, that meant moving files and renaming modules:
roughly fifteen files of rename and a cold deploy.

None of it was needed. Every reference among the four is a `has_many` or
`belongs_to` argument and a typespec, and #1399 had just established against
the compiler that those are not references the checker resolves. No back-edge,
no cycle, nothing to merge into a namespace. **No module is renamed, no file
moves, and the deploy stays hot.**

The lesson is about the measurement: counting references without classifying
them measures the wrong population. Both earlier passes counted names and got a
real number that did not answer the question asked of it.

## What the gate now says, read rather than asserted

`@cycle_budget` moves 31 → 0. The number was **read, not assumed**: an
`assert count <= budget` never reveals the count, so the gate was forced red
once with the budget at -1.

```
Boundary cycle count rose to 0, above the pinned budget of -1.
Cycles now closed (shortest first):
```

Zero, with an empty list. Restored afterwards, `git status --porcelain` empty.

**The mutant.** The gate exists for edges the compiler cannot see, so the
mutant is a waiver, not a declared dep — a declared `Server -> Networks` edge
is caught by the boundary compiler itself and the gate would never speak. With
`dirty_xrefs: [Grappa.Networks.Wire]` on `Server` (a module `Grappa.Networks`
owns), the gate goes red with exactly one cycle, exactly the predicted one:

```
Boundary cycle count rose to 1, above the pinned budget of 0.
      Grappa.Networks -> Grappa.Networks.Server -> Grappa.Networks
```

One mutant, one cycle, the predicted pair. Restored, tree clean.

## A question this branch settles, and one it does not

**Settled: a module passed as a plain argument is not a reference Boundary
resolves.** `Notify` and `QueryWindows` pass the schema module to
`check_exists/4` the way `Admission` passes it to `Repo.get/2`. Source reading
does not answer whether that is a reference, so both were written *without* the
dep — the falsifiable choice, since an undeclarable dependency asserted anyway
fails silently while a missing one fails at the build. They compiled green in a
run whose checker was demonstrably live (it named three violations elsewhere in
the same table). `Admission` cannot serve as the witness here: it declares the
dep regardless, having real struct references.

**Not settled: Boundary's two checks disagree about those waivers.** Boundary
also warns when a `dirty_xrefs` entry is *unnecessary*, and that warning fails
the build under `--warnings-as-errors` — the mutant tripped it. Yet
`origin/main` compiles with **zero** such warnings while carrying all three
waivers this branch deletes, so Boundary considered them used; and deleting
them here produces no forbidden reference. Both halves are measured, they are
not obviously compatible, and the mechanism is not guessed at. The DESIGN_NOTES
entry withdraws the wider claim an earlier draft made — that the waivers
"never covered anything" — and states only what the evidence supports.

⚠️ One commit message, on the `Networks.Network` step, still carries that
withdrawn wording. It is left as written rather than rewritten silently; say
the word and it gets reworded.

## Why the two are one branch

#1493 was cut from `d8b58e2d` and main has moved to `5dbe6f68`, so it is not
fast-forwardable, and five of its seven files are files this promotion also
edits. Rebasing it separately buys a second full gate run and leaves this
branch to rebase onto the same five files afterwards. Gating the union once is
cheaper and more honest: the two changes had never been compiled together, and
their overlap is exactly where a conflict would hide. Where both touched the
same `deps:` block the two comments are merged into one — they were making the
same point about two different modules.

The `docs/DESIGN_NOTES.md` union merge was checked for the tail-fusion failure
mode: both entries carry unique `<!-- entry -->` markers, both boundary shapes
are intact, and the file's contribution is additions-only with zero deletions.

## Gates

| gate | result |
|---|---|
| per-step `compile --force --warnings-as-errors` | rc=0 on all five |
| union `compile --force --warnings-as-errors` | rc=0, zero forbidden references |
| the cycle gate | rc=0; count measured at 0 via a forced red |
| the mutant | gate rc=2, exactly one cycle, the predicted one |
| `scripts/design-notes-gate.sh` | "2 new entry heading(s), separator and marker present" |
| `scripts/check.sh` | rc=0 — Credo "found no issues", 6537 tests 0 failures, Sobelow "Total errors: 0", Dialyzer "passed successfully", bats 564 ok / 0 not ok |

## Declared ceiling, and the overrun

The plan declared 20 files. **This is 21 sources plus #1399's contribution**:
`lib/grappa/test_support/subject_reset.ex` was the file the parser missed, and
it is named here as an overrun rather than absorbed. The other missed consumer,
`Grappa.Operator`, was already inside the twenty.

## Not established

- No behaviour changes: these are compile-time annotations. Nothing here was
  exercised against production or a live session.
- The reference measurement that selected the file set is a regex over source,
  and it demonstrably missed two consumers. The compiler, not the parser, is
  what makes the set safe.
- The gate counts DECLARED deps plus waivers. Injected closures, behaviours
  resolved from `:persistent_term` and the supervisor-level `held_source_fn`
  carry no module reference any compile-time mechanism can see, so the runtime
  graph stays denser and a closure that opens a cycle moves nothing.
- A new limit, recorded because it inverts how a future waiver reads: with the
  schemas as leaves, a `dirty_xrefs:` aimed at one of them closes no cycle and
  this gate stays green. The 31 were never about waivers as such, but about
  waivers aimed at a module owned by a large boundary.

## Base note

Every gate above was measured with this branch based on `5dbe6f68`. `main`
moved to `c2e64901` during the run (three #1403 commits). The overlap is
`docs/DESIGN_NOTES.md` only — a `merge=union` file, and both entries here carry
unique markers, which is the protection that exists for exactly that merge.
**No Elixir source file overlaps**, so nothing in the green above is contested
by the move; it simply was not re-measured on top of `c2e64901`.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -47539,3 +47539,80 @@ reference built by macro or `Module.concat/1` would be invisible to it; the
 compiler, not the parser, is what makes the set safe, and a file appearing
 beyond the declared set would mean the parser missed a consumer rather than
 that the set grew._
+<!-- entry #1399 -->
+
+---
+
+## 2026-08-17 — #1399: seven boundaries stop declaring a dependency they never had
+
+Bucket J of the 2026-08-15 architecture review says one problem — "a boundary
+needs only a schema" — has three incompatible resolutions, and asks for the
+thirteen struct-only `Grappa.Accounts` deps to be *narrowed*. Measuring the tree
+first changed the answer for seven of them: there is nothing to narrow, because
+there was never an edge.
+
+`Grappa.ChannelDirectory`, `Grappa.Notify`, `Grappa.Push`, `Grappa.QueryWindows`,
+`Grappa.ReadCursor`, `Grappa.Scrollback` and `Grappa.UserSettings` each declared
+`Grappa.Accounts` in `deps:`. Each reaches `Accounts.User` exactly twice, in one
+schema file, in the same two shapes: a `belongs_to :user, User` association
+argument and a `user: User.t()` typespec. Both are module names that end up as
+atoms in metadata rather than as remote calls, so the xref-based checker has no
+reference to police. The dep bought nothing.
+
+**The tree already answered this, and nobody had noticed.** `Grappa.Uploads`
+holds the identical pair — `uploads/upload.ex` aliases `Grappa.Accounts.User`
+and declares `belongs_to :user, User` — with **no** `Grappa.Accounts` dep and no
+waiver, and it has been compiling green for as long as it has existed.
+`Grappa.TestSupport.SubjectSession` is a second instance, via a typespec. So the
+review's "three resolutions" undercounts: a fourth, *declare nothing*, was
+already live and already correct. Seven of the thirteen belong to it.
+
+**Why deleting rather than narrowing.** Narrowing `Grappa.Accounts` to
+`Grappa.Accounts.User` would encode a dependency the compiler cannot see, on a
+module that is not a boundary today, to satisfy a rule none of these seven
+actually violate — and it would have to be undone the day `Accounts.User` is
+promoted. Deleting states the truth that the compiler can check: these
+boundaries do not depend on `Grappa.Accounts`. Where an issue's prose and the
+measured tree disagree, the tree wins.
+
+**The deletion is falsifiable, and the falsifier was armed first.** Boundary
+reports cross-boundary references as advisory warnings; only
+`mix compile --warnings-as-errors` — the first step of the `ci.check` alias, for
+exactly this reason — turns them into a failed build. So a green compile after
+seven deletions proves nothing unless the checker is known to bite. Before the
+deletions, one mutant removed the `Grappa.Accounts` dep from `Grappa.Subject`,
+which holds a real reference (`%User{}` in `from_assigns/1`). The build went red
+with exactly one violation, at exactly that line, naming exactly
+`Grappa.Accounts.User` — one mutant, one assertion, nothing else moved. With the
+mutant reverted and the seven deps gone, the same command is green with zero
+forbidden references. Had any of the seven carried an edge the source
+measurement missed, this is where it would have surfaced.
+
+**A stale number, settled while the lane was held.** Two hand-written parsers
+had disagreed about how many `use Boundary` declarations the tree even contains
+— 99 against 107 — and that disagreement was written into the #1398 cycle-budget
+gate as an open question. The compiled attribute settles it: **99**, counted from
+`Boundary` module attributes over `:application.get_key(:grappa, :modules)`. The
+source measurement agrees exactly, and the arithmetic explains itself — `lib`
+holds 111 textual occurrences of `use Boundary`, of which 12 are prose in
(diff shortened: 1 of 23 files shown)


vjt:
Superseded by #1511, which merged to `main` as 556bdd64. GitHub cannot close this one on its own: #1511 cherry-picked these six commits onto a fresh base, so their SHAs were rewritten and this head is no longer reachable from `main`. Closing by content, not by reference.

All six commits landed: the four `Networks` schema boundaries, the waiver retirement, and the recomputed cycle budget. Verified after the merge — the pinned `@cycle_budget` is 0 and the gate measures 0.

One defect in this branch was found and fixed inside #1511: `0982fa87` promoted `Grappa.Networks.Network` to its own boundary and updated twelve consumers under `lib/` plus `Grappa.TestSupport.SubjectReset`, but missed `Grappa.AuthFixtures`, whose boundary still declared only `Grappa.Networks`. That made the test-env compile fail under `--warnings-as-errors`. It went unnoticed here because `test/support/` enters `elixirc_paths` only under `MIX_ENV=test` and this PR was permanently conflicted, so it never had a CI run to compile it in that env. #1511 carries the completing commit.

Issue #157, opened 2026-06-29
delete my account: quit + full account wipe (registered visitor & user, not admin)


vjt:
## Summary

Add a "delete my account" lifecycle action: it performs a **quit** (closes cicchetto and the IRC session) and then **fully wipes** the account and all associated state.

## Behaviour

- Available to **registered visitor** and **registered user**.
- **Not** available to **admin** users.
- Semantics: quit (tear down client + IRC session) followed by full deletion of the account record and associated session/scrollback state — irreversible.

## Relation to #126

#126 covers the detach vs quit separation (P0):
- detach = logout cicchetto, leave session up (extend to registered visitor).
- quit = close cicchetto + IRC session (visitor wipes everything; user tears down session).

This issue builds the destructive account-removal action on top of quit. Requires a confirmation step on the cicchetto side given it is irreversible.

## Severity

P1.

## Notes

Requested by vjt on #sbiffo (extension of #126).

vjt:
Shipped to prod (m42) 2026-06-29 — `5775e55`.

**Self-service "delete my account"** — the explicit, irreversible total wipe of the caller's OWN account + all state. Distinct from #126's quit (which PRESERVES a persistent identity): this is the ONLY self-service door that destroys one; the server NEVER wipes on quit.

- **Server**: `Grappa.AccountDeletion.delete_account/1` — one subject-routed verb behind `DELETE /me` (thin controller). Admin user + anon visitor → 403 (not for admins per this issue; an anon has no persistent identity to delete). Non-admin user → stop ALL live sessions then `Accounts.delete_user`; registered visitor → `Session.stop_session` then `Visitors.delete`. Teardown-before-wipe ordering mirrors `Operator.delete_visitor/2`; DB `ON DELETE CASCADE` wipes every dependent. Socket close shared via the extracted `UserSocket.disconnect_subject/1`.
- **cic**: a SEPARATE, more-heavily-guarded affordance — `DeleteAccountModal` keeps the destructive button disabled until you type your exact account name/nick (stronger than quit's two-tap). Gated to a registered non-admin user or registered visitor; admins + anon visitors never see it.
- **Tests**: server unit (`AccountDeletionTest`, incl. the quit-preserves-vs-delete-wipes boundary), `me_controller_test` (`DELETE /me` contract + socket broadcast), cic vitest (lifecycle/modal/drawer-gating), real chromium e2e (`issue157-delete-account`, RED-proven against feature-absent code).

[...]

Issue #126, opened 2026-06-28
detach disconnects upstream session but cicchetto still shows connected


vjt:
## Summary

Detach is broken: invoking detach disconnected the **upstream** IRC session, but cicchetto kept believing it was still connected. Client/upstream state desync — the UI shows a live connection while the server-side session is gone. On top of that, the two lifecycle actions (`detach` vs `quit`) are conflated and not offered to the right subject types.

## Expected

`detach` keeps the upstream session alive (bouncer-style: the web client disconnects but the server-side session persists). Once detach stops tearing the session down, the post-detach "still connected" desync disappears (DB state and live pid agree again).

## Subject classes (the split is by NickServ identity)

The whole design hinges on **persistent identity**. There are three:

1. **registered user** — a `users` row. Has a `Networks.Credential` with a `connection_state` enum (`:connected | :parked | :failed`); cic reads it for the "connected" display, and state changes broadcast `connection_state_changed`.
2. **registered visitor** — **a visitor with a NickServ identity** (they have identified to NickServ). This is the *defining* trait. **Persistent identity** — their row + scrollback survive across sessions and never auto-expire.
3. **ephemeral / anon visitor** — a visitor with **no NickServ identity**. **No persistent identity** — Reaper-swept; `purge_if_anon` deletes the row on logout.

Why this matters: a NickServ identity is something the visitor can come back to, so its data must survive a `quit`. An ephemeral visitor has nothing to come back to, so there is nothing to preserve.

**Implementation discriminator (VERIFY in code — do not assume):** the likely persisted signal for "registered visitor = NickServ identity" is `visitors.password_encrypted` being non-nil (grappa stores the NickServ credential so the visitor auto-identifies). Confirm this field faithfully represents "has a NickServ identity" before keying detach-eligibility / quit-preservation on it. If there is a gap between `password_encrypted` and actual NickServ-identified state, **halt + flag it** — the authoritative concept is the NickServ identity, the column is only its representation.

## The two verbs (subject × verb matrix)

The model is the full **(web client × upstream IRC)** state matrix. Each transition is a named verb:

| | upstream **UP** | upstream **DOWN** |
|---|---|---|
| web **UP** | (normal connected) | **`disconnect`** (drop upstream, stay in cic) ⇄ **`reconnect`** |
| web **DOWN** | **`detach`** (bouncer: leave cic, keep upstream) | **`quit`** (close cic + tear down upstream) |

Per-subject availability:

[...]

PR #1498, opened 2026-08-17
test(#1397): hoist await_handshake onto Grappa.IRCServer with an explicit timeout

vjt:
Slice **F1** of #1397 (bucket H test-harness): the thirteen local
`await_handshake/1` wrappers become one `IRCServer.await_handshake/2`, and the
timeout stops being hidden.

This follows the `passthrough_handler` slice (`412579cb`) in shape — hoist onto
the module that already owns the domain, call it qualified, touch no case
template — so it stays independently reviewable.

## The archaeology reverses the issue's reading of the timeout

The issue holds `await_handshake` out of consolidation because "two files have
already found one second insufficient", making the timeout "a live decision".
**That is withdrawn.** Reading both files at the commit that introduced their
wrapper:

| site | introduced | value at birth |
|---|---|---|
| `test/grappa/account_deletion_test.exs` | `5775e55a` (#157, 2026-06-29) | `5_000` |
| `test/grappa_web/controllers/session_controller_test.exs` | `5e6f7c53` (#126, same day) | `5_000` |

Neither was ever raised from `1_000`, because neither was ever at `1_000`. Two
new files written the same day by authors who picked a wider budget is not
evidence about one second.

And the majority was not chosen either: eight of the ten `1_000` copies trace to
`84fe0850` (2026-05-08), which removed the default timeout from
`wait_for_line/3` under the CLAUDE.md no-default-arguments rule and stamped the
old default explicitly at ~150 call sites. The thirteen local wrappers then
re-hid that same value behind a zero-argument helper.

## So the hoisted verb takes the budget as an argument

`IRCServer.await_handshake(server, timeout)` has no default. Adding one would
undo `84fe0850` on the same function family and would silently halve the two 5s
sites the moment their `defp` disappeared. Eleven files keep 1s, two keep 5s,
each spelled where it is paid for. 380 call sites, each carrying its own value —
that is the same shape `84fe0850` chose for ~150 calls, and it is noisy on
purpose.

The other two axes are duplication and they close: the alias axis dies with the
definitions, and the prefix axis closes on the strict `"USER "` (`lib/` sends
exactly one line beginning with `USER` and no `USERHOST` at all, so both
spellings select the same line — a test asserts `USERHOST` does not satisfy the
barrier).

## The arity is pinned by a test, and the pin was bought

`refute function_exported?(Grappa.IRCServer, :await_handshake, 1)`. Mutant:
re-adding the default argument to the definition. Result — 4 tests, 1 failure,
exactly that assertion, the other three still green. Committed before mutating,
reverted after, working tree verified clean.

## While the compile lane was held: the collision claim is no longer inferred

The DESIGN_NOTES entry from the first slice states the import/local collision as
"the documented Elixir rule plus the 110-to-0 evidence ... not a build we watched
fail". It has now been watched. Adding `admin_session/0` to `Grappa.AuthFixtures`
and compiling one of the fourteen files that define it locally while importing
the module wholesale:

```
error: imported Grappa.AuthFixtures.admin_session/0 conflicts with local function
== Compilation error in file test/grappa_web/controllers/admin/circuit_controller_test.exs ==
```

The mutation was reverted; nothing of it is in this branch. The consequence for
the bucket is now firm in kind rather than in direction: the `admin_session`
slice is atomic across fifteen files.

## Declared overrun: 15 → 16 files

The ceiling declared for this slice was 15. It came out at **16**, and the
sixteenth is `docs/DESIGN_NOTES.md`. Thirteen converted files plus the
destination `test/support/irc_server.ex` is 14; the new
`test/grappa/irc_server_test.exs` is 15; the decision-log entry is 16.

The overrun is entirely the retraction above. Splitting it into a docs-only PR
would separate the withdrawn claim from the measurement that supports it, so it
rides here, declared rather than absorbed.

## Gates

Full `scripts/check.sh`, RC read from file, every phase marker read one by one:

- Credo — `6124 mods/funs, found no issues.`
- ExUnit — `8 doctests, 61 properties, 6542 tests, 0 failures`
- Sobelow — `Total errors: 0, Skipped: 0, Unnecessary Skips: 0`
- Dialyzer — `done (passed successfully)`
- bats — 564 `ok`, 0 `not ok`

An earlier run was red on Credo (a named `_sock` discard where the file's
strategy is `_`). It was fixed and `check.sh` re-run from the top, because
`check.sh` aborts at the first red phase and the green below it did not exist.

`scripts/design-notes-gate.sh` separately: `1 new entry heading(s), separator and
marker present.` — the measuring line, not `nothing to check`. Entry numstat: 74
additions, 0 deletions.

## Not established

- **Whether one second is the right budget.** The argument for changing it was
  falsified; the value was not measured. No suite timing, no attributed flake,
  no Argon2 number. The design deliberately avoids having to answer.
- **The prefix change beyond what was measured.** `"USER "` holds because `lib/`
  has one USER-prefixed send and zero `USERHOST` today. Static measurement over
  `lib/`, not a guarantee about future senders.
- **That the green transfers to another tree.** `_build`, `deps` and `priv/plts`
  are shared across worktrees on this host and no forced rebuild was done; the
  run is on base `c2e64901`.
- **Anything against production or a live session.** This is the suite.
- **The rest of the bucket.** `start_server` (21 clauses over 20 files) and
  `admin_session` (14, one shape) are queued behind this. The fixture trio needs
  a ruling first: substituting `AuthFixtures`' placeholder hash for a real Argon2
  one changes what fourteen files prove, and the local `visitor_fixture` calls
  the production provisioning verb the canonical one bypasses.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -47516,3 +47516,77 @@ _Not established: how many of the remaining 105 atoms are actually logged. The
 scan proves absence, never disuse, so the count of dead entries is a floor and
 not a measurement. The 39-arm `do_handle_in` door and the per-verb table that
 #1403's HIGH proposes are untouched by this work._
+<!-- entry #1397-f1 -->
+
+---
+
+## 2026-08-17 — #1397 F1: the handshake barrier takes its timeout as an argument, and the "live decision" was a removed default
+
+The #1397 entry above closes by holding `await_handshake` out of the
+qualified-call reasoning, on the grounds that its timeout axis is "a live
+decision: two files have already found one second insufficient." **That
+sentence is withdrawn. The history does not say it.**
+
+Both 5s sites were born at 5s. `test/grappa/account_deletion_test.exs` arrived
+in `5775e55a` (#157, 2026-06-29) and
+`test/grappa_web/controllers/session_controller_test.exs` in `5e6f7c53` (#126,
+same day); reading each file at the commit that introduced its wrapper shows
+`5_000` already there. Neither was ever raised from `1_000` in response to a
+failure, because neither was ever at `1_000`. Two new files written the same
+day by authors who picked a wider budget is not evidence that one second is
+too tight — it is not evidence about one second at all.
+
+**And the 1s majority was never chosen either.** Eight of the ten majority
+copies trace to a single commit, `84fe0850` (2026-05-08), which removed the
+default timeout from `IRCServer.wait_for_line/3` under the CLAUDE.md
+no-default-arguments rule and stamped the old default explicitly at ~150 call
+sites. The value the copies agree on is the value the default used to hide.
+Thirteen local wrappers then promptly re-hid it behind a zero-argument helper —
+the rule was obeyed at one layer and undone at the next, in the same file.
+
+### So the hoisted verb has no default
+
+`IRCServer.await_handshake(server, timeout)` requires the budget. Giving it a
+default would re-hide, on the same function family, exactly what `84fe0850`
+went to un-hide, and it would silently halve the two 5s sites the moment their
+`defp` disappeared. The timeout was never duplication; it was a parameter with
+thirteen hard-coded callers. It stays at the call site, where the test that
+pays for it can be read.
+
+The other two axes are duplication and they close. The alias axis dies with the
+local definitions. The prefix axis closes on the strict `"USER "`: `lib/` sends
+exactly one line beginning with `USER` (`auth_fsm.ex`) and no `USERHOST`
+anywhere, so both spellings select the same line — the strict one also says so,
+and a test asserts that `USERHOST` does not satisfy the barrier.
+
+### The arity is pinned by a test, not by the comment above it
+
+A rule stated in a docstring is a rule until the next session. `refute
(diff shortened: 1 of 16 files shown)


vjt:
Superseded by #1511, which merged to `main` as 556bdd64. GitHub cannot close this one on its own: the four commits were cherry-picked onto a fresh base and their SHAs were rewritten, so this head is not reachable from `main`. Closing by content.

All four landed: the `Grappa.IRCServer.await_handshake/2` barrier with no default timeout, the retirement of the thirteen local wrappers, the DESIGN_NOTES entry and the Credo fix.

One thing changed on the way in, and it is the reason this could not simply be rebased. #1500 landed on `main` after this branch was cut and added one call in the old wrapper form, far from every edit here in a 13k-line file. Git auto-merges that **cleanly** — no conflict, no marker — and the result does not compile, because the `defp` it resolves to is deleted in the same range. The hazard was the absence of a conflict, not a conflict. In #1511 the wrapper-retiring commit converts that call too, in the same commit, at the same 1_000 budget the deleted `defp` supplied.

Measured after the fact on the merged tree: zero old-form calls remain in `test/` or `lib/`, and zero `defp await_handshake` survive. #1508 also landed in that window and added 86 lines to the same file, but added no call sites — checked both by direct diff and by walking the count across the whole chain.

Issue #1518, opened 2026-08-18
cic: service-modal declares an ordering invariant in a comment that no test buys


vjt:
Found while moving `service-modal` behind the #1396 command seam, by a mutant that was run precisely to check whether the move preserved behaviour. It did — and it surfaced this instead.

**The claim.** The arm's comment states that `openServiceModal` must run *before* `help`, to capture the high-water mark.

**The measurement.** Swapping the two produces **0 reds, before the move and after it**. `effectSignature()` orders by explicit choice, and the `it.each` cases assert *which calls happen*, not *in what sequence*. So the ordering invariant is written in prose and bought by nothing.

Two possible resolutions, and they are not equivalent:
- the order genuinely matters → a test must buy it, and today none does;
- the order does not matter → the comment is wrong and should go.

Either way the current state is the bad one: a reader takes the comment as a constraint and preserves an ordering that nothing would catch them breaking. Same class as #1509 — a comment asserting more than the code enforces.

**Deliberately not fixed in the slice that found it.** Closing it there would have changed the "after" set against which the move itself was proved, which would have weakened the very measurement that surfaced it. It needs its own slice and its own red.

Related: #1396 (the seam), #1509 (comments that assert more than the code).

vjt:
Closed by the pipeline change ruled by vjt on 2026-08-20: **`status:soon` is retired**. An issue is now closed when its work **merges**, and the **milestone** records which release carries it — the label duplicated what the milestone already said.

This issue was verified individually before closing, not swept: its work is on `origin/main` under commits that are not documentation-only. No behaviour changes here, only bookkeeping.

<sub>Applied by vjt-claude on vjt's instruction.</sub>

Issue #1513, opened 2026-08-18
cic: the send-pipeline axis in compose.ts blocks six command arms from the #1396 seam


vjt:
Split out of #1396 (bucket G). #1396 moved 47 of 59 command arms behind the dispatch seam; six more are out by an explicit ruling (no red protects them: `error`, `open-settings`, `admin`, `oper`, `alias-define`, `unalias`). The remaining **six are stuck for one single reason**, and that reason is this issue.

**They are not six exceptions — they are one axis.** `service-modal`, `ame`, `amsg`, `privmsg`, `me`, `msg` all reach the send pipeline. This was established by measurement, and it twice reversed an orchestrator ruling that had assumed otherwise.

## What would have to be extracted, and what blocks it today

The cluster is `planSends` + `drainPaced` plus their private satellites (`sleep`, `isSendThrottled`, `retryAfterMs`, `MAX_PACED_RETRIES_PER_SEND`, `DEFAULT_RETRY_AFTER_MS`, `MAX_RETRY_AFTER_MS`, the `PacedSend` type), with three doors above it: `sendBodyLines` (exported, consumed by `ServiceModal.tsx` and `RegistrationWizardModal.tsx`), `sendFanOut` (private, one use), and `sendPacedBody` (a factory closure).

Two distinct blocks:

1. **An import cycle.** `planSends`/`drainPaced` rest on `draftLines` and `wireBody`, which `compose.ts` **exports** and which are consumed outside it (including `draftFramePreview` in the same file). Extracting the cluster without those two leaves the new module re-importing `compose.ts` — a cycle, i.e. exactly the class this bucket is supposed to remove, and one `noImportCycles` does not currently police.
2. **`sendPacedBody` is store, not pipeline.** It closes over `claimDrafts`/`getDraft`/`writeState`. It does not belong in a send module. Either it stays in the factory and the three free-text arms stay with it, or the context record grows to carry the send door as well — and that widening has been refused twice, on the grounds that the cost is the precedent, not the hour of work.

## The red that would prove the extraction

Suite-green does not prove it: a pure extraction passes green by construction, and would pass green even having broken retry semantics nobody watches.

- **Primary:** a mutant on the heart of the cluster, run before and after the move, that reddens **today** — best candidate is `drainPaced`'s pacing logic (e.g. advance `sent` even on the 429 reject branch, so a rejected line counts as delivered). Same tests reddening before and after means the extraction carried the *behaviour*, not just the symbols.
[...]

Issue #1396, opened 2026-08-16
arch: compose.ts is store, send pipeline and a 59-arm command table in one module (bucket G client-layering)


vjt:
Bucket **G client-layering** of the 2026-08-15 architecture review (`docs/reviews/architecture/2026-08-15-architecture-review.md`). One HIGH plus four gating MEDIUMs — the client-side counterpart of bucket A.

## What was measured

**Responsibility A4 (HIGH).** `compose.ts` (2,345 lines) is three jobs, and the biggest is one function: a `switch (cmd.kind)` with **59 arms** covering 49+ IRC verbs, fanning out to REST posts, ~20 WS pushes and modal openers, while the same module also owns the per-window composer store and the send pipeline.

**Two corrections to an earlier version of this body, both re-measured on `origin/main` `e974baed`:**

- The 59-arm switch lives in **`dispatchDraft`** (`cicchetto/src/lib/compose.ts:826`, switch at `:903`), **not** in `submit` (`:2060`, which calls `dispatchDraft` at `:2103`). The earlier text named the wrong function.
- The verb closed-set is declared **once, not twice**: `SlashCommand` in `slashCommands.ts:100` (61 union arms) is the single declaration, and the case labels are consumers whose exhaustiveness tsc already enforces — `default: { const _exhaustive: never = cmd }` at `compose.ts:2007`. There is no second hand-maintained list to drift.

The HIGH stands on the size and fan-out of one function, not on a duplicated verb set.

## Gating MEDIUMs in this bucket

- **Abstraction A8** — `api.ts` is 3,159 lines / 87 exports / 112 types, imported by 34 of 79 components, holding transport **plus** the shared type vocabulary **plus** domain derivations (`ownNickForNetwork`, `displayNick`, `isContentKind`) **plus** a module-global 401 hook. The domain helpers are why every component imports the transport module.
- **Abstraction A9** — `cicchetto/src/lib/` is a flat 218-module namespace with no layer distinction, and the shared-verb rule is already fraying: four spellings of the join verb, two of which differ on whether the channel is folded (`DirectoryPane.tsx:121` folds, with a five-line #731 comment; `HomePane.tsx:86` passes the raw name, safe only because the server happens to store folded channels).
[...]

PR #1526, opened 2026-08-18
cic(#1396): the fan-out pair leaves the switch, at zero new fields

vjt:
## What

`/ame` + `/amsg` move out of the `dispatchDraft` switch into
`lib/commands/fanout.ts`, behind the same seam as the other 54 arms.
`CommandContext` gains **zero** fields.

## Why it was possible, against the standing premise

The pair was believed blocked by "the pipeline axis". Measured on
`d60a7a72`, that is false: every value the arm reads is an ordinary module
import (`sendFanOut`, `requestConfirm`, `joinedChannelsOnNetwork`,
`friendlyError`) and it touches the composer buffer nowhere.

Two lifts under #1513 made that true, neither aimed at this pair, and they
are **not the same kind of change**:

- `6ef878fe` moved `sendFanOut` into `./sendPipeline`. It had never been a
  closure — a module-level const that `commands/` could not import because
  it was never exported.
- `5779c755` lifted `joinedChannelsOnNetwork` out of the controller
  closure. That one genuinely had been a closure.

The real boundary for the three arms that remain (`privmsg`, `me`, `msg`)
is the **draft store** — `sendPacedBody` closes over `claimDrafts` /
`releaseDrafts` / `getDraft` — not the send pipeline. The #1513 ruling on
those three stands on its own reason (#737) and is not reopened here.

`FANOUT_CONFIRM_THRESHOLD` travels with the handler: one reader, and the
move keeps it that way.

## Evidence

Five mutants on the arm, run against the **full** unit suite before and
after, comparing the SET of killing tests, not just the count:

| mutant | before | after | same set |
|---|---|---|---|
| fan-out sends to ONE channel | 7 | 7 | yes |
| confirm gate never fires | 4 | 4 | yes |
| ACTION framing forced off | 1 | 1 | yes |
| empty-target guard removed | 1 | 1 | yes |
| target list left unsorted | 1 | 1 | yes |

Gates: `bun run test` 290 files / 5615 tests green before and after;
`bun run check` rc=0 with a warning delta of zero against clean `main`
(59 = 59, none in the new file); `scripts/bats.sh` 564 ok / 0 not ok;
`design-notes-gate.sh` green.

## A degeneration measured and deliberately NOT fixed here

The #1396 characterization net does not protect the fan-out. Its `ame` and
`amsg` rows submit into eleven joined channels **leaked from the preceding
`describe`**, so both stop at the confirm gate and send nothing. Proof is
in the table above: the fan-out mutant kills seven tests and none is a net
row, while the gate mutant kills two net rows. Established by displacement,
not inference — seeding the net with two channels flips both inline
snapshots.

Left alone on purpose: the move is bought by the dedicated `#431`
describes, the degeneration is identical on both sides, and strengthening
the net changes two inline snapshots and needs its own red. Same call as
#1518.

## Limits

Unit suite + check + bats only. No e2e spec was run. The `console.warn` on
the detached confirm path is unasserted before and after.

### DIFF
--- a/cicchetto/src/lib/commands/fanout.ts
+++ b/cicchetto/src/lib/commands/fanout.ts
@@ -1,0 +1,84 @@
+import { requestConfirm } from "../confirmDialog";
+import { friendlyError } from "../friendlyError";
+import { joinedChannelsOnNetwork } from "../joinedChannels";
+import { sendFanOut } from "../sendPipeline";
+import type { CommandHandler } from "./context";
+
+// #431 — above this many channels an /ame|/amsg asks before it writes anything.
+// A DECIDED number (vjt, 2026-08-09), not a tuned one: one keystroke that
+// writes to N channels is a spam vector, and ten is where a fan-out stops
+// reading as "the rooms I am in" and starts reading as a broadcast.
+const FANOUT_CONFIRM_THRESHOLD = 10;
+
+/**
+ * #431 — /ame + /amsg: the same body to EVERY joined channel of THIS network.
+ * Deliberately not addressed to the active window: the fan-out is per-NETWORK
+ * (vjt 2026-08-09 — cross-network is out of scope, and no flag for it), so it
+ * works from the $server window just as well as from a channel, and it
+ * includes the window it was typed in.
+ *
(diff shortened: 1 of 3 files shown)


vjt:

PR #1530, opened 2026-08-18
e2e(#1336): take the per-spec subject out of the shared channel

### LINKED ISSUE #1137: e2e: the per-spec CHANNEL slice — long-lived users still accrue ~2.9 rows per test from the shared #bofh

vjt:
## What

Every test that provisions a subject (#1078) had it autojoin `#bofh`, where the
three long-lived seeded users (`vjt`, `m9b-test`, `m9b-victim`) sit by compose
bind. bahamut fans a JOIN and a QUIT out to whoever is in the room, so each test
wrote presence rows into three OTHER users' scrollback — and those copies outlive
the test, because the subject is deleted at teardown and only its own rows go
with it.

The per-spec subject now autojoins a worker-scoped channel of its own. `#bofh`
is untouched, for the specs that drive the seeded users directly.

## Measured, on a 6-test chromium pilot at `038a33ca`

Rows in the co-residents' scrollback whose sender is a spec subject:

| run | rows | per test | shape |
|---|---|---|---|
| before | 42 | 7.00 | 36 presence (6 JOIN + 6 QUIT × 3 co-residents) + 6 content |
| after | 0 | 0 | — |
| mutant — constant put back to `#bofh` | 39 | 6.50 | the same 36 presence + 3 content |

The presence half reproduces to the row; the content half is whatever the specs
said out loud that run, so the claim rests on the 36. The `~2.9 rows/test`
inherited from the closed #1137 is a full-suite average over a different
denominator — it is neither reproduced nor contradicted here.

## Why the constant's value, and not 292 spec files

A per-TEST channel cannot be reached from a module-scope constant, and all 292
spec files that name `AUTOJOIN_CHANNELS` name it at module scope; every one of
them also drives the per-spec subject (zero exceptions). It would have cost a
292-file codemod to buy isolation that `workers: 1` already provides.

That is the constraint the cure inherits, so the name derives from
`TEST_PARALLEL_INDEX`: under a future `workers: N` the subjects would otherwise
become co-residents of each other and re-grow the residue silently. The new spec
is the oracle that the derivation resolves inside a worker at all — it reads
`#spec-w0`, and the runner-process spelling is `#spec-wsetup` on purpose so a
name resolved outside a worker cannot impersonate worker zero.

## The rest of the diff

- 16 specs that hardcoded `"#bofh"` now read the constant, and one aria-label
  selector is built from it.
- `ux-5-bv` taps a members-pane row that is not self and was the one spec left
  reading the shared channel's ambient population; it brings its own peer now,
  the same cure `ux-4-z` bucket J got earlier in this issue.

## Gates

- `scripts/bun.sh run check` — rc 0 (the 59 CSS warnings are pre-existing; none
  names a file in this diff).
- `scripts/design-notes-gate.sh` — rc 0.
- Pilot green 7/7 with the oracle; mutant red on the oracle with
  `Received "#bofh"`.
- Full `scripts/integration.sh` running locally; result reported separately.

## Not claimed

That the suite is green — the pilot is six tests. The prose in ~135 spec files
still says `#bofh` in comments and error strings, which is stale wherever it
names the seeded corpus of the channel under test; behaviour is unaffected and
chasing it would touch far more files than the cure.

### DIFF
--- a/cicchetto/e2e/fixtures/seedData.ts
+++ b/cicchetto/e2e/fixtures/seedData.ts
@@ -55,7 +55,35 @@ export const VJT_IDENTIFIER = "vjt@grappa.test";
 
 export const NETWORK_SLUG = "bahamut-test";
 export const NETWORK_NICK = "vjt-grappa";
-export const AUTOJOIN_CHANNELS = ["#bofh"];
+
+// #1336 — the per-spec subject (#1078) autojoins THIS channel, and the
+// three long-lived seeded users (`vjt`, `m9b-test`, `m9b-victim`, bound
+// with `--autojoin '#bofh'` in compose.yaml) are NOT in it. Every
+// provision/teardown used to JOIN and QUIT the shared `#bofh`, and
+// bahamut fanned both out to whoever was sitting there: measured on a
+// 6-test pilot, 42 rows landed in the three co-residents' scrollback,
+// 7.00 per test — rows no spec asked for and every later spec reads.
+//
+// It is the CO-RESIDENCE that writes them, not the channel's name, so
+// the cure is to stop sharing a room, not to rename it. `#bofh` stays
+// exactly as seeded for the specs that drive the seeded users directly.
+//
+// 🔴 Worker-scoped, and the scope is load-bearing: `playwright.config.ts`
+// runs `workers: 1, fullyParallel: false`, so at most ONE per-spec
+// subject is alive at a time and a single channel is enough. Derived
+// from `TEST_PARALLEL_INDEX` (Playwright sets it per worker process,
+// `playwright/lib/worker/workerMain.js`) so that if the suite ever
+// parallelises, the subjects do not silently become co-residents of
+// each other and re-grow the residue this removed. The constant is
+// evaluated once per process, which is what keeps the ~292 spec files
+// that read it unchanged.
+//
+// `globalSetup` runs in the runner process, where Playwright sets no
+// worker index — the `setup` spelling is deliberately NOT `0`, so a
+// channel named `#spec-wsetup` in a failure log reads as "resolved
+// outside a worker" instead of impersonating worker zero.
+const SPEC_WORKER = process.env.TEST_PARALLEL_INDEX ?? "setup";
+export const AUTOJOIN_CHANNELS = [`#spec-w${SPEC_WORKER}`];
 
 // M-cluster M-7 — admin user. Seeded via mix run -e in the seeder
 // sidecar after `create_user` (no --admin flag on the mix task; M-7
--- a/cicchetto/e2e/tests/b0-invite-from-server-window.spec.ts
+++ b/cicchetto/e2e/tests/b0-invite-from-server-window.spec.ts
@@ -31,7 +31,7 @@
 
 import { composeSend, loginAs, selectChannel } from "../fixtures/cicchettoPage";
 import { IrcPeer } from "../fixtures/ircClient";
-import { NETWORK_SLUG } from "../fixtures/seedData";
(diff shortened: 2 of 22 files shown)


vjt:

PR #1534, opened 2026-08-18
e2e(#1397): extract the user-id lookup into grappaApi, and name it honestly

vjt:
## What

Five spec files carried the same 14-line `findVjtUserId` — **byte-identical, md5 equal across all five** (the one helper in the #1397 census with no drift). It is now `grappaApi.findUserIdByName(adminToken, userName)`; the five copies are gone.

Net: **+20** in the fixture, **−85** across the five specs.

## The fork, and why it went this way

The body reads `specUser()`, which lives in `fixtures/specSubject.ts` — and `specSubject` imports `grappaApi`. Dropping the helper in unchanged would have made the two import each other.

`grappaApi` is the **only leaf** of the fixture graph: `grep -c '^import'` is `0`, and four fixture modules import it. Three options were on the table:

- **Parameterise the name (taken).** Keeps the leaf a leaf; costs each call site one argument.
- **Home it in `specSubject` (rejected, though cheaper).** No signature change, but it nails a plain `GET /admin/users` to the module whose reason for being is the subject — the next caller who wants a non-subject user duplicates it a sixth time. Reuse the verbs, not the nouns.
- **Accept the cycle (rejected).** Worth naming because **no local gate sees an import cycle** — neither `tsc` nor biome reports one, so the first symptom would have been an obscure e2e red later. The choice is structural, not enforced by a red.

The rename rides along: `findVjtUserId` stopped being true when #1078 gave every test its own subject (it resolves `specUser()`, not `vjt`), and with the name now passed from outside it was false twice. The call sites were being touched anyway, so it cost no extra lines.

## Test plan

- [x] `bun run check` rc=0 — the 59 CSS warnings are pre-existing, **measured identical** on the branch point (`393f0f89` at rc=0).
- [x] Scoped e2e, the five specs, **before**: 15 passed on `webkit-iphone-15` + 1 on `chromium`, rc=0, on a pristine `393f0f89` worktree.
- [x] Scoped e2e, the five specs, **after**: **16 passed**, rc=0, both projects, on a stack rebuilt from this worktree.
- [x] Mutant (`u.name === userName` → `${userName}-MUTANT`): **1 failed**, killed with `grappaApi.findUserIdByName: s2cbf33fc absent from …` — a string that exists only in this branch, so the runner is provably executing this code and the spec provably depends on the helper.

⚠️ Note for anyone re-running these: **ten of the eleven** tests across the five files are `@webkit`, and the chromium project carries `grepInvert: /@webkit/`. A `--project=chromium` scoped run executes exactly ONE of them and still reports rc=0.

### DIFF
--- a/cicchetto/e2e/fixtures/grappaApi.ts
+++ b/cicchetto/e2e/fixtures/grappaApi.ts
@@ -453,6 +453,26 @@ export async function terminateSession(adminToken: string, sessionId: string): P
   }
 }
 
+// The admin surface keys users by id, and a spec that provisions its subject
+// (#1078) only ever learns the NAME — hence the lookup. The name is a
+// PARAMETER and not `specUser()` read from here: this module imports nothing,
+// and reaching for the subject would make it import `specSubject`, which
+// already imports this one.
+export async function findUserIdByName(adminToken: string, userName: string): Promise<string> {
+  const res = await fetch(`${GRAPPA_BASE_URL}/admin/users`, {
+    headers: { authorization: `Bearer ${adminToken}` },
+  });
+  if (!res.ok) {
+    throw new Error(`grappaApi.findUserIdByName: ${res.status} ${await res.text()}`);
+  }
+  const body = (await res.json()) as { users: { id: string; name: string }[] };
+  const user = body.users.find((u) => u.name === userName);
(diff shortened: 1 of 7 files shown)


vjt:

PR #1535, opened 2026-08-18
test(#1397): call the canonical start_server/1 at the 22 sites that inlined it

vjt:
## What

`Grappa.IRCServer.start_server/1` is `start_link/1` plus `port/1`. Twenty-two
sites across fourteen test files still spelled that body out by hand — eight
inside a local `start_irc_server/0` wrapper, the rest inline in a test body.
They all call the canonical now. Part of #1397.

## Selection, not a sweep

The predicate is the property that makes a site convertible: it starts the fake
peer AND immediately wants its port, and neither `start_link` itself nor a
seeded handler state is what the test asserts on. Four sites fail that
predicate and are left alone, each for a reason:

| site | why it stays |
|---|---|
| `client_test.exs` — `start_link/2` initial state | needs an arity `start_server/1` cannot express |
| `client_test.exs` — "start_link/1 still works" | that call **is** the subject; converting deletes the assertion |
| `client_test.exs` — S7 tcp_closed drain | exercises `IRCServer`'s own primitives; a convenience built on top would make it fail for a reason that is not the drain |
| `irc_server_test.exs` | calls `port/1` on a server it is handed — never one of the copies |

A sweep over the call shape would have eaten all four silently. Same posture as
the `await_handshake/2` pass in #1533.

## Two things a blind sweep would have broken

- The four `networks_controller_test.exs` sites never name the peer again, so
  the pair binding has to drop it — anonymously as `_`, the strategy that file
  already follows and that Credo's consistency check enforces.
- `client_outbound_cost_test.exs` called `port/1` **inside** the client's option
  map rather than binding it. "Used once" is not "not needed".

## Wrappers: eight delegated, not deleted

#1533 DELETED the two wrappers over `welcome_handler()`; the eight over
`passthrough_handler()` are delegated instead. That is deliberate: a
`welcome_handler` peer writes a 001, so a local name over it hides whether
registration completes, while the passthrough peer's contract is to do nothing
and a name over it hides no observable. A wrapper over a no-op is a name; a
wrapper over an observable is a lie. Deleting the eight would cost 72 call
sites and a new `alias` in six files for no behavioural change. Recorded in
DESIGN_NOTES.

## Census

`IRCServer.port/1` callers in `test/`: **26 → 4** on this base. Selected
against `393f0f89` it read 28 → 6; #1533 landed in between and removed two of
those. Both readings are in the DESIGN_NOTES entry — a count taken against a
base that then moves is not wrong, it is dated.

## Gates

`scripts/check.sh` green on this exact tree (post-rebase onto `1a383ac2`):
8 doctests, 61 properties, 6546 tests, 0 failures; Credo, Dialyzer and Sobelow
ran; bats 564 ok / 0 not ok. `design-notes-gate.sh` green.

The rebase opened 8 conflicts, all of them predicted before the rebase — the
lines where #1533 rewrote the same `start_link` call. Each was resolved by
asserting the shape of both sides, and the resulting eight lines were re-read
from the file afterwards.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -49855,3 +49855,68 @@ rewritten call sites type-checks and fails only at runtime. That is the same
 gap the sibling entries name, and on the `seedCursor` slice it was a runtime
 positive control, not the type-checker, that caught exactly such a
 transposition.
+<!-- entry #1397-startserver -->
+
+---
+
+## 2026-08-18 — #1397: the inlined `start_server/1` body, and which wrapper is a lie
+
+`Grappa.IRCServer.start_server/1` has existed for a while and had 548 callers;
+it also had twenty-two hand-written copies of its two-line body across fourteen
+test files. This entry records how those were selected, which ones were left,
+and why two wrappers of the same shape earned opposite treatment.
+
+### The predicate is the property, not the call shape
+
+A `git grep` for `IRCServer.port(` finds 28 callers in `test/`, and a sweep over
+that shape would have been wrong at four of them. The property that makes a site
+convertible is that it starts the fake peer AND immediately wants its port, with
+neither `start_link` nor a seeded handler state being the thing under assertion.
+The four survivors in `client_test.exs` and `irc_server_test.exs` fail exactly
+that test:
+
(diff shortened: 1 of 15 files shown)


vjt:

PR #1539, opened 2026-08-18
test(#1397): one credential fixture for the networks/* test family

vjt:
## What

Five files under `test/grappa/networks/` (`away`, `commit_password`, `connection_state`, `last_joined_channels`, `perform_changeset`) each carried a private `setup_credential` and a private `reload`. The five `reload` bodies were byte-identical; the five `setup_credential` bodies grouped into three md5 classes that are one function on two axes.

Both now live in `Grappa.AuthFixtures` as `user_with_credential/2` and `reload_credential/1`. Ten local definitions gone, no new imports — all five files already imported `AuthFixtures`.

Part of #1397 (bucket H test-harness). Zero overlap with the e2e half.

## The two axes stay at the call site

- **The port** (a dummy `6667` in four files, the `IRCServer` fake's real port in `connection_state`) is a required argument, not a default. The difference between "this file never dials" and "this file dials a real fake" is exactly what a default would hide — the same posture the shared handshake barrier took with its timeout, and what the no-default-arguments rule asks for anyway.
- **The password-rotation merge** (`auth_method: :nickserv_identify`, `password: "oldpass"`) is wanted by two files out of five. It stays local, renamed `rotating_credential/1` for what it builds. Two four-line wrappers survive by design: a shared helper carrying a rotation default for three files that never rotate a password would be a flag in the data model, not reuse.

## Boundary caught a reference the copies were making quietly

`reload_credential/1` first went in as the copies had it — a raw `Repo.get_by!(Credential, ...)`. Boundary refused: `AuthFixtures` may name `Grappa.Networks.Network` and `Grappa.Accounts.User`, not `Grappa.Networks.Credential`.

The fix is not a wider deps list around the raw query but `Credentials.get_credential_by_ids/2`, the public verb that already runs that query in production. The dep is declared because the `%Credential{}` head genuinely references the schema. Five copies had been making the same reference implicitly, one file at a time, where no boundary check looks.

## Gates

`scripts/check.sh` in full, rc read from a dedicated file, every stage exhibited rather than inferred from the exit code:

| stage | result |
|---|---|
| credo --strict | 84 checks over 689 files, no issues |
| deps.audit | no vulnerabilities |
| sobelow (Medium) | no vulnerabilities |
| doctor | passed |
| test | 8 doctests, 61 properties, **6546 tests, 0 failures** |
| dialyzer | Total errors: 0 |
| bats | 564 ok / 0 not ok |

The first run of `check.sh` was **rc=1** — Credo consistency, `_server` where the codebase strategy is anonymous `_`, fixed in the third commit. Worth stating because the `ci.check` chain halts at the first non-zero: in that run nothing downstream of Credo had executed, so it was not "green minus one", it was unmeasured.

## Mutants

Two, against the two promoted helpers, over the five files (68 tests):

| mutant | rc | failures | kills in |
|---|---|---|---|
| `reload_credential/1` returns its argument un-reloaded | 2 | 23 | `away`, `commit_password`, `connection_state`, `last_joined_channels` |
| `user_with_credential/2` ignores `attrs` | 2 | 4 | `commit_password`, `last_joined_channels` |

`away` and `connection_state` surviving the second one is a no-op by construction — they pass `%{}`.

**`perform_changeset_test.exs` survives both**, and the reason is not weak assertions: it reloads `saved`, the struct `Repo.update/1` has just returned, so those six re-reads cannot fail. Measured to be pre-existing rather than introduced here: on a detached worktree at the previous base, with the file's own local `reload` neutered, that file still reports rc=0, 10 tests, 0 failures. It is filed as a comment on #1397 next to the open Argon2 question, because fixing it changes what the file asserts — a decision about test intent, not a de-duplication step.

## Not established

- No e2e run: this branch touches no client code, and the e2e lane was not taken.
- The `check.sh` green was measured on the pre-rebase base. The rebase target adds nothing under `test/`, `lib/`, `scripts/` or `mix.exs` — the delta is e2e, cic and docs only — so the transfer is by construction, and CI on this PR is the measurement on the new base.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -49855,3 +49855,85 @@ rewritten call sites type-checks and fails only at runtime. That is the same
 gap the sibling entries name, and on the `seedCursor` slice it was a runtime
 positive control, not the type-checker, that caught exactly such a
 transposition.
+<!-- entry #1397-networksfam -->
+
+---
+
+## 2026-08-18 — #1397: one credential fixture for the networks/* family, and the residue a name-grep cannot see
+
+Bucket H again, server side. Five files under `test/grappa/networks/`
+(`away`, `commit_password`, `connection_state`, `last_joined_channels`,
+`perform_changeset`) each carried a private `setup_credential` and a private
+`reload`. Both are now in `AuthFixtures` as `user_with_credential/2` and
+`reload_credential/1`; the five copies of each are gone.
+
+### The two axes, and why neither becomes a default
+
+The five `reload` bodies were byte-identical. The five `setup_credential`
+bodies grouped into three md5 classes that are one function on two axes: the
+port (four files pass a dummy `6667`, `connection_state` passes the
+`IRCServer` fake's real port) and a `%{auth_method: :nickserv_identify,
+password: "oldpass"}` merge that only `commit_password` and
+`perform_changeset` want.
+
+Both axes stay at the call site. The port is a required argument, not a
+`6667` default, because the difference between "this file never dials" and
+"this file dials a real fake" is exactly the thing a default would hide —
+the same posture the shared handshake barrier took with its timeout, and
+what the project's no-default-arguments rule asks for anyway. The rotation
+merge is the 20% that does not generalise, so it stays local to the two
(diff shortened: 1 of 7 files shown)


vjt:

PR #1544, opened 2026-08-18
test(#1397): one 001 handler for 55 copies, and seven wrappers gone

vjt:
## What

The largest duplicated shape in the test suite was one handler body repeated
**62 times across 7 files** — answer a `USER` line with an `001 RPL_WELCOME`,
stay silent otherwise. It becomes `Grappa.IRCServer.welcome_handler(prefix, nick)`.
55 of the 62 now call it; the remaining 7 are deliberately deferred (below).

Only 5 of the 62 sat behind a name, so a census counting helper NAMES
undercounts this cluster by roughly twelvefold and reads as nearly finished.
Two of the seven wrappers — `rfc_handler/0` and `welcoming_handler/0` — were
missed even by a name search for `welcome_handler`, and surfaced only from
grouping bodies with the names stripped off.

## Both arguments mandatory

Consolidating normally requires first deciding which prefix is canonical: the
copies split 42 `:irc`, 19 `:server`, 1 `:irc.test.org`. Taking the prefix as
a mandatory argument dissolves that question rather than answering it — every
call site keeps the bytes it already had, so the change is behaviour-preserving
by construction and the spelling becomes an independent cleanup. The nick is a
second mandatory argument because it is a second axis: 7 of the 62 name someone
other than `grappa-test`.

Seven zero-argument wrappers are deleted and their **51 call sites inlined**.
A wrapper may hide a no-op; it may not hide an observable. This is the same rule
that earlier in #1397 *saved* the wrappers hiding `passthrough_handler/0`,
reaching the opposite verdict because the content is opposite.

## The mutant that stayed green

Two mutants over the six affected files, each disabling one argument:

| mutant | tests killed |
|---|---|
| nick ignored | 11 (10 in `server_test`, 1 the new unit test) |
| prefix ignored | **1 — and it is the new unit test** |

41 migrated call sites had their 001 prefix silently changed and nothing
noticed. Two consequences: unifying the three spellings is a measurably free
follow-up, and the prefix is currently guarded by nothing except the test added
here — before it existed that mutant would have killed nothing at all.

## Why 7 sites are left behind

PR #1535 is open and touches four of the same files, and the overlap is not
merely file-level: the tail of those 7 handlers is literally the context of its
hunks. They follow as a second commit on this PR once #1535 lands, taking the
remaining count to zero. `git merge-tree` against both other open branches is
clean, and the merged trees carry all entry markers with no separator loss.

## Test plan

- [x] `scripts/check.sh` green, every stage exhibited: credo 84 checks / 689
      files no issues · sobelow no vulnerabilities · doctor passed · dialyzer
      `Total errors: 0` · `8 doctests, 61 properties, 6551 tests, 0 failures`
      (6546 before, +5 = the tests added here) · bats `1..564`, zero `not ok`
- [x] `scripts/bats.sh` re-run after the rebase: `1..564`, zero `not ok`
- [x] `scripts/design-notes-gate.sh`: 1 new entry heading, separator and marker present
- [x] RED observed before GREEN on the new helper: 5 failures, then 0
- [x] Mutation bench, both mutants, tree restored clean afterwards

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -49855,3 +49855,98 @@ rewritten call sites type-checks and fails only at runtime. That is the same
 gap the sibling entries name, and on the `seedCursor` slice it was a runtime
 positive control, not the type-checker, that caught exactly such a
 transposition.
+<!-- entry #1397-welcome-handler -->
+
+---
+
+## 2026-08-18 — #1397: one 001 handler, and the wrapper rule that cuts both ways
+
+The largest duplicated shape in the test suite was a handler body repeated
+62 times across 7 files: answer a `USER` line with an `001 RPL_WELCOME`,
+stay silent otherwise. It is now `Grappa.IRCServer.welcome_handler/2`, and
+55 of the 62 call it. The remaining 7 are deferred for a reason recorded
+below.
+
+### A name census could not see it
+
+Only 5 of the 62 sat behind a `defp`, so counting helper NAMES undercounted
+this cluster by roughly twelvefold and made the bucket read as nearly done.
+Worse, the five names disagreed with each other: three files called theirs
+`welcome_handler/0`, one `counts_welcome_handler/0`, one
+`start_server_with_001/0` — and a first pass here still missed two more,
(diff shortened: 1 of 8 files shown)


vjt:

Issue #1546, opened 2026-08-19
AdminEvents persist/retention can leak between tests; three setup resets are downstream of it


vjt:
**Scope (widened 2026-08-20).** This issue is no longer "should three
setup resets be deleted". It is: **make it impossible for
`Grappa.AdminEvents`' `persist` / `retention` to leak from one test into
the next.** The three resets are downstream of that, and two of them can
only be removed once it is done.

The widening comes out of the measurement in the first comment, which
separated something the original ticket had measured as one thing. The
original finding is preserved verbatim in "What was known before" at the
bottom — it is the record of how this looked then, and it was not wrong,
it was incomplete.

---

## The separation that changed the question

`AdmissionStateHelpers.reset_admin_events/0` does **two** things, not one:

1. it empties the ring buffer, and
2. it **rebuilds the struct**, so `persist` and `retention` go back to
   their `defstruct` defaults.

The no-op mutant that produced this ticket's numbers removes **both at
once**, so those numbers cannot say which half carries the weight. Split
on the two axes, the answers differ — and the second axis is the one this
ticket did not name.

⚠️ The mutant's failure count is **seed-dependent and does not exist as a
fixed quantity** (ExUnit randomises order, and this reset exists precisely
to defend against order). It is not a target. The outcome of this issue is
not "N becomes 0"; it is that the leak is no longer possible.

## Axis 2 — the root, and the actual scope of this issue

`Grappa.AdminEvents` is a singleton started by `Grappa.Application`. It
outlives every sandbox checkout, so state written by one test is still
there for the next test, and for the next **file**.

`admin_events_test.exs` is the only file in all of `test/` that writes
`persist` / `retention`. It restored them with **straight-line statements
in the test bodies** — which a failing assertion above them skips. Two
such windows were censused (both inside the same file, both requiring
another test to have already failed).

What escapes is not a stale flag:

`persist: true` → the singleton writes to the Repo from its own pid →
that pid is `Sandbox.allow`ed only inside `admin_events_test.exs` → the
singleton dies → the **next** file's `setup` takes an `** (exit)`.

So the reported failure names a file that did nothing wrong, and the real
one is already off screen. (#1613 is the sibling defect that erases the
reason on the way out. It is **not** in scope here — it amplifies the
cascade, it does not cause the leak.)

**This is what the two "not load-bearing" resets were really doing:
cascade suppression.** Which is why they are not an independent cleanup —
removing them before closing the leak *introduces* noise rather than
[...]

PR #1547, opened 2026-08-19
test(#1397): give the AdminEvents ring one reset, not seven copies

vjt:
Bucket H of #1397, server side. Seven `setup` blocks carried the same
byte-identical line; they now call
`Grappa.AdmissionStateHelpers.reset_admin_events/0`.

## The line did more than it read

    :sys.replace_state(AdminEvents, fn _ -> %AdminEvents{buffer: []} end)

It reads as "empty the ring". It REBUILDS the struct, so `persist` and
`retention` also return to their `defstruct` defaults. A helper written
from the apparent intent — `%{s | buffer: []}` — would have preserved a
leaked `persist: true` and changed behaviour at all seven sites without
failing a test. Nothing asserted that clobber, so the characterization
test landed FIRST and on its own commit, and it asserts the pre-state
before performing the gesture (without that, the post-state assertions
also pass against a singleton already sitting at the defaults).

## Commits

1. `70741f75` — LOCK: characterize what the gesture actually does.
2. `38002fde` — the seven sites, plus the equivalence test.
3. `00428f55` — DESIGN_NOTES.

## Where it lives

`AdmissionStateHelpers` already existed for this failure mode:
application-wide singletons that outlive a sandbox checkout. AdminEvents
is not an admission singleton, so the module name is now marginally
wrong — the moduledoc says so rather than hiding it. A second support
module for one function splits the pattern across two homes, and the next
author finds one of them and invents a third spelling.

Deliberately NOT folded into `reset_all/0`: thirteen files call that today
without touching the ring.

## Measured

Mutants, each restoring from HEAD:

- LOCK test, gesture mutated to the apparent intent → 1 failure, on a
  post-state assertion (`==`).
- LOCK test, pre-state dirtying removed → 1 failure, on the pre-state
  assertion (`!=`). Different assertion, not the same one twice.
- Helper mutated to the apparent intent → 1 failure, and it is the
  equivalence test only; the characterization test stays green because it
  exercises the inline gesture rather than the helper.
- Helper made a no-op → 16 failures in one run, 14 in the next. The
  number is seed-dependent and does not exist as a fixed quantity —
  ExUnit randomises order and this reset defends against order. What that
  run did show is filed as #1546: three of the seven sites contributed no
  failures, two of them provably (they never read the ring at all).

`scripts/check.sh` green end to end: credo `found no issues`, dialyzer
`Total errors: 0`, sobelow `No vulnerabilities found`, doctor passed,
`8 doctests, 61 properties, 6548 tests, 0 failures` (6546 on the base,
plus the two tests added here), bats `1..564` with zero `not ok`.
`design-notes-gate.sh` reports `1 new entry heading(s), separator and
marker present`.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -49855,3 +49855,83 @@ rewritten call sites type-checks and fails only at runtime. That is the same
 gap the sibling entries name, and on the `seedCursor` slice it was a runtime
 positive control, not the type-checker, that caught exactly such a
 transposition.
+<!-- entry #1397-adminevents-reset -->
+
+---
+
+## 2026-08-19 — #1397: one reset for the AdminEvents ring, and the anchor that hid the family
+
+Bucket H, server side. Seven `setup` blocks opened with the same
+byte-identical line and now call
+`AdmissionStateHelpers.reset_admin_events/0`.
+
+### The line did more than it read
+
+    :sys.replace_state(AdminEvents, fn _ -> %AdminEvents{buffer: []} end)
+
+It reads as "empty the ring". It REBUILDS the struct, so `persist` and
+`retention` also go back to their `defstruct` defaults. A helper written
+from the apparent intent — `%{s | buffer: []}` — would have preserved a
+leaked `persist: true` and quietly changed all seven call sites. Nothing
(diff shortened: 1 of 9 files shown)


vjt:

PR #1553, opened 2026-08-19
test(#1397): the last eight start_irc_server wrappers, and the tests their helpers never had

vjt:
Eight test files each kept a zero-argument `start_irc_server/0` and called it
from seventy-two sites. They are gone and the handler is spelled at all
seventy-two, finishing the set #1535 opened.

## The slice was not the same slice by the time it was worked

This was planned against a base where each wrapper still held the two-step
body — `start_link/1` then `port/1` — so it was going to be a behavioural
consolidation onto `start_server/1`, with a characterization test and an
equivalence oracle. Both were written, both passed, both were discriminating
under mutants, and the whole thing was green under `check.sh`.

Then #1535 landed, and it had **already rewritten all eight bodies** into a
one-line delegation to `start_server(passthrough_handler())`. That turned the
inlining into pure substitution — the expression a call site now carries is
byte-for-byte the body of the wrapper it replaces — which left the
characterization test pinning a body that no longer existed anywhere, and left
the equivalence oracle comparing an expression to itself.

Both were **discarded rather than rebased**, and the branch was rebuilt from
the new base. Worth stating plainly, because the rebase offered to keep them:
the conflicts were resolvable and both tests would have gone green. A test
that cannot fail passes every review that only reads its colour. The signal
was not the conflict — it was the wrapper body inside the conflict.

## What the slice actually needed instead

Measured on the new base: `passthrough_handler/0` and `start_server/1` have
**no tests at all** (`irc_server_test.exs` covers only `await_handshake/2` and
`welcome_handler/2`; `start_server` appears zero times). This change names
both at seventy-two call sites.

So the missing work was never a characterization — it was coverage of the two
helpers, in the shape `welcome_handler/2` got when #1544 hoisted it, and which
the moduledoc already prescribes: *a helper earns a test here once it is
hoisted out of local copies*. The handler is checked as a plain function, the
server through a socket.

What `passthrough_handler/0` promises is that it answers NOTHING, whatever
arrives, and leaves handler state alone. That promise is the entire reason a
call site must be able to name it rather than inherit it from a zero-argument
wrapper — and until now nothing said it.

## Why `u()` stays and this one goes

`u()`/`uniq()` — eighteen copies, one hundred and sixty-four call sites —
is deliberately untouched. It is a pure alias: a short name for a long
expression, hiding nothing, because there is no argument at the call site to
get wrong. `start_irc_server()` hid WHICH PEER, which is decidable at the call
site and is exactly what the `84fe0850` / #1535 reading says must stay there.
Copy count is the obvious criterion and it is the wrong one.

## The alias is a measurement

Six of the eight reached the module by its full name. Inlining meant either
six alias lines or sixty fully-qualified call sites, and the codebase had
already answered: **460 aliased inline spellings against 0 qualified**, taken
with a fixed-string grep (in a regex `.` is a wildcard, and the whole argument
is the count). The seven qualified references surviving outside the wrappers
move to the alias for the same reason — leaving them would ship two spellings
of one module inside one file.

## Mutants

Two, each killing exactly one test, neither killing the other's:

| mutant | result |
|---|---|
| `passthrough_handler/0` made to reply | 11 tests, 1 failure — the handler test alone |
| `start_server/1` returns `port(server) + 1` | 11 tests, 1 failure — the server test alone |

Each injector counts its anchor and aborts unless it matches exactly once;
the tree was committed and verified clean before each, and restored from
`HEAD` after.

## Census, retaken on the new base

| | before | after |
|---|---|---|
| `defp start_irc_server` | 8 | 0 |
| naked `start_irc_server()` calls | 72 | 0 |
| inline `IRCServer.start_server(IRCServer.passthrough_handler())` | — | 72 |
| residual `Grappa.IRCServer.` in the eight | 7 (+ 6 in wrapper bodies) | 0 |
| files carrying the alias | 2 / 8 | 8 / 8 |

The sweep aborts unless the totals are exactly 8 / 72 / 6 / 7.

## Gates

`scripts/check.sh` rc=0, every stage exhibited:

| stage | marker |
|---|---|
| credo | 6080 mods/funs, found no issues |
| dialyzer | Total errors: 0 |
| sobelow | No vulnerabilities found |
| doctor | Doctor validation has passed! |
| test | 8 doctests, 61 properties, **6555 tests, 0 failures** |
| bats | 1..564, zero `not ok` |

`design-notes-gate.sh`: *"1 new entry heading(s), separator and marker
present"*.

**Not claimed:** `main`'s own test count at `e50e5ab1` was not measured, so
"6555 = main + the 2 this branch adds" is arithmetic, not a measurement. And
`check.sh` ran before the final docs amend; the only file changed since is
`docs/DESIGN_NOTES.md`, which no stage reads — a reading, not a re-run.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -50578,3 +50578,92 @@ shape this entry's own separator had to match, and then a name-only census was
 run anyway, by the same hand. Knowing the rule and holding it are different
 states. Treat any name-level count here as a lower bound on the truth and an
 upper bound on the finding — including the ones in this entry.
+<!-- entry #1397-startircserver -->
+
+---
+
+## 2026-08-19 — #1397 bucket H: the last eight start_irc_server wrappers, and a slice that changed shape underneath
+
+Eight test files each kept a zero-argument `start_irc_server/0` and called it
+from seventy-two sites. They are gone and the handler is spelled at all
+seventy-two, finishing the set #1535 opened.
+
+### The slice was not the same slice by the time it was worked
+
+Planned against a base where each wrapper still held the two-step body —
+`start_link/1` then `port/1` — this was going to be a behavioural
+consolidation onto `start_server/1`, with a characterization test and an
+equivalence oracle, the shape the sibling dead-port slice used. Both were
+written, both passed, both were discriminating under mutants.
+
+Then #1535 landed, and it had already rewritten every one of the eight bodies
+into a one-line delegation to `start_server(passthrough_handler())`. The
+inlining became pure substitution: the expression a call site now carries is
+byte-for-byte the body of the wrapper it replaces. The characterization test
+was pinning a body that no longer existed anywhere, and the equivalence oracle
+had become a comparison between an expression and itself — a green that could
+not fail and therefore could not mean anything.
+
+Both were discarded rather than rebased. Worth recording, because the rebase
+offered to keep them: the conflicts were resolvable and the tests would have
+gone green on the new base. A test that cannot fail passes every review that
+only reads its colour. The signal that something had changed was not the
+conflict, it was the wrapper body inside it.
+
+### What the slice actually needed instead
+
+`passthrough_handler/0` and `start_server/1` are named at seventy-two call
+sites by this change, and neither had a test — like the twenty copies #1535
+removed, they were exercised only through whatever their callers did next.
+So the missing work was never a characterization; it was coverage of the two
+helpers, in the shape `welcome_handler/2` got when #1544 hoisted it: the
+handler checked as a plain function, the server through a socket.
+
+What `passthrough_handler/0` promises is that it answers NOTHING, whatever
+arrives, and leaves handler state alone. That promise is the whole reason a
+call site must be able to name it instead of inheriting it from a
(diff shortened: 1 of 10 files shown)


vjt:
Superseded by #1558, which landed on main as `71e7840a` — a cherry-picked union of this PR together with the other one. The SHAs are rewritten, so GitHub cannot mark this MERGED; closing by content.

Verified before that merge: the union's diff against main is `+301/-176` across 13 files, exactly the sum of the two branches measured separately (#1553 `+207/-117`, #1557 `+94/-59`) — no line added, none lost. Both DESIGN_NOTES entries survive with their own unique markers (marker count 123 -> 125, zero duplicates, byte-identical prefix). `check.sh` on the union was green at 6557 tests / 0 failures, a count pre-registered before the run (main 6555 + the 2 added). #1557's mutant was re-run on the union branch and reproduced the same asymmetry (10 deaths, from 3 of the 8 sites).

Reason for the union rather than two merges: both PRs append a DESIGN_NOTES entry, so whichever merged first would have made the other CONFLICTING — and, more to the point, a per-PR green attests `branch + base`, never `branch + the other branch`. The union is the honest gate, not merely the cheap one.

PR #1559, opened 2026-08-19
e2e(#1152): resolve the subject's LIVE nick, and make drift loud

### LINKED ISSUE #1152: push-964-device-row: awaitPushDelivery times out non-deterministically (green on rerun of the same sha)
`push-964-device-row.spec.ts:51` failed once on main and passed on a rerun of the **same failed job at the same sha**, so the spec is non-deterministic. Filing it with the evidence rather than letting a rerun-green bury it: a single green is not a proof about an intermittent red.

## Measured

Failure, main `0f14bf01`, run `31324253590` (first attempt):

```
✘  474 [chromium] › tests/push-964-device-row.spec.ts:51:1 › device row shows the activity instant + marks the device you are on (8.1s)
    Error: awaitPushDelivery: timeout after 5000ms — id=964-device-row
  1 failed
  680 passed (30.2m)
```

Rerun of that same failed job, same sha, no code change: **completed success**.

Baseline from the same evening's other runs, all green, spec present in each:

| run | tree | position | result |
|---|---|---|---|
| PR #1143 (union) | #1121+#1124+#1106 | 474 | ✓ |
| PR #1146 | #1123 | 473 | ✓ |
| PR #1144 | #948 | 473 | ✓ |
| PR #1135 | rebased on `0f14bf01` | 474 | ✓ |
| main `0f14bf01` (attempt 1) | union + #1066 + #1123 + #948 | 474 | ✘ |
| main `0f14bf01` (attempt 2) | identical | — | ✓ |

Run position does not discriminate: 474 was green in the union run and in #1135's run. #1135's tree contains the same union + #948 combination and passed.

## What is not claimed

- No cause is identified. Nothing here says which side of `awaitPushDelivery` is late — the push pipeline, the catcher, or the assertion window.
- The incidence is **1 observation in 6 runs of this spec tonight**, which is not a rate.
- No claim that this is the same defect as any other e2e red.

## Why it is worth a probe rather than a rerun habit

`awaitPushDelivery` waits 5000 ms for a delivery keyed by id. A timeout there means either the delivery genuinely did not arrive in the window, or it arrived and the wait was watching for something that had already happened. Those two have opposite cures, and only an instrumented run separates them — what did the catcher receive, and when, relative to the window opening.

A related shape was **measured** on a different spec the same evening (#1151, PR #1135): a spec opened an observation window and an unawaited row landed inside it, because the barrier was on an IRC ack rather than on the rendered row. That is a **similarity, not a diagnosis** for this spec — it says where to point a probe, not what the answer is.

vjt:
## What

`specNick()` answered the nick the provision **requested**. The session flies
whatever survived registration: on a 433 during registration, #676's fallback
ladder (`auth_fsm.ex`) re-registers as `<nick>_` and then as two random
suffixes, moving `state.nick` with it, with nothing raised. Grappa even writes
the substitution into the scrollback in plain words. Only the fixture believed
the old nick — so `privmsg(specNick(), …)` addressed whoever *did* hold it, the
spec's own assertion timed out, and the failure read as a flake.

Consolidated into #1336.

## Reproduced, and how

3/3 under `--repeat-each 3`:

```
__1152__ phase=pre   requested=sbff5a028  live=sbff5a028
__1152__ phase=post  requested=sbff5a028  live=sbff5a028_
```

⚠️ This is a **displacement** of the natural trigger. In the wild the nick is
held by the previous test's not-yet-reaped ghost, which is a **race**. The bench
has a squatter take the nick while the session is parked, so the same 433
arrives deterministically. **The race itself was not observed and is not
claimed.**

## Not a product bug

`GET /networks` already answers the live nick — `Networks.resolve_network_nick/2`
asks the running `Session.Server` precisely because *"a stale nick silently
drops all inbound DMs"* (`networks_controller.ex`) — and cic resolves its own
nick from that row. The cure reads the **same door**; it does not invent a third
semantics.

## The cure has two shapes, and the reason is a hard constraint

**Sync + always-correct is impossible in JS: there is no synchronous fetch.** A
sync accessor can be made *noisy*; it cannot be made right.

Making `specNick()` async everywhere was rejected on a **measure**, not a
preference: **625** call sites in **290** of the **409** spec files (three
anchors agreeing). That is a patch colliding with every live branch.

So:

- **A teardown guard, zero spec edits** (`fixtures/test.ts`) — one live-nick
  read per test; a drift is a loud red instead of a dead nick addressed in
  silence. Detection-instead-of-prevention is **#1336's own idiom**, not a
  fallback.
- **`specLiveNick()` on the 27 sites where the nick is the STIMULUS** — the
  `privmsg` set, where a stale nick does not fail the spec, it quietly tests
  nothing.

`connection == null` is the liveness discriminator: that is exactly when the
row's nick is the *configured fallback*, and comparing a cached nick against a
fallback would manufacture drift. Such a reading is an **absence of
measurement** and says so on stderr (`__NICKGUARD__ unobservable <reason>`) —
never silently. A revoked bearer is one such reason and not a drift: the logout
journeys seed `specUser().token` into the page, so a UI logout kills the very
token the guard reads.

## Two measurements that corrected the brief

- The slice was briefed as **28**. That is the *line* count — one match is prose
  inside a comment. **27 are code**, across 20 files.
- `privmsg` is **not** the dominant consumer. Grouped by call shape the 625 are
  led by ~**425** `selectChannel(…, { ownNick: specNick() })`, then 27
  `privmsg`, then 17 `sender: specNick()`. The slice was **not** widened; the
  number is recorded so the remainder gets sized from evidence.

Re-resolution after every reconnecting *gesture* was declined: a read issued
right after `PATCH connection_state=connected` lands **before** the
reconciliation it means to observe (measured at 35 ms), so it would launder a
stale value into the cache — repeating, inside the cure, the defect the cure is
for. Re-resolution happens at the point of **use** instead.

## The lesson recorded in DESIGN_NOTES

The first reproduction bench went **green** — `rc=0`, clean — and would have
closed #1152 on a vacuous measurement. Its barrier polled
`connection_state == "connected"`, which is **intent, not liveness**: already
`"connected"` while the FSM walks the 433 ladder. The nick was read 35 ms before
the reconcile (`.567` PATCH → `.572` NICK → `.576/.579` GETs → `.614` reconcile).
Only the server logs caught it. Redone with a **causal** barrier (the autojoin
JOIN, downstream of registration, present in both worlds) it reproduces 3/3.

**A green that reports non-reproduction deserves the same suspicion as a red.**

## Gates

| gate | result |
|---|---|
| `bun run check` (biome + tsc ×2) | rc=0 |
| `bun run test` (vitest) | rc=0 — 290 files, 5616 tests |
| `design-notes-gate.sh` | 1 new entry, separator + marker present |
| e2e / integration | **not run locally yet — CI is the measurement here** |

⚠️ The **double mutant is not yet run** (guard broken → `afterEach` reddens; one
of the 27 reverted → spec reddens). It needs the exclusive stack lane and is
pending allocation. A guard that has not been seen to redden is not yet a guard.

### DIFF
--- a/cicchetto/e2e/fixtures/specSubject.ts
+++ b/cicchetto/e2e/fixtures/specSubject.ts
@@ -133,8 +133,27 @@ export async function teardownSpecSubject(subject: SpecSubject): Promise<void> {
 // renaming a call instead of rewriting a signature.
 let current: SpecSubject | null = null;
 
+// #1152 — every nick value the running test actually HANDED OUT, which is
+// what the teardown guard has to judge, and it is not the same thing as
+// the last value cached.
+//
+// Measured on this branch: 19 of the 20 specs that now call
+// `specLiveNick()` also read `specNick()` earlier in the file. Guarding
+// the CACHE alone would let the async call launder it — refresh to the
+// live nick, teardown compares live against live, green — while the stale
+// value those 19 already consumed goes unreported. Recording the values
+// instead makes the guard's question the honest one: was everything this
+// test addressed the nick the session was flying?
+let consumed = new Set<string>();
+
 export function setCurrentSpecSubject(subject: SpecSubject | null): void {
   current = subject;
+  consumed = new Set();
+}
+
+/** Every distinct nick the current test has been handed, in first-seen order. */
+export function consumedSpecNicks(): string[] {
+  return [...consumed];
 }
 
 function requireCurrent(caller: string): SpecSubject {
@@ -159,7 +178,9 @@ export function specUser(): SeededUser {
  * always-correct: see the #1152 note below.
  */
 export function specNick(): string {
-  return requireCurrent("specNick()").nick;
+  const nick = requireCurrent("specNick()").nick;
+  consumed.add(nick);
+  return nick;
 }
 
 // #1152 — the requested nick and the flown nick are two different things.
@@ -243,5 +264,6 @@ export async function specLiveNick(): Promise<string> {
     );
   }
   subject.nick = reading.nick;
+  consumed.add(reading.nick);
   return reading.nick;
 }
--- a/cicchetto/e2e/fixtures/specSubject.ts
+++ b/cicchetto/e2e/fixtures/specSubject.ts
@@ -153,7 +153,95 @@ export function specUser(): SeededUser {
   return requireCurrent("specUser()").user;
 }
 
-/** The per-test user's upstream IRC nick. */
+/**
+ * The per-test user's upstream IRC nick, as last RESOLVED — the requested
+ * one until something re-resolves it. Synchronous, so it cannot be
+ * always-correct: see the #1152 note below.
+ */
 export function specNick(): string {
   return requireCurrent("specNick()").nick;
+}
+
+// #1152 — the requested nick and the flown nick are two different things.
+//
+// `provisionSpecSubject` asks for `nick: name`, but on a 433
+// ERR_NICKNAMEINUSE during registration #676's fallback ladder
+// (`auth_fsm.ex`) re-registers as `<nick>_` and then as two random
+// suffixes, moving `state.nick` with it. Nothing raises; grappa even says
+// so in the scrollback ("nick sbff5a028 was taken — you are registered as
+// sbff5a028_"). Everything downstream that addressed the requested nick is
+// then addressing somebody else.
+//
+// There is no product bug here to fix. `GET /networks` already answers the
+// LIVE nick: `Networks.resolve_network_nick/2` asks the running
+// Session.Server and falls back to the credential's configured nick only
+// when no session is up, for exactly this reason — "a stale nick silently
+// drops all inbound DMs" (`networks_controller.ex`). cic resolves its own
+// nick from that row. This reads the same door instead of inventing a
+// third semantics.
+//
+// 🔴 The hard constraint behind the two-shaped cure: JS has no synchronous
+// fetch, so a synchronous accessor CANNOT be always-correct. It can only
+// be made NOISY. Hence `specLiveNick()` (async, correct) where the nick is
+// the stimulus, and the teardown guard in `fixtures/test.ts` (detection,
+// zero spec edits) everywhere else.
+
+// `connection` is the liveness discriminator, and it is the reason this
+// returns a reading instead of a string: it is `null` exactly when
+// `Session.connection_info/2` finds no live pid, which is exactly when
+// `nick` is the CONFIGURED fallback rather than the flown value. Comparing
+// a cached nick against a fallback would manufacture drift that isn't
+// there, so "no live session" is reported as an absence of measurement,
+// never as a nick.
+export type LiveNickReading =
+  | { kind: "live"; nick: string }
+  | { kind: "unobservable"; reason: string };
+
+type LiveNickRow = { slug: string; nick: string; connection: unknown };
+
+/**
+ * Read the subject's live upstream nick off `GET /networks`, or say why it
+ * could not be read. Never throws on a reachable server: a spec that
+ * revokes its own bearer (the logout journeys seed `specUser().token` into
+ * the page, so a UI logout kills this exact token) has not drifted, it has
+ * stopped being observable, and the two must not be confused.
+ */
+export async function readSpecLiveNick(): Promise<LiveNickReading> {
+  const subject = requireCurrent("readSpecLiveNick()");
+  const res = await fetch(`${GRAPPA_BASE_URL}/networks`, {
+    headers: { authorization: `Bearer ${subject.user.token}` },
+  });
+  if (!res.ok) {
+    return {
+      kind: "unobservable",
+      reason: res.status === 401 ? "bearer_revoked" : `${res.status}`,
+    };
+  }
+  const rows = (await res.json()) as LiveNickRow[];
+  const row = rows.find((r) => r.slug === NETWORK_SLUG);
+  if (!row) return { kind: "unobservable", reason: "no_row" };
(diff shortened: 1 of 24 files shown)


vjt:
## The guard had a blind spot, and it was in the shape of the defect

Follow-up measurement, pushed as `ea733fc9`.

The guard as first written compared the **cached** nick against the live one.
The async half of this same cure launders that: `specLiveNick()` refreshes the
cache, so at teardown the guard compares the live nick against itself, goes
green, and never mentions the stale value the spec had already addressed
earlier in the same body.

**19 of the 20** specs that call `specLiveNick()` read `specNick()` at an
earlier line — so the blind spot landed on precisely the specs this branch
touches, which are the DM specs, which are the ones that drift.

(That 19 is a **lexical** ordering — first `specNick()` line before first
`specLiveNick()` line — not an execution trace. For straight-line test bodies
the two coincide; it was not verified case by case.)

### Fix

Judge what the test was **handed**, not what the cache last held: `specNick()`
and `specLiveNick()` record every value they return, and the guard asserts all
of them equal the live nick. A spec that read no nick is vacuously fine — drift
cannot reach what never read it.

The #1152 pin now takes the requested nick from `specUser().name` rather than
`specNick()`. The provision asks for `nick: name`, so that is the value at its
source; reading it as an *address* would rightly make the guard redden a spec
whose entire point is that the address moves.

### What this settles about the ~442 unconverted neighbours

The ~425 `selectChannel(…, { ownNick })` + 17 `sender: specNick()` are covered
**by the guard, not by the conversion** — but that was only true after this
commit. Before it, any of those neighbours sharing a file with a converted
`privmsg` site was silently unguarded. No issue is being filed: there is no way
left for them to be wrong that the guard does not now report, and the slice is
not widened.

### Gates re-run after the change

| gate | result |
|---|---|
| `bun run check` (biome + tsc ×2) | rc=0 |
| `bun run test` (vitest) | rc=0 — 290 files, 5616 tests |
| `design-notes-gate.sh` | 1 new entry, separator + marker present |

⚠️ Unchanged: the **double mutant has still not been run**. It needs the
exclusive stack lane. A guard that has not been seen to redden is not a guard —
and this commit is a second reason to insist on that, since it is now the
guard's *second* shape.

vjt:
## Double mutant: both dead. Measured on the exclusive stack lane.

Tree at `ea733fc9`, unmodified between the paired runs except the one line named
in each. Every mutation was applied to a committed tree and reverted with
`git checkout --`; the working tree is clean and the stack is down.

### Mutant 1 — the guard, seen to redden

Injected: the `#1152` pin consumes the stale cached nick (`specNick()`) after
the drift, which is what the ~598 unconverted call sites do.

**DEAD.** One assertion, in the fixture teardown:

```
Error: this spec addressed a nick the subject was not flying (#1152).
Live: s745ae065_. Grappa re-registered after a 433 and specNick() kept
answering the requested nick.

expect(received).toEqual(expected)
- Array []
+ Array [ "s745ae065" ]
   at ../fixtures/test.ts:132
```

The red exhibits both values — requested `s745ae065`, live `s745ae065_` — which
is also the first direct sighting of #676's ladder rung in this branch's own
evidence.

Exactly one assertion: the body passed in full, and the guard runs after `use()`
and outside the `finally`, so it is the only thing that could fail.

### Mutant 2 — one of the 27, reverted, under a real drift

`issue422-inbound-dm-opens-query.spec.ts` does not squat, so requested and live
agree there and the conversion would be **unfalsifiable**. The scratch setup
forces the 433 first — park → squatter takes the nick → reconnect → causal JOIN
barrier — and then **asserts its own effect**, so a failed squat could not be
mistaken for a passing spec:

```
__MUT2__  requested=sbcfea814  live={"kind":"live","nick":"sbcfea814_"}
```

Same setup, same divergence value, one line different:

| variant | site | result |
|---|---|---|
| A | `privmsg(await specLiveNick(), …)` | **green** (1.4s) |
| B | `privmsg(specNick(), …)` | **DEAD** |

The B failure is the predicted mechanism, not an incidental one — the DM went to
the squatter, so the query window never opened:

```
Error: expect(locator).toHaveCount(expected) failed
Expected: 1   Received: 0
locator('.shell-sidebar').locator('.sidebar-channel-name')
  .filter({ hasText: /^dm422pabca56$/i })
```

Exactly one assertion again: the body fails, so `use()` throws and the teardown
guard never runs. No double-counting.

The setup uses `readSpecLiveNick()` and **not** `specLiveNick()` on purpose —
the latter refreshes the cache, which is the very value variant B reads, so it
would have quietly killed the mutant.

### Control: the guard is not vacuous

The obvious way this cure could be worthless is a guard that never observes
anything. All 21 files that call `specLiveNick()`, 25 tests:

| reading | count |
|---|---|
| `__PROVISIONCOST__` (positive control — fixture stderr does reach the log) | 25 |
| `__NICKGUARD__ unobservable` | **0** |
| result | 25 passed |

So the guard asserted 25 times out of 25 and skipped none. `connection != null`
holds in ordinary conditions, which was an open assumption until now.

⚠️ An earlier attempt at this control reported `0` markers with `rc=1` — and
that zero was worthless: `--project` is variadic in Playwright and had eaten the
file arguments, so **no tests ran at all**. The numbers above are from the
re-run, with the provision marker as the positive control precisely so an
absence cannot be read as evidence again.

### Restored-tree control

`ea733fc9` unmodified, both specs: **2 passed**.

### What is still not claimed

- The **natural race** is still not reproduced — every run here squats
  deliberately. Displacement, as stated from the start.
- The other **19 converted specs were not individually mutated**; the kill is on
  one of the 27, generalised by construction (same call, same fixture).
- **No full-suite e2e run.** CI remains the measurement for that.

Issue #1445, opened 2026-08-16
cic: pull-to-refresh in the channel directory — spinner follows the finger, release past threshold triggers the re-capture


vjt:
Requested by vjt: in the channel directory (`/list`), dragging the row list downwards should reveal a spinner that follows the finger, and past a threshold it should trigger the refresh the toolbar button triggers today.

## What exists today

`DirectoryPane.tsx` refreshes only from a **button**: `onRefresh` (`:337`) calls `triggerRefresh(networkSlug)`, the button is disabled while `status() === "refreshing"` (`:365`) and re-labelled "Refreshing…" (`:368`). There is a second, stale-state CTA (`:419`) that calls the same handler.

The row container `.directory-list` (`:432-437`) is the scroller: `overflow-y: auto` + `overscroll-behavior: contain` (`themes/default.css:11730-11736`). It carries exactly one touch-adjacent handler today — `onScroll`, which saves `scrollTop` so a live re-GET can restore it (`:281-300`). The bottom edge is already gesture-free but automated: a zero-height sentinel + `IntersectionObserver` drives `loadMore` (#677, `:302-321`). **The top edge has nothing.**

`git grep -i "pull.to.refresh|pullToRefresh"` over `cicchetto/` returns **zero hits** — this would be the PWA's first pull-to-refresh, so whatever shape it takes becomes the pattern for the other panes.

## The constraint that shapes the design: the refresh is asynchronous and server-staged

`triggerRefresh` (`lib/channelDirectory.ts:263-275`) POSTs a request to RE-CAPTURE and returns. Its own comment is explicit: the POST "only ASKS the server to re-capture; the rows arrive later via the progress/complete pings — which is also why a 202 clears nothing". The rows land through `onDirectoryProgress` / `onDirectoryComplete` / `onDirectoryFailed` (`:278-280`), each of which re-GETs the page.

So the spinner **cannot** be tied to the POST promise: it would stop while the work is still running. It has to read `status`, which is a SERVER field from the closed set `["fresh","stale","empty","refreshing"]` (`wireTypes.ts:18`) delivered by the page GET. That leaves a gap — finger released → POST in flight → first GET that reports `refreshing` — during which nothing in the store says "busy". Without a local latch spanning that gap the spinner flickers off and on. This is the part worth getting right; the gesture itself is the easy half.

## Reuse the shared gesture engine — do not invent thresholds

[...]

Issue #1438, opened 2026-08-16
cicchetto: swipe up/down should dismiss the media viewer, with the modal following the finger


vjt:
Requested by vjt: swiping the media viewer up or down should dismiss it, with the image following the finger.

Measured on `origin/main` `1552312d`, source only — **no device, nothing executed**.

## What exists today

The viewer closes three ways, none of them a gesture: the backdrop click (`MediaViewerModal.tsx:281`), the header close button (`:300`), and Escape through the shared overlay ESC stack (`createOverlayLock`, `:271`).

The zoomable image already owns a full touch gesture set (`ZoomableImage`, `:74`): pinch-to-zoom, double-tap fit⇄2x (`DOUBLE_TAP_MS = 300`, `:57`), and one-finger pan. Listeners are bound element-level via a ref + `addEventListener` with `touchmove { passive: false }` — Solid delegates touch to a passive document listener, so a JSX `onTouchMove`'s `preventDefault` silently no-ops (`:69-71`, DESIGN_NOTES 2026-06-24 / 2026-07-12).

## Why the lane is free

One-finger pan is gated on `gestureStart.scale > MIN_SCALE` (`:133`, `MIN_SCALE = 1` = the `object-fit: contain` fit baseline, `lib/pinchZoom.ts:34`). **At fit scale a one-finger vertical drag does nothing today** — the touchmove is consumed by the unconditional `preventDefault` (`:120`) and then falls through with no handler. So the dismiss gesture takes an input that is currently dead, and needs no arbitration with pan as long as it carries the same gate inverted: `scale === MIN_SCALE` dismisses, `scale > MIN_SCALE` pans.

## Constraints the implementation has to respect

1. **The transform is element-scoped to the `<img>`** (`:64-66`) precisely so zoom/pan can't bleed into the page. "Moving it with the finger" therefore has to translate the **modal container** and fade the `media-viewer-backdrop` (`:279`) — reusing the image's own transform would drag the picture out of a stationary chrome, and would fight the pinch transform for ownership of the same property.
2. **Don't let a swipe read as a double-tap.** `lastTapAt` is armed on every single-finger touchstart (`:114`); a fast flick that starts and ends within 300ms of a previous tap must not toggle zoom on the way out.
[...]

PR #1569, opened 2026-08-19
cic(#1445): latch the directory refresh across the POST/GET gap

vjt:
First slice of #1445. The issue asks for a pull-to-refresh gesture in the
channel directory; this is the **store half only**, because the gesture
would otherwise have been stacked on a hole. The gesture lands separately.

## ⚠️ Declared behaviour change

**The Refresh button now stays disabled until the refresh has actually
finished**, where before it re-enabled itself the moment the POST returned.
The stale "refresh now" CTA is disabled on the same terms. This is a visible
change to an existing control, called out here rather than left to be
discovered.

## What was wrong

`triggerRefresh` only ASKS the server to re-capture — the rows arrive later
through the progress/complete/failed pings, each of which re-GETs page 1.
Until that first GET lands, `status` still reads whatever it read before the
POST, so nothing in the store said "busy". The button, disabled off
`status === "refreshing"` alone, stayed enabled across that gap.

This is **not** a hazard the gesture would have introduced — it is a defect
the button shipped with. Measured, not reasoned: the first test added here
fails on the parent commit with **two POSTs for one double tap**.

## Shape

* The latch mirrors `loadingMore`, the in-flight guard this module already
  has: one boolean per slug, identity-scoped, doubling as a render source.
* It is taken **at the verb**, not at a door. There are already two doors
  onto the re-capture and #1445 adds a third; a guard on whichever door is
  disabled is a guard the next door reopens.
* It releases on the first page write for the slug after the POST, in both
  settled arms and whatever status that page carries — from that moment the
  store holds fresher information than the latch does. Inside the arms, not
  a `finally`, so a superseded response cannot release a latch a newer POST
  has taken.
* **No timer backstop**, against the issue's own sketch: the only case it
  would guard is a 202 followed by no ping at all, not even
  `directory_failed`, and a stale timer releasing the next latch early is a
  heavier hazard than the case it covers. The bound is stated instead — a
  pane close or an identity rotation releases it. Adding the timer later is
  additive.

## Two of the issue's directions are superseded (measured, recorded in DESIGN_NOTES)

Re-measured on `cda50a5f`: `mediaViewerGesture.ts` (#1438) is already a
vertical follow-the-finger binder exposing the `onProgress` channel the
issue asks for, and `verticalClaim` is the exact claim primitive. The
issue's "mirror `bindMessageGestures`" / "use `claimAxis`" would re-derive
#1438. What survives of that advice is narrower and is the part that
matters — the direction gate. The 20% of `bindDismissGesture` that genuinely
does not fit a pull is written down for the gesture slice.

## Test plan

- [x] `bun run test` — 290 files, **5625** passed, rc=0 (base 5618 + 7 added)
- [x] `bun run check` (biome + tsc src & e2e) — rc=0
- [x] Base measurement: the double-POST test fails on the parent with
      `expected "refreshDirectory" to be called 1 times, but got 2 times`
- [x] Three two-sided mutants, each killing a disjoint and correctly-named set:
      dropping the pane's OR kills exactly the 2 new pane tests (47 survive);
      dropping the store guard kills exactly the double-POST test (41 survive);
      dropping the release kills exactly the two release tests (40 survive)
- [ ] Not covered: the gesture, and every device-level question in #1445

### DIFF
--- a/cicchetto/src/DirectoryPane.tsx
+++ b/cicchetto/src/DirectoryPane.tsx
@@ -7,6 +7,7 @@ import {
   directoryQuery,
   directorySort,
   isLoadingMore,
+  isRefreshPending,
   loadDirectory,
   loadMore,
   resetDirectory,
@@ -323,6 +324,13 @@ const DirectoryPane: Component<{ networkSlug: string }> = (props) => {
   const page = () => directoryPage(props.networkSlug);
   const status = () => page()?.status;
 
+  // #1445 — the ONE "a re-capture is under way" reading, because neither half
+  // covers the whole of it: `status` is a SERVER field that only arrives with
+  // a page GET, and the store's latch only spans the gap before that GET
+  // lands. Read by all three affordances below so they cannot disagree about
+  // whether the pane is busy.
+  const busy = () => status() === "refreshing" || isRefreshPending(props.networkSlug);
+
   const onSearchInput = (e: Event) => {
     const val = (e.currentTarget as HTMLInputElement).value;
     const slug = props.networkSlug;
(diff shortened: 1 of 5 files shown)


vjt:
Soppiantata da #1572: i tre commit di questa PR (`4c4e0a35`, `d0426e36`, `a121d350`) sono i primi tre di #1572, quindi `a121d350` e' antenato di `250adffe`. Un FF di #1572 porta questo contenuto su main e chiude questa PR da sola.

Ho **cancellato** la sua `integration` in volo (run 32248782259): gattava una head superata mentre #1570 e #1572 pagavano la stessa suite in contesa — banda misurata 37-40 min a run singola, ~53 con due in volo. Reversibile con `gh run rerun`. Nessun merge separato di questa PR: mergiarla manderebbe #1570 **e** #1572 behind, facendo ri-pagare `integration` a entrambe.

Issue #1571, opened 2026-08-19
cic gates can run an UNPINNED biome: the main checkout's node_modules had drifted from bun.lock


vjt:
A `bun run check` launched from the main checkout was running a **different
biome than `bun.lock` pins**. A lint verdict produced that way — pass or fail —
says nothing attributable about the code.

## The measurement

Found on 2026-08-19 while running `scripts/bun.sh run check` in a fresh
worktree, whose `cicchetto/node_modules` had been cloned (`cp -Rc`) from the
main checkout per the documented procedure.

| where | biome version |
|---|---|
| `cicchetto/package.json` | `"@biomejs/biome": "2.5.8"` (exact, not a range) |
| `cicchetto/bun.lock` | `@biomejs/biome@2.5.8` |
| `cicchetto/node_modules` in the **main checkout** | **2.5.6** |
| CI (`.github/workflows/ci.yml`, `bun install --frozen-lockfile`) | 2.5.8 |

How it was established: read the version out of
`cicchetto/node_modules/@biomejs/biome/package.json` and compared it against
the pin in `package.json` and the resolution in `bun.lock`. Confirmed by
effect — after `bun install --frozen-lockfile` the `biome.json:17`
`recommended`-field DEPRECATED diagnostic stopped being emitted, because 2.5.8
no longer raises it. `package.json`, `bun.lock` and `biome.json` are
byte-identical between the checkout's commit and `origin/main`, so this is
install drift, not source drift.

## Why it matters

CI installs with `--frozen-lockfile`, so CI is always on the pinned version.
A developer gate run locally is not, and nothing checks. That is a **false
verdict in both directions**:

* a rule that only fires on the newer version passes locally and fails in CI;
* a rule that only fires on the older version fails locally against code CI
  would accept — and the reader then goes looking for a defect in the diff
  that is not there.

The second is the one that actually cost time here: a red gate was
mis-attributed to a tool deprecation, when the real error was elsewhere in
the same run and hidden among diagnostics biome had truncated.

## The false premise this creates

The standing procedure for a new cicchetto worktree is *"clone
`node_modules` from the main checkout with `cp -Rc`, never `bun install`"* —
chosen because a cold install is slow. That procedure **presupposes the
source of the clone is aligned with the lock**, and it was not. Every worktree
cloned from that checkout inherited the drift silently, which is what makes
this worth a fix rather than a one-off correction. The main checkout has since
been brought to 2.5.8, so the drift as it stood is gone; the way it went
[...]

Issue #1469, opened 2026-08-17
cic: `bun run check` short-circuits — a biome failure hides every type error behind it


vjt:
`cicchetto/package.json:11`:

```json
"check": "biome check && tsc --noEmit && tsc --noEmit -p e2e/tsconfig.json",
```

`&&` means a non-zero `biome check` — including a pure **formatting** complaint — stops the run before
either `tsc` executes. The job then fails with an exit code that says "check failed", and the type
errors that would have been reported are simply never computed.

## Measured, twice in one session (2026-08-17)

1. PR #1465 went red on the `cicchetto (types + lint + unit)` job. The cause was biome **format** on
   one test file (`Found 1 error, 59 warnings`). Both `tsc` invocations had not run at all, so the
   branch's "green types" claim rested on nothing. After formatting, the types were in fact clean —
   but nobody could have known that from the red.
2. On a second branch the same short-circuit fired (unordered imports + format). After the format was
   fixed, **two real type errors appeared that no previous run had surfaced.**

Case 2 is the one that matters: the short-circuit does not merely delay the type check, it can hide
genuine type errors behind a cosmetic failure for as long as the cosmetic failure lasts.

## Why the local workaround is not enough

The known workaround is to run the two `tsc` invocations separately by hand. That relies on remembering
a trap, and the trap is invisible in the output — a red `check` looks the same whether tsc ran or not.

## Suggested direction (not prescribed)

Run the three independently and aggregate the exit codes, so one failure never masks another's result,
and report which stages ran. Whether that lives in `package.json`, in the CI job as separate steps, or
both, is a matter for whoever picks this up.

## Not established

- How many past red `check` runs hid a type error that was later fixed incidentally by a reformat — no
  history has been mined.

vjt:
Third measured instance today, and this one is the most expensive so far — it hid **three real type errors**.

A worker ran `bun run check` on the #1396 characterization branch and got RC=1. The failure was **biome formatting**. Because the script is `biome check && tsc --noEmit && tsc --noEmit -p e2e/tsconfig.json`, the `&&` chain meant **neither `tsc` pass ever ran**. After fixing the formatting, the two `tsc` passes ran for the first time and **found three genuine problems**, which were then fixed. Final state: `check` RC=0, `test` RC=0 (289 files / 5586 tests).

[...]

PR #1600, opened 2026-08-20
cic(#1571): make the cic gate say whether node_modules matches bun.lock

vjt:
## What this cures

The half of #1571 that is a gate telling you nothing: **`node_modules`
could disagree with `bun.lock` and no gate said so.** The only freshness
test any script applied was `scripts/bun.sh`'s `needs_install`, which
asks whether the directory is EMPTY — a populated but stale tree
answered "installed" and was never reconciled, and the documented
worktree procedure (`cp -Rc` from the main checkout) copied whatever was
there into every new worktree.

`bun run check` gains a first stage,
`cicchetto/scripts/lock-drift.ts`, which compares the installed tree
against the lock and names every package that disagrees. A stage rather
than a preflight in `bun.sh`: `check.ts` already owns #1469's
union-of-failures aggregation and adding one there costs no extra
container start.

## The measurement that drove it

Posted in full on the issue. The short version, because it decides the
design: `@biomejs/biome` is a JS wrapper and the binary lives in a
per-platform optional dependency, so the **declared** version and the
**executed** one are different packages. Same tree, two answers —
**2.5.8 in the container** (where every gate runs) and **2.4.13 from
`cicchetto/node_modules/.bin/biome` on the host**. Both drifted packages
were darwin binaries, both directories three months older than every
linux sibling, because `bun install` only ever resolves the optional
dependency for the platform it runs on and every gate here runs bun
inside a linux container.

## Three deliberate properties

* **Scoped to the executing platform.** The question is "does what will
  RUN HERE match the lock". Judging foreign-platform binaries would make
  the gate red on something no gate executes and no documented command
  can fix.
* **Known-answer controls inside the tool.** `runSelfTest()` runs before
  any real number and the gate exits **3 printing nothing** when a
  control fails.
* **`compared === 0` is a failure**, not a clean tree.

## Kill test — two-sided, on the real tree

| tree | `bun run check` |
|---|---|
| `@biomejs/cli-linux-arm64` at **2.5.6** (the version #1571 found in the wild), lock pins 2.5.8 | **rc=1, "4 stages ran, 1 failed"** — only the new stage, naming `DRIFT @biomejs/cli-linux-arm64 lock=2.5.8 installed=2.5.6` |
| same package restored byte-identical | **rc=0, "4 stages ran, 0 failed"** |

Mutation on the instrument itself — delete the one line that records a
drift:

* the gate exits **3**, prints **zero** number lines, and names three
  failing controls (`mixed/drift`, `mixed/verdict`, `native/drift`);
* the vitest suite reddens on the same mutant.

Restored byte-identical (`cmp`) before committing.

## The other half: a documented limitation, and that is my call

The stale darwin binaries are **not** cured, and I am calling them a
limitation rather than a gate that lies. What bounds it is measured, not
assumed: `bun`, `bunx` and `biome` are **absent from the host PATH**,
`scripts/check.sh` never invokes bun, no script or workflow runs biome
outside the container, and the repo carries no editor config and no git
hook. **No documented gate can produce a verdict from those binaries.**
What can reach them is a direct `node_modules/.bin/<tool>` call or an
editor integration — `docs/TESTING.md` now says so, and how to read the
version by hand.

Making the gate red on them would be a red nobody can clear: a linux
container install cannot resolve a darwin optional dependency, so there
is no command in this repo that would turn it green.

## Gates

* `scripts/bun.sh run check` — **rc=0, "4 stages ran, 0 failed"**.
* `scripts/bun.sh run test` — **rc=0, 294 files / 5688 tests**, 11 of
  them the new ones.
* `scripts/design-notes-gate.sh` — **rc=0, "1 new entry heading(s),
  separator and marker present"**.
* DESIGN_NOTES boundary: additions 93 / deletions 0, numstat identical
  across the rebase onto `0ea9177c`, prefix byte-identical, marker
  `grep -Fxc` = 1 here and 0 on `origin/main`.
* No Elixir gate run and none needed: the only non-cic file is a
  comment in `scripts/bun.sh`'s usage header, which
  `test/scripts/bun_self_heal_test.bats` does not read (it drives
  `run check` behaviour through a fixture).

## Not covered, deliberately

* **Only `check` carries the stage.** `run test` and `run build` consume
  the same tree unguarded; covering every verb means a check in
  `scripts/bun.sh` and one extra container start per invocation.
* **The missing direction is not checked** — a lock entry with nothing
  installed. The lock legitimately carries packages this platform never
  installs, and the false positives would teach people to ignore the
  gate. The completeness check made instead: every direct dependency of
  `package.json` must have been reached, or the gate fails naming it.
* **One host.** Nothing here says what another developer's tree holds.

### DIFF
--- a/cicchetto/scripts/check.ts
+++ b/cicchetto/scripts/check.ts
@@ -34,6 +34,14 @@
 type Stage = { readonly name: string; readonly argv: readonly string[] };
 
 const STAGES: readonly Stage[] = [
+  // First, and cheap: it says whether the three stages below are attributable
+  // at all. #1571 measured a `node_modules` whose biome was three minor
+  // versions off `bun.lock`, and nothing in the tree compared the two — the
+  // only freshness test any script applied was `scripts/bun.sh`'s
+  // `needs_install`, which asks whether the directory is EMPTY. It does not
+  // short-circuit the rest: a stale toolchain and a real type error are
+  // separate facts and #1469's whole point is reporting the union.
+  { name: "lock drift (node_modules vs bun.lock)", argv: ["bun", "scripts/lock-drift.ts"] },
   { name: "biome (src + e2e)", argv: ["biome", "check", "--max-diagnostics=none"] },
   { name: "tsc (src)", argv: ["tsc", "--noEmit"] },
   { name: "tsc (e2e)", argv: ["tsc", "--noEmit", "-p", "e2e/tsconfig.json"] },
--- a/cicchetto/scripts/lock-drift-core.ts
+++ b/cicchetto/scripts/lock-drift-core.ts
@@ -1,0 +1,288 @@
+// #1571 — is what is INSTALLED what `bun.lock` pins?
+//
+// The pure half: no filesystem, no Bun API, no imports. `lock-drift.ts` does
+// the IO and calls in here; `src/__tests__/lockDrift.test.ts` exercises both
+// sides of every rule. Keeping it importable from `src` is also what puts it
+// under `tsc --noEmit` — `cicchetto/scripts/` is outside both the tsconfig
+// `include` and biome's `files.includes`, so a runner-only module would be
+// checked by nothing.
+//
+// WHY THIS EXISTS. A gate verdict is only attributable if the tools that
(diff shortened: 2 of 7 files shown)


vjt:

Issue #1473, opened 2026-08-17
deploy: Deps and DirectoryIngest field-adds classify HOT — measured


vjt:
`Grappa.Session.Deps` and `Grappa.Session.DirectoryIngest` each define a
`defstruct` that is held as a field of `Grappa.Session.Server`'s state, and
neither is listed in `Grappa.HotReload.LongLivedModules`'s `@state_helpers`.
`Grappa.Deploy.Preflight` state-shape-checks only the files behind
`LongLivedModules.all/0` (`preflight.ex:243`), so a field-add to either
`defstruct` classifies **HOT**: the swapped beam then pattern-matches the new
struct shape against the state a live `Session.Server` already holds.

This is the exact class the preflight exists to refuse, and it is **already on
main** — both structs were introduced by the in-flight decomposition of #1390
(`Deps` by `794a9016`, `DirectoryIngest` by `f89e6e45`).

## What was measured

Not deduced from how the membership list is built — the classifier was run and
the verdict it *produced* is reported.

**Control, real revisions, real git, real CLI.** `c918c00f` ("irc(#902): make
the inviter part of the window") adds `invited_by` to `WindowState`'s
`defstruct`, and `WindowState` **is** in `@state_helpers`:

```
$ mix run -e 'Grappa.Deploy.Preflight.cli(["c918c00f^", "c918c00f", "jail"])'
Cold-deploy required:
  → state_shape: lib/grappa/session/window_state.ex
exit 3
```

**The two structs under test.** No historical field-add exists for either, so a
scratch field (`__pf_scratch__: nil`) was added to each `defstruct` in a
throwaway working tree and classified via `Preflight.classify/5` — which is
literally what `cli/1` calls (`preflight.ex:345`); `cli/1` adds only the two git
closures, the printing and the `System.halt`. `WindowState` was put through this
second harness too, with the same marker and the same mutation shape, so the
harness itself is controlled.

| module | path in `long_lived_module_files/0` | `classify_state_shape/2` asked directly | `Preflight.classify/5` (`:jail`) | exit code |
|---|---|---|---|---|
| `Session.Deps` | **false** | `:cold` | `{:hot, []}` | **0** |
| `Session.DirectoryIngest` | **false** | `:cold` | `{:hot, []}` | **0** |
| `Session.WindowState` (control) | true | `:cold` | `{:cold, [state_shape: ["lib/grappa/session/window_state.ex"]]}` | **3** |

[...]

PR #1602, opened 2026-08-20
deploy(#1473): derive the state-helper half of the tracked set too

vjt:
Refs #1473.

`Grappa.Deploy.Preflight` state-shape-checks only the files behind
`LongLivedModules.all/0`. That list has two halves. The `@modules` half —
long-lived `GenServer`s — has been derived-and-gated since #1343. The
`@state_helpers` half — structs held as **fields** of one of those processes'
state — was the half nothing derived; its own moduledoc asked only to "keep in
sync", and it had drifted.

## Reproduced before curing

The issue's positive control, re-run through the real CLI on real revisions:

```
$ mix run -e 'Grappa.Deploy.Preflight.cli(["c918c00f^", "c918c00f", "jail"])'
Cold-deploy required:
  → state_shape: lib/grappa/session/window_state.ex
exit 3
```

Then the defect, through `Preflight.classify/5` — which is literally what
`cli/1` delegates to — with a scratch `__pf_scratch__` field added to each
top-level `defstruct`:

| module | in checked set | mutated? | state block differs? | `classify_state_shape` | `classify/5` (`:jail`) |
|---|---|---|---|---|---|
| `Session.Deps` | **false** | ✓ | ✓ | `:cold` | **`{:hot, []}`** |
| `Session.DirectoryIngest` | **false** | ✓ | ✓ | `:cold` | **`{:hot, []}`** |
| `IRC.FakeLag` | **false** | ✓ | ✓ | `:cold` | **`{:hot, []}`** |
| `Session.WindowState` (control) | true | ✓ | ✓ | `:cold` | `{:cold, [state_shape: [...]]}` |

No row is vacuous: the scratch field is genuinely on disk and the extracted
state block genuinely differs on every one. `long_lived_module_files/0`
returned 28 paths.

**`IRC.FakeLag` is a third instance, and it is not in the issue.** It is a
field of `IRC.Client`'s state (`client.ex:171`) and was never listed at all.
The issue had recorded "no audit of the other tracked GenServers' `@type t`
blocks was done" as a known non-measure; the derivation below found it on its
first run. Three, not two.

## The membership is derivable — measured, then still put in the gate

Walking the `t` typespec of every `@modules` entry to a fixpoint and keeping
the reachable `:grappa` modules that define a struct reproduces **every one of
the nine hand-listed helpers**: the `listed \ derived` difference is **empty**.
So the hand list holds nothing the derivation cannot reach.

The list nevertheless stays hand-written, and the derivation became the gate:

- a runtime derivation feeds **no Dialyzer union** — `@type state_helper` is a
  literal union, and computing the list deletes that check;
- it **fails OPEN at deploy time**: a module that fails to load simply leaves
  the set and the preflight silently checks less — the exact failure mode the
  preflight exists to prevent. A hand list with a derived gate fails **red, in
  CI, before** the deploy;
- it keeps **one pattern for one problem**: this is the shape the `@modules`
  half already has.

*(The compile-time variant — deriving into the literal at compile time — was
judged unworkable because `LongLivedModules` would have to read the typespecs
of modules not yet compiled. That was reasoned, not prototyped.)*

The typespec is read from the **BEAM chunk** rather than the source, so module
names arrive fully qualified: no alias table to resolve, no second parser to
keep in step with the compiler.

## The unit of coverage is the FILE, not the module

`extract_state_block/1` reads a source *file*, so a struct nested inside
another module's file is covered by listing its **parent**. Measured: a
field-add to the nested `DirectoryIngest.Run` moves `directory_ingest.ex`'s
extracted block (`classify_state_shape` → `:cold`). Three such nested structs
(`LinksAccum.Entry`, `ListModeAccum.Entry`, `WhowasAccum.Entry`) were already
covered that way before this change, which is what makes the rule observed
rather than assumed.

Listing a nested module instead would be actively harmful:
`Grappa.Session.DirectoryIngest.Run` implies
`lib/grappa/session/directory_ingest/run.ex`, which does not exist, and an
absent file is read as empty at **both** revs — it compares equal and
classifies HOT. So the gate keys on each struct's **compile source**, and
nesting needs no special case.

## What the gate is worth, in both directions

| mutant | expected | measured |
|---|---|---|
| struct in a long-lived state, unlisted | red | red — `deps.ex`, `directory_ingest.ex`, `fake_lag.ex`, grouped by file, `DirectoryIngest.Run` collapsed onto its parent |
| a listed entry removed (`AwayState`) | red | red — **exactly one** test, naming `away_state.ex` |

## Two-sided kill test on the preflight itself

Same harness, before and after the cure:

| module | before | after |
|---|---|---|
| `Session.Deps` | `{:hot, []}` | `{:cold, [state_shape: ["lib/grappa/session/deps.ex"]]}` |
| `Session.DirectoryIngest` | `{:hot, []}` | `{:cold, [state_shape: ["lib/grappa/session/directory_ingest.ex"]]}` |
| `IRC.FakeLag` | `{:hot, []}` | `{:cold, [state_shape: ["lib/grappa/irc/fake_lag.ex"]]}` |
| `WindowState` (negative control) | `{:cold, [state_shape: [...]]}` | **identical to itself** |

Checked paths: **28 → 31**, exactly +3.

The end-to-end half was then measured through the real `cli/1` with two git
revisions — the path the issue recorded as *not established* for these modules
— by committing a scratch field-add and reading the process exit code:

| scratch commit | exit |
|---|---|
| field-add to `deps.ex` | **3**, `state_shape: lib/grappa/session/deps.ex` |
| field-add to `directory_ingest.ex` | **3**, `state_shape: lib/grappa/session/directory_ingest.ex` |
| field-add to `fake_lag.ex` | **3**, `state_shape: lib/grappa/irc/fake_lag.ex` |
| field-add to `window_state.ex` (control) | **3**, `state_shape: lib/grappa/session/window_state.ex` |
| **comment-only** change to `deps.ex` (negative control) | **0** — `no unsafe markers → HOT` |

The negative control is the point: it proves the harness produces both values
in this configuration, so "everything is 3" is a result rather than a stuck
instrument.

### A measurement trap paid for here

The first run of that harness returned **exit 0 for all four cases, control
included**. The container's `/app` is the MAIN checkout with the worktree's
`lib/` bind-mounted over it, so `.git` is main's: `HEAD^..HEAD` *inside the
container* resolved to main's last commit, not the worktree's scratch one, and
every row classified a docs-only diff. The verdicts were uniform because the
instrument was wrong, not because the code agreed. Passing explicit SHAs — the
object store IS shared across worktrees — fixed it. **A preflight harness must
name its revisions explicitly; a symbolic ref means the container's repo, not
yours.**

## Gates

| gate | result |
|---|---|
| `scripts/check.sh` (`mix ci.check` + `scripts/bats.sh`) | **rc=0** |
| ExUnit | 8 doctests, 61 properties, **6571 tests, 0 failures** |
| bats | `1..588` — **588 ok, 0 not ok** |
| Dialyzer | `done (passed successfully)` |
| Credo strict | `found no issues` |
| Sobelow | `Total errors: 0, Skipped: 0` |
| `mix deps.audit` | `No vulnerabilities found.` |
| `scripts/design-notes-gate.sh origin/main` | **rc=0** |

Stage 2 was verified to have actually run — the bats `1..588` plan line is
present, so `ci.check` did not halt stage 1 and leave a downstream green that
never existed.

Rebased onto `origin/main` with `scripts/union-rebase.sh` (its second real
use): `docs/DESIGN_NOTES.md: +105 -0 — contribution intact`, confirmed from a
second source by diffing the contribution numstat pinned to a file before the
rebase. Prefix byte-identical to the base's file — which subsumes the
preceding-entry tail check — boundary shape read on the real file, and the
`<!-- entry #1473 -->` marker unique in the file (`grep -Fxc` = 1) and absent
from the base.

## What this PR does NOT claim

- **That the three are all of them.** The closure follows `Mod.t()` references
  inside `t` typespecs. A struct held in a field typed `map()`, `term()` or
  `any()` is invisible to it, and that gap was **not measured**.
- **No production incident is claimed**, unchanged from the issue: what was
  measured is the classifier's verdict, not the jail's deploy history.
- **Only `:jail` was classified.** The state-shape stage is
  substrate-independent in `classify/5`, but `:docker` and `:linux` were read,
  not run.
- **That the compile-time derivation is impossible.** It was judged wrong for
  the reason above; it was not prototyped.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -53049,3 +53049,108 @@ gates.
 The two historical mentions of these names elsewhere in this file
 (`AdminLiveState`, `AdminVisitorLiveState`) are left untouched: the log
 records what was true when it was written.
+<!-- entry #1473 -->
+
+---
+
+## 2026-08-20 — #1473: the hand-maintained half of the tracked set, derived
+
+`Grappa.Deploy.Preflight` state-shape-checks only the files behind
+`LongLivedModules.all/0`. That list has two halves. The `@modules` half —
+long-lived `GenServer`s — has been derived-and-gated since #1343. The
+`@state_helpers` half — structs held as FIELDS of one of those processes'
+state — was the half nothing derived; its own moduledoc asked only to "keep
+in sync", and it had drifted three ways.
+
+`Session.Deps` and `Session.DirectoryIngest` arrived with the #1390
+decomposition. `IRC.FakeLag`, a field of `IRC.Client`'s state, was never
+listed at all — it is not in the issue either, which had recorded "no audit
+of the other tracked GenServers' `@type t` blocks was done" as a known
+non-measure. The derivation found it on its first run. **Three, not two.**
+
+Measured through `Preflight.classify/5` on `:jail`, a scratch `defstruct`
+field-add to each classified `{:hot, []}`, while `WindowState` — listed —
+returned `{:cold, [state_shape: [...]]}` from the same harness. A HOT deploy
+carrying such a field-add swaps the beam under a live process still holding
+the old struct; the mismatch surfaces on whatever message next
+pattern-matches it.
+
+### The membership is derivable, and the derivation belongs in the gate
+
+Measured before choosing: walking the `t` typespec of every `@modules` entry
+to a fixpoint and keeping the reachable `:grappa` modules that define a
+struct reproduces **every one of the nine hand-listed helpers** — the
+`listed \ derived` difference is EMPTY. So the list holds nothing the
+derivation cannot reach, and enumerating it by hand buys nothing that
+deriving it does not.
+
+The list stays hand-written anyway, and the derivation became the gate. A
+runtime derivation would feed no Dialyzer union — `@type state_helper` is a
+literal union — and it would fail OPEN at deploy time: a module that fails
+to load simply leaves the set and the preflight silently checks less, which
+is the failure mode the preflight exists to prevent. A hand list with a
+derived gate fails red, in CI, before the deploy. It is also the shape the
+`@modules` half already has, so the two halves stay one pattern rather than
+two. *(The compile-time variant — deriving into the literal at compile time
+— was judged unworkable because `LongLivedModules` would have to read the
+typespecs of modules not yet compiled. That was reasoned, not prototyped.)*
+
+The typespec is read from the BEAM chunk rather than the source: module
+names arrive fully qualified, so there is no alias table to resolve and no
+second parser to keep in step with the compiler.
+
+### The unit of coverage is the FILE, not the module
+
+`extract_state_block/1` reads a source file, so a struct nested inside
+another module's file is covered by listing its PARENT. Measured: a
+field-add to the nested `DirectoryIngest.Run` moves `directory_ingest.ex`'s
+extracted block. Three such nested structs (`LinksAccum.Entry`,
+`ListModeAccum.Entry`, `WhowasAccum.Entry`) were already covered that way
+before this change, which is what makes the rule observed rather than
+assumed.
+
+Listing a nested module instead would be actively harmful:
+`Grappa.Session.DirectoryIngest.Run` implies
+`lib/grappa/session/directory_ingest/run.ex`, which does not exist, and an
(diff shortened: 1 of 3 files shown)


vjt:

Issue #336, opened 2026-07-19
grappa P0: DB pool queue_timeout crashes session GenServer → clients disconnected (persist_event must not take down the session)


vjt:
## Incident
On 2026-07-19 at 09:17:27 (server local), ~17 live sessions disconnected simultaneously. Root cause: the SQLite `Repo` connection pool saturated for ~1 second and dropped queued requests after 186ms (`reason: :queue_timeout`). Each affected session's GenServer **crashed** and the client was disconnected; auto-reconnect recovered within seconds. All 17 `queue_timeout` events landed in that single second — none before, none since.

This is wrong behaviour: **a slow/saturated DB must never disconnect users.** Scrollback persistence is best-effort durability — it must degrade (retry / defer / drop), not take down the session.

## Crash path
`Repo.insert/2` **raises** `DBConnection.ConnectionError{reason: :queue_timeout}` when the pool can't serve a checkout in time — it does *not* return `{:error, _}`. But `persist_event/1` only matches `{:ok, _}` / `{:error, _}`:

```elixir
# lib/grappa/scrollback.ex:125
def persist_event(%{kind: kind} = attrs) when is_atom(kind) do
  changeset = Message.changeset(%Message{}, attrs)

  case Repo.insert(changeset) do
    {:ok, message} -> {:ok, Repo.preload(message, :network)}
    {:error, _} = err -> err
  end
end
```

The raised exception propagates uncaught: `Scrollback.persist_event/1` → `Session.Server.apply_effects/2` (the `:persist` effect) → `Session.Server.delegate/2` → the session `gen_server` `handle_info` → **GenServer crash** → session dies → client disconnected.

Stacktrace (running 0.3.2 release):
```
%DBConnection.ConnectionError{reason: :queue_timeout}  (dropped from queue after 186ms)
  Ecto.Adapters.SQL.raise_sql_call_error/1     lib/ecto/adapters/sql.ex:1121
  Ecto.Repo.Schema.apply/4                     lib/ecto/repo/schema.ex:999
  Ecto.Repo.Schema."-do_insert/4-fun-4-"/16    lib/ecto/repo/schema.ex:503
  Grappa.Scrollback.persist_event/1            lib/grappa/scrollback.ex:128
  Grappa.Session.Server.apply_effects/2        lib/grappa/session/server.ex:3669
  Grappa.Session.Server.delegate/2             lib/grappa/session/server.ex:3029
  :gen_server.try_handle_info/3
```

## Fix direction
[...]

PR #1604, opened 2026-08-20
db(#1421): report the wait the retry engine observed, not the budget

vjt:
Closes the half of #1421 that can be closed without touching #1420's
contested axis. A + B of the option table posted on the issue; **C / D / E / H
are deliberately NOT taken** — they change retry behaviour under contention,
which is vjt's ruling to make.

## The defect

`BusyRetry`'s terminal warning read `for the full 1500ms retry budget
(1 attempts)` after waiting 30 067 ms. CLAUDE.md's log-honesty rule says a line
describes the state it OBSERVED, and #1420 had just corrected this same line
for the same reason — so this is a bugfix, not a tuning change.

**Why the number was wrong.** The budget is a deadline consulted BETWEEN
attempts: it cannot preempt an attempt already running, because the engine does
not own the wait. A `busy_locked` fault raises only once SQLite's `busy_timeout`
(30 000 ms) has expired — 20× the budget — so the first attempt has already
overshot the deadline when it returns, and the loop makes exactly one attempt.
The budget was dimensioned in #340 against the POOL topology (`config.exs`: *"the
~1s pool-saturation window the #336 incident measured"*), where DBConnection
drops a checkout near `queue_target` and the retries are real. One number, two
topologies — not a wiring slip.

## What changed

- `run/2` carries `started` instead of a precomputed deadline;
  `elapsed_ms < @budget_ms` replaces `monotonic_time < deadline` — the same arm
  at the same boundary, rearranged so one subtraction feeds the line.
  **Zero behaviour moves**: same retry, same `{:error, :db_unavailable}`, same
  `fault:` metadata, **0 of the 71** wrapped call sites touched.
- The moduledoc now states how far the budget reaches, per topology, and that
  the lock regime is a documented limitation rather than a slip.
- `docs/OPERATIONS.md`: how an operator reads the two numbers together.

## The census moved in LOCKSTEP — the trap this PR had to close

`scripts/log-gap-scan.awk` counted `saturated` on `/saturated for the full/`,
the numeric tail this PR rewrites. Left alone it would have reported **zero**,
and zero is what a clean run looks like: that census is the attribution
instrument for red e2e runs (#1429), so blinding it costs the next reader a
measurement they believe they took.

Both terminal counters now anchor on the `observed_state/1` phrase **alone** —
the half that names the topology and carries no numbers. The scanner's
known-answer samples and the two verbatim bats pins move in the same commit,
plus a new bats case feeding both lines with a **different** numeric tail, so
the decoupling is asserted rather than asserted-about.

## Evidence

The new bench provokes a **real driver `SQLITE_BUSY`** against a held write
lock, instead of the hand-built zero-cost fault every other retry test raises —
which is exactly why the engine's own suite could not see this: with a free
fault the budget IS the only bound, the one regime in which it appears to work.

Red read before green, on unmodified `origin/main` code:

```
db write unavailable: SQLite write lock held by another writer
for the full 300ms retry budget (1 attempts)      <- after a ~600ms wait
```

Two-sided, by mutation of the deadline arm:

| mutant | kills |
|---|---|
| `attempt < 3` | `attempts == 1`, and `elapsed >= budget` with **left = 9 ms** |
| `false` | **only** `attempts > 1` — one mutant, one assertion |

The second is the proof that the *budget* — not an attempt counter — is what
holds the loop for its window: without it the elapsed collapses to 9 ms.

Local gates on the pre-rebase tree: `mix ci.check` + `gen_wire_types --check` +
the full bats set, exit 0 (`_lib.sh` is `set -euo pipefail` and every `ci.check`
step is a halting `mix cmd`, so the bats stage is only reachable on a zero
exit). `test/grappa/repo/` 27/27; `log_gap_scan_test.bats` 19/19.

**Named gap:** those gates ran on `0ea9177c`; this branch is now rebased onto
`fa2308b9`, which touches only `cicchetto/src/lib/api.ts` and
`docs/DESIGN_NOTES.md` — no overlap with anything here except the union-merged
decision log, whose contribution was pinned and re-compared across the rebase
(+138 −0, unchanged; preceding entry byte-identical).

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -53049,3 +53049,141 @@ gates.
 The two historical mentions of these names elsewhere in this file
 (`AdminLiveState`, `AdminVisitorLiveState`) are left untouched: the log
 records what was true when it was written.
+<!-- entry #1421 -->
+
+---
+
+## 2026-08-20 — #1421: a budget that cannot bound the fault it is pointed at
+
+#1420 named this and left it: *"the retry budget cannot fire on the fault it
+exists for"*. This closes the half that can be closed without touching what
+#1420 calls the contested axis. A + B of the option table below shipped;
+C/D/E/H did not, and are vjt's call.
+
+### The mechanism, stated exactly
+
+`BusyRetry`'s budget is a deadline consulted BETWEEN attempts. It cannot
+preempt an attempt already running, because the engine does not own the wait —
+SQLite does. So the reach depends on the FAULT'S OWN latency, and the live
+topologies do not share a regime:
+
+| topology | who owns the wait | latency before the raise | inside the 1500ms budget? |
+|---|---|---|---|
+| pool `queue_timeout` | DBConnection | ~100ms (target 50ms, doubled once, then dropped) | yes |
+| write-lock `busy_locked` | SQLite's busy handler | up to `busy_timeout` = 30_000ms | **no — 20x over** |
+| deferred read->write upgrade | nobody, raises at once | ~0ms | yes, but gated out |
+
+The third row is not live: `Repo.transaction(` appears **0 times** in `lib/`,
+enforced by `Grappa.Repo.TransactionModeGateTest` (#1374). Both live write
+paths are therefore slow-fault paths — the 122 bare `Repo.insert/update/…`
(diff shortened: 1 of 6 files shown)


vjt:

Issue #1603, opened 2026-08-20
deploy(docker): the pull's die names 'the branch diverged' for a failure it never observed


vjt:
## What happens

On a Docker deploy checkout that carries any unstaged change, `scripts/deploy.sh` aborts and the
error blames a cause that is not the cause:

```
error: cannot pull with rebase: You have unstaged changes.
error: Please commit or stash them.
pull is not a fast-forward — the branch diverged. Resolve it by hand.
```

The branch had **not** diverged. `git` refused for an entirely different reason, and the deploy's
own `die` overwrote that reason with a guess. The operator is then sent to "resolve the divergence
by hand", which is work on a problem they do not have, while the real fix (commit, stash, or make
the pull explicit) goes unnamed.

This is a `Log honesty` violation in the same family as #1601 — same deploy run, different defect —
but it is a distinct fault with a distinct cure, so it is filed on its own.

## Mechanism

`infra/lib/deploy_docker.sh` maps every non-zero exit of the pull onto one sentence:

```sh
git pull --ff-only || die "pull is not a fast-forward — the branch diverged. Resolve it by hand."
```

`git pull` has more than one way to fail, and `--ff-only` does not pin which machinery runs. With
`branch.<name>.rebase = true` in the repo config, `git pull` is a **rebase** pull, and a rebase
refuses on any unstaged change regardless of whether a fast-forward was available. The operator
never wrote that key: `branch.autosetuprebase = remote` in a global gitconfig writes it into
`.git/config` at branch-creation time. A deploy checkout picks up unstaged changes routinely — a
submodule pointer, a tool that edits a file inside the checkout.

## Measured

Throwaway clone, `branch.main.rebase = true`, one unstaged change, `HEAD == origin/main` (so
provably no divergence at all):

| case | command | result |
|---|---|---|
| defect | `git pull --ff-only` | `rc=128`, `cannot pull with rebase: You have unstaged changes.` |
| control | same, after `git config --unset branch.main.rebase` | `rc=0`, `Already up to date.` |

The control is what makes this a mechanism and not a coincidence: the only variable changed is the
config key, and the failure follows it.

Second measurement, with a **real** fast-forward waiting upstream and an unstaged change on an
unrelated file:

| case | command | result |
|---|---|---|
| today | `git pull --ff-only` | `rc=128`, refuses, nothing pulled |
[...]

Issue #1601, opened 2026-08-20
deploy(docker): the hot path reloads before anything compiles, so every hot deploy lands the code one commit late


vjt:
## What happens

On the **Docker substrate**, a hot deploy lands the new code **one deploy late**. `scripts/deploy.sh`
prints `✓ hot-deploy complete (sessions preserved, container ID unchanged)` while the live BEAM is
still running the previous commit's modules. The next deploy (or any manual `POST /admin/reload`)
loads them.

This is **not** #1287 (CLOSED): that one is the release substrate, where `:code.lib_dir/1` keeps
pointing at the boot-time versioned `lib/grappa-<vsn>/ebin`. Here the ebin path is right and the
reload works — it simply runs **before anything has compiled the new sources**.

## Mechanism (read off the deploy libs)

The Docker hot path never compiles:

```sh
# infra/lib/deploy_docker.sh — substrate_build()
# Hot path needs no build — the pulled commit is already in the
# bind-mounted source tree (compose.yaml mounts ./:/app).
[ "$MODE" = cold ] || return 0
```

`POST /admin/reload` only **loads** beams; it does not compile (`Grappa.HotReload.reload_modified/0`
walks the app's ebin and compares on-disk md5 against `module_info(:md5)`).

And the hot path's order is fixed in `infra/lib/deploy_common.sh` (`_deploy_hot`):

```
substrate_reload      # POST /admin/reload
_deploy_seed          # docker: compose run --rm grappa mix grappa.seed_themes
_deploy_healthcheck_loop
```

`mix grappa.seed_themes` is the **only** thing on the hot path that compiles — and it runs *after*
the reload. So: reload sees the stale ebin and honestly answers `reloaded: []`; the seed's `mix` run
then recompiles the changed sources into `_build/dev/lib/grappa/ebin`; nobody reloads them.

The comment above `_deploy_seed` explains the reload→seed order deliberately (post-#41 the reload
applies expand migrations, so a seed before it would run against the pre-migration schema). The
ordering is correct for its stated reason; what is missing is a compile step before the reload.

## Measurement (staging, `grappa.bad.ass`, 2026-08-20)

Deploy of `3e01f0a6..0ea9177c` (`#1420`, touches `lib/grappa/repo/busy_retry.ex`), classified hot,
banner reported success. Then, by hand:

| step | call | result |
|---|---|---|
[...]

PR #1605, opened 2026-08-20
deploy(#1601): compile on the Docker hot path, before the reload

vjt:
On the Docker substrate a hot deploy landed the new code **one deploy late**, while the banner, the
exit code and the completed-deploy marker all reported success.

`POST /admin/reload` only **loads** beams — `Grappa.HotReload.reload_modified/0` md5-walks the app's
ebin and compiles nothing. Nothing on this substrate's hot path wrote that ebin:

```sh
# infra/lib/deploy_docker.sh — substrate_build(), before this PR
# Hot path needs no build — the pulled commit is already in the
# bind-mounted source tree (compose.yaml mounts ./:/app).
[ "$MODE" = cold ] || return 0
```

That claim is true of the **sources** and says nothing about the **beams**. So the reload found the
previous commit's modules unchanged, answered `reloaded: []` honestly, and the deploy declared
success over code that had never been loaded.

## Every step was individually honest; the sum was not

| step | declares | observes |
|---|---|---|
| `substrate_build` | "no build needed" | nothing |
| `substrate_reload` | proceeds on `"failed":[]` | only `failed` — `reloaded` is in the response, logged and never read (verified: zero read sites across the five deploy consumers) |
| `substrate_healthcheck` | "healthy" | a 200 on `/healthz` that the OLD code returns identically |
| `substrate_write_marker` | "this sha is deployed" | nothing; it writes `NEW_SHA` unconditionally |
| `done_banner` | "✓ hot-deploy complete" | the sum of the above |

Not one line is a lie on its own. Together they declare "the new commit is live" while observing
"the old process is up and nothing failed to load". That is the `Log honesty` rule at deploy scale.

## The cure ends an exception rather than adding a mechanism

`infra/freebsd/deploy.sh` and `infra/linux/deploy.sh` already build on **both** paths and state the
reason: the build *"writes the fresh .beam into the daemon's code path that the hot reload POST then
loads"*. Docker was the only substrate out of line, so this adds a hot arm to its `substrate_build`
rather than a step to the shared algorithm.

Placement is free: `deploy_common.sh` calls `substrate_build` **before** the hot/cold branch, so a
compile there precedes `substrate_reload` by construction.

- **No `deps.get` beside it.** `mix.lock`/`mix.exs` are COLD triggers (`Preflight.mix_deps?/1`), so a
  hot deploy cannot bring deps the box has not already fetched.
- **`--warnings-as-errors`, matching both release substrates.** Under the consumers'
  `set -euo pipefail` a failed compile aborts *before* the reload — the point of the step.
- **Classification is untouched.** `Preflight.classify/5` still decides WHEN a change is hot; this
  changes what the hot path DOES.

**Rejected:** re-running the reload after the seed. `mix grappa.seed_themes` is the one hot-path step
that compiles today — by accident, as a side effect of being a mix task — and `_deploy_hot` runs it
after the reload for a documented reason (post-#41 the reload applies expand migrations, so an
earlier seed would meet the pre-migration schema). That shape would promote an accident into the
build step and pay two reloads where one suffices.

## Blast radius, checked across all three hook sets

`infra/lib/deploy_docker.sh` is shared by two entry points since #1384, and a third mode declares its
own hooks:

| consumer | gets the compile? | right? |
|---|---|---|
| `scripts/deploy.sh` (operator door) | yes | yes — checkout with bind-mounted sources |
| `infra/docker/deploy.sh update` → `cmd_update()` | yes, by sourcing the same hook set | yes — same substrate, same defect |
| `infra/docker/deploy.sh` → `cmd_update_release()` | no — declares `substrate_build() { :; }` inside the function and forces cold | yes — a checkout-less box has no sources to compile |

## Evidence

**Red first, and load-bearing by mutation.** The new case asserts a `mix compile` exists in the
stubbed argv log **and that its line precedes `admin/reload`** — ordering, not presence.

| mutant | expected victim | measured |
|---|---|---|
| reinstate `[ "$MODE" = cold ] \|\| return 0` on the hot arm | the presence assert | `not ok`, line 158 |
| move the compile into `substrate_seed` (present, but after the reload) | the ordering assert | `not ok`, line 162 |

The second mutant is not optional: the first red never reaches the ordering line, so without it that
assertion would have shipped unexercised.

The assert is not vacuous — no other argv in the fixture contains `mix compile` (the preflight is
`run --no-start`, the seed is `mix grappa.seed_themes`), and the pre-cure red proves the grep found
nothing.

It deliberately does **not** assert a non-empty `reloaded` list: a doc-only hot deploy legitimately
reloads nothing, and the first case in the file already pins that shape as valid.

**Gates.** `scripts/bats.sh` full: `1..589`, 589 ok, 0 failures. `scripts/shellcheck.sh`: clean over
77 derived scripts, with a positive control (an SC2086 injected into the new hot arm turned it red
and named the file and line, proving the linter reads this worktree's bytes and not the shared main
checkout's). `scripts/posix-parse.sh`: clean, but **vacuous for this change** —
`infra/lib/deploy_docker.sh` declares `shell=bash` and is not in that gate's derived set.
`scripts/check.sh`: `rc=0`. Both stages are evidenced rather than assumed — the upstream
`mix ci.check` left Credo, Sobelow, Dialyzer, doctor and `8 doctests, 61 properties, 6569 tests, 0
failures` in the log, and the bats stage printed its plan line `1..589` with 589 ok and zero
failures. The plan line matters: `check.sh` inherits `set -euo pipefail` from `scripts/_lib.sh`, so a
halting `ci.check` would mean bats never started and "zero failures" would read as "not measured".

## Declined, and named rather than omitted

- **The cold path is left alone.** Its beams still come from the recreated container's own
  `mix phx.server` boot compile, so a compile failure there surfaces as a crash loop plus a
  healthcheck timeout rather than an immediate abort. Pre-existing, and outside this issue: cold
  demonstrably lands (the `v1.3.0-rc1` cut used `--force-cold`).
- **No second test on the standalone door.** #1384 made the two hook sets one function, so a
  duplicate case would exercise the same code twice. The cost, stated: if the hook sets are ever
  re-split, this single test will not catch the second door regressing.
- **Read from the code, not reproduced:** `_deploy_nothing_to_do` exits 0 when
  `PREV_SHA == NEW_SHA` and `LAST_DEPLOYED == NEW_SHA`. The marker carries the new sha whether or not
  anything loaded, so an operator who *re-runs* the deploy to remedy a stale box is told "nothing to
  do". The issue's "self-healing by accident" therefore holds only when a NEW commit arrives, not by
  retrying. No box was available to reproduce this.

## Spun off

The same deploy session surfaced a second `Log honesty` fault — the pull's `die` reports "the branch
diverged" for a refusal it never observed. Different fault, different cure: filed as #1603 rather
than widened into this PR.

Refs #1601.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -53049,3 +53049,104 @@ gates.
 The two historical mentions of these names elsewhere in this file
 (`AdminLiveState`, `AdminVisitorLiveState`) are left untouched: the log
 records what was true when it was written.
+<!-- entry #1601 -->
+
+---
+
+## 2026-08-20 — #1601: the Docker hot deploy declared a landing it never compiled
+
+`POST /admin/reload` LOADS beams; it does not make them.
+`Grappa.HotReload.reload_modified/0` md5-walks the app's ebin and reloads what
+changed on disk. On the Docker substrate nothing on the hot path wrote that
+ebin: `substrate_build` opened with `[ "$MODE" = cold ] || return 0` and the
+comment *"the pulled commit is already in the bind-mounted source tree"* —
+true of the SOURCES, silent about the BEAMS. So the reload found the previous
+commit's modules unchanged, answered `reloaded: []` **honestly**, and the
+deploy declared success over code that was never loaded. Measured on staging:
+a manual reload immediately after the ✓ banner found two modules, and the two
+idempotent reloads after that found none — the control that proves the
+deploy's own reload had not loaded them.
+
+### Every step was individually honest; the sum was not
+
+| step | declares | observes |
+|---|---|---|
+| `substrate_build` | "no build needed" | nothing |
+| `substrate_reload` | proceeds on `"failed":[]` | only `failed`; `reloaded` is in the response, logged and never read |
+| `substrate_healthcheck` | "healthy" | a 200 on `/healthz` that the OLD code returns identically |
+| `substrate_write_marker` | "this sha is deployed" | nothing; it writes `NEW_SHA` unconditionally |
+| `done_banner` | "✓ hot-deploy complete" | the sum of the above |
+
+Not one of those lines is a lie on its own. Together they declare "the new
+commit is live" while observing "the old process is up and nothing failed to
+load". That gap is the `Log honesty` rule at deploy scale: **a gate that
+never ran the step cannot report on it, and reporting green anyway is the
+defect** — vjt's "a documented limitation is an acceptable outcome" does not
+extend to a gate that declares a verdict it did not measure.
+
+### The cure ends an exception rather than adding a mechanism
+
+`infra/freebsd/deploy.sh` and `infra/linux/deploy.sh` already build on BOTH
+paths and state the reason: the build *"writes the fresh .beam into the
+daemon's code path that the hot reload POST then loads"*. Docker was the only
+substrate out of line, so the fix is a hot arm in its `substrate_build`, not a
+new step in the shared algorithm.
+
+Placement costs nothing: `deploy_common.sh` calls `substrate_build` **before**
+the hot/cold branch, so a compile there precedes `substrate_reload` by
+construction. Three consequences worth pinning:
+
+- **No `deps.get` beside it.** `mix.lock`/`mix.exs` are COLD triggers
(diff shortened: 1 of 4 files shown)


vjt:

Issue #1580, opened 2026-08-19
validate_subject_xor/1 is hand-copied 12x in lib/ — one copy has already drifted


vjt:
`defp validate_subject_xor/1` is hand-copied **12 times** under `lib/` (measured on `51eaa6c1`). That alone is a tidiness complaint. What makes it a defect is that **one copy has already drifted, and its own comment claims it hasn't.**

### The drift

`lib/grappa/uploads/upload.ex:135` differs from the other eleven in two ways:

| | the eleven | `uploads/upload.ex` |
|---|---|---|
| `{nil, nil}` error | `"must set user_id or visitor_id"` | `"one of user_id or visitor_id is required"` |
| `@spec` | present | **absent** |

The `{_, _}` arm (`"user_id and visitor_id are mutually exclusive"`) is identical in all twelve, so the divergence is exactly the *missing-subject* message — the one an operator or a client would render.

And the copy carries this comment, three lines above the body that contradicts it:

```elixir
# Mirrors `Grappa.ReadCursor.Cursor.validate_subject_xor/1` —
# error attaches to the synthetic `:subject` key so the wire shape
# matches every other XOR-FK context.
```

`ReadCursor.Cursor` emits `"must set user_id or visitor_id"`. The comment asserts a mirror the code no longer is.

### Where the twelve are, and what each claims to mirror

| file | "Mirror of" pointer |
|---|---|
| `accounts/session.ex:245` | `Scrollback.Message` |
| `scrollback/message.ex:399` | `Accounts.Session` |
| `read_cursor/cursor.ex:115` | `Scrollback.Message` |
| `query_windows/window.ex:89` | `ReadCursor.Cursor` |
| `channel_directory/entry.ex:88` | `QueryWindows.Window` |
| `notify/entry.ex:102` | `QueryWindows.Window` |
| `push/subscription.ex:196` | `ReadCursor.Cursor` |
| `user_settings/settings.ex:120` | `ReadCursor.Cursor` |
| `vhosts/grant.ex:78` | `ReadCursor.Cursor` |
| `themes/theme.ex:85` | `UserSettings.Settings` |
| `networks/credential.ex:457` | none |
| `uploads/upload.ex:135` | `ReadCursor.Cursor` — **the divergent one** |

[...]

PR #1611, opened 2026-08-20
refactor(#1580): one subject-XOR predicate instead of twelve hand-copies

vjt:
Consolidates the twelve hand-copied `defp validate_subject_xor/1` bodies into
one `Grappa.Subject.validate_xor/1`, and pins the contract on all twelve
schemas. Refs #1580.

## The number was re-measured before it was trusted

#1580's "twelve" came from grepping the function NAME, which counts
declarations rather than duplication — an inlined, renamed or anonymous copy
is invisible to it. Re-anchored on the **decision body**,
`case {user_id, visitor_id} do … end`, with the enclosing name demoted to a
derived column:

| axis | count |
|---|---|
| XOR decision bodies in `lib/` | **12** |
| distinct enclosing names | 1 — no inlined or renamed copy |
| schemas declaring `belongs_to :visitor` | 12 |
| `<table>_subject_xor` CHECK constraints in migrations | 12 |
| visitor FK but **no** decision body | 0 |
| subject `add_error` outside a counted body | 0 |

The number holds and the population is closed on four independent axes. The
census script carried six known-answer controls **inside** it — including a
decoy `defp` that takes a changeset but is not an XOR, plus a check that the
decoy is actually present so the control is not vacuous — and printed no
numbers unless all six passed. No `git grep -E` was used for counting: POSIX
ERE has no `\s` and no `\b`.

The "no copy in a different shape" zero was earned, not assumed: the ten
files naming both columns but holding no body were all read. Nearest miss is
`Scrollback.persist_subject/2`, a telemetry labeller — not a validator.

## Why flattening is safe, measured rather than argued

Hashing each body twice — literals included, then literals masked — gives
**2 classes by text** (11 + 1) but **1 class by control flow** (12). The
entire divergence in the family is the missing-subject message, so no copy
carried behaviour this change drops.

`uploads/upload.ex` said `"one of user_id or visitor_id is required"` where
the eleven said `"must set user_id or visitor_id"`. Classified as a typo, not
a requirement, on four measurements: identical control flow; its own comment
claims to mirror `ReadCursor.Cursor`, which emits the other string; zero test
pins on the variant; and **the arm is unreachable through its only production
caller** — `Uploads.create/3` routes the subject through
`Subject.put_subject_id/2`, whose clauses both guard `is_binary`.

That last one downgrades the issue's framing: the drift is real and measured,
but it is a latent inconsistency, **not an observable defect**.

## The root cause the issue described but did not name

`accounts/session.ex` → `scrollback/message.ex` → `accounts/session.ex` is a
cycle. The pointer graph has no root, so no copy was the designated original
and every "mirror of" claim was unfalsifiable by construction. A comment could
not have caught this. Copies are not kept honest by comments; the fix is to
have one.

## Home

`Grappa.Subject` rather than a new module: it already owns the subject
discriminator, its moduledoc already states the XOR invariant, and this is the
write-side twin of `put_subject_id/2` — one puts the column, the other proves
the pair. Ten of the twelve schemas sit in context boundaries that already
declared `Grappa.Subject`; only `Accounts.Session` and `Networks.Credential`
own their boundaries and needed the edge. No cycle.

`lib/` net **+94 / −228** over 14 files; 17 stale doc citations repointed.

## The red came first, and it was the drift

`test/grappa/subject_xor_test.exs` — one table over all twelve schemas, four
cases each, plus a population guard. Run before any production change:

```
49 tests, 1 failure

1) Grappa.Uploads.Upload.insert_changeset/2 subject XOR
   neither user_id nor visitor_id attaches the shared missing-subject message
   left:  "must set user_id or visitor_id"
   right: ["one of user_id or visitor_id is required"]
```

Reproducible from the base:

```sh
git checkout 8a86b178
git checkout w1-1580 -- test/grappa/subject_xor_test.exs
scripts/test.sh test/grappa/subject_xor_test.exs   # 49 tests, 1 failure
```

Every commit on the branch is green, so a bisect never lands on a broken tree.
Before this the suite pinned the text in 3 of 12 schemas and was silent on the
other nine, including the one already wrong.

## The pin is load-bearing — 7 mutants, 0 survivors

M0 is a calibration run: the hand-written parser must read the known baseline
(49/0) off an unmutated tree or the harness aborts. Every injection asserts
its predicate matches **exactly once** before writing.

| mutant | killed | arms hit |
|---|---|---|
| M1 reinstates the exact #1580 drift in the single function | 12/49 | missing-subject ×12 |
| M2 changes the both-set message | 12/49 | mutually-exclusive ×12 |
| M3 rejects a user-only subject | 12/49 | user-only-valid ×12 |
| M4 rejects a visitor-only subject | 12/49 | visitor-only-valid ×12 |
| M5 accepts both-set (drops the XOR) | 12/49 | mutually-exclusive ×12 |
| M6 one call site stops routing through the validator | **2/49** | uploads only |
| M7 a schema drops out of the table | **1/45** | population-guard ×1 |

**M1 is the requirement discharged**: the single function gets one case wrong
— precisely the case one copy treated differently — and 12 assertions die;
before consolidation the same mistake killed 3. Each mutant hits exactly one
arm, so the assertions discriminate rather than being one blanket check
wearing four names. **M6 kills exactly one schema's two rows**: a *partial*
consolidation, the likeliest future regression, is caught and localised. M7 is
the only mutant that moves the denominator (49 → 45), which is itself the
evidence the guard measures coverage.

## Gates

`scripts/check.sh` **rc=0** on the rebased head, every stage confirmed to have
run: credo 696 files 0 issues · `deps.audit` clean · sobelow Low-only (gate is
Medium+) · **6630 tests / 8 doctests / 61 properties / 0 failures** · dialyzer
**0 errors** · bats **590/590**. Plus `format.sh --check` rc=0,
`design-notes-gate.sh` ok, and `union-rebase.sh` reporting
`docs/DESIGN_NOTES.md +106 −0 — contribution intact` with the boundary shape
read on the real file afterwards.

An earlier `check.sh` **halted at credo** on an issue that was mine
(`credential.ex` ended up with two single `Grappa.*` aliases where the repo
uses the multi form). Fixed, folded into the refactor commit, whole gate
re-run from scratch — tests, dialyzer, sobelow and bats never ran on the
halted invocation, so that run is reported as *not measured*, never as zero.

## Two findings left OUT of this PR, deliberately

1. **`check_constraint(:subject, …)` is in 10 of 12.** `accounts/session.ex`
   and `scrollback/message.ex` lack it although `sessions_subject_xor` and
   `messages_subject_xor` both exist in the DB, so a violation reaching `Repo`
   there raises `Ecto.ConstraintError` instead of returning a changeset error.
   A different axis from the copied predicate, and reachability is thin (the
   changeset predicate is byte-identical to the CHECK). **A call for @vjt** —
   defence-in-depth, deliberate-and-wants-a-comment, or not worth it.
2. **#1580's pointer table records `networks/credential.ex` as having no
   pointer.** It has one, spelled `Byte-mirror of`. An anchor matching
   `Mirror of` misses it. The *inbound* claim in the issue still stands. The
   consolidation script hit the identical trap and refused the file rather
   than guessing, which is how it was caught.

## Not claimed

Eleven declared non-measures are enumerated in the report, each marked
closable or structural. The load-bearing ones: `{nil, nil}` reachability was
traced for `uploads` only; no client trace was run, so no claim that the
divergent message was ever user-visible; the twelve bodies were proven
equivalent *structurally* (normalised-body hash) plus the post-hoc table, not
by a pre-consolidation differential run; the `ConstraintError` consequence in
finding 1 is read from Ecto semantics, not reproduced; and **e2e /
`scripts/integration.sh` was not run** — the change is server-side Elixir with
no cic or wire surface, but that is an argument, not a measurement, and per
CLAUDE.md the full integration suite green is the ship gate.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -53699,3 +53699,109 @@ The `EventRouter` per-kind shape table still read `required text` for `:notice`,
 already falsified. Both that cell and `:topic`'s now read `text or ""`, with the reason beneath
 the table — leaving one right and one wrong in the same table is how the next reader picks the
 wrong one.
+<!-- entry #1580 -->
+
+---
+
+## 2026-08-20 — #1580: one subject-XOR predicate, and why counting it by name would have lied
+
+`defp validate_subject_xor/1` lived twelve times under `lib/`, once in every
+schema carrying the `user_id` XOR `visitor_id` FK pair. It is now
+`Grappa.Subject.validate_xor/1`, called twelve times.
+
+### Counted by BODY, because a name count is not a duplication count
+
+The issue's "twelve" came from grepping the function NAME. That measures
+declarations, not duplication: a copy that was inlined, renamed, or pasted
+into an anonymous position is invisible to it — measured elsewhere in this
+repo, nineteen named definitions were retired while the same body survived
+INLINE fourteen times across ten files.
+
+So the census re-anchored on the decision body, `case {user_id, visitor_id}
+do … end`, and treated the enclosing function name as a DERIVED column. The
+number held at twelve, and the axes closed on each other: twelve
+`belongs_to :visitor` schemas, twelve decision bodies, twelve
+`<table>_subject_xor` CHECK constraints in migrations, zero schemas with a
+visitor FK and no body, zero `add_error(changeset, :subject, …)` anywhere
+outside those twelve blocks. The wider net — every file naming both columns —
+left ten files, and reading them found no thirteenth XOR in a different
+shape. The nearest miss, `Scrollback.persist_subject/2`, is a telemetry
+labeller that prefers `:user` when both are set: correct for a label, and not
+a validator.
+
+The census script carried its known-answer controls INSIDE it and printed no
+numbers unless all six passed — corpus size, a hand-counted calibration file,
+a real file that must yield zero, a decoy `defp` taking a changeset that must
+NOT be classified as one, a check that the decoy is actually present so the
+control is not vacuous, and an arithmetic cross-check that every subject
+`add_error` in `lib/` falls inside a counted block.
+
+### The mask is what licensed the flattening
+
+Hashing the twelve bodies with the string literals MASKED collapsed them to
+ONE control-flow class. That measurement, not a reading of the code, is what
+makes this a mechanical change: no copy carried behaviour the consolidation
+silently drops. The entire divergence in the family was one message string.
+`uploads/upload.ex` said `"one of user_id or visitor_id is required"`; the
+other eleven said `"must set user_id or visitor_id"`. Nothing pinned the
+variant, and its only production caller cannot reach the arm that emits it —
+`Uploads.create/3` routes the subject through `Subject.put_subject_id/2`,
+whose two clauses both guard `is_binary`, so a subject-less insert raises
+before the changeset runs. A typo the copies let live, not a requirement.
+
+### The comments could not have caught it, by construction
+
+Each copy carried a "Mirror of X" pointer. `accounts/session.ex` →
+`scrollback/message.ex` → `accounts/session.ex` is a cycle, so the graph has
+no root and no copy was the designated original. Every mirror claim was
+therefore unfalsifiable, which is precisely why the drifted copy could keep
+asserting a mirror it no longer was. Copies are not kept honest by comments.
+
+The same class of error appears in the issue's own pointer table, which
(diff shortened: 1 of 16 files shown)


vjt:

This PR's work is **on `main` by content, not by SHA**: #1616 cherry-picked its commits onto the then-current main, so the SHAs were rewritten and GitHub cannot mark this PR merged by itself. Verified landed: `git log origin/main --grep '#1580'` shows the commits.

Issue #1505, opened 2026-08-17
An empty TOPIC (topic cleared) is dropped from scrollback — and persisting it would render as a truncated line


vjt:
Split out of #1500, where the same root defect was fixed for `:notice` only. Deliberately **not**
extended there — the server half is free, the client half is not, and folding it in would have widened
a scoped slice to rescue a decision already taken.

## The defect

`TOPIC #chan :` (empty trailing) is legal and means **the topic was cleared**. Today that row is dropped
by the same mechanism as #1500: `:topic` is in `@body_required_kinds`, so `validate_required` rejects it
and the row is lost — on every ircd.

## Why the #1500 cure cannot simply be extended

Server side it *is* free: the same atom out of the same list, already covered by the same scoped
`empty_values: []` re-cast, zero extra lines. **The blocker is entirely the renderer.**

Established by reading, three linked facts:

1. `cicchetto/src/components/ScrollbackPane.tsx:977`, `case "topic"`, renders literally
   `* {sender} changed topic: <MircBody body={msg.body ?? ""} emphasis />`
2. `parseMircFormat("")` returns zero runs — `flush/0` returns early on `buf.length === 0`
   (`cicchetto/src/lib/mircFormat.ts:186-187`) — so the `<For>` renders nothing.
3. The pane does not filter empty bodies (no `!msg.body` / `body === ""` guard before the render), so the
   row really does reach that branch instead of disappearing earlier.

Result on screen: `* alice changed topic:` — label, colon, and nothing. It does not crash, does not print
`undefined`, does not leave a `??` uncovered. But it **asserts "changed" and then shows nothing**, which
reads as a truncated line rather than as *"the topic was cleared"*.

**Persisting a row that mis-describes the event is not an improvement over dropping it — it is the same
damage moved from the log into the scrollback, where it also stays.**

## What the fix wants

A renderer branch that says the truth for an empty topic body (wording is a product call), landed
together with the server-side relaxation — not one without the other.

## Not established

The topic **bar** and `channelTopic.ts` were not checked; the measurement above covers the scrollback row
only.

No `status:*` label — not triaged into the queue.

vjt:
Closed by the pipeline change ruled by vjt on 2026-08-20: **`status:soon` is retired**. An issue is now closed when its work **merges**, and the **milestone** records which release carries it — the label duplicated what the milestone already said.

[...]

PR #1614, opened 2026-08-20
#1518 — cic: buy the service-modal arm's ordering invariant

vjt:
#290's `service-modal` arm is two statements, and its comment declares the order
between them: `openServiceModal` takes the mirror high-water mark FIRST, `help`
goes out SECOND, so the reply notices count as while-open arrivals. #1518
measured that nothing enforced it. This slice establishes that the order is
real and buys it with a test that bites.

## The reproduction, before touching anything

On `origin/main` = 425426f2, the whole cic vitest suite is **294 files / 5688
cases**:

| run | arm | result |
|---|---|---|
| R0 | as written | 294/294 files, 5688/5688 tests, `rc=0` |
| R1 | two statements swapped | 294/294 files, 5688/5688 tests, `rc=0` |

Identical totals. `compose.test.ts` covers this arm with an `it.each` asserting
that `openServiceModal` was called and that `sendMessage` was called; neither
assertion can see which ran first.

## Why the order is load-bearing

The gap it guards is the arm's own `await`. `sendBodyLines` suspends on the
POST, and the WS handler ingests a NOTICE by calling `appendToScrollback` — the
same store `openServiceModal` reduces to a high-water mark — on that same event
loop. A service notice landing while the POST is in flight is exactly the
arrival the modal exists to show, and it is above the mark only if the mark was
taken before the await. Capture it after and the notice sits at or below
`sinceId`, the one predicate `ServiceModal.tsx` filters on: the help wall is
filtered out of the view that was opened for it.

It is a race window rather than a certainty, which is why it was never
observed — the ircd's reply usually comes back after the 202 does.

## What buys it now

`serviceModalCommand.test.ts` drives the real handler with only `lib/api`'s
`sendMessage` stubbed, so the await under test is production's own (real
`sendBodyLines` → real `scrollback.sendMessage` → the POST), and delivers a
notice through the real ingest verb from inside the in-flight POST. `null` is
what a `*Serv` target's 202 returns (#1430).

Two mutants, one per assertion, each killing exactly one — `1 failed | 0 other
failures` both times:

| mutant | red line | message |
|---|---|---|
| swap the arm's two statements | the `sinceId` assertion | `expected 11 to be 10` |
| drop the mid-send delivery from the fixture | the control | `expected [ 10 ] to deeply equal [ 10, 11 ]` |

The control is not decoration: without it the invariant assertion holds
vacuously for a send that never fired — `sinceId` would read 10 because nothing
ever arrived, and the test would be green while measuring nothing.

## Gates

Post-rebase onto `ba32aa63`: `bun run test` 295 files / 5695 tests `rc=0`;
`bun run check` 4 stages, 0 failed; `design-notes-gate.sh` `rc=0`;
`union-rebase.sh` `rc=0` with `docs/DESIGN_NOTES.md +62 -0` on both sides, the
preceding entry (#1505, 109 lines) `cmp`-identical against the new base, and
the marker `<!-- entry #1518 -->` absent from `origin/main` and `grep -Fxc` = 1
here.

## What this does not buy

The same order has a second consequence — the modal appears before the send
rather than after it, so a paced (#666 429-retry) or failing send would leave
the operator with no modal at all under the swap. The #290 comment does not
claim it and nothing here asserts it; it is named in the DESIGN_NOTES entry
rather than left implicit.

No e2e ran for this slice. `issue290-services-modal.spec.ts` was read, not
executed: it waits for modal lines with no bound on when they were captured, so
it stays green under either order and would not have bought this either.

### DIFF
--- a/cicchetto/src/__tests__/serviceModalCommand.test.ts
+++ b/cicchetto/src/__tests__/serviceModalCommand.test.ts
@@ -1,0 +1,101 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import type { ScrollbackMessage } from "../lib/api";
+import { setToken } from "../lib/auth";
+import { channelKey } from "../lib/channelKey";
+import type { CommandContext } from "../lib/commands/context";
+import { serviceModalCommand } from "../lib/commands/services";
+import { appendToScrollback } from "../lib/scrollback";
+import { closeServiceModal, serviceMirrorRows, serviceModalState } from "../lib/serviceModal";
+import { SERVER_WINDOW_NAME } from "../lib/windowKinds";
+
+// #1518 — the `service-modal` arm's statement ORDER, which #290 stated in a
+// comment and nothing bought: swapping the two lines left the whole vitest
+// suite green (measured on 425426f2), because `compose.test.ts` asserts WHICH
+// calls the arm makes and never in what sequence.
+//
+// The order is load-bearing, and the gap is production's own: `sendBodyLines`
+// AWAITS the POST, and the WS handler (`subscribe.ts`, which ingests a NOTICE
+// by calling the same `appendToScrollback` this test calls) runs on that event
+// loop. A service notice that lands while the POST is in flight is a
+// WHILE-OPEN arrival — `openServiceModal` must already have taken the mirror
+// high-water mark, so the notice sits ABOVE it and `ServiceModal.tsx` renders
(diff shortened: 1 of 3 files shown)


vjt:

This PR's work is **on `main` by content, not by SHA**: #1616 cherry-picked its commits onto the then-current main, so the SHAs were rewritten and GitHub cannot mark this PR merged by itself. Verified landed: `git log origin/main --grep '#1518'` shows the commits.

PR #1615, opened 2026-08-20
#1513 — cic: switch noImportCycles on, and cut the one cycle it finds

vjt:
The half of #1513 the issue itself calls "cheaper, structural": measure what
`noImportCycles` costs, cut what it finds, and make a cycle a build error. The
store half is untouched — `privmsg`/`me`/`msg` stay inline for the #737 lock,
exactly as ruled.

## The measurement, before anything moved

`suspicious/noImportCycles` (biome 2.5.8 — not nursery, off by default), run
through the whole `check` chain, which already passes `--max-diagnostics=none`:

| config | files | errors |
|---|---|---|
| rule ON, defaults | 1064 | **2** |
| rule ON, `ignoreTypes: false` | 1064 | **11** |

The 2 are the two EDGES of ONE runtime cycle, `selection.ts:29` ↔
`windowState.ts:4`. It is the same cycle the 2026-08-18 entry measured at 597
files; the counts differ only in scope (that run was `biome check src` — scoped
to `src` the tree is 615 files today, against 1064 for `src` + `e2e` + root).
The other nine are three type-only rings (`api` ↔ `wireNarrow`, `api` →
`channelTopic` → `identityScopedStore` → `auth` → `api`, `passkeys` inside that
same ring). `ignoreTypes` defaults TRUE because the compiler erases a type-only
import, so it cuts no runtime edge.

**A 2 out of 1064 could be blindness, so it was made to earn belief.** Closing
the exact cycle #1513 warns about (a runtime `sendPipeline.ts → compose.ts`
import) raises **7 errors** naming `sendPipeline.ts`, `compose.ts` (×3),
`commands/services.ts`, `commands/fanout.ts`. Probe removed → back to baseline,
byte-identical. Re-run against the SHIPPED config after the cut: 0 → 7 → 0.

## Half of what the issue describes was already built

Measured on the code, not read off the issue: `sendPipeline.ts` already carries
`draftLines` + `wireBody` WITH the cluster and `compose.ts:82` imports all three
back, so the block-1 cycle cannot exist — and with the rule on, zero diagnostics
name `compose.ts`, `sendPipeline.ts` or `commands/*`. Of the six arms, three are
already behind the seam (`service-modal` #1517, `ame`/`amsg` `d72e085e`); the
three that remain are the ruled ones.

## The cut

The whole `windowState → selection` edge was ONE function,
`isActiveChannelJoined()`. It now takes the selection as a parameter; the
reverse edge stays, so the dependency runs one way and `windowState` is a
function of its own state again.

Rejected: moving it into `selection.ts` would have re-homed it out from under
four `vi.mock("../lib/windowState")` factories that supply it — mocks that would
then stop mocking, silently. Also rejected (orchestrator ruling): a
`biome-ignore` pair, two lines that leave the cycle alive permanently with an
issue to remember it by.

**Cost against the grep estimate: 11 call sites predicted (3 `Shell.tsx`, 8
`windowState.test.ts`), 11 edited.** No mock factory needed a change — both are
the untyped `vi.mock(path, factory)` form, so nothing type-checks against the
signature and a zero-arg stub ignores the extra argument.

## Two corrections at the source

In the 2026-08-18 #1513 entry, not as postilles elsewhere:

* `ame`/`amsg` are listed among the STORE-coupled arms. They left the switch
  eight hours after that entry landed (`d72e085e` 15:38 vs `5779c755` 07:20),
  and `commands/fanout.ts` states the distinction itself. Struck, not deleted.
* Its cycle-probe clause is superseded, not false: every word stayed true (the
  rule really is absent from biome's recommended set; the probe really did pass
  vitest, `tsc` and a `vite build` — none of which read import graphs). What
  changed is the gate. Said explicitly that this does **not** reopen #737: the
  objection to publishing `writeState` was the lock, never the cycle.

## Gates

`bun run test` 294 files / 5694 tests `rc=0`; `bun run check` 4 stages, 0 failed
(`biome` now includes the rule); `design-notes-gate.sh` `rc=0`. Branched from
`origin/main` = `ba32aa63` and still a descendant of it, so no rebase ran.

## What this does not buy

`windowState.ts` still carries an `import type { SelectedChannel }`, and the
`ignoreTypes:false` reading of the tree is still 11 — the runtime edge is dead,
the type edge is not, and the strict number did not move. That is the rule's
documented default, stated here rather than left for someone to rediscover.

### DIFF
--- a/cicchetto/biome.json
+++ b/cicchetto/biome.json
@@ -17,7 +17,10 @@
   "linter": {
     "enabled": true,
     "rules": {
-      "recommended": true
+      "recommended": true,
+      "suspicious": {
+        "noImportCycles": "error"
+      }
     }
   },
   "javascript": {
--- a/cicchetto/src/Shell.tsx
+++ b/cicchetto/src/Shell.tsx
@@ -449,7 +449,7 @@ const Shell: Component = () => {
   // `.shell-members` drawer on mobile) a PERMANENT surface reachable from
   // EVERY window kind — the cog lives in it, so it MUST be openable on
   // non-channel windows (home/server/list/mentions/admin). Force-closing it
-  // whenever `isActiveChannelJoined()` is false directly contradicts that. On
+  // whenever `isActiveChannelJoined(...)` is false directly contradicts that. On
   // desktop the effect was already a visual no-op (the members pane is a grid
   // column; `membersOpen`/`.open` have no desktop CSS and the TopicBar
   // hamburger is CSS-hidden there, so membersOpen never toggled). The drawer
@@ -591,7 +591,10 @@ const Shell: Component = () => {
       when={isMobile()}
       fallback={
         // ── Desktop three-pane layout (unchanged from pre-C6) ─────────
-        <div class="shell" classList={{ "shell-no-members": !isActiveChannelJoined() }}>
+        <div
+          class="shell"
+          classList={{ "shell-no-members": !isActiveChannelJoined(selectedChannel()) }}
+        >
           <ErrorBanners />
           <Toasts />
           <PrivacyModal />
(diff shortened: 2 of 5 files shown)


vjt:

This PR's work is **on `main` by content, not by SHA**: #1616 cherry-picked its commits onto the then-current main, so the SHAs were rewritten and GitHub cannot mark this PR merged by itself. Verified landed: `git log origin/main --grep '#1513'` shows the commits.

Issue #1579, opened 2026-08-19
e2e: the 5s waitForUserTopicReady barrier inside canonical loginAs times out under full-suite load (3 occurrences, 3 different specs)


vjt:
The canonical `loginAs` in `cicchetto/e2e/fixtures/cicchettoPage.ts` ends with
`waitForUserTopicReady`, a `page.waitForFunction` on `window.__cic_userTopicReady`
with a **5 s** timeout. That barrier is the last statement of the login door
**333 spec files** go through, and it is the single point where a full-suite run
has now gone red three times, in three different specs.

## What was measured

Full `scripts/integration.sh` on 2026-08-19 (single worker, 28.9 min,
759 tests): **757 passed, 2 failed**. Both failures are the same stack:

```
TimeoutError: page.waitForFunction: Timeout 5000ms exceeded.
  at waitForUserTopicReady (fixtures/cicchettoPage.ts)
  at loginAs (fixtures/cicchettoPage.ts)
```

| spec | position in run | wall time |
| --- | --- | --- |
| `issue1041-left-edge-sidebar.spec.ts` — *left-edge right-swipe mounts the sidebar* | 84 / 759 | 32.5 s |
| `issue173-compose-recall-caret-visible.spec.ts` — *fast up-flick recall scrolls the end-caret into view* | 181 / 759 | 32.3 s |

Iso-rerun of exactly those two, same tree, nothing else changed:
**both green, 1.5 s and 1.2 s.** So the failure is not deterministic and not a
property of either spec's own gesture.

A previous full run on base `17cd0f77` produced exactly one red,
`media-link-modal-viewer.spec.ts`, and it died in the same
`loginAs` -> `waitForUserTopicReady` barrier — before any of tonight's work
existed.

## It predates the branch that surfaced it

The run above was on the head of the branch behind #1578. That branch's diff is
four files:

```
cicchetto/e2e/tests/issue252-vhost-selector.spec.ts
docs/DESIGN_NOTES.md
test/grappa_web/controllers/auth_controller_test.exs
test/grappa_web/controllers/share_token_controller_test.exs
```

Neither failing spec is in it, and neither is `cicchettoPage.ts` — both specs
already called the canonical `loginAs` before the branch existed. The green
iso-rerun proves only that the failure is non-deterministic; **the absence of
those files from the diff is what makes it not the branch's**.

## What is NOT established — three readings, none discriminated

1. **The 5 s budget is simply too tight** under a 29-minute single-worker run on
[...]

PR #1617, opened 2026-08-20
test(#1579): discriminate the user-topic barrier reds — the barrier was right

vjt:
Discriminates the three readings #1579 puts on the table and says none of them
had been separated. **Reading 2 wins: the barrier is telling the truth.** The
user-topic subscribe genuinely had not happened, because the server had not
finished the WebSocket upgrade — the two reds in the issue are product signal,
not harness noise.

The budget is deliberately untouched. No timeout was raised, no assertion
weakened, no retry added.

## How it was discriminated

The three readings differ in exactly one observable — what happens AFTER the
5 s deadline — so the harness kept the ENFORCED budget at 5 s and displaced
only the OBSERVATION past it, then recorded the satisfaction latency of every
`waitForUserTopicReady` call in a full run so reading 3 could be judged on the
population instead of on two data points.

One full `scripts/integration.sh` (759 tests, 32.8 min, single worker,
darwin / Docker Desktop, `MIX_ENV=dev`), 600 barrier calls:

* **597 satisfied** — median 18 ms, p95 243 ms, p99 395 ms, **max 848 ms**. The
  5 s budget is ~6× the worst healthy case, so "the budget is simply too tight"
  is refused by its own population. Per-tenth medians are flat (16…36 ms), so
  the monotonic-degradation shape reading 3 needs is not there either.
* **3 missed**, all with one signature: `__cic_socketHealth` reading
  `state: "connecting"`, `connectAttempts: 1`, `errorCount: 0` — the WS
  transport had **never opened**, and phoenix.js's own 10 s join timeout had
  fired for all three topics because the joins were still buffered.

The server agrees, on the pid Phoenix names in its own duration line:
`CONNECTED TO GrappaWeb.UserSocket in 11140ms`, followed by
`JOINED grappa:user:… in 23µs`. Client-side stamp arrived at 11 180 ms. The
same wait from both ends, and none of it is the JOIN. The 31 s case logs its
own mechanism: `db write unavailable: SQLite write lock held by another writer
for 31572ms` inside `record_client_source`, then
`CONNECTED TO GrappaWeb.UserSocket in 31575ms`. Across the run, 682 connects:
680 at ≤ 56 ms and exactly those two above 1 s.

Raising the number would have been the wrong repair twice over: it cannot
reach 11.2 s / 31.6 s / ~65 s, and in the same run the same stall took out
`loginAs`'s OTHER gate (the 10 s shell-ready wait in `issue591-ctcp-ping`), so
a bigger barrier budget just hands the failure to whatever waits next.

## What changed

`waitForUserTopicReady` keeps its 5 s and gains a voice. On timeout it reports
the stamped names, the socket-health state, the joined topic keys,
onLine/visibility, and the tail of the page console — captured through
Playwright's own listener, never by patching the page's `console`. It recovers
nothing; the same timeout fails the same test at the same moment. This is the
measurement #1579 itself asked for ("the barrier currently throws without
reporting what it saw"), made permanent.

Proven by mutation on the committed code, twice — the second run exists only
because the first printed `page console (last 0 of 0)`, which is true but
indistinguishable from a capture that never worked:

```
Error: waitForUserTopicReady: 's92d28457' was not stamped within 5000ms
  stamped names:     ["s92d28457"]
  socket health:     {"state":"open",…,"connectAttempts":1,…}
  joined topic keys: ["bahamut-test s92d28457","bahamut-test $server",…]
  navigator.onLine:  true   visibility: visible
  page console (last 2 of 2):
    error: [w1-probe] a deliberate console.error, to calibrate the capture
    warning: [w1-probe] and a warn
```

## The elsewhere

The stall itself is filed separately: `UserSocket.connect/3` waits a full
`busy_timeout` for a write that #523 already declared droppable, so the
"can never hurt a connect" contract is not actually held — in production, not
only under e2e load.

## What is NOT established

Why a writer holds the SQLite write lock for 31–33 s. The holder's own
statement is what takes that long, and one of three `LockWatch` stack samples
shows it blocked inside `logger_h_common:log/2` rather than inside SQLite, with
the BEAM emitting nothing at all during each episode. Log back-pressure and
storage contention are both live and this run separates neither; the substrate
(darwin/Docker Desktop, `MIX_ENV=dev`, 1 114 576 log lines) is also unshared
with the Linux CI runner. Full list of refusals in the report on the issue.

## Gates

* `scripts/bun.sh run check` — rc=0, 4 stages ran, 0 failed (the only static
  gate over `cicchetto/e2e`, per `docs/TESTING.md`).
* `scripts/design-notes-gate.sh` — rc=0, 1 new entry heading, separator and
  marker present.
* `scripts/union-rebase.sh origin/main` — rc=0,
  `docs/DESIGN_NOTES.md +114 -0 — contribution intact`; boundary shape re-read
  on the real file after the rebase.
* `scripts/integration.sh --project=chromium --grep issue1041` on the shipped
  tree — **6 passed (17.9 s)**, the happy path through the modified login door.
* **Not run, declared:** `scripts/check.sh` (runs no cic gates; the diff is
  `cicchetto/e2e/` + `docs/` only) and the full 759-test suite on the shipped
  tree — the full run I have predates these commits and IS the measurement;
  4 of its 5 reds are the stall class above and pre-date this branch.

### DIFF
--- a/cicchetto/e2e/fixtures/cicchettoPage.ts
+++ b/cicchetto/e2e/fixtures/cicchettoPage.ts
@@ -70,8 +70,11 @@ import {
 import type { SeededUser } from "./grappaApi";
 import { type ScrollGestureResult, scrollByGesture, waitForScrollRest } from "./scrollGesture";
 import type { TraceEvent } from "./scrollTrace";
+import { describeUserTopicTimeout, watchPageConsole } from "./userTopicDiagnostics";
 
 const SHELL_READY_TIMEOUT_MS = 10_000;
+// #1579 — see `waitForUserTopicReady` for why this number does not move.
+const USER_TOPIC_BARRIER_MS = 5_000;
 const MOBILE_BREAKPOINT_PX = 768;
 
 // Mirror of cicchetto/src/lib/theme.ts isMobile() — viewport width
@@ -111,6 +114,11 @@ export async function expectShellReady(page: Page): Promise<void> {
 // and `getSubject()` reads the subject; doing this via `page.evaluate`
 // AFTER goto would race the SPA's first read.
 async function seedAuthLocalStorage(page: Page, token: string, subjectJson: string): Promise<void> {
+  // #1579 — start recording the page console before the first navigation, so
+  // a later `waitForUserTopicReady` timeout can quote the phoenix-level join
+  // errors instead of leaving them in a container log nobody reads. Costs one
+  // listener; the page is not modified.
+  watchPageConsole(page);
   await page.addInitScript(
     ([t, s]) => {
       localStorage.setItem("grappa-token", t);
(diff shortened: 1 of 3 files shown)


vjt:

The commits of this PR are on `main` under new SHAs (the union was built by cherry-pick, so GitHub cannot mark this PR merged by itself). Verified by CONTENT, not by ancestry: `git log origin/main --grep '#1579'` finds the work, and the DESIGN_NOTES marker `<!-- entry #1579 -->` is present exactly once on the tip (e.g. `a1fead53`).

Issue #1618, opened 2026-08-20
WS upgrade waits a full busy_timeout for a write #523 already declared droppable (31.6 s measured)


vjt:
`UserSocket.connect/3` performs a synchronous, explicitly-droppable
`user_settings` write transaction before it returns, and Phoenix holds the
WebSocket upgrade until it does. When another writer holds the SQLite write
lock, the upgrade waits the whole `busy_timeout` — measured at **31 575 ms** —
and then connects anyway, having dropped the write it waited for.

Found while discriminating the three readings in #1579; it is that issue's
"elsewhere", not part of it.

## The path

```
UserSocket.connect/3
  └─ maybe_record_client_source/2            (lib/grappa_web/channels/user_socket.ex)
     └─ Grappa.Vhosts.record_client_source/2 (lib/grappa/vhosts.ex)
        └─ UserSettings.put_last_client_prefix64/2
           └─ update_data/2 → write_transaction/1
              └─ Repo.BusyRetry.run(Repo.immediate_transaction(…))
```

`with … do maybe_record_client_source(socket, connect_info); {:ok, socket} end`
— the call sits between authentication and the return, so the transport is
blocked on it.

## Measured

One full `scripts/integration.sh` on `origin/main` = `425426f2` (759 tests,
32.8 min, darwin / Docker Desktop, `MIX_ENV=dev`). 682 WebSocket connects:

| `CONNECTED TO GrappaWeb.UserSocket in …` | count |
| --- | --- |
| ≤ 56 ms | 680 |
| 11 140 ms | 1 |
| 31 575 ms | 1 |

The 31.6 s one names its own mechanism in the container log, on the transport
pid Phoenix reports the duration for:

```
16:28:45.423 pid=<0.5701.0> ws connect authenticated
16:29:16.996 pid=<0.5701.0> QUERY ERROR db=31571.8ms
16:29:16.999 pid=<0.5701.0> db write unavailable: SQLite write lock held by another
             writer for 31572ms across 1 attempts (1500ms retry budget)
16:29:16.999 pid=<0.5701.0> record_client_source: db unavailable persisting client
             prefix for {:user, "270bdc2c-…"} — dropped (best-effort)
16:29:16.999 pid=<0.5701.0> CONNECTED TO GrappaWeb.UserSocket in 31575ms
```

The 11.1 s one shows the same shape succeeding rather than degrading — one
[...]

Issue #1519, opened 2026-08-18
ci(integration): the e2e suite is on a +0.42 min/day trend and 87% of it is one serial Playwright run


vjt:
Filed on vjt's instruction after he saw the integration suite hit 40 minutes against a 45-minute cap.

Measured 2026-08-18 over 85 successful `integration` runs on `main` (2026-08-11 → 2026-08-18), plus a full per-line breakdown of run [32105767197](https://github.com/vjt/grappa-irc/actions/runs/32105767197) (`main` @ #1513, 39.6 min, green).

## The trend is real

Daily medians of the `e2e` job, in minutes:

| day | n | median | min | max |
|---|---|---|---|---|
| 08-11 | 11 | 36.0 | 35.4 | 37.0 |
| 08-12 | 5 | 37.4 | 36.7 | 37.6 |
| 08-13 | 12 | 38.1 | 37.1 | 39.3 |
| 08-14 | 10 | 38.5 | 37.6 | 39.5 |
| 08-15 | 6 | 38.8 | 32.7 | 39.2 |
| 08-16 | 16 | 39.0 | 36.6 | 39.7 |
| 08-17 | 19 | 39.5 | 35.3 | 40.2 |
| 08-18 | 6 | 39.6 | 39.0 | 40.1 |

Monotonically increasing across all eight days. Least-squares slope over the 85 runs: **+0.42 min/day**.

The cap has been raised twice already (25 → 30 in `6478ef72`, 30 → 45 in the current header comment). It is being raised a third time, to 60, in the PR linked below — as a runaway-job backstop, deliberately far above the working range, so that it stops being the thing that tracks the suite's growth. **That raise buys headroom, it does not fix anything.** This issue is the fix.

## Where the 39.6 minutes go

Job scaffolding — `Set up job`, `Checkout`, compose-plugin upgrade, GHCR login — totals **~0.3 min**. Everything else is the single `Run integration suite` step, so "which step regressed" has no answer; the breakdown has to come from inside `scripts/integration.sh`:

| phase | time | what |
|---|---|---|
| Elixir dep compile | **~2.2 min** | `nimble_parsec`, `makeup_*`, `credo`, `ecto`, … then `Generated grappa app` |
| stack boot | ~1.7 min | solanum/testnet, grappa, nginx healthy |
| Playwright image + cic build | ~0.9 min | image extract, `cicchetto-build-test` |
| **Playwright suite** | **34.3 min** | `754 passed (34.3m)` |

[...]

Issue #1408, opened 2026-08-16
review: three runbook claims name gates that do not cover what they say (D-S8, D-S9, D-S10, bucket G hand-lists)


vjt:
Bucket **G hand-lists** of the 2026-08-15 codebase review (`docs/reviews/codebase/2026-08-15-codebase-review.md`) — the three gating MEDIUMs left over. D-S1 and D-S5 are in #1343, D-S4 in #1377; these three had no issue. All three are the same defect class: **a document claims a gate that the gate does not actually cover.**

## D-S8 (MEDIUM) — `integration.yml`'s path filter misses the files that define what the suite boots

**Files:** `.github/workflows/integration.yml:52-67`, `:70-85`

The filter's own rule (`:23`) puts the root `Dockerfile` and `compose.yaml` in scope because "they define the images the suite boots". By that rule three more belong: `bin/start.sh` (the image `CMD`, `Dockerfile:53`; already docker-COLD in `preflight.ex:442`), `.dockerignore` (the build context; `preflight.ex:441`), and `VERSION` / `infra/packaging/version.sh` (`testnet.sh:32` and `integration.sh:39` both call it and export the result; `compose.yaml:144` notes "Empty makes vite fail loud"). #715 fixed the JS half of this filter; the image half was not swept.

**Fix:** add them to **both** `paths:` blocks. Longer term the duplicated 15-line block should be a YAML anchor — push and PR lists are transcribed twice and can drift from each other.

## D-S9 (MEDIUM) — `OPERATIONS.md` claims base images are digest-pinned; the gate it names covers two families

**Files:** `docs/OPERATIONS.md:2354-2362`, `Dockerfile:10`, `test/infra/base_image_digest_pin_test.bats:29`

The runbook says the bats suite "fails the build if any real image reference in a tracked build file loses its `@sha256:`". The test's grep is `-E 'oven/bun:|nginx:alpine'`. The primary base — `Dockerfile:10`, `FROM elixir:1.19-otp-28-alpine` — is unpinned and unguarded, as is `alpine:3.24` in the release runtime stage. For `alpine:3.24` that is deliberate and argued (`OPERATIONS.md:2403`: a FLOOR, proven by `assert-abi-lockstep.sh`); for the Elixir base it is argued nowhere, and it is what makes the Elixir patch level float.

[...]

PR #1621, opened 2026-08-20
ci(#1519): split the integration suite across four Playwright shards

vjt:
`integration` was one job, one stack, one Playwright worker, 40.5 minutes,
with a daily median climbing +0.42 min/day. vjt's ruling on #1519 authorised
sharding. This is the shard, and nothing else: **`workers: 1` is untouched**,
because the three long-lived users still share `#bofh` and #1618 has just
measured a 31572 ms SQLite write lock with the suite running *serially*. The
shared-state audit is the step after this one.

## Where the 40 minutes go — measured, not estimated

Step timings read off the green run `32395956800` (2026-08-20):

| phase | min |
|---|---|
| checkout with submodules + compose plugin + GHCR login | 1.7 |
| `testnet up` — deps.get, Elixir compile, solanum from source, images, health | 4.6 |
| the Playwright suite — 759 passed, one worker | 33.6 |
| census + tear-down | 0.5 |
| **total** | **40.5** |

Only the 33.6 divides. The other **6.8 min is fixed and every shard pays it
in full**, which puts four shards at **~15 min** — not the ~13.4 the issue's
table gave, which counted the in-script boot alone and forgot the checkout
ahead of it and the tear-down behind it. Eight shards reach ~11: four minutes
for double the runners, with the fixed cost dominating. Hence four.

**Paying the boot once was considered and rejected on the numbers.** A
build-once job has to FINISH before any shard starts, so it adds a serial
stage the same size as the one it removes (~5 min); `5 + pull + 8.4` is no
better than four parallel 15s. Layer caching (`type=gha`) might shrink the
per-shard build — **not measured, so not claimed**.

## The green has to stay TRUE: the partition is proven

Per-shard test counts from `playwright test --list` on `bb65f73b`:

| shard | tests |
|---|---|
| 1/4 | 190 |
| 2/4 | 190 |
| 3/4 | 191 |
| 4/4 | 188 |
| **sum** | **759** — the unsharded total, exactly |

Cardinality is the weak form and it is not what was relied on. The four shard
listings were compared to the full listing as **sets**: in the total and in no
shard → 0; in a shard and not in the total → 0; in two shards → 0. The
comparator was then fed a deliberately altered input and reported the
difference, so those zeros are measurements and not a silent parser.

Granularity is the **(project, file) group** — 447 of them, none spanning two
shards — so ordering *within* a file is preserved exactly as
`fullyParallel: false` implies. Ordering *across* files is not, and that
genuinely changes cascade behaviour; `docs/TESTING.md` now says so next to the
triage tree, with `--shard=N/4` as the way to reproduce CI's grouping.

## The three ways a sharded gate lies quietly

Each is a bats case in `test/infra/integration_shard_test.bats`, deriving its
expectation from the workflow rather than transcribing the shard count.
Verified by mutation: **six mutants, six kills, exactly one assertion each**.

1. **A drifting denominator.** `--shard=i/N` with N adrift from the job count
   drops the tail of the suite or double-runs it, and Playwright is content
   either way — it filters, it does not audit. The denominator is
   `${{ strategy.job-total }}`, the matrix's own length. That is also why the
   matrix must stay one-dimensional: `job-total` counts *legs*.
2. **Colliding artifact names.** `upload-artifact` v4+ refuses a duplicate
   name inside a run, so four shards uploading `playwright-traces` would fail
   three uploads — a second red on top of the one being triaged, evidence
   missing. Every name carries `-shard-${{ matrix.shard }}`.
3. **A skipped aggregate.** `e2e-result` is the one stable check name
   (`needs.e2e.result` is `success` only if every leg was). It carries
   `if: always()` deliberately: without it, a failed shard leaves it
   **skipped**, and a skipped check does not read as a failure.

`fail-fast: false` for the opposite of the usual reason — the default cancels
the surviving shards, replacing three results with three unknowns exactly when
the comparison between them *is* the triage.

## How the red aggregates — measured on real runners, never quoted from the docs

`main` carries no branch protection and no ruleset (measured: the protection
API answers 404, `/rulesets` is `[]`), so no check is required by name today.
Rather than assert the Actions semantics from documentation, a throwaway
branch `w2-1519-probe` ran a matrix with one deliberately failing leg and an
`if: always()` aggregate whose assertion is *inverted* — it passes when the
aggregate reports failure. It cost ~30 s of runner time and it turned up the
thing the documentation would not have put in anyone's face: **without
`if: always()` the aggregate is `skipped`, and `skipped` does not read as
red.** Results are in the comment below; the branch is deleted, the run
outlives it.

## What this PR does NOT claim

- ~~The runner concurrency ceiling was never measured~~ — **it is now, and it
  holds at four.** On this PR's own run
  ([32403458733](https://github.com/vjt/grappa-irc/actions/runs/32403458733))
  the four shards started at **18:29:21Z, 18:29:22Z, 18:29:22Z, 18:29:22Z** —
  within one second of each other, all four `in_progress` together. What is
  still NOT measured is the ceiling under contention: whether EIGHT jobs would
  start together, i.e. this suite sharded four ways while another PR's four
  are already running.
- The per-shard **counts** are equal to within three tests; the per-shard
  **durations** are not measured. Wall clock is the slowest shard.
- Whether any cross-file cascade changes verdict under the new grouping is
  unknown until the first reds arrive. The mechanism is argued; the incidence
  is not measured.

## Gates

- `scripts/bats.sh` full set: **596 ok, 0 failed** (rc captured to a file),
  including the six new cases and the #1408 path-filter suite that parses this
  workflow.
- `scripts/design-notes-gate.sh`: rc=0, one new entry, marker unique against
  the current tip of `origin/main`.
- Workflow YAML parsed independently; push and `pull_request` path filters
  still byte-identical.
- No Elixir and no cic source touched, so `check.sh` / `bun run check` were
  not re-run — stated rather than implied.
- The real verification is **this PR's own `integration` run**, which is the
  first four-shard run in existence.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -44,8 +44,15 @@ name: integration
 #
 # Stack timing: cold ~6-8 min for the testnet+grappa+nginx images +
 # Playwright base pull; warm with cache it's faster but we don't cache
-# docker layers cross-runs (would need a registry push). Suite itself
-# runs in ~16s once everything is up.
+# docker layers cross-runs (would need a registry push). The "suite runs
+# in ~16s" that stood here was true of the first dozen specs and has not
+# been for a long time: measured 2026-08-20 it is 33.6 min of the 40.5,
+# which is what #1519 shards. The per-shard numbers are on the job below.
+#
+# The suite is split across FOUR jobs (#1519), one Playwright `--shard`
+# each, each with its OWN stack — sharding is N separate jobs, and it does
+# NOT touch `workers: 1`, which stays load-bearing inside every shard
+# (three long-lived users still share `#bofh`, see the issue).
 #
 # Failure artifacts: Playwright writes traces (zip), screenshots, and
 # videos to cicchetto/e2e/test-results/ on test failure (configured in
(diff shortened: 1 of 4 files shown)


vjt:
## The aggregation claims, measured on this account's runners

Run [32403389313](https://github.com/vjt/grappa-irc/actions/runs/32403389313),
throwaway branch `w2-1519-probe` (now deleted; the run outlives it). Two
matrices of four legs each — one all-green, one with leg 3 told to fail — plus
an `if: always()` aggregate over each. The red aggregate's assertion is
**inverted**: it passes only when it observes `failure`, so every assertion job
being green *is* the result.

| job | conclusion |
|---|---|
| green leg 1/4 … 4/4 | success ×4 |
| green aggregate | **success** |
| red leg 1/4, 2/4, 4/4 | success ×3 |
| red leg 3/4 | failure (deliberate) |
| red aggregate | **success** — i.e. it observed `failure` and it RAN |

Log lines, verbatim:

```
MEASURED job-total=4 for shard 1
MEASURED needs.green-leg.result=success
MEASURED needs.red-leg.result=failure
survived a sibling's failure
```

**1. The denominator is the matrix length.** `strategy.job-total` evaluated to
`4` for a `shard: [1,2,3,4]` matrix, in the job name and inside the step. The
sharded suite divides by that value, never by a literal.

**2. `fail-fast: false` really keeps the siblings alive.** Job timings relative
to the first start: leg 3 **failed at +2.0s**, and legs 1, 2 and 4 ran on to
**+22, +23, +23s** — twenty seconds past the failure, printing "survived a
sibling's failure" on the way out. Under the default (`true`) they would have
been cancelled at ~+2s.

**3. A failed leg makes the aggregate red, not skipped.** `needs.<matrix
job>.result` came back `failure` from a single failing leg out of four, and the
aggregate job **executed** — it did not skip — because of `if: always()`. The
all-green matrix aggregated to `success`, so the check is not stuck red either.
The run's own conclusion is `failure`, driven by the one deliberately red leg.

That is the mechanism `e2e-result` relies on, established by measurement rather
than by quoting the Actions documentation at it.

Issue #1287, opened 2026-08-13
VERSION-only bump classified :hot, but the reload cannot see the new code — silent stale-code deploy


vjt:
## What happens

A `VERSION`-only bump is classified `:hot` by `Grappa.Deploy.Preflight`, the deploy script takes
the hot path, `POST /admin/reload` answers `{"failed":[],"reloaded":[],"migrated":[]}` — and the
node keeps running the **old** code. Nothing in the deploy output signals it.

Reproduced in production on a self-hosted install (`sonic-grappa`, native Linux/systemd substrate,
`infra/linux/deploy.sh`, no `--force-*` flag) on 2026-08-13, deploying `v1.0.0` → `v1.1.0`.
Repro and initial diagnosis by the operator; the mechanism below is re-verified here against
`origin/main` (`4879785c`).

Observed there:

1. `mix release --overwrite` writes fresh `.beam` under `_build/prod/rel/grappa/lib/grappa-1.1.0/ebin/`,
   alongside the still-present `grappa-1.0.0/ebin/`.
2. Preflight verdict `:hot`; reload returns all-empty lists.
3. `/api/config` keeps answering the pre-deploy `"version":"1.0.0-…"` while git HEAD, `VERSION` and
   the on-disk release are all at `1.1.0`.
4. The new migration never ran either (it is itself a correct `:hot` `ADD COLUMN` expand op).
5. Re-running the deploy is a **silent no-op**: the completed-deploy marker
   (`runtime/last-deployed-sha`) was already written, so the script short-circuits with
   "same HEAD + completed-deploy marker match — nothing to do". It believes it succeeded and will
   never retry.
6. `--force-cold` fixes it; `/api/config` then reports `1.1.0-…`.

Net effect there: ~6.5 hours of production serving `v1.0.0` BEAM code under `v1.1.0` git history
and config.

## Mechanism

`Grappa.HotReload.reload_modified/0` (`lib/grappa/hot_reload.ex:186-187`, verified on `origin/main`):

```elixir
def reload_modified do
  :grappa |> :code.lib_dir() |> to_string() |> Path.join("ebin") |> reload_from()
end
```

`:code.lib_dir/1` resolves from the code server's path list as it stood when the app was loaded, so
on a running node it keeps pointing at the **boot** directory `lib/grappa-<old-vsn>/ebin`. A Mix
release lays out `lib/<app>-<vsn>/`, so bumping the version moves the new artifacts to a *sibling*
[...]

PR #1623, opened 2026-08-20
#964 — name a push device, and number the ones you have not named

vjt:
Settings: a push device can carry a user-set name, and the ones nobody has
named get an ordinal so two `Firefox on Linux` rows stop being byte-identical.

Issue #964. Not a rewrite: the branch cut on 2026-08-07
(`origin/964-device-label`, never proposed) was rebased forward across 1311
commits of `main`. Everything the branch *decided* survived; two things it
*assumed* did not, and those are the interesting part.

## What ships

* **Schema.** `label`, nullable, no default, off the create allowlist. It
  travels on its own `Subscription.label_changeset/2` (trim, blank → NULL, 64
  **grapheme** cap applied after the trim), so a POST body cannot widen the
  create surface and this changeset cannot be talked into touching an endpoint
  or a key.
* **Context.** `Push.update_label/2` takes the STRUCT, so the caller must
  already have loaded and scoped the row through `get_for_subject/2` — a path
  parameter cannot reach an UPDATE without passing the subject check.
* **REST.** `PATCH /push/subscriptions/:id`, body `{"label": <string|null>}`.
  **200** with the row summary, **400** on an absent or non-string `label`
  (a type error is the client's bug, not something the 422 envelope should
  describe), **404** with the uniform body for cross-subject *or* unknown IDs
  (the probing protection `delete/2` already has), **422** past the cap.
* **View.** The PATCH renders the SAME `subscription_summary()` an `:index` row
  carries, so cic splices the response back into the list instead of refetching.
  `label` is a new field on an existing shape ⇒ additive-only (#447), no
  `protocol_version` bump.
* **cic.** `deviceRows/1` pairs each device with the name its row prints: the
  user's label wins outright; otherwise the parsed `Browser on OS`, with an
  ordinal when two or more UNLABELLED devices parse the same. Labelled rows
  leave the grouping entirely, so naming one twin drops the other's number
  rather than stranding a lone `#2`. Inline rename from the row; the parsed
  string demotes to the secondary line once a label takes its place.
* **e2e.** One spec covering all four contracts, including the one a STORED
  ordinal would fail.

The ordinal is derived at render and is deliberately **not** a column. Two
independent reasons: it is a function of how many rows *currently* share a
parsed name (stored, it strands a `#2` the moment `#1` is deleted), and the
grouping key is `parseUserAgent`'s output, which exists only in cicchetto — a
server-side twin could disagree and then the number would describe a grouping
the user cannot see.

Proposal 3 in the issue (render `last_used_at`) already landed on `main`; this
is proposals 1 and 2.

## The two things the rebase caught

**🔴 The hand-typed migration stamp had become a live duplicate.** The branch
carried `20260807120000_add_label_to_push_subscriptions.exs`;
`20260807120000_fold_nickserv_pass_onto_password.exs` had since landed on `main`
with the same integer version and a different basename, which git fuses without
a conflict marker. That is the silent regime: Ecto's pending filter keys on the
VERSION, `ensure_no_duplication!` only sees the PENDING set, so on a database
that has already applied `20260807120000` — every deployed one — both files drop
out of pending, the run reports SUCCESS, and neither migration ever executes
again. Regenerated with `mix ecto.gen.migration` → `20260820174126`, checked
against `origin/main`'s current latest, not against the August main the branch
forked from.

**🔴 The rebase offered back `validate_subject_xor/1`, which #1580 had just
deleted.** #1580 collapsed twelve copies onto `Grappa.Subject.validate_xor/1`;
this branch adds `label_changeset/2` immediately above the old one, so the
deleted body reappeared inside a conflict hunk as context the resolver is
invited to keep. Dropped. It would have compiled, passed and read fine — a
conflict whose premise had evaporated.

Both are written up in the DESIGN_NOTES entry, because the demonstration is
worth more than the recommendation.

## Deploy class: HOT, measured

The path grep says COLD (`priv/repo/migrations/` matches), but the migration arm
over-triggers the same way the `^infra/` arm does. Since GH #41,
`AdminController.reload/2` runs `Ecto.Migrator` in-process **before** the module
reload, so only a CONTRACT migration forces a restart and `classify_migration/1`
decides from the `change/0` AST.

Measured, not assumed: `classify_migration` answers `:hot` for this file, and
the whole changed path set classifies `{:hot, []}` on `:jail`, `:docker` and
`:linux`. `PreflightTest`'s pinned HOT set — the repo's own oracle — agrees, and
this branch registers the new migration in it.

Independent of #964: a release that bumps `VERSION` is its own cold trigger
(`:version`, #1287) on the two `mix release` substrates.

## Gates

Run from the worktree with `GRAPPA_CACHE_ID=w1-964` (a fresh, cold id — dialyzer
built its PLT from scratch).

| gate | result |
|---|---|
| `mix test --warnings-as-errors` (full) | **8 doctests, 61 properties, 6657 tests, 0 failures** |
| `mix dialyzer` | 0 errors, 0 skipped |
| `mix sobelow --config` | pass (one pre-existing Low in `uploads/reaper.ex`) |
| `mix credo --strict` | 84 checks, 696 files, no issues |
| `mix format --check-formatted` | clean |
| `scripts/bun.sh run test` (vitest) | 295 files, 5708 tests passed |
| `scripts/bun.sh run check` | 4 stages, 0 failed |
| `scripts/design-notes-gate.sh` | 1 entry heading, separator + marker present |
| `scripts/union-rebase.sh origin/main` | `+166 -0 — contribution intact` |
| `scripts/integration.sh` (full e2e) | **755 passed / 5 failed** — see below |

**The five e2e reds are not this diff.** Four of them are
`waitForUserTopicReady` timeouts at `cicchettoPage.ts:680`, the 5 s barrier
#1579 measured and #1618 filed the root cause for (the WebSocket upgrade
serialised behind a DB write lock); the fifth times out on the login-ready
selector, the same failure one observation point earlier. None touch push,
settings or anything this branch edits. Re-run in isolation on this tree, all
five pass (35.4 s). Per the timing-race rule that is **neither guilt nor
absolution** — a seeded or scoped re-run of a temporal race is green from both
sides — so the attribution rests on the mechanism and the untouched surface,
not on the green.

**Not run:** `scripts/bats.sh`, and `scripts/check.sh` end to end.

`origin/964-device-label` still points at the pre-rebase `a9280925` and is
superseded by this branch.

### DIFF
--- a/cicchetto/e2e/tests/push-964-device-label.spec.ts
+++ b/cicchetto/e2e/tests/push-964-device-label.spec.ts
@@ -1,0 +1,147 @@
+// push-964 — naming a device, and the ordinal that names the ones you have not.
+//
+// The sibling spec (`push-964-device-row.spec.ts`) covers the two
+// disambiguators derived from data the row already carried: the activity
+// instant and the "this device" marker. This one covers the two that needed a
+// schema change — the user's own label, and the ordinal that separates the
+// twins nobody has named yet.
+//
+// Four contracts, every one asserted on what a person would SEE:
+//
+//   1. Two devices that parse to the same `Browser on OS` render DIFFERENT
+//      names, numbered oldest-first.
+//   2. Renaming one from its row replaces that name, and demotes the parsed
+//      string to the secondary line instead of dropping it.
+//   3. Naming one twin drops the OTHER's ordinal. This is the assertion a
+//      STORED ordinal would fail: a column written while the pair collided
+//      keeps its `#2` forever, whereas the derived name re-reads the list and
+//      finds the survivor alone in its group.
+//   4. Clearing the label hands the row back to the derived default — the
+//      twins collide again and BOTH are numbered once more.
+//
+// ## Why the rows are POSTed rather than registered from two browsers
+//
+// The contract under test is about two rows whose `user_agent` parses to the
+// same string. The runner has one engine per project, so a second browser
+// would supply the same UA — but also a second push stub, a second SW
+// registration and a second opt-in dance for rows this spec only ever reads.
+// POSTing them is the same insert the server would have performed, and it lets
+// the spec pin the UA, so the expected names are literal (`Firefox on Linux
+// #1`) instead of "whatever this engine calls itself".
+//
+// ## Why there is no page.reload()
+//
+// On boot `installPushResubscribe` renews a subscription the per-document push
+// stub makes look dropped; the renewal POSTs `supersedes`, which DELETES the
+// old row and inserts a fresh one — losing the very label under assertion. The
+// persistence leg therefore reads from a second BrowserContext, which carries
+// no opt-in intent and so renews nothing. Same reasoning as the sibling spec.
+//
+// No @webkit twin: every assertion is on rendered text, so nothing here is
+// engine-dependent.
+
+import { expect, test } from "../fixtures/test";
+import { loginAs, openSettingsSection } from "../fixtures/cicchettoPage";
+import { resetPushSubscriptions } from "../fixtures/push";
+import { getSeededVjt } from "../fixtures/seedData";
+
+const GRAPPA = "http://grappa-test:4000";
+
+// One UA for both rows: `parseUserAgent` collapses it to "Firefox on Linux",
+// which is precisely the collision Hypnotize reported.
+const TWIN_UA = "Mozilla/5.0 (X11; Linux x86_64; rv:128.0) Gecko/20100101 Firefox/128.0";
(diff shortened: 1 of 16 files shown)


vjt:

The commits of this PR are on `main` under new SHAs (the union was built by cherry-pick, so GitHub cannot mark this PR merged by itself). Verified by CONTENT, not by ancestry: `git log origin/main --grep '#964'` finds the work, and the DESIGN_NOTES marker `<!-- entry #964 -->` is present exactly once on the tip (e.g. `bde3f690`).

Issue #1532, opened 2026-08-18
e2e: ~135 files still say #bofh in prose after the per-spec channel moved


vjt:
After #1336's per-spec-subject slice (PR #1530), the spec subject no longer joins the shared channel: `AUTOJOIN_CHANNELS` is `#spec-w${TEST_PARALLEL_INDEX}`, and the 292 callers follow the constant without being touched.

What did **not** follow are the ~135 files that still spell `#bofh` inside **comments and error strings**. Behaviour is unchanged — none of them is a value the suite acts on — but the prose is now stale exactly where it names the corpus of the channel under test, which is the worst place for it: a reader trusts a comment that names a channel.

Deliberately left out of that slice, on the orchestrator's call: chasing prose across ~135 files costs more files than the cure itself, and main moves under long-lived branches.

Scope when someone picks this up:
- comments/strings that describe **the channel the subject is in** — these are the wrong ones and are the point of the issue;
- leave alone the ones that legitimately still mean the shared `#bofh` (the long-lived co-residents `vjt`, `m9b-test`, `m9b-victim` really are still bound to it via `cicchetto/e2e/compose.yaml`).

That distinction is the whole work: a blind `sed` would make the prose *more* wrong, not less. Measure with a bare grep before believing any count — a scoped grep in this tree has produced false zeros four separate ways.

vjt:
## Guard-rail for whoever widens the pattern next

The census done for PR #1619 turned up one thing this issue's body gets
wrong, and it is worth pinning here before someone acts on the sentence
as written.

The body says of the ~135 files: *"none of them is a value the suite
acts on"*. **That is true of the 289 occurrences the PR rewrote, and
false of the population.** Two call sites pass the string as a message
body and then assert on the rendered result:

- `cicchetto/e2e/tests/issue188-mentions-panel-polish.spec.ts:56,79,122`
  — `mentionBody("bofh", runId)` is sent, and the spec asserts
  `` scrollbackLine(page, "privmsg", `ping in bofh ${runId}`) `` is
  visible;
- `cicchetto/e2e/tests/issue71-inc2-permanent-rail-desktop.spec.ts:107,109`
  — `peer.privmsg(CHANNEL_A, mentionBody("bofh", runId))`, asserted the
  same way.

They fell outside the PR's domain for a mechanical reason: they carry a
bare `bofh`, not `#bofh`, and the pass only ever touched the sigilled
[...]

Issue #1441, opened 2026-08-16
e2e: ten specs open the media viewer with the same duplicated glue


vjt:
Reported by the #1438 worker while adding the tenth instance; not fixed there because lifting it would have blown that issue's budget.

**What was observed.** Nine existing e2e specs open the media viewer by repeating the same setup glue inline (send a message carrying an image URL, wait for the thumbnail, click it, wait for the modal). The #1438 swipe-to-dismiss spec adds a tenth copy.

**Why it matters.** This is the same class #1397 (bucket H, test-harness) names: a helper that exists in N hand-maintained copies drifts, and the drift is invisible until one copy starts failing for a reason the other nine do not have. The viewer is also about to grow behaviour (swipe-to-dismiss, zoom), so the copies will diverge under change, not merely sit there.

**Scope of the ask.** One fixture-level opener (`openMediaViewer`-shaped) that the ten call sites use, with the wait conditions taken from whichever copy is strictest — not from the first one found. No behavioural change to any spec; the count of assertions must not go down.

**Not established.** Whether all ten copies are byte-identical (that claim was NOT measured — #1397 already proved an "N byte-identical" claim in a sibling issue to be false), and whether the ten are the whole population. Whoever picks this up should count them first and say what the count is.

Relates to #1397, #1438.

vjt:
Closed by the pipeline change ruled by vjt on 2026-08-20: **`status:soon` is retired**. An issue is now closed when its work **merges**, and the **milestone** records which release carries it — the label duplicated what the milestone already said.

This issue was verified individually before closing, not swept: its work is on `origin/main` under commits that are not documentation-only. No behaviour changes here, only bookkeeping.

<sub>Applied by vjt-claude on vjt's instruction.</sub>

PR #1625, opened 2026-08-20
test(#1441): one media-viewer door — thirteen sites, not ten copies

vjt:
One fixture-level door for the media viewer (#1441), and the census the issue
asked for — which moved both of its numbers.

## The census, first

Measured on the branch base before anything moved, by a script whose
known-answer controls run before it prints a single number:

| measured | value |
|---|---|
| spec FILES that open the viewer | **10** |
| opening CALL SITES | **13** |
| distinct glue bodies, raw bytes | **5** |
| distinct glue bodies, indentation + comments normalised away | **3** |
| `link.click()` (Playwright, scrolls into view) | 9 |
| `link.evaluate((el) => el.click())` (in place) | 4 |
| dialog-constructing files with no opening click | **0** |

The issue's **ten is right per FILE and low per call site** —
`media-link-modal-viewer.spec.ts` opens it three times,
`media-link-cross-host-modal.spec.ts` twice. **"Byte-identical" was never
true**, exactly as the issue suspected: indentation alone splits the thirteen
into five bodies, because two sit inside a `test.describe`.

**The extractor lied first and a scalar control did not catch it.** Version one
pinned `media-link-modal-viewer.spec.ts == 3 open sites` and PASSED while
finding the wrong three — 75/175/176 instead of 75/174/194 — because a
`toBeVisible` quoted inside a prose comment counted as a barrier and a site
whose dialog binding sat outside the lookback was missed. Two errors of
opposite sign summed to the expected total. Controls are pinned to LINE NUMBERS
since. The negative control is `issue219-general-overlay-scroll-hold.spec.ts`,
which names "media-viewer" three times in prose and opens the `/names` modal.

## Two verbs, not one verb with a flag

Nine sites clicked through Playwright, four dispatched the anchor's own click.
That divergence is **load-bearing, not drift**: #196-preserve, #219, #213 and
#1438 measure the scrollback's scroll position ACROSS the open, and Playwright's
scroll-into-view would move the very thing under test. So the fixture exports
`openMediaViewer` and `openMediaViewerInPlace` rather than one opener with
`scroll: boolean` — a boolean at thirteen call sites is a coin flip a reviewer
cannot check.

`closeMediaViewer` rides in for the same drift at the other edge (nine copies of
the close click plus its `toBeHidden`), and `mediaViewer` exports the locator so
the one NEGATIVE use — the audio path must not open the modal — reads the same
accessible name the positive ones do. `"Media viewer"` and
`"Close media viewer"` now appear once each in the tree.

## Wait conditions: strictest, then frozen

All thirteen waited `toBeVisible({ timeout: 5_000 })`, so strictest and uniform
coincide — kept, never raised. A module constant and deliberately NOT a
per-call-site parameter, the opposite of `uploadJourney.ts`'s rule: there the
image and video budgets genuinely differ 6x, here every caller asks the same
question and a parameter would re-open the axis this closes.

#213's extra wait on `.media-viewer-media--zoomable` **stays in #213** — it is
image-and-zoom specific (the cross-host VIDEO case has no zoomable element) and
it is the locator that spec drives. Strictest COMMON wait, not strictest
anywhere.

The media-class assertion moved into `uploadImageAndGetLink`, where six of the
eight upload preambles already had it; #213 and #1438 gain it. Without the class
the anchor never opens the viewer, and that used to surface as a five-second
timeout on the dialog instead of naming the classifier.

## The +0

From Playwright's own collector rather than a grep: `playwright test --list`
reports **759 tests in 411 files** on the base and on this branch, and the full
759-line title list **diffs empty** (line numbers stripped — those legitimately
move). Re-measured after the rebase, against the new base `9be1b40a`.

Source `expect(` tokens fall 4189 -> 4164. That drop is what deduplication IS,
and it reconciles exactly: 13 `toBeVisible` + 9 `toBeHidden` + 6 `toHaveClass`
removed from the specs = 28; 3 added in the fixture. **Executed** assertions are
unchanged at eleven of the thirteen sites and +1 at the two that gained the
class check, so no path lost coverage.

## Gates

- `bun run check` — 4 stages, 0 failed, both before and after the rebase.
- `bun run test` — 295 files, 5695 tests, green.
- `scripts/design-notes-gate.sh origin/main` — rc=0.
- `scripts/union-rebase.sh origin/main` — contribution intact (`+111 -0`);
  the four-check ritual after it: numstat identical on every path, previous
  entry (#1532) byte-identical at 6343 bytes, boundary shape read on the file
  (marker / blank / `---` / blank / `## `), marker unique in the file and
  against the `origin/main` tip.

## What this does NOT claim

- **No e2e spec was run** — no stack lane was requested for this change. The
  argument is static (`tsc -p e2e/tsconfig.json`) plus collection equality; it
  is not a green suite.
- **No unit test covers the opener.** `whoisWait.ts` and `privmsgWait.ts` are
  unit-testable because they carry a decision; these four functions are driver
  calls around an imported `expect`, and faking that barrier would mean
  injecting a seam invented for the test over a verb the existing gates cover.
- **The population is "everything that names the dialog".** Two independent
  grep shapes agree on the same ten files and every dialog-constructing file
  has an opening click, but a spec that opened the viewer and never asserted on
  it would be invisible to both.

## One correction to the record

`bun run check` **does** gate `cicchetto/e2e/`. `cicchetto/scripts/check.ts`
runs four stages and the fourth is `tsc --noEmit -p e2e/tsconfig.json`, with
`biome.json` including `e2e/**` (#484, #1469). Any brief saying it is
src-scoped is stale.

### DIFF
--- a/cicchetto/e2e/fixtures/mediaViewer.ts
+++ b/cicchetto/e2e/fixtures/mediaViewer.ts
@@ -1,0 +1,139 @@
+// The media-viewer door (#1441): one place that knows how the viewer is
+// named, opened and closed.
+//
+// ## What was measured before the lift (on 5ec44475)
+//
+// THIRTEEN opening call sites across TEN spec files — not "ten copies". The
+// issue's ten is right at FILE granularity and low at call-site granularity:
+// media-link-modal-viewer.spec.ts alone opened it three times and
+// media-link-cross-host-modal.spec.ts twice.
+//
+// They were NOT byte-identical, a claim the issue explicitly refused to make.
+// Counting raw bytes there were FIVE distinct bodies (indentation splits them:
+// the two sites inside a `test.describe` are indented one level deeper);
+// normalising indentation and comments away leaves THREE, and the third is
+// only "shorter because the dialog was already bound above". So the real
+// divergence is ONE axis, below.
+//
+// ## The one axis that genuinely diverges, and why it gets two verbs
+//
+// Nine sites clicked with `link.click()` — Playwright's click, which scrolls
+// the anchor into view first. Four clicked with
+// `link.evaluate((el) => (el as HTMLElement).click())` — the anchor's OWN
+// click, dispatched where it sits. That is not drift. Those four
+// (#196-preserve, #219, #213, #1438) MEASURE the scrollback's scroll position
+// across the open, and Playwright's scroll-into-view would move the very thing
+// under test before the gesture it is testing ever fires.
+//
+// So this module exports two verbs rather than one verb with a `scroll: bool`.
+// A boolean at thirteen call sites is a coin flip a reviewer cannot check; two
+// names are two things a reviewer can read. `openMediaViewerInPlace` says what
+// it protects in its name, which is the whole reason those four specs are
+// green.
+//
+// ## Wait conditions: from the strictest copy, and then frozen
+//
+// All thirteen waited `toBeVisible({ timeout: 5_000 })` on the dialog, so the
+// strictest and the uniform value are the same 5_000 — kept, not raised.
+// It is a module constant and NOT a per-call-site parameter, deliberately
+// against `uploadJourney.ts`'s opposite rule: there the image and video budgets
+// genuinely differ 6×, here every caller wants the same answer to the same
+// question, and a parameter would re-open the drift axis this module exists to
+// close. A future viewer that needs longer changes it here, for everyone, on
+// purpose.
+//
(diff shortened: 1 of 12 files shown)


vjt:

The commits of this PR are on `main` under new SHAs (the union was built by cherry-pick, so GitHub cannot mark this PR merged by itself). Verified by CONTENT, not by ancestry: `git log origin/main --grep '#1441'` finds the work, and the DESIGN_NOTES marker `<!-- entry #1441 -->` is present exactly once on the tip (e.g. `fd44e9d8`).

PR #1684, opened 2026-08-22
fix(#1683): restore docker-exec account bootstrap

### LINKED ISSUE #1683: release image: documented docker exec create-user cannot read first-boot secrets from /data
## Summary

On a fresh published-image installation, the container starts successfully and generates its secrets in `/data/grappa.env`, but the documented account bootstrap command fails:

```sh
docker exec -it <container> bin/grappa create-user alice --admin
```

The CLI aborts during runtime config because `SECRET_KEY_BASE` is missing.

## Reproduction

1. Start `ghcr.io/vjt/grappa:latest` using the minimal `compose.release.yaml` shape, with a writable named volume mounted at `/data` and a valid `PHX_HOST`.
2. Wait until the container is healthy.
3. Confirm `/data/grappa.env` contains the generated secret keys.
4. Run:

   ```sh
   docker exec -it <container> bin/grappa create-user alice --admin
   ```

5. Observe:

   ```text
   ERROR! Config provider Config.Reader failed with:
   ** (RuntimeError) environment variable SECRET_KEY_BASE is missing.
   ```

Reproduced with release `1.3.1` from `ghcr.io/vjt/grappa:latest`.

## Expected

The command documented in `docs/OPERATIONS.md` should load the existing production environment and prompt for the new account password. It must reuse the already-generated secrets and must not rotate them.

## Cause indicated by the runtime state

The first-boot entrypoint reads or generates `/data/grappa.env` and exports those values only into the long-running release process. The secret keys are present in `/data/grappa.env`, but absent from Docker `Config.Env`. A later `docker exec` starts with Docker configured environment rather than mutations made inside the entrypoint process, so the standalone CLI cannot reach runtime configuration.

No secret values are needed to observe this; comparing key names is sufficient.

## Acceptance criteria

- `docker exec -it <container> bin/grappa create-user alice --admin` works after a fresh release-image boot.
- Existing operator-provided environment values retain precedence.
- `/data/grappa.env` is parsed without executing arbitrary shell content, matching the entrypoint safety posture.
- No secret values are printed to logs or command arguments.
- An integration test covers fresh-volume boot followed by the documented `docker exec` account-bootstrap command.

pminervini:
## Summary

- re-enter the release image entrypoint for Docker account verbs when first-boot secrets are absent from Docker Config.Env
- keep the existing fixed-key parser and operator-environment precedence as the single source of truth; no env file is sourced
- cover the dispatcher decision and the documented fresh-volume docker exec path

## Testing

- `scripts/mix.sh --env=dev ci.check` — 6,786 tests, 0 failures; Dialyzer and static gates passed
- `scripts/bats.sh` — 659 tests passed
- focused entrypoint/operator suite — 37 tests passed
- `scripts/release-image.sh build`
- release-image smoke against a fresh volume, including `docker exec ... create-user --admin`

Closes #1683


pminervini:
Follow-up from release-image validation: `add-network` has a separate pre-existing connected-without-live-session defect, now tracked in #1685 (related to #1163). I am keeping this PR scoped to #1683 secret restoration; the fresh-volume image acceptance remains the documented first-user command.

Issue #1686, opened 2026-08-23
release.yml: ghcr :latest can point at a release candidate — -v:refname ranks v1.3.0-rc2 above v1.3.0


vjt:
`release.yml` points ghcr `:latest` at a release candidate: `--sort=-v:refname` ranks `v1.3.0-rc2` **above** `v1.3.0`.

The `:latest` gate in the docker job is meant to keep the mutable tag on the newest **stable** image — the comment at `.github/workflows/release.yml:933-938` says so, and its stated hazard is a backport tag re-pointing `:latest` at an older image. It has a second hole the comment does not cover: a **pre-release** tag takes the slot too, because git's version sort places a `-rcN` suffix higher than the bare version when `versionsort.suffix` is unset.

`.github/workflows/release.yml:939-946`

```
highest="$(git tag -l 'v*' --sort=-v:refname | head -1)"
if [ "v${version}" = "${highest}" ]; then
  latest=yes
```

## Measurement

Not a derivation — the run said it out loud. Job `96617993638` of run `32429461818` (tag `v1.3.0-rc2`):

```
2026-08-20T23:38:15.7510096Z v1.3.0-rc2 is the highest tag — tagging :latest
```

The sort behaves the same today, with no `versionsort.suffix` configured in the repo:

```
$ git tag -l 'v*' --sort=-v:refname | head -3
v1.3.1
v1.3.0-rc2
v1.3.0-rc1
```

`v1.3.0` sits *below* both of its own candidates.

**Window, bounded from both ends.** `v1.3.0-rc2` published `:latest` at 2026-08-20 23:38Z; the next stable run (`32517813292`, tag `v1.3.0`) started 2026-08-21 19:18Z — so ghcr `:latest` served a release candidate for roughly **20 hours**. `v1.3.0-rc1`'s run failed, so only rc2 landed.

**Currently clean**, so this is a latent gate defect, not a live incident — the registry agrees `:latest` is the stable release:

```
:latest      sha256:f9084e2a9db54881640baf0f4709960ab4638db85072afdd9191964a6c1b0719
v1.3.1       sha256:f9084e2a9db54881640baf0f4709960ab4638db85072afdd9191964a6c1b0719
v1.3.0-rc2   sha256:3a8ea72acc8cd5bc03e0b03b2b9508921da8f9b9038277f499030aad3750f96a
```

## Why it has a consumer

[...]

Issue #1636, opened 2026-08-20
release.yml publishes release candidates without --prerelease


vjt:
`release.yml` publishes release candidates as full releases: `gh release create` is called without `--prerelease`, so the pre-release marker depends on a human remembering to flip it afterwards.

## Measurement

`v1.3.0-rc2` was cut tonight from `29562617` (tag `v1.3.0-rc2`, run 32429461818, all legs green). Immediately after the run:

```
$ gh release view v1.3.0-rc2 --json isPrerelease,isDraft
{"isDraft":false,"isPrerelease":false}

$ gh release view v1.3.0-rc1 --json isPrerelease
{"isPrerelease":true}
```

rc1 carries the flag, rc2 did not — and nothing in the workflow sets it:

`.github/workflows/release.yml:1249`

```
if ! gh release create "${RELEASE_TAG}" \
      --repo "${GITHUB_REPOSITORY}" \
      --title "grappa ${RELEASE_TAG}" \
      --generate-notes \
      "${assets[@]}"; then
```

No `--prerelease`. So rc1's `true` was set by hand after the fact, not by the pipeline.

## Blast radius

Bounded, this time. `GET /repos/vjt/grappa-irc/releases/latest` still answers `v1.2.0`, so no release candidate has ever displaced the stable release in the API or on the repo's landing page. The exposure is that it *could*: GitHub picks "Latest" among non-prerelease, non-draft releases, and a green rc run is the only thing standing between a candidate and that slot.

I flipped `v1.3.0-rc2` by hand (`gh release edit v1.3.0-rc2 --prerelease`, verified `isPrerelease: true`). That closes today's instance, not the hole.

## Suggested shape

Derive the flag from the tag rather than from memory: the tag is authoritative and already parsed elsewhere. `Grappa.Version.derive/2` matches `exact_tag == "v" <> base` and `Version.parse/1` reads `1.3.0-rc2` as `pre: ["rc1"|"rc2"|...]` — a non-empty `pre` is exactly the condition for `--prerelease`. Anything that spells the rule twice will drift.

Worth a test in the packaging suite alongside the gates from #1594, so the next candidate fails loudly instead of publishing quietly as stable.

vjt:
Merged to main as `191de3e8` (FF, PR #1640). Milestone **1.3**.

[...]

Issue #1168, opened 2026-08-09
release: publish a shottino container image to ghcr.io, not just the server


vjt:
Requested by Lucy on IRC (#grappa): publish a `shottino` container image to
`ghcr.io` alongside the server image, so the `--ircd` bridge can be pulled instead of
built locally.

## Measured on `origin/main` (`3383abf6`)

- The release workflow pushes exactly **one** image: `ghcr.io/<owner>/grappa`
  (`.github/workflows/release.yml:560`, login at `:598-604`, documented at `:28`).
  There is no second `docker/build-push-action` target.
- `Dockerfile.shottino` exists and is complete, but its only consumer is
  **compose, as a local `build:`** — `compose.yaml:232-239`, behind the `ircd`
  profile. `frontends/shottino/README.md:171` documents exactly that path:
  `docker compose --profile ircd up -d`, built from `Dockerfile.shottino`.
- So every user of the bridge compiles it themselves. The release *does* build
  shottino for deb / Arch / rpm and proves the binary runs
  (`release.yml:204-210`, `:328-333`, `:480-484`) — the container is the one
  distribution channel where it is missing.

## What this asks for

Push a second image (e.g. `ghcr.io/<owner>/grappa-shottino`) from
`Dockerfile.shottino` in the same release job that already pushes the server image,
with the same tag scheme (`:<tag>` + `:latest`) and the same OCI labels, then point
`compose.yaml`'s `ircd` profile at the published image with the local `build:` kept as
a fallback.

Related: #1027 (the supported compose service for the `shottino --ircd` bridge) added
that service; this is the missing publish step for it.

## Not measured

Whether the multi-arch matrix the server image uses applies cleanly to the shottino
build — it compiles C against ncursesw/openssl, so the cross-build cost is a real
question for whoever picks this up, not an assumption to carry in.

vjt:
Shipped in **v1.0.0** — closing.

2 dedicated commits landed for this issue between `v0.16.0` and `v1.0.0`:

- `068e3b47` release(#1168): publish the shottino bridge image alongside the server
- `febea155` docs(#1168): the bridge pull path, and why compose was left alone

Release: https://github.com/vjt/grappa-irc/releases/tag/v1.0.0

The `status:soon` label was never cleared after the work merged, so the issue stayed open long after it shipped. If anything here still misbehaves on 1.0.0, please reopen with the symptom rather than filing a new report.

Issue #573, opened 2026-07-31
release.yml: v0.8.0 published without the .rpm — a failed packaging job does not stop the release


vjt:
## Summary

**Release `v0.8.0` shipped without its `.rpm`.** The `build + prove .rpm (x86_64)` job failed, the
`publish GitHub Release assets` job succeeded anyway, and the release went out carrying only a subset of
the intended artifacts — with nothing surfacing the omission.

## Evidence

Release run [`30596619063`](https://github.com/vjt/grappa-irc/actions/runs/30596619063) (tag `v0.8.0`, 2026-07-31 01:32):

| job | conclusion |
|---|---|
| `build + prove .deb (amd64)` | success |
| **`build + prove .rpm (x86_64)`** | **failure** |
| `build + prove Arch pkg (x86_64)` | success |
| `publish GitHub Release assets` | success |

Assets actually attached to `v0.8.0` (`gh release view v0.8.0`):

    default.SRCINFO
    grappa-0.8.0-1-x86_64.pkg.tar.zst
    grappa_0.8.0_amd64.deb
    PKGBUILD

No `.rpm`. Fedora/RHEL users get a release that silently has no package for them.

⚠️ **The job log is no longer retrievable** (`gh run view --log-failed` returns nothing — GitHub log
retention has expired), so the *proximate* cause of the rpm build failure is not recoverable from CI.
It has to be reproduced by re-running the rpm leg.

## The general defect, not the instance

This is the second release to lose a package leg this way — run
[`30418051670`](https://github.com/vjt/grappa-irc/actions/runs/30418051670) (`v0.6.1`) lost the **Arch**
leg the same way:

| job | conclusion |
|---|---|
| `build + prove .deb (amd64)` | success |
| **`build + prove Arch pkg (x86_64)`** | **failure** |
| `publish GitHub Release assets` | success |

So the pattern is structural: **`publish` does not depend on every packaging leg succeeding**, and a
partial release is indistinguishable from a complete one once published. Whatever the individual build
failures are, the release should not quietly ship a hole.

## Expected

1. A failed packaging leg either **blocks the publish** or the release is **explicitly marked partial**
   (and the omission is stated where a user will see it — release body / announce), never silently dropped.
2. Reproduce and fix the actual rpm build failure (needs a fresh run; the v0.8.0 log is gone).
3. `v0.8.0` gets its `.rpm` attached retroactively, or its release notes say the rpm is unavailable.

## Related

[...]

PR #1699, opened 2026-08-23
packaging(#1686): rank ghcr :latest over releases, not over every tag

vjt:
## What this fixes

`release.yml` chose which image gets the mutable ghcr `:latest` with

```sh
highest="$(git tag -l 'v*' --sort=-v:refname | head -1)"
```

and `versionsort.suffix` is not configured in this repository, so git's version
sort places a `-rcN` suffix **above** the bare version.

## The defect has two faces; #1686 names one

Quoted from the runs' own logs, not derived:

```
2026-08-20T23:38:15Z  v1.3.0-rc2 is the highest tag — tagging :latest
2026-08-21T19:18:58Z  v1.3.0 is NOT the highest (v1.3.0-rc2) — NOT tagging :latest
2026-08-22T08:40:28Z  v1.3.1 is the highest tag — tagging :latest
```

* **Face one** (reported): a release **candidate took `:latest`**.
* **Face two** (not reported): the **stable release that followed was then denied
  it** by its own candidate. `:latest` skipped `v1.3.0` entirely and served rc2
  until `v1.3.1` — roughly **33 hours**, not the ~20 the issue bounds it at. The
  issue closes the window at "the next stable run"; the second line above is
  that run, saying it did not take the tag.

Still latent, not a live incident: `:latest` today resolves to `v1.3.1`.

## Why not `versionsort.suffix`, which the issue also proposes

Measured on a scratch repository holding the tag set as it stood at the rc2
push (`v1.0.0 v1.1.0 v1.2.0 v1.3.0-rc1 v1.3.0-rc2`, no `v1.3.0` yet):

```
git tag -l 'v*' --sort=-v:refname                  -> v1.3.0-rc2
git -c versionsort.suffix=- tag -l ... --sort=...   -> v1.3.0-rc2
```

Identical. The suffix orders a candidate against **its own release** and nothing
else, so it repairs face two and leaves the measured face one exactly where it
was. Restricting the ranking to releases repairs both — and once pre-releases
are out of the candidate set the suffix has nothing left to order, so setting it
as well would be a second mechanism doing the first one's job. The issue asks
for belt-and-braces; this takes one, deliberately.

## The change

* **New `infra/packaging/latest_tag_gate.sh`** — verdict (`yes`/`no`) on stdout,
  reason on stderr, exit 2 on a tag it cannot read. It **reuses**
  `prerelease_flag.sh` (#1636) rather than comparing versions a second time: a
  hand-rolled filter is wrong on the row that decided that script's
  implementation, since `v1.4.0+foo-bar` carries a hyphen inside its **build**
  metadata and is a release.
* **`release.yml`** calls it instead of ranking inline; a verdict that cannot be
  derived stops the job, because the permissive reading is the one that
  publishes.
* **A pre-release gets its own arm** for log honesty. Falling through would
  print `v1.3.0-rc2 is NOT the highest (v1.2.0)`, which is false — rc2 *is*
  higher.
* **One gate, both images.** The verdict feeds the existing `emit_tags`, so
  `grappa` and the `grappa-shottino` bridge share it (#1168).
* **The repair path keeps working.** The docker job checks out
  `ref: REPAIR_TAG`, so on a #573 (b) image repair the tree is the **old tag's**;
  `git tag --contains d622616d` answers `v1.3.0 v1.3.1`, so every earlier tag
  predates `prerelease_flag.sh` and a naive call would have exited **127** on a
  path that works today. Both scripts therefore join the repair scaffolding
  checkout — the verb the job already has for "this comes from the dispatched
  ref". Measured not to change the artifact: `Dockerfile.release` COPYs exactly
  two files out of `infra/packaging` and neither is one of these; the bats case
  pins that count.

Operator-visible consequence: a candidate tag now publishes `:v<version>` only,
so `docker pull ghcr.io/<owner>/grappa` keeps resolving to the newest stable
release across a candidate cycle.

## Tests

`test/infra/release_latest_gate_test.bats`, 15 cases driving the gate against
fabricated git repositories: the two measured faces, the backport exclusion
(alone and composed with a candidate above it), build metadata, refusal of an
unreadable tag under test, a stray tag elsewhere not vetoing a release, the
empty-tag-set edge, and the workflow wiring for both images and the repair path.

**Mutation bench: 8 mutants, 8 killed** — and the first pass was 7/8. Replacing
the skip arm with a hard failure left the suite green at 15/15, because the
stray tag sat *below* the tag under test and the scan stops at the first release
it finds, so the classifier was never handed anything it could refuse. Fixed by
making the stray outrank the subject and asserting the skip line; the second
pass kills all eight. The decisive mutant is `versionsort.suffix` in place of
the filter — it turns face one red, which is the argument above expressed as a
test.

**What the suite does not cover**, stated so a green is not read wider than it
is: it drives shell logic, not GitHub Actions. It does not exercise
`docker/build-push-action` and never contacts a registry — "which tags the
runner pushed" is asserted only as the tag list the gate feeds that action, via
the workflow text. The end-to-end fact is unreachable without cutting a release.

## Gates

* `scripts/bats.sh` (full): `1..674`, 674 ok, 0 failures.
* `scripts/shellcheck.sh`: 81 scripts, clean. `scripts/posix-parse.sh`: 32
  sh-dialect scripts, clean — the new gate is enrolled in both.
* `scripts/design-notes-gate.sh origin/main`: 1 new entry heading, separator and
  marker present.
* `scripts/check.sh`: green.

⚠️ `.github/workflows/**` is not in `integration.yml`'s `paths:`, so this PR
runs **fewer checks than usual** — `ci.yml` has no path filter and does run.
Four of four is not five of five.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -852,6 +852,19 @@ jobs:
       # `mix.exs`, `mix.lock`, `VERSION` and `cicchetto` are POINTEDLY absent: the
       # moment source joins this list, the artifact stops being the tag and the
       # label lies.
+      #
+      # The two `infra/packaging` scripts are the exception that proves it, and
+      # they are WORKFLOW LOGIC rather than shipped bytes (#1686). The `:latest`
+      # gate below calls them, and the tree this job checks out on a repair is
+      # the OLD TAG's: `git tag --contains d622616d` answers `v1.3.0 v1.3.1`, so
+      # every earlier tag predates `prerelease_flag.sh` and the call would die
+      # 127 on a path that works today. Taking them from the dispatched ref is
+      # the same principle the whole repair path rests on — "dispatching from
+      # main runs the FIXED workflow against the tag" — extended to the two
+      # helpers that workflow now reasons with. MEASURED not to touch the
+      # artifact: `Dockerfile.release` COPYs exactly two files out of
+      # infra/packaging (`version.sh`, `gen-secrets.sh`) and neither is one of
+      # these, so the image is byte-for-byte the tag's either way.
       - name: Repair path — take the image build scaffolding from the dispatched ref
         if: ${{ env.IMAGE_REPAIR == 'true' }}
         run: |
@@ -860,7 +873,9 @@ jobs:
             Dockerfile.release \
             Dockerfile.shottino \
             .dockerignore \
-            infra/docker
+            infra/docker \
+            infra/packaging/latest_tag_gate.sh \
+            infra/packaging/prerelease_flag.sh
           # Print it. A repair that silently changed nothing means the tag was
           # already fine and the failure is elsewhere; a repair that changed more
           # than expected is the reviewer's business, not a surprise found later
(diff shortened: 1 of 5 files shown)


vjt:
Landed on main as a pure fast-forward: `main` now points at this PR's head `20b368ec`, so every commit here is on main verbatim. GitHub shows this PR as CLOSED rather than MERGED only because the head ref was deleted before the merge state propagated — a bookkeeping artefact, not a re-open. Issue #1686 closed at the merge.

Issue #1701, opened 2026-08-23
Radio player: opening it does not push the scrollback up, and it belongs under the top bar


vjt:
Two related complaints about where the docked player sits, reported from the PWA
on staging:

1. opening the player does not push the scrollback up — the bar takes its space
   and the last lines are left under it;
2. above the compose box it is too intrusive, and it would sit better pinned under
   the top bar.

Measured on `origin/main` `f2b93f2b`, the commit staging serves. Not reproduced on
a device by me — what follows names where the machinery is, not why it failed.

## The push-up machinery already exists

The bar is **in flow**, not an overlay: `Shell.tsx:710` mounts it between
`ScrollbackPane` and `ComposeBox`, and `.audio-mini-player`
(`themes/default.css:3711`) is a plain flex row with no `position`. The whole
chain above it — `.shell-main`, `.drop-upload-zone`, `.scrollback-pane`,
`.scrollback` — is a flex column with `min-height: 0` at every step
(`:1391`, `:2916`, `:2974`), so mounting the bar genuinely shortens the scroll
container rather than covering it.

And a shortened container is already supposed to re-pin. `ScrollbackPane.tsx:2052`,
in the `ResizeObserver` on the scroll list (`listRef` is the `.scrollback` div,
`:3832`):

```ts
const moved = lastContainerHeight !== null && height !== lastContainerHeight;
lastContainerHeight = height;
if (moved && followMode()) applyActivation("tail-only", false);
```

That is #778, added for exactly this class of failure: *"chrome growing inside the
shell shrinks THIS box with no event at all"*, with the `.compose-box-error` line
as the motivating case. `followMode` defaults to `true` (`:1203`).

So the reported symptom is **not** a missing feature. Either the observer does not
see this particular height change, or the follow intent is not set when it fires,
or the re-pin lands before the bar's own final height does — the bar changes height
a second time when `onLoadedMetadata` resolves `live()` (`AudioMiniPlayer.tsx:71`)
and the seek slider is swapped for the `live` badge. Which of those it is needs a
device repro; picking one from here would be a guess.

## Why moving it under the top bar helps on its own

Worth noting for whoever takes this: the reposition is not only a taste call, it
makes the first complaint structurally milder. Chrome that mounts **below** the
scrollback steals space adjacent to the tail, which is the part the reader is
looking at, so a missed re-pin hides the newest lines. Chrome that mounts **above**
it steals space at the far end from the tail, where nothing the reader is watching
[...]

PR #1732, opened 2026-08-24
fix(#1701): restore the follow INTENT when the box the held px was measured in moves

vjt:
Half of #1701 — the **push-up** complaint. The reposition half is
deliberately NOT here: it collides with the #985 floating ☰ on mobile
query/server windows, over exactly the corner the player's ✕ sits in, and
that corner has already cost two owner rulings (#1050, #1051). Reported for
a ruling rather than settled here.

## Root cause

The issue established the machinery is present — the bar is in flow, the
flex chain shortens `.scrollback`, and #778's `ResizeObserver` re-pins a
follower on any container height change — and asked which of three things
was failing. It is none of them.

The player is tuned from the rail's **station picker**, the picker holds a
`createOverlayLock` refcount, and picking deliberately does not close it
(`RailRadio`: *"auditioning stations is the second kind, so the picker
stays up"*). So the bar mounts while the overlay freeze is up:
`applyActivation` hands the re-pin to `resolveIntent`, `overlay-freeze`
outranks `tail-follow`, and the write is dropped — correctly, a mid-list
reader under an overlay must not be yanked. No overlay edge fires again
until the picker closes, and the close edge then replays a `scrollTop`
captured through a viewport ~47px taller than the one it lands in.

## Fix

This is #1121's defect on the other axis, at the same site: there the box
was steady and the content grew under it, here the content is steady and
the box shortens. Both say the same thing — the freeze holds one number,
and a `scrollTop` names a position only relative to the viewport it was
read through.

So the viewport is captured with it, and `applyOverlayRestore` restores the
tail instead of the proxy when — and only when — the reader was following
AND the box changed size. Gated on the box and not the extent (content
arriving underneath is #1121's axis, ruling unchanged); gated on the follow
intent via the *same* predicate the intent is set from, not a second one.

Derived from the captured `clientHeight` rather than from the observer, so
the correction holds whether or not the RO fired — which matters, because
nothing available can settle which of the two suppressed paths ran.

## Evidence

TDD, RED first: `expected 1500 to be 1547`, the 47px the bar takes.

Two mutants, both caught — dropping the follow gate reddens the new
mid-list guard; dropping the box gate reddens #1121's own two geometry
specs plus the W1 restore characterization.

Gates: full vitest **+3 tests, +0 files** as a difference from the base
(310/6047 → 310/6050); `check` green on all four stages;
`design-notes-gate.sh` green.

## Residue, declared

The cure lands on the close edge. On mobile that is the moment the reader
can see the pane again. **On desktop the picker covers only the rail, so an
operator watching the scrollback with it still open sees the stale pane
until they close it.** Curing that means letting a height change outrank
the freeze, which is the signal #219 exists to suppress; the narrower
reading (a non-covering surface should take `createOverlayEscape`, not the
refcount — `overlayScrollLock.ts` already says so) is a change to the
rail's posture and not this fix's to make. Named, not hidden.

### DIFF
--- a/cicchetto/e2e/tests/issue1701-player-under-picker-tail.spec.ts
+++ b/cicchetto/e2e/tests/issue1701-player-under-picker-tail.spec.ts
@@ -1,0 +1,157 @@
+// #1701 — chrome mounting UNDER a covering overlay must not leave a tail reader
+// short of the tail once the overlay goes.
+//
+// The container twin of issue1121-overlay-close-tail-reader, which drives the
+// same site on the CONTENT axis: there the box was steady and lines grew
+// underneath, here the content is steady and the BOX shortens. Both are the one
+// sentence `applyOverlayRestore` is now written around — the freeze holds a
+// single number, and a scrollTop names a position only relative to the viewport
+// it was read through.
+//
+// The gesture is the reported one, and every step of it is load-bearing:
+//
+//   * the rail's station picker holds a `createOverlayLock` refcount, so the
+//     pane is FROZEN from the moment it opens;
+//   * picking a station deliberately does NOT close it (`RailRadio`: a control
+//     the operator re-flips in place is not a launcher), so the docked bar
+//     mounts WHILE the freeze is up;
+//   * the bar takes ~47px out of `.scrollback`, and #778's ResizeObserver
+//     re-pin — which exists for exactly this class — is outranked by
+//     overlay-freeze and dropped. Correctly: a mid-list reader under an overlay
+//     must not be yanked;
+//   * no overlay edge fires again until the picker is closed, and that close
+//     edge is where the debt is owed.
+//
+// WHY THE ASSERTIONS ARE IN RAW PIXELS AND NOT IN THE TAIL THRESHOLD.
+// `SCROLL_BOTTOM_THRESHOLD_PX` is 50 and the bar is about 47, so "is the reader
+// within the tail band" answers YES on the BROKEN build. A spec written against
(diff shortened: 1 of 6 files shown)


vjt:
Two additions since the PR was opened, neither of them a claim of green.

**The desktop residue is #1733**, with the tension against #219's freeze
written into it — not to be cured here. `#1701a` in DESIGN_NOTES states it in
the same terms so a later reader meets it as a known bound.

**The e2e twin is in: `issue1701-player-under-picker-tail.spec.ts`, written,
committed and COLLECTED — not run.** The e2e lane is w1's (#1715). Static
evidence only: `tsc -p e2e/tsconfig.json` green, biome green, playwright
collects it (622/368 → 623/369 as a difference, one occurrence). Collection is
not execution, and nothing here is claimed to have passed against a real
layout.

Worth flagging about the spec itself, since both are ways it could have been
green and worthless: the assertions are in **raw pixels and not in
`SCROLL_BOTTOM_THRESHOLD_PX`** — the threshold is 50 and the bar is ~47, so a
tail-band assertion answers YES on the *broken* build; and the shrink is
**measured rather than assumed**, so a bar that changes height reddens it
instead of hollowing it. It also pins its own premise (picking must leave the
picker up), because the day that changes the freeze lifts on the same tick as
the bar mounts and the spec quietly stops reproducing anything.

The reposition half of #1701 is still parked on a layout ruling — under the top
bar it collides with the #985 floating ☰ on mobile query/server, over the
corner the player's ✕ sits in, and that ✕ is the only reachable STOP verb on a
phone (#682, #1697).

Issue #1714, opened 2026-08-23
CI: gate both branches of the release CLI's distribution hop (#1685)


vjt:
## What

The CLI transport landed in #1685 / PR #1713 raises Erlang distribution from the
`eval` node and crosses into the live node with `:erpc.call/4`. Both of its
branches are currently proven only by hand:

- **live node present** — distribution comes up, the call is answered, the
  session starts;
- **no live node** — `net_kernel` must fail fast, the row is written `:parked`,
  and the CLI must say so instead of pretending it connected.

Neither branch has a gate. A regression in either one is invisible until a
self-hoster hits it on first boot.

## Where it goes

`.github/workflows/release.yml`, job `deb`, right after
**"Install both packages (postinstall runs openssl secrets + migrate)"**.

That job is already the right substrate and needs no new scaffolding: it
installs the real `.deb` and the step **"Prove the artifact reports the bare tag
version"** already drives the packaged CLI with

```
sudo grappa eval 'Application.load(:grappa); IO.puts(...)'
```

which is exactly the entry point the new transport goes through. Two sibling
steps assert the two branches.

## Assertions

1. **No live node** — with the service down, the CLI must exit non-zero (or
   print the parked marker, whichever the PR settled on), and the credential row
   must read `:parked`. Assert on a sentinel-prefixed stdout line, matching the
   convention the version step already uses, so a stray boot line cannot be
   mistaken for the result.
2. **Live node up** — with the service running, the same invocation must reach
   the live node and report the connected outcome.

## Reference measurement

From the FreeBSD bench jail on m42 (measured by the orchestrator, not by me):
`net_kernel` fails fast at **21 ms** with no live node, connects in **54 ms**
with one, and all three `erpc` calls were answered. Useful as an order-of-
magnitude sanity bound for a CI timeout — not as an assertion, since runner
timings differ.

## Declared limitation — FreeBSD is not coverable here

Every `runs-on:` in `ci.yml`, `integration.yml` and `release.yml` is
`ubuntu-latest`. GitHub Actions offers no native FreeBSD runner, so CI can cover
the Linux / `.deb` / Docker substrates and **not** the FreeBSD jail. The jail
stays covered by the manual probe only. This is accepted, not an oversight.

## Open question for whoever picks this up — not measured

The "live node up" branch needs the `grappa` service actually running inside the
`deb` job. It is not established here whether the postinstall leaves it up, or
whether the step must start it explicitly first. Verify before writing the
[...]

Issue #1685, opened 2026-08-23
release CLI: add-network records connected but does not spawn a live session


pminervini:
## Summary

On a running published-image deployment, the release CLI `add-network` command reports success and persists the credential as `connected`, but it does not spawn a session in the already-running node. The web client then sees a network that claims to be connected while no live session exists.

This is the same invariant violation described in #1163, but through the release CLI path rather than the admin-console path fixed there. #1163 explicitly requested reopening if the symptom remained after v1.0.0; I cannot reopen an issue filed by a maintainer, so this follow-up records the remaining path.

## Reproduction

Reproduced with release image v1.3.1 while validating #1683.

1. Start the published release image and create an account.
2. While the container is running, execute the release account verb (the entrypoint form below is the v1.3.1 workaround for #1683):

   ```sh
   docker exec -i <container> /app/release-entrypoint.sh \
     add-network alice azzurra \
     --server irc.azzurra.chat:6697 \
     --nick alice --auth none --services-flavor azzurra
   ```

3. The command exits successfully:

   ```text
   alice can now use azzurra (server irc.azzurra.chat:6697)
   ```

4. Log in as the account. `/networks` renders the binding, but there is no live session. A directory refresh returns HTTP 400 with `not_connected`, rendered as:

   ```text
   That network is not in a state to connect or disconnect right now.
   ```

5. Invoke the normal bootstrap routine on the live node:

   ```sh
   bin/grappa rpc "IO.inspect(Grappa.Bootstrap.run())"
   ```

   It reports `spawned: 1`, after which the same credential connects successfully. This demonstrates that the server, TLS configuration and credential are dialable; the missing step is adopting the new row in the live node.

## Expected

A successful `add-network` against a running release deployment must not leave `connection_state: connected` without a live session. It should either:

- spawn the session through the canonical runtime orchestration path; or
- return an honest non-live outcome/state with explicit operator instructions.

The UI should not receive a state that claims connected while every live-session operation returns `not_connected`.

## Relationships

- #1163 fixed the equivalent admin-console bind path and documents this exact connected-without-pid failure class.
- #1158 introduced the release-image account verbs.
- #1683 only restores first-boot secrets to later `docker exec` account verbs; this live-session defect is independent and should not expand that fix.

vjt:
Triaged. The report is accurate and the mechanism is confirmed in source; I did not reproduce it myself — you ran it, I read the code.

[...]

Issue #1675, opened 2026-08-22
connection_state reports 'connected' while the upstream link is failing: the row tracks the spawn, not the IRC registration


vjt:
`connection_state` says `connected` while the upstream link never registered: the row records that a session process was *started*, not that IRC came up. The UI reads the row, so a network that is failing every connect attempt is presented as connected.

Found on prod on 2026-08-22 while adding three new networks (ircnet / efnet / undernet). All three showed as connected in cicchetto; none of them ever completed registration. The endpoints were misconfigured — that part is on me and is now fixed — but the state reporting is the defect: the operator had no way to tell "connected" from "hammering a dead upstream in backoff".

## What happens

`PATCH /networks/:slug {connection_state: "connected"}` runs the post-U-0 order (`networks_controller.ex:395-419`):

1. `resolve_plan/3`
2. `NetworkSpawn.orchestrate/4` → `SpawnOrchestrator.spawn/4`
3. `Networks.connect/1` — writes `connection_state: :connected` **only on spawn success**

That ordering was the U-0 stop-swallow fix, and it correctly refuses to report success when the *spawn* is rejected (cf. #642). What it does not cover: the spawn succeeds, `Session.Server` and `Irc.Client` are alive, and the **upstream TCP/TLS/registration** then fails. The row was already committed to `:connected` at step 3; nothing walks it back when the client dies, and the reconnect backoff keeps a session process alive, so the row stays `:connected` indefinitely.

`Credential.connection_states/0` is `[:connected, :parked, :failed]`, so `:failed` exists and is unused for this class.

## Measured on prod

From `runtime/log/erlang.log.5` and `.6` (log timestamps are Europe/Rome; DB columns are UTC):

- **ircnet** — `irc.ircnet.com:6697` presents a certificate for `ircnet.tngnet.nl`. With `verify_peer` (#89): `{:bad_cert, {:hostname_check_failed, {:requested, "irc.ircnet.com"}, {:received, [dNSName: "ircnet.tngnet.nl"]}}}` → `tls_alert handshake_failure` → `session disconnected duration_ms=30075` → `session reconnect backoff delay_ms=5660 failure_count=1`.
- **efnet** — `irc.efnet.org` round-robin: first leaf `{:tls_alert, {:bad_certificate, "selfsigned_peer"}}`, the remaining four leaves all `:timeout`, one "upstream leaf connect failed — rotating to next leaf" every 30s.
[...]

PR #1740, opened 2026-08-24
ci(#1714): gate the release CLI's distribution hop on the real substrate

vjt:
Three arms at the tail of `release.yml`'s `deb` job hold the distribution
hop #1685 introduced. Refs #1714, #1685.

## How many arms, and the criterion

The issue asked for "both branches". `Grappa.Release.LiveNode.reach/2`
distinguishes **five** transport outcomes, so the count came from the code
instead:

| outcome | gated where | why |
|---|---|---|
| `:no_distribution` (no epmd) | **CI arm 1** | state of the deployment |
| `:unreachable` (epmd up, node down) | **CI arm 2** | state of the deployment |
| `{:ok, :started}` | **CI arm 3** | needs systemd + cookie + derived node name |
| `:no_release_node` | `live_node_test.exs:66` | refusal before any I/O, on one env var |
| `:distribution_disabled` | `live_node_test.exs:72` | refusal before any I/O, on one env var |

**Criterion: gate the outcomes the substrate owns; leave to the unit test
the ones a unit test already owns.** An arm for the bottom two would
re-read the same `System.get_env` through a slower door — CI minutes, no
extra falsifying power.

## Two places the spec was wrong, both measured

**"Right after Install both packages."** That runs the live arm — which
must START the service — before every existing proof in the job, leaving a
daemon nobody asked for running underneath them. A proof that runs with
unrequested state up no longer proves what it says it proves. The arms go
last among the proofs, still ahead of the uploads, so a failed arm keeps
the artifacts unpublished.

**"Needs no new scaffolding."** The postinstall `enable`s the unit and
deliberately does NOT start it (`postinstall.sh:68-74` — a real `PHX_HOST`
must be set first; the env file ships `REPLACE_ME`), and no workflow in
this repo had ever run `systemctl` or probed `/healthz`: zero hits across
`.github/workflows/*.yml`. The offline arms are free; the live arm is
genuinely new scaffolding.

## Two measurements that shaped the arms

**`systemctl stop` does not stage the second offline shape.**
`grappa.service` sets no `KillMode`, so systemd's default `control-group`
takes epmd down *with* the node and lands back in arm 1 — a green tick on
the wrong claim. Arm 2 starts `epmd -daemon` from the installed release's
own ERTS instead, outside the unit's cgroup, which is how epmd outlives
the service on a real host anyway.

**Each arm asserts its own premise** (`pgrep -x epmd` before arm 1 and
after arm 2, `/healthz` before arm 3). An arm that silently found itself
in a neighbour's state would otherwise report green for a claim it never
made.

## What is deliberately NOT asserted

- **The exit status.** `Grappa.Release.CLI` returns 0 in every branch on
  purpose (`cli.ex:263-268`), so `create-user && add-network && start the
  service` does not break on the case that has no live node by definition.
- **`connection_state = 'connected'`** on the live arm. Since #1675 that
  means *registered upstream*, and the arm points at `127.0.0.1:6667` with
  nothing listening, so the row settles to `:failing` on its own — pinning
  either value races. The arm asserts the row **left `:parked`**, which is
  what the hop does and what nothing on the CLI side can fake.

## Limits, stated rather than discovered later

- ~~`release.yml` fires on a tag, so these arms will not run on this PR.~~
  **No longer true.** A `deb_validation` dispatch now runs the deb job — and
  its arms — from a branch, with nothing published. They have been executed:
  [run 32706000296](https://github.com/vjt/grappa-irc/actions/runs/32706000296),
  all three green, `publish` skipped, no release touched. See the comment
  below for the output lines.
- GitHub offers no FreeBSD runner (every `runs-on:` here is
  `ubuntu-latest`), so the bastille jail keeps its manual probe. Accepted
  in the issue, restated in the docs.

## Verification

`shellcheck` clean on all four new `run` blocks; YAML parses and the steps
land in the intended order; `scripts/design-notes-gate.sh` green; full
`scripts/check.sh` reported below once it settles.

— w2

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -312,6 +312,159 @@ jobs:
           [ "$got" = "$want" ] || {
             echo "::error::client package is stamped '$got', expected '$want'"; exit 1; }
 
+      # ── #1714: the release CLI's distribution hop, on the real substrate ──
+      #
+      # #1685 taught `grappa add-network` to cross out of its transient `eval`
+      # VM and into the running bouncer as Erlang terms. Every branch of that
+      # hop was proven by hand exactly once, and the branch a self-hoster meets
+      # FIRST — no live node at all — had no gate.
+      #
+      # WHICH OUTCOMES GET AN ARM, AND THE CRITERION USED
+      #
+      # `Grappa.Release.LiveNode.reach/2` distinguishes five transport
+      # outcomes before the far side is even reached: `:no_release_node`,
+      # `:distribution_disabled`, `:no_distribution`, `:unreachable`, and the
+      # `{:ok, :started}` round trip. Three are gated below and two are not.
+      # The split is NOT the issue title's "both branches" — it is which
+      # outcomes only THIS substrate can falsify:
+      #
+      #   gated here    `:no_distribution`, `:unreachable`, `{:ok, :started}`
+      #                 — each is a STATE OF THE DEPLOYMENT (is epmd up? is the
+      #                 node up?) and turns on systemd, epmd, the release
+      #                 cookie and the node name ERTS derives. Nothing
+      #                 in-process can make any of them true or false.
+      #   left to unit  `:no_release_node`, `:distribution_disabled`
+      #                 — both are refusals taken BEFORE any I/O, on the value
+      #                 of one environment variable, and
+      #                 test/grappa/release/live_node_test.exs already covers
+      #                 them. An arm here would re-read the same
+      #                 `System.get_env` through a slower door: CI minutes, no
+      #                 extra falsifying power.
+      #
+      # WHY THE TAIL OF THE JOB AND NOT WHERE #1714 PUT IT
+      #
+      # The issue says "right after Install both packages". That would run the
+      # live arm — which must START the service — BEFORE every existing proof
+      # in this job, leaving a daemon nobody asked for running underneath them.
+      # A proof that runs with unrequested state up no longer proves what it
+      # says it proves. So the arms go last among the proofs, and still ahead
+      # of the uploads: a failed arm must keep the artifacts unpublished.
+      #
+      # NOT asserted: the exit status. `Grappa.Release.CLI` returns 0 in every
+      # branch ON PURPOSE (cli.ex — the first-run script `create-user &&
+      # add-network && start the service` must not break on the case that has
+      # no live node BY DEFINITION). The bind is reported through stdout and
+      # the row, so that is what these steps read.
+      - name: Create the account the distribution-hop arms bind against
+        run: |
+          sudo grappa create-user gate --password 'ci-not-a-secret' | tee /tmp/hop-user.log
+          grep -q '^created user gate ' /tmp/hop-user.log || {
+            echo "::error::create-user did not report the account it created"; exit 1; }
+
+      # ARM 1 — no epmd at all. This is a box that has never started the
+      # service, which is the FIRST of the two no-live-node shapes #1685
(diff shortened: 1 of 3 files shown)


vjt:
## The arms have been executed. Run: https://github.com/vjt/grappa-irc/actions/runs/32706000296

The earlier body said they would not run on this PR. That is no longer
true, and the body has been corrected rather than left standing.

**Why this changed.** A gate that cannot be executed is not a gate, and
these arms sit AHEAD of the uploads — so the first person to learn one was
broken would have been whoever cut the next release, holding a blocked cut
and nothing to explain it with. Writing the limit down honestly did not pay
for it.

**Why the existing dispatch was not the answer** (measured before building
anything): the repair path checks out the TAG's tree, the newest tag is
`v1.3.1`, and `git show v1.3.1:lib/grappa/release/live_node.ex` → absent. It
would have built a package with no transport in it and failed all three arms
for a reason that is not a defect. It also reaches `publish`, which runs even
on a red leg, and would have re-attached assets to a release that is already
out. That is not something to do for convenience.

**What was added:** `deb_validation`, the twin of the existing
`docker_validation` — a zero-publication dry run of the deb job *itself*,
from the dispatched ref. Five points, all inside `release.yml`. The deb job's
own `if:` needed none: that path sets `docker_validation=false`, which the
condition already admits.

### The run

`deb` **success**; `arch`, `rpm`, `docker`, `smoke` and `publish` all
**skipped**; `gh release list` unchanged, v1.3.1 still Latest.

```
Verify the tag …   deb-validation dispatch: no tag involved; version from VERSION (1.3.1)
                   RELEASE_TAG: w2-1714   DEB_VALIDATION: true

arm 1/3            epmd: not running (arm premise holds)
                   Protocol 'inet_tcp': register/listen error: econnrefused
                   gate can now use gate-no-epmd (server 127.0.0.1:6667)
                     the binding is PARKED: nothing is listening on epmd, so no bouncer is running here.
                   gate-no-epmd connection_state: 'parked'

arm 2/3            epmd: running, no node registered (arm premise holds)
                   gate can now use gate-no-node (server 127.0.0.1:6667)
                     the binding is PARKED: the live node did not answer (not running, or a different RELEASE_COOKIE).
                   gate-no-node connection_state: 'parked'

arm 3/3            service: healthy (arm premise holds)
                   gate can now use gate-live (server 127.0.0.1:6667)
                     started a session on the live node — nothing else to do
                   gate-live connection_state: 'connected'
```

Three things that run confirmed beyond the ticks:

1. **The second offline shape is real and distinct.** Arm 1 shows
   `net_kernel` failing outright (`register/listen error: econnrefused`);
   arm 2, with epmd up, gets all the way to a connect that answers no. Two
   different park reasons, as the code says.
2. **Arm 3 came back `connected`, not `failing`** — the row had not settled
   yet. That is exactly why the assertion accepts both: a slower runner sees
   the other value, and pinning either would have been a race dressed as a
   check.
3. **`publish` skipping is the load-bearing exclusion**, not an incidental
   one. `deb` runs on this path and *does* upload its artifact, so the
   "nothing downloaded" guard would not have caught a leak — that single
   condition is the whole distance between a validation run and branch-built
   packages on a real release.

The tag-push path is unchanged by construction: `github.event_name ==
'push'` is the first disjunct of every condition touched, so it
short-circuits before any new term is evaluated.

Docs corrected in the same push — both `OPERATIONS.md` and the DESIGN_NOTES
entry had stated the now-obsolete limit. The one limit that stands is
FreeBSD: no runner exists, so the jail keeps its manual probe.

— w2

Issue #1702, opened 2026-08-23
Radio player: no Media Session metadata — iOS lock screen shows only the app name


vjt:
On an iOS lock screen the playing radio shows up as "Cicchetto" with no station, no
track, no artwork — the OS knows an app is making noise and nothing else.

Measured on `origin/main` `f2b93f2b`: cicchetto never tells it. `navigator.mediaSession`
and `MediaMetadata` appear **nowhere** in `cicchetto/src` — zero hits across the tree.
The `<audio>` element in `AudioMiniPlayer.tsx` is mounted and played with no session
metadata attached, so the lock screen falls back to the app name from the manifest,
which is exactly what is on screen.

## What is missing

`navigator.mediaSession.metadata = new MediaMetadata({...})` set whenever
`activeAudio()` changes, with:

* `title` — the station name for radio, and for an upload the only thing we have is
  the link, so the honest value is the slug rather than an invented one;
* `artist` / `album` — empty until #1698 lands the now-playing fetch; once it does,
  `songs[0].title` / `.artist` from `api.somafm.com/songs/<id>.json` is the same data
  that feeds `/np`, so the two should share one source rather than fetch twice;
* `artwork` — the station logo, which is the same asset #1696 is fixing the extension
  of. Note `albumArt` in the SomaFM payload is frequently an empty string, so a
  fallback to the station logo is required, not optional.

`navigator.mediaSession.setActionHandler` for `play`/`pause` should be wired at the
same time, so the lock-screen transport drives the same element the in-app bar does
instead of whatever the OS guesses.

## Ordering

This is the surface that #1698 feeds. Landing #1698 first and then this one means the
metadata has something to say the day it is wired; landing this one first gives a
lock screen that finally names the station, with the track arriving later. Either
order works — they should not be done as one change.

Related: #1698 (now-playing + `/np`), #1696 (station logos), #682 (the player).

vjt:
Picked this up. Three things in the spec need correcting before code, all
measured on `origin/main` `fe5c5e6e` — recorded here so the record survives the
branch.

**The defect reproduces as written:** zero occurrences of `mediaSession` /
`MediaMetadata` across `cicchetto/src` **and** `cicchetto/e2e`.

## 1. The spec is stale — both prerequisites have landed

It says `artist`/`album` are "empty until #1698 lands the now-playing fetch" and
devotes an **Ordering** section to the choice. **#1698 and #1696 are both CLOSED
[...]

Issue #1698, opened 2026-08-23
Radio player: show now-playing metadata and add /np (gated on #1695)


vjt:
Two related wants for the radio player: show what is playing, and let the user say it
in channel with `/np`.

## Where the metadata comes from — measured

**ICY in-band metadata is not an option.** Requesting the stream with
`Icy-MetaData: 1` returns `icy-name`, `icy-genre`, `icy-br` and friends, but **no
`icy-metaint`** — so there is no in-band title track to read. And even where it were
offered, a browser `<audio>` element does not surface ICY metadata to the page.

The workable source is SomaFM's per-channel song feed:

```
https://api.somafm.com/songs/<id>.json
```

`songs[0]` is the current track, newest first, each entry carrying `title`,
`artist`, `album`, `albumArt` and a `date` epoch:

```json
{ "title": "A Land Unknown", "artist": "Trestal",
  "album": "Ceramic Illumination", "albumArt": "", "date": "1787500712" }
```

Measured: it answers an identical 200 (2,447 bytes for `groovesalad`) from both
`api.somafm.com` and the bare domain. **Use `api.somafm.com`** — see the CSP note
below.

## Dependency

This is a `fetch`, so it hits `connect-src` and is **blocked today**. It needs
#1695, and it must address `api.somafm.com`: the wildcard granted there does not
cover the bare `somafm.com`.

`albumArt` is frequently an empty string in the feed, so the display needs a
sensible fallback rather than a broken image.

## 1. Show the metadata in the player

Artist and title of the current track, in the player surface. Needs a refresh
cadence — the feed is polled, not pushed. Keep it modest and stop polling when
nothing is playing.

## 2. `/np`

A slash command that emits a `/me`-style action with the current track, e.g.

```
* nick is now playing: Trestal — A Land Unknown [Groove Salad]
```

Commands live in `src/lib/slashCommands.ts`; this should be registered there like
any other rather than special-cased in the composer.

Open questions worth settling before building:

- What `/np` does when nothing is playing, or when the feed has not answered yet —
  silence, an error to the user, or an action naming just the station.
- Whether the station name belongs in the line (proposed above) or only the track.

## Verification

- With #1695 in, confirm the fetch to `api.somafm.com` succeeds with no CSP violation.
- Confirm the displayed track matches `songs[0]` of the tuned channel.
- Confirm `/np` sends an ACTION, not a PRIVMSG, into the current channel.
- Confirm polling stops when playback stops.

[...]

Issue #1696, opened 2026-08-23
Radio player: 10 of 14 station logos 404 — we hardcode .png where upstream serves .jpg


vjt:
Several stations in the radio player show a broken logo (Drone Zone among them).
This is ours, not an upstream gap: the logos exist, we are asking for the wrong
filename.

## Measured

`radioStations.ts` hardcodes every logo as `<id>120.png`:

```
logoUrl: "https://api.somafm.com/logos/120/dronezone120.png"
```

Fetching all 14 hardcoded logo URLs just now: **10 return 404, 4 return 200.**

The four that work (`groovesalad`, `defcon`, `reggae`, `u80s`) are the ones whose
120px asset genuinely is a PNG. The ten that 404 are served as **JPEG** upstream —
`channels.json` gives, for example:

| station | what we request | what upstream serves (`image`) |
|---|---|---|
| `dronezone` | `.../120/dronezone120.png` | `.../120/dronezone120.jpg?v=1781033109` |
| `lush` | `.../120/lush120.png` | `.../120/lush120.jpg?v=1674955397` |
| `spacestation` | `.../120/spacestation120.png` | `.../120/spacestation120.jpg?v=1781038488` |

So the extension is per-station, and the URL also carries a `?v=` cache-buster we
drop. The comment at `radioStations.ts:26` claims the versionless logo URLs were
checked — the measurement above contradicts it for 10 of 14.

## Fix

The correct fix is to stop deriving the URL by convention and read `image` (120px),
`largeimage` (256px) or `xlimage` (512px) straight from `channels.json`. That is
gated on #1695, which is what makes `channels.json` fetchable at all.

Until #1695 lands, the stopgap is to correct the ten hardcoded extensions to `.jpg`
so the curated list renders. Note this is a stopgap by nature: the extension is
upstream's choice and can change again, which is precisely the argument for reading
the catalogue instead of guessing.

`img-src 'self' data: https:` already admits all of these, so no CSP change is needed
for the logos themselves.

## Verification

- Fetch every logo URL the client will use and confirm each returns 200.
- Confirm Drone Zone, Lush and Space Station render a logo in the picker.

vjt:
Landed on `main` in PR #1719 (`2171d707`).

[...]

PR #1742, opened 2026-08-24
feat(#1702): Media Session metadata — the lock screen learns the track, station and artwork

vjt:
⚠️ **DRAFT, and honestly so: the e2e spec RAN and lands partly red.** (a)–(c)
pass on the defect's own platform; (d) fails on this spec's instrumentation,
not on production. Details below — nothing here is claimed green that is not.

On an iOS lock screen the radio showed "Cicchetto" and nothing else. Measured:
`mediaSession` / `MediaMetadata` appeared **nowhere** in `cicchetto/src`.

## What lands

| commit | |
|---|---|
| `5d58def0` | 16 unit cases, RED-first, over a module that did not exist |
| `5833b2ce` | `lib/mediaSession.ts` — the projection |
| `12373bc8` | `AudioMiniPlayer` — play/pause split out of the toggle, three effects |
| `b0865589` | the `@webkit` e2e spec |

Unit suite: **6076/6076 green**. cic gates (biome + tsc src + tsc e2e + lock
drift): **4/4 green**.

## Three spec corrections, agreed on the issue before building

1. **The spec was stale.** It gates `artist`/`album` on "#1698 landing" and
   offers an ordering choice. **#1698 and #1696 are both CLOSED and in the
   tree**, so the metadata is complete on day one and no title-only interim was
   built.
2. **Taken literally, the field mapping buries the track.** The issue asks for
   `title` = station with the track in `artist`/`album`; on iOS `title` is the
   primary line, so that leaves half the original complaint ("no station, **no
   track**") standing. Shipped instead with the platform convention — `title` =
   track, `artist` = artist, `album` = station — which is **already this repo's
   spelling** in `nowPlayingLine`: `<artist> — <title> [<station>]`.
3. **The `albumArt` fallback has nothing to fall back from.** #1698's
   `parseSongsFeed` returns only `{artist, title}`; the artwork is just
   `station.logoUrl`. What DOES survive from #1696: the table is 10 `.jpg` and 4
   `.png`, so the mime is **read off the URL**, never hardcoded.

## e2e: run, and what it proved

Eight runs on `webkit-iphone-15` (WebKit + iPhone 15 profile, `@webkit`), not
the default chromium project — the bug is an iOS lock screen and a chromium
green would be a green off the defect's platform.

**Green:** metadata exists at all; the field mapping above; `image/png` read off
a `.png` station's logo URL; the station's exact feed URL requested.

**Red — (d), the action handlers.** `calls=[none] spyStillInstalled=false
mediaSessionPresent=true`. The spy this spec installs on
`navigator.mediaSession.setActionHandler` is gone by read time, on a document
where the init script demonstrably ran. Two hypotheses already killed (plain
assignment, then `Object.defineProperty`). Leading unverified hypothesis:
`navigator.mediaSession` returns a different object per access. **Production's
action handlers are therefore UNPROVEN, not known broken** — that distinction is
the point of leaving this visible rather than deleting the assertion.

## Findings worth more than this issue

🔴 **`page.route` does not intercept a third-party `fetch()` once the service
worker has claimed the page.** Measured: the first runs reached the real
api.somafm.com and returned a track SomaFM was actually playing. cic ships a
Workbox SW; Playwright does not intercept through one. It is a **race**, which
is why `issue1695-…` (which touches a third party early) is fine and this one —
after a login, a channel select and a rail interaction — was not. Cure used:
stub `window.fetch` above the worker. `serviceWorkers: "block"` was tried and
rejected — cic then raises a registration-failed banner that blocks the rail tap
and would mean testing a degraded app.

## Not asserted, by platform limit

That iOS actually renders what WebKit was handed is not observable from any
browser automation. That stays a device check, and is not claimed here.

## Left for the next session

Resolve (d), then undraft. `docs/DESIGN_NOTES.md` entry still to write
(marker `#1702`, verified free against the tip at time of writing).

— w1

### DIFF
--- a/cicchetto/e2e/tests/issue1702-media-session-metadata.spec.ts
+++ b/cicchetto/e2e/tests/issue1702-media-session-metadata.spec.ts
@@ -1,0 +1,287 @@
+// #1702 — Media Session metadata: the OS is told what is playing.
+//
+// THE DEFECT: on an iOS lock screen the radio showed "Cicchetto" and nothing
+// else — no station, no track, no artwork. `navigator.mediaSession` was never
+// set anywhere in cic, so the OS fell back to the manifest's app name.
+//
+// 🔴 WHY THIS SPEC IS `@webkit` AND NOT THE DEFAULT CHROMIUM PROJECT. The bug
+// is an iOS lock screen. A green on desktop Chrome would be a green OFF the
+// defect's platform, which proves the code ran somewhere rather than that it
+// works where it failed. The suite already has the right project —
+// `webkit-iphone-15`, WebKit on an iPhone 15 device profile — and Playwright
+// 1.59.1's WebKit was measured to implement the whole API before this spec was
+// written: `navigator.mediaSession`, the `MediaMetadata` constructor,
+// `setActionHandler`, `playbackState`, and read-back of every metadata field.
+// So the platform gap is closed, not papered over.
+//
+// WHAT REMAINS OUT OF REACH, stated rather than hidden: that iOS actually
+// RENDERS on its lock screen what WebKit was handed is not observable from any
+// browser automation. That is a device check and it stays one. What this spec
+// does buy is the same engine, an iPhone device profile, and an assertion on
+// the exact API the OS reads — which is the whole of what cic controls.
+//
+// WHAT IT ASSERTS, and why each is an outcome rather than a mirror:
+//   (a) metadata EXISTS at all — the regression guard for the defect itself,
+//       which was precisely that it was always null;
+//   (b) the FIELD MAPPING: `title` is the TRACK, `artist` its artist, `album`
+//       the STATION. The issue as filed asked for `title` = station with the
+//       track fed into `artist`/`album`; taken literally that buries the track
+//       on the primary line, leaving half the original complaint standing. The
+//       platform convention won, and it is already this repo's spelling in
+//       `nowPlayingLine`. This assertion is what pins that decision;
+//   (c) the artwork mime is READ off the logo URL — `groovesalad` is a `.png`
+//       and 10 of the table's 14 rows are `.jpg`, so a hardcoded mime would
+//       ship wrong for either group. #1696 is the issue that lesson came from;
+//   (d) the lock-screen transport drives THE SAME element the in-app bar does.
+//       Asserted by capturing the registered handlers and invoking `pause`,
+//       then checking the element actually stopped — a registration count
+//       alone would pass with the handlers pointed at nothing.
+//
+// 🔴 NO THIRD-PARTY NETWORK, AND `page.route` IS NOT ENOUGH FOR THE FEED.
+// Measured, not assumed: with the feed on `page.route` this spec reached the
+// REAL api.somafm.com and the lock screen came back naming "Kaya Project —
+// Desert Phase (Hibernation Remix)" instead of the canned track below.
+//
+// cic ships a Workbox service worker (`VitePWA`, `registerType: "autoUpdate"`),
+// and Playwright does not intercept requests that pass through a service
+// worker — only ones the page issues directly. The `<audio>` stream is a media
+// load that does not go through it, which is why #682's stream intercept works
+// and hides the problem; the now-playing `fetch()` does, and escaped. It is
+// also a RACE rather than a constant: the worker only intercepts once it has
+// CLAIMED the page, so a spec touching a third party early
+// (`issue1695-somafm-connect-src-perimeter`) is intercepted fine, while this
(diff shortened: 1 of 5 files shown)


vjt:
**Measured, and it upgrades this from a loose end to a blocker:** the `@webkit` spec does run in
CI, and assertion (d) reddens **shard 4/4** — so `integration` is RED on this PR.

Recording it here because the cost is not confined to #1742. If this landed red it would redden
`integration` on every unrelated PR, which is the exact bill #1715 already made this project pay
three times: a shared job people learn to read past, until it hides a real regression. **Staying
draft until (d) is settled.**

To be precise about what is and is not known: **(d) failing does NOT mean the action handlers are
broken.** The last diagnostic reads `calls=[none] spyStillInstalled=false
mediaSessionPresent=true` — the spy this spec installs on
`navigator.mediaSession.setActionHandler` is gone by read time, on a document where the init
script demonstrably ran. So production's `setMediaSessionHandlers` is **unproven, not known
broken**, and nothing in `lib/mediaSession.ts` should be changed before that is measured.

Two hypotheses already killed: a plain assignment dropped in sloppy mode, and its replacement
with `Object.defineProperty` — the log came back `(never called)` both times. The leading
remaining idea is that `navigator.mediaSession` may hand back a different object per access, so
the patch lands on one instance and the app calls another. **That is NOT verified** and is a lead
only.

One caveat worth writing down before someone leans on it: the evidence that effect #3 runs is
weak. WebKit maintains `playbackState` on its own, so the `"playing"` assertion may have passed
without our code. The only effect provably running is the metadata one.

— w1

vjt:
## Ready for review — (d) is green, and the red was the instrument

### The blocker, resolved

Assertion (d) failed eight runs with `calls=[none] spyStillInstalled=false mediaSessionPresent=true`. It was the spec's own spy, not `setMediaSessionHandlers`.

`navigator.mediaSession` is `[SameObject]`, and JSC honours that by **caching** a wrapper rather than pinning one. `metadata` and `playbackState` are state on the platform object underneath and outlive the wrapper; a JS own property has nowhere to live but the wrapper, so once nothing in JS references it the wrapper is collected and the property goes with it.

Two-arm probe, same WebKit build and iPhone 15 profile, ~250 MB of allocation churn between install and read-back:

| arm | spy | own property | `metadata` | calls seen |
|---|---|---|---|---|
| retain nothing | `false` | `false` | `"PIN"` | `[]` |
| retain the object | `true` | `true` | `"PIN"` | `["play:fn"]` |

The first row reproduces the failure exactly, and the asymmetry is the whole diagnosis: it is why (a)–(c) passed throughout while (d) did not. The fix is one line — hold the object in a global from the init script. The assertion was **not** deleted or weakened.

It also explains why the two earlier hypotheses died identically: neither a plain assignment nor an `Object.defineProperty` survives losing the object it was written on. The same init script's stub on `window.fetch` never had the problem, because `window` is not collectible.

Full rationale, the falsified alternatives, and the two limits that stay stated rather than hidden (WebKit maintains `playbackState` itself; production handlers surviving a collected wrapper is *inferred*, not measured) are in the DESIGN_NOTES entry and on the issue.

### Gates

| gate | result |
|---|---|
| cic unit | 6076 / 6076 |
| cic check (lock drift · biome · tsc src · tsc e2e) | 4 / 4, re-run after the rebase |
| `design-notes-gate.sh` | green, marker `<!-- entry #1702 -->` unique against `origin/main`'s current tip |
| `union-rebase.sh` onto `a108389d` | green — `+139 -0`, contribution intact, boundary shape read on the real file |
| **full `scripts/integration.sh`** | **768 passed, 2 failed (30.4 m)** — and `issue1702-media-session-metadata.spec.ts` is **green on `webkit-iphone-15`** |

### The two reds are foreign, and they are one class

Stated plainly rather than waved through. Neither is in this branch's blast radius (both `@chromium`; this branch adds a `@webkit` spec, a lib, a component's effects and a docs entry), and both are a **`loginAs` barrier timing out under full-suite load**:

- `issue168-scroll-authority:140` — `.sidebar-network-header` not visible within the 10 s login-ready barrier (`element(s) not found`). Tracked class: #552, #1080, #1078, epic #653.
- `push-trigger-channel-mention:56` — `waitForUserTopicReady` not stamped within 5000 ms, with socket health `{"state":"connecting","errorCount":0}`. Tracked class: **#1579**, which is verbatim this barrier, this timeout, "3 occurrences, 3 different specs". This is a 4th spec.

Supporting evidence that neither is broken on main: the last eight `integration` runs on `main` are all `success`, three of them today. Every tracking issue above is **closed**, so these are fresh sightings of a closed class — handing that call to the orchestrator rather than re-filing.

— w1

Issue #1584, opened 2026-08-19
e2e: a spec that is first-in-project runs cold — its green is not evidence, and any earlier spec silences it


vjt:
## The fact

A spec whose failure needs the **cold first-test-of-the-project** boot goes
**green** the moment any other spec that sorts before it joins the same
Playwright project. Nothing in the report says so. The reader sees a green
tick and has no way to tell it apart from a green that means the contract
holds.

Measured on `issue310-scroll-to-bottom-btn-cursor.spec.ts` while diagnosing
#1543 (that issue carries the full measurement and the mechanism):

| body | position in the `webkit-iphone-15` project | result |
|---|---|---|
| the spec as written | first `@webkit` spec collected | RED 4/4 |
| a bare open, no assertions, no tap | first `@webkit` spec collected | RED |
| the same bare open | after that spec | green |
| the spec's own pre-tap body | after that spec | green |

And the swap, inside a single run: inserting **one** `@webkit` spec before it
turned `#310 … @webkit tapping the button advances the cursor…` **green**
(2.7 s) and moved the red onto the newly-first spec. In the **full**
`scripts/integration.sh` (759 tests, 28.5 min) that spec **passes** — it sits
behind dozens of earlier `@webkit` files.

## Why it is a problem beyond that one spec

The e2e suite partitions across two projects (`chromium`, `webkit-iphone-15`),
`workers: 1`, `fullyParallel: false`. Exactly one test per project pays the cold
browser launch and the cold bundle/SW load; every later test in that project
runs warm. So **"is this test first in its project?" is an input to the test**,
it is decided by filename sort order plus which files the invocation collected,
and it is invisible in the result.

Consequences, in the order they bite:

1. A green from such a spec is **not evidence** that the behaviour it names is
   correct — it can be evidence only that something else was collected first.
   `feedback_landed_claim_evidence` asks for a measured green; this is a class
   of green that cannot carry that weight and does not announce itself.
2. The condition is **silently changed by unrelated work**: adding any
   `@webkit` spec whose filename sorts earlier flips it, with no edit to the
   spec and no line in any diff pointing at it.
3. It cuts the other way too, and that is how #1543 was born: a hand-picked
   `--grep`/file subset can *promote* a spec to first-in-project and make it
   red, which reads as a regression in the code under test rather than as a
   property of the subset. `docs/TESTING.md`'s cascade-vs-flake tree does not
   cover this case — the iso-rerun it prescribes reproduces the red, because an
[...]

Issue #682, opened 2026-08-02
cic: internet radio player — station picker in the RailActions drawer, player embedded in the right rail


vjt:
## Idea

Requested by Wollino on #grappa (KiwiIRC has one), kept alive by vjt: an internet-radio
player inside cicchetto. **Shape decided in channel by vjt (2026-08-23): the station picker
is reachable from the RailActions drawer, and the player is embedded in the right rail** —
not a Settings menu, not a new modal.

Reference implementation looked at (CosmicIRC's Kiwi bolt-on, `cosmicirc.com/chat/`): worth
nothing as code — it is a `<div>` injected next to Kiwi's DOM that finds its anchor by
searching every `button, a, div` for the visible text "Add Network" and re-positions itself
from a `MutationObserver` plus two 1-second `setInterval`s. Its only reusable part is the
station list, and even that is better taken from source (below).

## Most of the machinery already exists (#115)

Measured on `origin/main` `390bd56e`:

- `cicchetto/src/lib/audioPlayer.ts` — module-scope, identity-scoped store holding one
  `{ href }`. `playAudio(href)` starts or **swaps** the source: ONE player instance, not N.
  Identity change / logout clears it.
- `cicchetto/src/AudioMiniPlayer.tsx` — the docked transport bar pinned above the compose
  box, mounted at Shell root (`Shell.tsx:710`), inside `.shell-main` and therefore **outside**
  the rail. Explicitly non-modal and persistent: switching the active channel does not kill
  playback. The `<audio>` element is mounted unconditionally; only the chrome is gated.

So "keep it playing while I move around the app" is **already solved**. A station is just
another `href` handed to `playAudio`. That is what makes this cheap.

## Where the new shape lands, and what it costs

The right rail is `.shell-members`, PERMANENT since #71 INC-2 (R1) and mounted by BOTH
branches of Shell's `isMobile()` split. Its stack, top to bottom (`Shell.tsx:765-787`):
`MembersPane` (channel-only) → `RailContext` (per-window-kind context) → `RailActions`
(floored by `margin-top: auto`).

Three constraints fall out of that, and none of them is cosmetic:

1. **The radio panel is NOT another `RailContext` arm.** `RailContext` grafts content *by
   the active window's kind* and renders nothing for kinds it has no arm for. A radio that
   disappears when you switch to a query is the exact opposite of what the docked player
   already guarantees. It has to be an unconditionally-mounted sibling inside
[...]

PR #1757, opened 2026-08-24
#1701: move the docked player below the compose box

vjt:
The repositioning half of #1701, on vjt's ruling relayed from IRC (2026-08-24):
*"lasciamo il player sotto ma mettiamolo SOTTO la compose box — tra compose box e
bottom bar"*. The re-pin half of that issue landed separately in #1732 and is not
touched here.

It is **none of the three options** that had been argued. The debate had narrowed
to who wins the bottom corner on a mobile query window — the ☰ float or the
docked bar's controls — and the ruling dissolves the question by moving the bar
out of the corner. Consequences, all of them good:

- the ☰ is never covered, so the premise of **#1051's ruling is not contested and
  it stands as written** — no reopening;
- the four e2e specs that pin ☰ on query windows (`issue1039-hamburger-corner`,
  `issue1051-card-close-above-float` ×2, `issue985-mobile-floating-opener`) are
  **untouched** — nothing here is hidden or gated by window kind;
- the bar's ✕, the only STOP verb reachable from a phone while the rail is a
  drawer off-screen (#682), **stays in the main view**.

The superseded local commit that implemented option ② (dock under the top bar,
yield the corner) was **discarded, not rebased** — it never left this host.

## The measurement the ruling asked for

The ruling named one thing as not-yet-measured: whether the placement fights the
on-screen keyboard (#209) or the scroll-to-bottom / next-window buttons.
**Neither.**

**The floats.** Both live in ONE container — `.scrollback-float-stack`,
`position: absolute; bottom: 0.75rem` — inside `.scrollback-pane`. That is
**#280's root-cause fix**, and its own comment states the property that answers
this: anchoring to the message container rather than the viewport keeps their
position *"CONSTANT relative to the container regardless of keyboard state"*.
The reorder cannot move that box — the pane is the `flex: 1` item of a column
whose other children take natural height, so swapping two natural-height siblings
leaves the grown item's border box identical.

Falsifiers checked rather than assumed: no sibling combinator / `:nth-*` /
`:last-child` in the sheet touching `.audio-mini-player` or `.compose-box`;
`order:` appears **zero** times in the entire stylesheet; no TS reads DOM order
near these; one stylesheet only.

**The keyboard.** `.shell-mobile` is `height: var(--viewport-height)`, written
from `visualViewport.height`. The keyboard shrinks the var and the pane absorbs
it through a `min-height: 0` chain. The player stays inside that same chain, so
the fixed pixels below the pane are unchanged — only their order is.

## What is pinned, and what is only declared

The one real delta: while the radio plays, the float pair used to be separated
from the send button by the bar's own row, and now it is adjacent — which is
already the shipped geometry whenever the radio is **not** playing.

That convergence was at first left as an argument, and **#278 cannot be its
witness: it runs with no audio**, so a player that began displacing that geometry
again would sail past it green. Rather than clone #278 with the radio on —
repeating its overlap constants buys no coverage and creates a second place for
them to drift — the **identity** is asserted:

> the floats are anchored inside the pane, so the pane's bottom edge is their
> frame of reference, and the send button rides the compose box; therefore if
> mounting the bar leaves the **pane→compose gap unchanged**, the with-audio
> float-vs-send relationship is *by construction* the one #278 already pins.

Measured before the link click and again with the bar up, inside
`media-link-modal-viewer.spec.ts` (which already has both on screen). Polled, not
read once — the bar changes height a second time when `onLoadedMetadata` swaps
the seek slider for the live badge. The failure signal is the bar's own height
against a 1px tolerance: **≈47px, derived from the sheet** (`min-height: 2.5rem`
controls at `1rem = 14px`, plus `0.8rem` vertical padding and a `1px` border) —
so the threshold is not a tight one.

⚠️ **Declared, not pinned.** #209 says it itself — *"soft keyboard physics don't
reproduce under chromium e2e"* — and #278/#280 accordingly drive the app's own
focus-derived signal (`.shell-mobile:has(textarea:focus)`), not a real keyboard.
What CI holds is that branch. **The device leg is vjt's**, and no chromium green
here should be read as a claim about an iPhone.

## Shape: last child of the zone, not a sibling of the main section

`<BottomBar />` is a sibling of `.shell-main` while the player lives inside it, so
the ruling's words admit two implementations. The bar became the **last child of
`DropUploadZone`**, after `<ComposeBox />`. They render identically where the
player mounts; hoisting it out would spend two things nobody asked for — #351's
whole-pane drop target would stop covering the transport strip, and the bar would
begin mounting on *every* mobile window kind instead of the scrollback kinds it
mounts on today, a behaviour change disguised as a reposition.

`border-top` needed **no flip**, unlike the discarded option-② commit: the edge
it separates on is still its top one; the neighbour above it is simply the
compose box now.

## The stale comment, and its class

`.audio-mini-player`'s header justified an **absent** declaration by naming a
**neighbour** — *"Sits above compose (and above compose's own safe-area
handling), so it needs no bottom inset."* The conclusion survived the move and
the reason did not: the **shell** owns that inset (base `.shell` carries
`env(safe-area-inset-bottom)`, `.shell-mobile` deliberately zeroes it per #1127).

Load-bearing rather than pedantic, because #1127 measured the failure: with the
keyboard up iOS reports `env(safe-area-inset-bottom)` against the DEVICE, not the
shrunken visual viewport, so an inset on this edge double-counts against
`--viewport-height` and reopens a ~34px gap. The bar is now the bottom-most row
of that column, so the warning is recorded on its rule.

**The general shape, worth more than the instance:** a comment that justifies an
ABSENT declaration by naming a NEIGHBOUR is falsified silently by a reorder — the
declaration it describes never changes, and no gate reads prose. Two more said
"above compose" and were corrected in the same commit.

## Tests

The unit pin asserts a **pair**, because either half alone passes a wrong shape:
document order alone ("later in the tree") is satisfied by a player hoisted below
the bottom bar; shared parent alone ("same column") is satisfied by the pre-move
mount above compose. Together they say "inside the compose box's own flex column,
after it". jsdom computes no layout, so the step that turns document order into
VISUAL order — the column is `flex-direction: column`, and `order` appears
nowhere in the sheet — is read off the CSS text, the technique
`audioMiniPlayerLayout` already uses on this bar.

Measured RED before the move on the order assertion, both form factors, with the
parent assertion already green.

## Gates

| gate | result |
|---|---|
| `bun.sh run check` — lock drift, biome (src+e2e), tsc (src), tsc (e2e) | 4 stages, 0 failed |
| `bun.sh run test --run` — full vitest | 313 files, 6145 tests green |
| `design-notes-gate.sh` | green — 1 entry, separator + marker `#1701b` |
| e2e — new identity assertion | **RED→GREEN, both demonstrated** |
| e2e — `issue278` + `issue280` (regression witnesses) | 4 passed, webkit-iphone-15 |
| `scripts/integration.sh` run 1 | 770 passed, 1 failed — `issue981` |
| `scripts/integration.sh` run 2 | 770 passed, 1 failed — `issue152`; **`issue981` green** |

The RED was constructed, not assumed: restoring the pre-move `Shell.tsx` from the
base makes the identity assertion fail with `Expected: < 1 / Received: 47.1875` —
the bar's own height, matching the ~47.2px derived from the stylesheet beforehand.

⚠️ **The full suite has not gone green, and neither red is this branch's.**

`issue981-no-badge-at-tail` is the one that needed discriminating, because it has
a plausible link to this change (a badge on a window read at the TAIL, while the
bar moves toward the tail). It was measured, not waved off:

- the failure is in the **setup**, not the badge — `.sidebar-network-header` never
  found, 10s timeout, sidebar showing only Home;
- **no causal path** — the branch touches neither `Sidebar.tsx` nor
  `lib/networks.ts`, which own that element;
- **not deterministic** — same branch, same suite, same roster position 458:
  ✘ 33.6s in run 1, ✓ 4.0s in run 2;
- **3/3 green in isolation** on this branch *and* on a detached control worktree
  at clean `31f9502f`;
- run 2's single red was a **different** spec (`issue152-ident-realname:37`),
  itself 3/3 green in isolation.

By `docs/TESTING.md`'s decision tree that is branch 1 — CASCADE, *"filed as a
follow-up bucket. DO NOT change production code."* The rotating single red across
two runs is that section's own listed cascade signature. The #1584 cold-start
exception was checked and does not apply (`issue981` was not a cold seat).

**Attribution, stated precisely because it is easy to garble:** BOTH reds are on
this branch — run 1 red `issue981`, run 2 red `issue152`. The only clean-base
measurement taken was a detached control worktree at `31f9502f` running
`--grep "#981" --repeat-each 3` (3 passed); `issue152` was not in that roster, so
**no clean-base full run was performed** and none should be inferred.

**What was NOT done, declared rather than left silent:** the targeted 5×5
repetition of `issue981` (five iso runs each side) was not run. It was judged the
weaker instrument here and the reason is structural — isolation removes the
upstream poisoner *by construction*, so a cascade returns 0/5 on both sides and
the comparison discriminates nothing. The evidence relied on instead is the
same-load, same-position full-suite flip (✘ 33.6s → ✓ 4.0s at roster position
458) plus the absent causal path. Anyone who wants the 5×5 anyway should know it
is expected to come back green on both sides regardless of cause.

What is **not** established at all is which upstream spec is the poisoner. That
is a bisect, and by the runbook's own instruction it belongs in its own bucket
rather than in this slice. Observed environmental rate: ~1 red per 770 with a
rotating target, measured over two full local runs.

**Local e2e informs; it does not decide.** The gate that matters is this PR's CI,
on different hardware under a different load. If CI reproduces either red, it
should be triaged there against real data rather than against a single local run.

— w2

### DIFF
--- a/cicchetto/e2e/tests/media-link-modal-viewer.spec.ts
+++ b/cicchetto/e2e/tests/media-link-modal-viewer.spec.ts
@@ -102,12 +102,50 @@ test("🎵 upload link click opens the docked mini-player, NOT the modal (GH #11
   const { row, link } = await mediaScrollbackRow(page, "🎵", slug);
   await expect(link).toHaveClass(/scrollback-media-link/);
 
+  // #1701 — the bar now docks BELOW the compose box, and what makes that safe
+  // for the float stack (scroll-to-bottom + next-active, anchored `bottom:
+  // 0.75rem` INSIDE `.scrollback-pane`) is that mounting it no longer comes
+  // between the pane and the compose box: the pane's bottom edge, which is the
+  // floats' frame of reference, keeps the same relationship to the send button's
+  // row whether or not audio is playing.
+  //
+  // `issue278-next-active-send-overlap` pins the float-vs-send geometry, but it
+  // runs with NO audio — so a player that started displacing that geometry again
+  // would sail past it unseen. This is the witness for the case #278 cannot see,
+  // and it asserts the IDENTITY rather than the numbers: repeating #278's
+  // overlap arithmetic with the bar up would buy no coverage and cost a second
+  // place to maintain the same constants.
+  //
+  // Form-factor independent on purpose — both Shell branches mount the same
+  // `.drop-upload-zone` column, so this holds wherever the spec's project runs.
+  const paneToComposeGap = async (): Promise<number> => {
+    const pane = await page.locator(".scrollback-pane").boundingBox();
+    const compose = await page.locator(".compose-box").boundingBox();
+    if (pane === null || compose === null) {
+      throw new Error("pane and compose box must both be laid out to compare the gap");
+    }
+    return compose.y - (pane.y + pane.height);
+  };
+  const gapWithoutBar = await paneToComposeGap();
+
   const cicUrl = page.url();
   await link.click();
 
   // The docked bar appears; the media viewer modal stays closed.
   const player = page.getByTestId("audio-mini-player");
   await expect(player).toBeVisible({ timeout: 5_000 });
+
+  // …and it did NOT insert itself between the pane and the compose box. Polled,
+  // because the bar changes height a second time when `onLoadedMetadata`
+  // resolves and swaps the seek slider for the live badge — measuring once on
+  // first paint would read a layout that is still settling.
+  await expect
+    .poll(async () => Math.abs((await paneToComposeGap()) - gapWithoutBar), {
+      message:
+        "#1701: mounting the docked bar must not change the pane→compose gap — a bar that lands between them moves the float stack's frame relative to the send button, which #278 cannot see because it runs with no audio",
+      timeout: 5_000,
+    })
+    .toBeLessThan(1);
   await expect(mediaViewer(page)).toBeHidden();
   expect(page.url()).toBe(cicUrl);
 
--- a/cicchetto/src/AudioMiniPlayer.tsx
+++ b/cicchetto/src/AudioMiniPlayer.tsx
@@ -19,8 +19,9 @@ import {
 } from "./lib/mediaSession";
 import { nowPlayingLabel } from "./lib/nowPlaying";
 
-// Docked audio mini-player (GH #115) — a slim transport bar pinned above
-// the compose box. Non-modal: scrollback stays scrollable + readable
+// Docked audio mini-player (GH #115) — a slim transport bar pinned BELOW
+// the compose box (#1701: between it and the mobile bottom bar; it used to
+// sit above compose). Non-modal: scrollback stays scrollable + readable
 // while audio plays (CLAUDE.md "IRC stays text only" — audio routes here
 // instead of MediaViewerModal). Persistent: switching the active channel
 // doesn't kill playback; a new audio link swaps the source on the single
--- a/cicchetto/src/DropUploadZone.tsx
+++ b/cicchetto/src/DropUploadZone.tsx
@@ -13,8 +13,10 @@ import { dragHasFiles, dropUpload } from "./lib/dropUpload";
 // Layout: `.drop-upload-zone` is a transparent pass-through flex column
 // (`flex: 1`) occupying the exact slot the stack used to fill directly, so
 // `.scrollback-pane`'s `flex: 1` still grows within it and TopicBar /
-// AudioMiniPlayer / ComposeBox keep their natural heights. `position:
-// relative` anchors the overlay.
+// ComposeBox / AudioMiniPlayer keep their natural heights (#1701 put the
+// player last, below compose — the zone still wraps it, so a drop on the
+// transport strip uploads like a drop anywhere else in the pane).
+// `position: relative` anchors the overlay.
 //
 // Drag-depth counter: dragenter/dragleave fire once per child element the
 // cursor crosses, so a naive boolean flickers the overlay off every time the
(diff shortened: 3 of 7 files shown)


vjt:
## Rebased onto `713736f9`, and one caveat on the numbers above

The branch was rebased onto current main via `scripts/union-rebase.sh` (main had
moved 4 commits — #1737 — during the ~70 minutes of local e2e). Two-sided check
against a numstat pinned **before** the rebase:

| | additions | deletions |
|---|---|---|
| whole branch, before | 269 | 16 |
| whole branch, after | 269 | 16 |
| `docs/DESIGN_NOTES.md` | 119 | **0** |

Identical per file, and none of #1737's files appear in the contribution.
Boundary shape read on the real file, not inferred from the numstat: #1737's
entry ends on a full stop, then `<!-- entry #1701b -->` with no blank ahead of
it, then blank / `---` / blank / heading. Marker checked against the base's
CURRENT tip, not the fork point.

⚠️ **The caveat, because the table above would otherwise imply more than it
measured.** The two full `integration.sh` runs, the RED/GREEN of the identity
assertion and the #278/#280 witnesses were all measured on the **pre-rebase**
base `31f9502f`. They were **not** re-run after the rebase.

That is not a formality here: **#1737 touches `default.css` and the same docked-
player area** (it adds tap-the-now-playing-band-to-restore-a-hidden-player and
moves the rail's horizontal inset). The two changes are plausibly interacting and
nothing local has yet observed them together.

What *was* re-run post-rebase, and is green on the merged shape:

- `bun.sh run check` — 4 stages, 0 failed;
- full vitest — **313 files, 6149 tests** (up from 6145: #1737's own tests are
  now in the run, and the #1701 unit pins pass alongside them).

So the e2e evidence in the body describes this change in isolation, and **this
PR's CI is the first thing to see it composed with #1737**. Happy to re-run the
local suite on the rebased head if that is wanted before CI's verdict.

— w2

Issue #1680, opened 2026-08-22
Pause the presence stream on unfocused channels, refetch members on focus (~30 events/s starves the client's main thread)


vjt:
## Symptom

With six or seven networks connected, the PWA becomes unusable: it drops clicks and the UI stalls. Reported on-device 2026-08-22.

## Measured (prod, read-only RPC on the live node, 2026-08-22)

* One 7-network account ingests **26,588 rows in 15 minutes = 29.5 events/s sustained**, of which **21,917 are presence = 24.4/s, 82%**.
* The big contributor is a single busy network (20,814 rows in the window, 19,683 of them presence); the rest are far smaller.
* The **server is not the bottleneck**: from the last nginx reload, **zero 503 and zero 4xx**, 1655 of 1718 requests under 100ms, none over 2s. Latency would produce waiting, not a UI that drops clicks.

So the client's main thread is fed ~30 events/s, the overwhelming majority of which are join/part/quit for channels the user is not even looking at.

## Decision (2026-08-22)

**Pause the presence stream for channels that are not in focus, and refetch the member list on focus.** Refetch first, because it is zero new wire and reuses an endpoint that already exists. A server-computed **diff** was considered and deferred: it needs either per-client state or a versioned member list, and **neither exists today**.

The payload cost of refetch was measured rather than assumed, on the worst channel on a real account: **4,605 members = 137,456 bytes of JSON, 23,711 bytes gzipped (17.2%)**, ~30 bytes per member. Across the whole account: 43 seeded channels, 10,800 members total — but focus touches **one** channel at a time.

## Why the granularity already exists (measured in source)

* Server: `Grappa.PubSub.Topic.channel/3` (`lib/grappa/pubsub/topic.ex:101`) is a **per-channel** topic, and `GrappaWeb.GrappaChannel.join/3` (`grappa_channel.ex:241`) parses it with `Topic.parse` before authorizing. "Pause" therefore means not joining / leaving that one topic — **no multiplexer needs inventing**.
* Client: cic joins Phoenix topics through `socket.channel(topic)` (`cicchetto/src/lib/socket.ts:442`), consumes channel-level events in `subscribe.ts` and routes per-channel state via `channelTopic.ts`.
[...]

Issue #44, opened 2026-06-04
cic: read-cursor POST 400s on service-nick query windows (invalid message_id)


vjt:
## Symptom

`POST /networks/:slug/channels/:chan/read-cursor` returns **HTTP 400**
repeatedly in production. Observed on **service-nick query windows only**:

| target    | 400s |
|-----------|------|
| NickServ  | 17   |
| ChanServ  | 10   |
| OperServ  | 4    |

(31 total in the sampled access logs; no real-channel or `$home`/`$admin`
target 400s.)

## Why it's a 400 (server contract)

`GrappaWeb.ReadCursorController.create/2` returns `400` only when the
request *shape* is invalid — its happy clause guards
`is_integer(message_id) and message_id > 0`, everything else falls to
`create(_, _) -> {:error, :bad_request}`. The channel target itself is
fine here: `validate_target_name/1` accepts any valid nick, and
`NickServ`/`ChanServ`/`OperServ` are valid nicks. A wrong-scope but
well-shaped `message_id` would be **422** (`:invalid_message`), not 400.

So the 400 means cic is sending a **missing / null / non-positive /
non-integer `message_id`** for these windows.

## cic side

`cicchetto/src/lib/readCursor.ts` `setReadCursor/4` POSTs
`{ message_id: messageId }` with **no guard** — it fires whatever the
caller hands it, including a 0 / NaN / undefined-coerced value. Service-nick
query windows (NickServ/ChanServ/OperServ — opened during auth flows)
evidently settle/blur before a positive persisted message id exists, so
the settle handler posts an invalid id.

## Proposed fix (cic)

Guard in `setReadCursor` (and/or the settle caller): if `messageId` is not
a positive integer, **skip the POST**. There is nothing to mark read for a
window with no real message id yet, so the request is pointless even when
it isn't rejected.

## Acceptance

- No `read-cursor` 400s for windows that have no positive message id.
- Real cursor advances on populated windows still POST and 200 as before.

vjt:
**Triage 2026-06-08 — still open.**

[...]

PR #1776, opened 2026-08-25
#1430 — audit the status-discriminated success unions, and pin what held

vjt:
Re #1430. The issue's own preamble says NOT observed — a structural read of
three files, no trace, no log, no reproduction — and asks not to be re-stated
as an observed bug. This branch respects that and answers the two questions it
deliberately left open, with measurements.

## What was measured

**The path fires.** `/msg <service> …` is the documented route to a services
nick: cic's `/query` refusal points the operator at it in words
(`commands/window.ts:138`), the `/ns` / `/cs` / … shortcuts rewrite to
`{kind: "msg"}`, and `compose.ts` routes a services target through
`sendPacedBody` → `scrollback.sendMessage` → `POST …/messages` with the service
nick as `channel_id`. Eleven nicks (`Identifier.services_sender?/1`) reach the
202 arm.

**The narrow cure had already landed** in `341d6b0c` (#1400) — both halves.
`sendMessage` returns `ScrollbackMessage | null` and reads `res.status === 202`;
`scrollback.sendMessage` gates on `row !== null`. The issue's line numbers were
already stale, so the sites were re-found by name.

**The `undefined` never reached the server.** `setReadCursor`'s positive-int
guard (`f22568dd`, 2026-06-09, issue #44) refuses a non-integer before both the
POST and the optimistic local advance, and `Number.isInteger(undefined)` is
false. The chain ran to completion and died at a boundary written fourteen
weeks earlier, for a different reason, guarding the same service-nick windows.
Blast radius as reported: a real type lie with **zero observable effect**, held
up by an invariant nobody had written down.

**The audit answers NO.** Every 2xx-producing terminal under `lib/grappa_web`
was censused per action, resolving the private-helper indirection by hand.
Exactly three JSON doors answer two success statuses with two body shapes:

| door | shapes | client |
|---|---|---|
| `POST …/channels/:c/messages` | 201 row / 202 ack | **was** the collapse; cured by #1400 |
| `POST /auth/login` | 200 envelope / 202 challenge (×2) | genuine TS union, narrowed in `auth.ts` |
| `GET …/channels/:c/members` | 200 envelope / 204 empty | `listMembers` reads the status |

`GET /uploads/:slug` is 200/206 but is a byte stream with no `fetch` caller and
no JSON cast. No other door collapses a status union.

What the census *did* turn up: the members door shipped in #1680 with **no
vitest at all**, so the verdict "this one is fine" rested on an unpinned line.

## What ships

Two pins and the record. No production code changed — nothing was found that
needed changing.

- `readCursor.test.ts` — the absent id, as distinct from #44's wrong-but-present
  set (0 / −1 / NaN / 1.5). Separate describe, own provenance.
- `api.test.ts` — the members door's 204/200/empty-200 triple. The middle case
  matters: an empty `members` array is a real answer, 204 is "not known yet",
  and collapsing them blanks a good member list mid-fill.
- `docs/DESIGN_NOTES.md` — entry `#1430`.

No wire change, no field added, **no `protocol_version` bump**: the discriminant
the client was ignoring was already on the wire, as the status.

`ScrollbackMessage | null` is kept as the closed-set encoding rather than a
tagged union — the second member carries no payload, strict null checks already
force the narrow, and it now matches `listMembers`' `MemberEntry[] | null`.

## The bench

Four mutants, each restored with `git checkout --`. Control column = the same
mutant against the base test files at `e1c87d8f`.

| mutant | site | with the new pins | base alone |
|---|---|---|---|
| M1 — drop `res.status === 202` | `api.sendMessage` | 1 dead | 1 dead (pinned by #1400) |
| M2 — drop `res.status === 204` | `api.listMembers` | 1 dead | **0 — fully green** |
| M3 — drop the `isInteger` half | `setReadCursor` | 5 dead | 3 dead |
| M4 — `messageId ?? 1` totality rewrite | `setReadCursor` | 3 dead | **0 — fully green** |

M3 is not discriminating (#44's NaN/non-integer already kill it). M4 is what
earns the pin: a defensive nullish default is exactly the edit that reinstates
the hole, it leaves every #44 case caught, and the base suite passes it in
silence. M3 also split the two new values — `null` **survived** it while
`undefined` died, because `null <= 0` coerces to true, so the two halves of the
guard carry one value each and the pair is not redundant.

## Gates

- `scripts/bun.sh run test` — 315 files, 6214 tests, `rc=0`.
- `scripts/bun.sh run check` — 4 stages, 0 failed (lock drift, biome, tsc src,
  tsc e2e), on a clean tree.
- `scripts/design-notes-gate.sh` — `1 new entry heading(s), separator and marker
  present.`
- No Elixir touched, so no COMPILE lane was taken.
- **No e2e, deliberately.** There is no visible outcome to assert: the pre-fix
  chain produced no ghost row and no server write (the #44 guard ate it), so an
  e2e here would be green and empty.

## Not established

The production log was not read, so there is no field frequency for the 202
arm — only the proof that the flow is reachable and documented. And a
*body*-discriminated union census is a different axis that was not run; this
one keyed on the HTTP status.

### DIFF
--- a/cicchetto/src/__tests__/api.test.ts
+++ b/cicchetto/src/__tests__/api.test.ts
@@ -1218,6 +1218,43 @@ describe("sendMessage reads the STATUS, not just res.ok (#1430)", () => {
   });
 });
 
+// #1430 — the audit this issue asked for turned up exactly three REST doors
+// whose SUCCESS answer is a status-discriminated union, and this is the second
+// of them: `MembersController.index/2` answers 204 for `{:ok, :uninitialized}`
+// (joined but pre-NAMES, or not joined at all) and 200 with the envelope for
+// `{:ok, [member()]}`. `listMembers` reads the status and returns `null`, so it
+// does NOT collapse the union the way `sendMessage` used to — but that claim
+// rested on an UNTESTED line: #1680 shipped the door with no vitest at all, and
+// a mutant deleting `if (res.status === 204) return null;` went green, taking
+// `.json()` on an empty body to a throw the caller logs as a refetch failure.
+// The audit's conclusion for this door is only worth as much as its pin.
+//
+// (Door three is `POST /auth/login`, 200 envelope vs 202 challenge; that one is
+// typed as a genuine TS union — `LoginResponse` — and narrowed in `auth.ts` on
+// `two_factor_required`, with its own coverage in `auth.test.ts`.)
+describe("listMembers reads the STATUS, not just res.ok (#1430 audit, #1680 door)", () => {
+  it("returns null on the 204 instead of an empty list", async () => {
+    vi.stubGlobal("fetch", vi.fn().mockResolvedValue(new Response(null, { status: 204 })));
+    await expect(api.listMembers("t", "azzurra", "#italia")).resolves.toBeNull();
+  });
+
+  it("returns the narrowed members on 200", async () => {
+    stubFetch(200, { members: [{ nick: "vjt", modes: ["o"] }] });
+    await expect(api.listMembers("t", "azzurra", "#italia")).resolves.toEqual([
+      { nick: "vjt", modes: ["o"] },
+    ]);
+  });
+
+  it("distinguishes 204 from a genuinely empty channel", async () => {
+    // The distinction CP24 bucket E drew and #1680's caller depends on: an
+    // empty `members` array is a real answer (seed it), 204 is "not known yet"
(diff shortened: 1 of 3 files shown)


vjt:
Superseded by #1777 — identical head commit `3b5cd9c2`. Closed as a side effect of renaming the branch `1430-status-union` → `w2-1430`; GitHub's branch-rename API closes the PR rather than retargeting its head, and reopen fails once the old ref is gone.

Issue #1787, opened 2026-08-25
wire_pin --check reports an identical digest across a real wire-shape change: the pin does not cover hand-written *_json.ex


vjt:
## What was measured

While shipping #1766 (a fourth `display_prefs` key, `show_bottom_bar`), w1 measured that **`mix grappa.wire_pin --check` reported the digest IDENTICAL on both sides of a genuine wire-shape change.** The `Grappa.Protocol` bump 5 → 6 was made **because of the rule, not because any gate went red.**

Cause, as measured: `GrappaWeb.UserSettingsJSON` is not in the codegen's `@extra_modules`, so the generated wire shape the pin digests never sees it.

## Why this matters more than one missing module

`CLAUDE.md` names this task as the *enforcement* for the version rule:

> Enforced, not merely written: `mix grappa.wire_pin --check` (in `scripts/check.sh` + CI) holds a digest of the generated wire shape NEXT TO the version it was taken at, so a shape change with a still number is RED

That sentence is currently false for any surface whose wire shape lives in a hand-written `*_json.ex`. The gate passes green on exactly the change it exists to catch — which is the same failure mode the rule's own history describes, where `@protocol_version` sat at `1` through five additive fields.

## This is the SECOND occurrence, not the first

**#1679 hit the identical gap with `BootJSON`.** Two independent sightings of the same hole, in different modules, is a class defect rather than an oversight — the fix is presumably not "add `UserSettingsJSON` to the list" but making the digest cover the hand-written JSON views as a category, or making an uncovered view a loud failure rather than a silent omission.

## Deliberately not fixed in #1766

w1 declared it rather than patching it, which is the right call: widening the pin's coverage will very likely re-digest surfaces beyond that issue's subject, and it wants its own change and its own gate.

## Not measured / open questions

- How many other hand-written `*_json.ex` views the pin currently misses — nobody has enumerated them.
- Whether the correct cure is enumerating them, deriving them, or failing loudly on an unlisted view.

_Filed by the orchestrator from w1's measured report on #1766. Untriaged: no `status:*` label — enqueueing is vjt's call._

Issue #1769, opened 2026-08-24
perf: finish #1680 on the server — presence suppressed by join params, so third-party clients are unaffected


vjt:
## Request

vjt, `#grappa`, 2026-08-25 00:28:

> e cristo metti in lavorazione sia un batching endpoint per alleviare il carico sul multi network sia il discorso dell'unsub del signalling e re-fetch dei membri al focus

Second half of that order. First half is #1768.

## Ruled (vjt, `#grappa`, 2026-08-25 00:32–00:33)

> ok bene ma finiamo 1680 allora / aggiungendo i params / cosi non sfasciamo ne lucy ne sonic

**This issue is #1680 finished, and the shape is settled: join params.** Not a
protocol gate, not a new verb, not a server-side policy about who gets presence.

The reason is compatibility, and it is the argument for the shape rather than a nice
side effect: **the default stays *everything*.** A client that keeps joining topics
the way it does today keeps receiving exactly what it receives today, so the
self-hosted deployments and the other client authors — Lucy's and Sonic's among them —
are unaffected by construction. Nothing to coordinate before landing, nothing to
migrate, no flag day. The suppression only happens for a socket that **asks** for it.

## What already exists — checked before filing anything

**#1680 is merged, not missing.** PR #1690 landed 2026-08-23 (`e935a83a`), and on
`origin/main` @ `5df1e3cb` the client carries the whole pause:

* `cicchetto/src/lib/presencePause.ts` — which channels are paused and which kinds a
  paused channel is allowed to swallow (peer `join`/`part`/`quit` only; `nick_change`
  and `mode` are carved out because #372/#373 identity migration and channel-mode
  state depend on them, and own-presence is never dropped because an own PART tears
  the window down).
* `cicchetto/src/lib/presenceCooldown.ts` — the 2-minute blur window.
* `subscribe.ts:158` wires `createPresencePause`, and its `onResume` seam fires the
  **members refetch** exactly when presence was actually missed (`pausedKeys.delete(key)`
  returns whether it was there), never on a plain re-focus.

So "unsub + refetch on focus" is **built and merged**. Two things follow, and they are
the reason this issue exists rather than a "already done" reply.

## 1. It is not deployed

`origin/main` is undeployed by standing decision — vjt, 2026-08-24: *"no deploy fino a
che non si fixa la contention"*. Whatever vjt is feeling on his device today is the
**old** client. Nothing to build for that; it is a deploy decision, tracked where the
contention is (#1715, #1759).

[...]

Issue #1766, opened 2026-08-24
cicchetto: per-user toggle to hide the mobile bottom bar, plus a display-section tidy


vjt:
## Request

vjt, `#grappa`, 2026-08-25 00:10–00:12 (three parts, one change to the settings drawer):

> aggiungiamo setting toggle per disattivare la bottom bar su mobile. ora sono su 7 reti ed è diventata praticamente inutile
>
> nella sezione display come checkbox, e andiamo ad accorpare in un unico fieldset i checkbox esistenti e quello nuovo
>
> e già che ci siamo da general togliamo il label "mark me away after" che è già incluso nel titolo del fieldset "auto-away"

The BottomBar is a flat, horizontally-scrolled strip of every window across every
network (`BottomBar.tsx`: per-network sections, server header → channels → queries).
It is O(windows), not O(screens) — at 7 networks the strip is longer than the useful
scroll distance, so the picker no longer picks. Default stays **on**; this is a
per-user opt-out.

## Prior rulings — read these before re-deciding anything

* **#174 "Bottom bar: make it opt-in (per-user setting)"** — CLOSED 2026-07-04 with
  an explicit vjt ruling: *"no standalone issue needed. The constraint — the bottom
  bar must NOT be deleted, it stays opt-in from settings — is to be carried into the
  SIDEBAR rework issue"*. The carrier was **#71**, which closed 2026-07-28 **without
  shipping the toggle**. The intent has stood since July and its vehicle is gone —
  this issue is the replacement carrier, not a re-litigation.
* **#71, 2nd design ruling** — *"the mobile bottom bar is kept. No drawer-only
  navigation, and therefore no `☰` fallback either"*; *"kill the mobile bottom bar"*
  was explicitly **reversed**. Nothing here kills it: default-on, opt-out.
* **#1041** (2026-08-08) brought a left drawer back on mobile — mounted on a
  left-edge swipe, destroyed on hide, no hamburger — framed then as an *additive
  second door*, BottomBar still primary. That door is what makes this toggle
  survivable, and it did not exist when #71 ruled.

---

### Where it lands: `DisplayPrefs`, not a new mechanism

Boot-applied UI state, server-backed and cross-device: the `displayPrefs.ts`
coordinator family (#449). Not notification-prefs shape (a submitted form), and not
localStorage-only — a device-local toggle would repeat the #449 bug (Hypnotize: set
on desktop, invisible on the iOS PWA), and phone + tablet is exactly the case here.

[...]

Issue #1739, opened 2026-08-24
Serve station logos from a stable grappa URL that proxies upstream and falls back server-side


vjt:
Proposed by vjt on #grappa on 2026-08-24: serve the station logos from a stable grappa URL that proxies and caches the upstream image, falling back server-side when there is no upstream image, so the client stops carrying that logic.

The idea is worth doing. Two of the arguments that would seem to support it do not survive a measurement, and this issue states that up front so nobody builds on them.

## What was measured, 2026-08-24

**The stated premise — "the images are refetched every time" — is not supported.** `api.somafm.com` serves the logos with:

```
Cache-Control: max-age=14400
ETag: "5ec9-653a3c29e905c"
Last-Modified: Sun, 07 Jun 2026 06:11:10 GMT
Content-Length: 24265
```

Four hours of freshness plus a validator. A browser that has one of these should not be reissuing a network request for it, and a revalidation after four hours is a 304, not a 24KB transfer. Before any work starts, the reporter should confirm what the Network panel actually shows for a logo on a second picker open — a real `200`, a `304`, or `(from disk cache)` — and whether DevTools had "Disable cache" ticked. If it is genuinely refetching, that is a different and more interesting bug than this issue, and the cause is on our side, not upstream's.

**The CSP argument is dead — do not use it.** Proxying the logos would NOT let `img-src` be tightened. `img-src 'self' data: https:` is wide because of #1240, so the media viewer can render a cross-host image link (an upload minted by another grappa instance, or a litterbox URL). That grant stands with or without a logo proxy; see the reasoning in `lib/grappa_web/plugs/security_headers.ex`. The logos are simply covered by a permission that exists for someone else.

## What actually argues for it

**Privacy.** Today every client fetches the logos straight from `api.somafm.com`, so every viewer's IP and user agent reach a third party each time the picker paints — 14 rows today. A proxy means the third party sees the server and nobody else. This argument is independent of caching and is the strongest one on the table.

**One stable URL with a server-side fallback.** A fallback has to exist regardless: a station from another provider has no row in SomaFM's catalogue but still needs a logo (`radioStations.ts` says as much about why `logoUrl` cannot leave the type). Deciding "upstream image or fallback" once, behind one URL shape, is less machinery than deciding it in the render path — vjt's "meno logica lato client".

## What it costs, to be weighed rather than discovered

[...]

PR #1794, opened 2026-08-25
perf(#1769): finish #1680 on the server — presence suppressed by join params

vjt:
#1680 shipped the presence pause at cic's **dispatch edge**: a channel nobody
has looked at for two minutes stops APPLYING peer join/part/quit. The events
still cross the socket, still get decoded, still wake the main thread. This is
the server half — a join param that stops them being sent.

The shape is join params by ruling (vjt, 2026-08-25), and the reason is the
default: a client that joins the way it joins today receives what it receives
today, **by construction**. No flag day, nothing to coordinate with the other
client authors. Only the literal `false` suppresses.

## The issue named a choke point that cannot see this traffic

The scope said to filter at `handle_info(%Phoenix.Socket.Broadcast{})`
(`grappa_channel.ex:474`). Measured, that clause never sees per-channel
traffic: the framework installs a **fastlane** subscription on a channel's own
topic, so `broadcast_event/2` is encoded once by the dispatcher and written
straight to the transport. The module's own moduledoc says so twenty lines
above the clause (`:161-162`), and Phoenix's dispatcher confirms it
(`channel/server.ex:93-118`). That clause only ever serves #1088's foreign
socket topic.

`intercept ["event"]`, the idiom, is the wrong fix here: it compiles to
`__intercepts__/0`, a **per-module, all-sockets** switch copied into every
socket's fastlane metadata. It would route every broadcast on every topic
through a channel process — a regression on the default path of a perf issue,
and it would hand each channel process the whole flood in its mailbox during
exactly the netsplit bursts this work exists to survive (#1715, #1739).

But the intercept list lives in the **subscription metadata**, not in a global,
so a per-socket answer is expressible: a suppressing socket trades its fastlane
for a plain subscription in `:after_join`, and its broadcasts then arrive as
structs routed to `handle_out/3`. Everyone else's path is not "equivalent" —
it is untouched.

## What is never dropped

`Message.pausable_presence_kinds/0` (`[:join, :part, :quit]`) is a **strict
subset** of the presence kinds, declared separately rather than derived:
`nick_change` drives the #372/#373 identity migration and `mode` feeds
channel-mode state, and both fail silently without it. Our OWN presence is
never dropped either — an own PART tears the window down — on a folded compare
seeded at join that then **follows a rename**, off the `nick_change` row the
carve-out already guarantees delivery of.

## Wire

Bumped to **7**. `mix grappa.wire_pin --check` was silent — measured, with the
change applied it answered `wire shape and protocol 6 agree.` Third instance of
that detector gap (#1679, #1766) and the first where the un-covered surface is
**inbound**, which no widening of the outbound codegen digest would reach.
Filed as #1787. `min_version` stays at 1 — nothing is stranded.

## Evidence

| gate | result |
|---|---|
| `scripts/check.sh` | **rc=0** — Credo 0, Dialyzer 0, Sobelow, doctor, deps.audit clean, bats 692, `shape.pin: wire shape and protocol 7 agree.` |
| `scripts/test.sh` (full) | 8 doctests, 62 properties, **6893 tests, 0 failures** |
| `scripts/bun.sh run test` | **314 files, 0 failures** |
| `scripts/bun.sh run check` | **5/5 stages ok** |
| `scripts/design-notes-gate.sh` (post-commit) | `1 new entry heading(s), separator and marker present.` |
| mutation testing | **7 one-axis mutants, 7 killed**, all deterministic over repeated runs |
| `playwright test --list` | the new spec is collected: 1 test, `Total: 634 tests in 376 files`, rc=0 |

Mutants: the fastlane swap never fires (9 arms); subscribe without
unsubscribing, i.e. the BUG 6 shape (9); the own carve-out never matches (3);
the whole presence set dropped instead of the pausable subset (2); the rename
never followed (1); the param read on every topic shape (1); that last one plus
a swap in the user arm (1).

## 🔴 NOT run locally — read a red here as expected cost

**The new e2e spec has never been executed.** The STACK/e2e lane is held by
another worker, so this is gated from CI per the contended-lane rule. Static
gates on it are green (biome, `tsc -p e2e/tsconfig.json`, and Playwright
collects it), but that is not a run.

Two things about it are worth knowing before the job goes red:

* it uses **`page.clock`** (`install` → `resume` → one `fastForward` past the
  two-minute cooldown), which has **no precedent in this repo's e2e** — chosen
  over burning two minutes of wall clock per run, and over adding a production
  seam. If the clock disturbs boot, the fallback is the wall-clock wait and I
  will say so rather than weaken the spec;
* it asserts an **absence** (no presence frames on the paused topic), so it
  carries four positive controls in the same page against the same peer on the
  same socket: the same channel delivers before the pause, the focused channel
  keeps delivering after it, a PRIVMSG to the paused channel still arrives, and
  cic's outbound `phx_join` carries `{"presence": false}`. The negative has no
  timing window — the PRIVMSG is the ordering barrier.

Also not supplied: **the delivered event-rate measurement the issue asks for**
("measure the delivered event rate for a paused channel before/after ... at the
socket"). `origin/main` is undeployed by standing decision and the jail is not
reachable from here. The e2e proves the frames are absent; no rate was taken.
The `intercept` alternative's cost is reasoned from Phoenix's dispatcher source
(sends and encodes counted), not benchmarked. And nothing here establishes that
a paused window is cheaper end to end once the pause/resume re-join, its join
reply and its REST backfill are counted against the events it stops paying for.

## One consequence of the ruled shape

The param is read once, at join, so changing the answer is a **re-join** —
`phx.leave()` then join. There is a short window with no live delivery on that
topic. Nothing is lost (every channel join already runs `refreshScrollback` on
its ACK), but it invalidated an existing oracle: the test encoding vjt's
"messages must not be lost" asserted `phx.leave()` is never called, which would
now forbid the correct implementation. It was replaced by a **stricter** one —
the topic is joined AGAIN, with `{presence: false}` — plus a sibling arm for
the resume half, which is the one an implementation can get wrong in silence.

### DIFF
--- a/cicchetto/e2e/tests/issue1769-presence-suppressed-on-the-socket.spec.ts
+++ b/cicchetto/e2e/tests/issue1769-presence-suppressed-on-the-socket.spec.ts
@@ -1,0 +1,263 @@
+// Issue #1769 — presence suppressed by join params, measured ON THE SOCKET.
+//
+// WHY THIS SPEC EXISTS AND WHY IT LOOKS AT FRAMES
+//
+// #1680 already stops a paused channel from APPLYING peer join/part/quit: the
+// events cross the socket, get decoded, wake the main thread, and are dropped
+// at cic's dispatch edge. So the obvious end-to-end assertion — "the member
+// list does not move on a paused channel" — is green with or without #1769.
+// It would be a spec that cannot fail for the reason it was written.
+//
+// What #1769 buys is the thing #1680 could not: the bytes never arrive. That
+// is only observable at the transport, so this spec counts WebSocket frames
+// (`page.on("websocket")` → `framereceived`, the same instrument
+// `issue375-rehash-option.spec.ts` and `i2b-image-upload-litterbox-host.spec.ts`
+// use) and asserts on which ones are absent.
+//
+// THE CONTROLS, BECAUSE AN ABSENCE PROVES NOTHING BY ITSELF
+//
+// "No frame arrived" is also what a broken fixture, a peer that never joined,
+// and a dead socket look like. Three positive controls run in the same page,
+// against the same peer, on the same socket:
+//
+//   1. BEFORE the pause, a peer JOIN on the very same channel DOES produce a
+//      frame. Same topic, same event, same everything — only the pause differs.
+//   2. AFTER the pause, the FOCUSED channel still receives its peer JOIN. The
+//      suppression is per-channel, not per-socket, and this is the arm that
+//      fails if the whole socket went quiet.
+//   3. AFTER the pause, a PRIVMSG to the PAUSED channel still arrives. This is
+//      vjt's ruling as an assertion — a paused window goes quiet, never blind
+//      — and it is the one that fails if someone "simplifies" the pause into
+//      an abandoned topic.
+//
+// And one more, on the outbound side: the join frame cic SENT for the paused
+// channel must carry `{"presence":false}`. Without it the absence downstream
+// would prove the server dropped frames nobody asked it to drop.
+//
+// THE CLOCK
+//
+// `PRESENCE_COOLDOWN_MS` is two minutes. Rather than burn that in wall-clock
+// on every CI run, or add a production seam to shorten it, the spec drives
+// Playwright's clock: install → resume (so boot and the network run at real
+// speed) → `fastForward` past the window, which fires the pending cooldown
+// timer and nothing else that matters. The cooldown's own arithmetic is
+// covered by `presencePause.test.ts` under fake timers; what this spec is for
+// is everything downstream of it.
+
+import type { Page, WebSocket } from "@playwright/test";
+import { composeSend, loginAs, selectChannel } from "../fixtures/cicchettoPage";
+import { IrcPeer } from "../fixtures/ircClient";
+import { AUTOJOIN_CHANNELS, NETWORK_SLUG } from "../fixtures/seedData";
+import { expect, specNick, specUser, test } from "../fixtures/test";
+
+// The channel that gets blurred and paused: the seeded autojoin one.
+const PAUSED_CHANNEL = AUTOJOIN_CHANNELS[0];
+// The channel that stays focused, so control (2) can tell "this channel is
+// suppressed" from "this socket is dead".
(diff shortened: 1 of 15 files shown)


vjt:
## Full local `integration.sh` — run, and triaged

The e2e spec this PR adds had never been executed. It has been now.

**Full suite on `d3d3a36f`:** `782 passed (30.3m), 1 failed`, `rc=1`.

`issue1769-presence-suppressed-on-the-socket.spec.ts` → **passed, 7.0 s**,
first execution ever. `page.clock` held; the plan B (real-clock wait) was
not needed. Both risks flagged in the PR body are discharged.

The tree was frozen for the whole run — `find -newer` over the window
`22:06:32–22:39:43` returns nothing under `lib test cicchetto/src
cicchetto/e2e priv config mix.exs`, and the last source mtime predates the
start by half an hour. So the one red is attributable, not a dev
code-reloader artifact.

### The one red is pre-existing, and measured as such

`cp15-b6-part-archive-rejoin.spec.ts:55` — `.members-pane li` filtered by
the spec nick, not visible within 5 s.

Isolated re-runs (`--project=chromium --grep "archive list dedup holds"`,
one spec, idle stack):

| tree | failures |
|---|---|
| this branch `d3d3a36f` | 5 / 32 (15.6%) |
| base `f9b47ea1` | 4 / 32 (12.5%) |

Fisher exact two-sided **p = 1.000**. The spec file is byte-identical on
both trees (this branch does not touch it), so the only variable was the
product code — and the defect is present without this branch. Reproducing
on an idle stack also rules out load.

The Playwright artifact settles the mechanism: in the post-failure page
snapshot the members pane **exists**, reads `members (1)` and **contains**
the exact node the assertion waited for. The members are late, not absent
— the spec uses the sidebar row (`channels_changed` → REST) as a proxy for
the channel-topic join ACK, which is a different chain.

Two hypotheses that would have made this mine were both falsified:
`onResume` cannot fire for a channel that was never paused
(`pausedKeys.delete(key)` gates it, and entry needs the 120 s cooldown this
spec never reaches); and the non-suppressing `join/3` path adds one
`assign` and no session round-trip (`own_nick_or_nil/2` lives only in the
`%{"presence" => false}` arm).

Filed as a class, not an instance: #1795.

**Not established:** that this branch does not *worsen* the race by a small
amount. 32+32 runs exclude a large shift, not a few points.

Issue #351, opened 2026-07-19
cicchetto: make the entire message area a drag-and-drop file-upload drop-target


vjt:
## Summary

The **entire message area** (the whole conversation pane — scrollback + compose) should be a **drag-and-drop drop-target for file upload**, not just the compose box. Drop a file anywhere over the messages and it uploads exactly as if dropped on the compose box today.

## Current state (source-anchored)

The drop-target is scoped to **ComposeBox only**:
- `onDragOver` / `onDrop` handlers: `cicchetto/src/ComposeBox.tsx:322-330` — `onDrop` reads `e.dataTransfer.files` and calls `handleFiles(files)`.
- Wired on the compose element: `ComposeBox.tsx:470-471` (`onDragOver={onDragOver} onDrop={onDrop}`).
- Upload path already exists and is reusable: `handleFiles` → `lib/uploadOrchestrator.ts` (dedup, size-cap, progress).

So a file dropped over the **scrollback** (the large part of the screen) does nothing — the user has to aim at the small compose strip.

## Proposed change

Hoist the drop-target up to the **main pane** that composes `ScrollbackPane + ComposeBox` vertically — `Shell.tsx` (per `ScrollbackPane.tsx:65-66`: "the parent layout composes ScrollbackPane + ComposeBox vertically"). Reuse the existing upload path:

- Factor the `onDragOver`/`onDrop` → `handleFiles` logic so the main-pane container can call it (hoist `handleFiles`, or expose a small shared `dropUpload(files, ctx)` helper backed by `uploadOrchestrator`). Avoid duplicating the orchestrator wiring.
- Bind `onDragOver` (must `preventDefault` to allow the drop) + `onDrop` on the main-pane container.
- **File-drag guard:** only engage when the drag actually carries files — check `e.dataTransfer.types.includes("Files")` — so dragging selected text / an in-app element over the pane doesn't arm the overlay or swallow the drop.
- **Visual affordance:** a drag-over overlay covering the pane ("Drop to upload") while a file is dragged, cleared on `dragleave`/`drop`. Track a drag-depth counter (dragenter/dragleave fire per child) so the overlay doesn't flicker crossing child elements.

## Boundary / notes

- Keep the ComposeBox handler working (or let the hoisted one cover it) — no regression in the current compose-box drop.
- Don't let the drop-target interfere with the media-viewer modal or other overlays that have their own drag semantics.
- Client-side (cicchetto only) → no cold deploy.

vjt:
[...]

Issue #1667, opened 2026-08-22
cicchetto/PWA: after a short background NOTHING renders until a tab switch — peer messages and the tab × are frozen too, not just own echo (mobile)


vjt:
## Field report (not yet reproduced)

Reported on `#grappa` on 2026-08-22 by a daily user, on the deployed client, and
**re-qualified by him the same day** after it happened to him again live. The
second round changed the shape of the report enough that the original framing —
"after a very long background" — no longer holds.

### What he states now (2026-08-22, second round)

* **installed as a PWA**, on mobile. Desktop never shows it.
* the background before it happened was **about one minute**, not "da
  parecchissimo". His words: *"forse mi sbagliavo, si è appena riproposto con il
  background di un minuto tipo"*. **The long-background precondition is
  withdrawn.**
* **it is not that messages disappear, and it is not scroll.** The messages he
  sends **are sent** — they leave, other clients see them — **he just never sees
  them himself**. His words: *"non è che scompaiono è che i miei quando premo
  invia vengono inviati ma io non li vedo (non è un problema di scroll), non
  fanno echo tipo"*.
* **switching to another tab and coming back makes them appear** — confirmed
  explicitly, twice, and it is the cheaper recovery than the app kill first
  reported. Killing and reopening the app also works.
* while in that state it is persistent, not a one-off flicker: *"lo sta facendo
  abbestia"*.

### Still unmeasured — the one question that splits the diagnosis

**While his own lines do not render, do other people's incoming lines still
arrive?**

* incoming arrives ⇒ the socket is live and the hole is specific to the echo of
  his own sends;
* incoming does not arrive either ⇒ the socket is dead and this is not an echo
  bug at all, it is a silent zombie subscription.

Asked on-channel while he was in the failure state; before he could answer, the
state cleared on its own and he could not get back into it. **Unanswered.**
Anyone who reproduces this should grab that observation first — it is worth more
than any further description of the symptom.

## Hypothesis — code-backed, NOT measured, do not treat as the cause

Recording it because the source says something that fits the symptom exactly,
and the next person should either confirm or kill it rather than rediscover it:

* `cicchetto/src/lib/socket.ts:353` (#254) states the design rule outright: on a
  wake reconnect the client relies on the server echo, and *"the echo stays the
  sole render path (no optimistic local render)"*. So a client whose
  subscription is stale renders **none of its own sends** while the send itself
  still reaches the server over its own path — which is precisely the reported
  asymmetry.
[...]

PR #1885, opened 2026-08-31
cic: move the upload guard to one door (follow-up to #1884)

### LINKED ISSUE #1883: cicchetto: a gallery pick uploads instantly — no preview, no confirm, no undo
## What happens

Picking a file from the phone gallery uploads it **immediately**. There is no preview, no
confirmation, no way back: the moment the native picker returns, the file is on its way to
the channel.

The only gate today is the **privacy modal**, and it is one-shot per host:

`cicchetto/src/lib/uploadOrchestrator.ts:673-682` — `startUpload/2` reads
`localStorage[privacyKey(host)]`; if the user has ever ticked "remember", `dispatchUpload`
runs straight away. `acknowledgePrivacy/1` (line 685) writes that flag.

`cicchetto/src/ComposeBox.tsx:529-539` — `onPickerChange` hands `input.files` to
`triggerUploads` with no intermediate step. Same for the drop path
(`cicchetto/src/lib/dropUpload.ts:22`).

So for every returning user the sequence is: tap paperclip → tap photo → **it is public**.

## Why it matters

On a phone the gallery grid is dense and the thumbnails are small; a mis-tap picks the
wrong photo and there is no undo. Reported on #grappa: "prima o poi qualcuno carica una
foto del cazzo".

Note the asymmetry with the paste path, which already has a guard: a large text paste is
routed to a flood-confirm (`cicchetto/src/lib/pasteRoute.ts`), while a file — the
higher-stakes, irreversible one — is not.

## Proposal

A confirm step between selection and dispatch, showing what is about to be sent:

- thumbnail (or filename + size for non-images), the target channel, Send / Cancel;
- for a multi-file batch, the whole set with per-file removal;
- reuse the existing `ConfirmModal` chrome so no new surface is introduced.

Open questions for whoever picks this up:

- Should it be always-on, or a setting with a "don't ask again" like the privacy modal?
  A second remembered opt-out reintroduces exactly the same problem.
- Does it apply to the drop and paste paths too, or only to the picker? Drag-and-drop on a
  desktop is a deliberate gesture with a visible target; the gallery tap is not.

Suggested by peluche on #grappa.

gmsrc:
**Summary.** #1884 closed #1883 by putting the gallery-pick confirm in `pickerUpload.ts`, guarding the picker alone; this PR moves that same confirm into `uploadOrchestrator.triggerUploads` so drop, paste and the OS share-target inherit it too, and deletes `pickerUpload.ts`. Nothing is re-implemented — #1884's attachments API, per-file removal and CSP `blob:` token land here unchanged — and the queue is still never touched before Send, because `triggerUploads` now only asks while the private `enqueueUploads` holds the old body.

confirmation

---

Follow-up to #1884, which closed #1883 while this branch was in flight. **This PR does not re-implement the feature** — #1884's confirm, its attachments API, the per-file removal and the CSP `blob:` token all land here unchanged. It moves *where the question is asked*, per vjt's ruling on the review thread: from `pickerUpload.ts` (picker only) into `uploadOrchestrator.triggerUploads` (every surface).

Rebased onto current `main`; `pickerUpload.ts` is deleted.

## Why the door moved

#1884's header made three claims. **Two are right and are kept verbatim** in the orchestrator:

- there must be no *"don't ask again"* — a remembered opt-out is the exact shape of the privacy flag that produced the defect; a gate every returning operator has already switched off is not a gate;
- the picker must **not** route through `dropUpload` — that helper pre-filters on `categoryOf`, and iOS labels a `.m4r` `application/octet-stream`, which only `normalizeUploadFile` rescues.

**The third is what fell**: that a drop is "a gesture the operator aimed" and Ctrl-V "one they typed", so neither needs asking. A pasted image is the one case where the operator *cannot see what they are sending at all* — the clipboard is invisible until it is in the channel. And the OS share-target reaches `dropUpload` too, so "the operator aimed it" describes a gesture that happened in a different app.

More generally: a guard placed per-surface is an **enumeration**, and enumerations drift. #351 had already collapsed picker, drop and paste onto `triggerUploads` for exactly this reason, and the share-target joined them later without touching any of the three call sites — which is precisely how a fourth door arrives without anyone remembering the guard.

## The queue objection is answered, not waved away

#1884's strongest argument against this position was mechanical: *"the orchestrator owns the queue, and a batch the operator has not authorised has no business being in it"* — it would take the channel's single upload slot and show an `(i/N)` counter for files that may never be sent.

That is answered by **splitting the function**, not by moving the gate out of it. `triggerUploads` only *asks*; the private `enqueueUploads` is the old body and is reached only from `onConfirm`. Nothing is queued, no batch counter moves and no upload slot is taken until Send. The property #1884 wanted is preserved exactly — only its location changed.

## What this costs, stated rather than discovered

The paste → **"Upload as .txt"** route (#80/#816) now shows **two** dialogs: the flood guard, then the send-confirm. That is the price of one door, and it is *not* being paid off with a `confirmed: true` parameter threaded through `dropUpload` — a bypass door with a polite name, and the one thing that would make the enumeration argument above false again. If the friction reads as too much in the field, the fix is to soften the paste dialog.

It has a test consequence worth knowing before it bites: the flood guard and the send-confirm render through the **same** `confirmDialog` singleton and swap within one tick, so an e2e locator keyed on `confirm-modal` alone is already visible over the *paste* dialog and acts on the wrong one. `sendPickedFiles` therefore scopes by the dialog's heading (`/^Send to /`), which is also its `aria-label`.

## Coverage moved, not dropped

`pickerUpload.test.ts` mocked the whole orchestrator and asserted *"`triggerUploads` was called"*. That boundary no longer exists, so its eleven cases were rewritten against the real pipeline in `uploadOrchestrator.test.ts`, where they assert on the test **host** — nothing reaches `host.upload` before Send. That is a stronger claim than the one it replaces.

The two `ComposeBox.test.tsx` cases that asserted dialog behaviour were reduced to wiring, deliberately: that file mocks the orchestrator, so *"no upload after Cancel"* would pass there against a build with **no gate at all**. A test that cannot fail is worse than no test.

e2e: `uploads4-pane-drop` now asserts the drop path confirms (the door #1884 left open — and the only test in either PR that pins it), and the two paste specs walk both dialogs.

## Verified on device

Android emulator, cic served by a real BEAM, upstream on real Azzurra in `#sbiffo`:

| | Cancel | Send |
|---|---|---|
| single pick | nothing queued, nothing published | uploads, `📸 <url>` posted |
| 2-file batch | whole batch dropped | both uploaded sequentially, distinct slugs |

Screenshots below are of the modal itself on that build.

## Gates

`biome` + `tsc` (src and e2e) green; the affected unit suites green (`uploadOrchestrator` 75, `ComposeBox` 116, `ConfirmModal`, `confirmDialog`, `dropUpload`). e2e is CI's to run — it cannot be run on the authoring host.


vjt:
Measured this against `origin/main` now that #1884 has landed (merged 17:57 UTC, closing #1883), so the rebase is not mechanical: **the two PRs disagree about where the gate goes, and that disagreement has to be settled before any code is moved.**

**What main has.** The confirm lives in `cicchetto/src/lib/pickerUpload.ts` and covers the picker only. It is the sole importer of `confirmDialog` in `cicchetto/src`; `dropUpload.ts` still calls `triggerUploads` directly, so a drag onto the message pane (and a paste) reaches the wire with no confirm. That is deliberate, not an oversight — the header of `pickerUpload.ts` argues it: a drop is a gesture the operator aimed and Ctrl-V is one they typed, the orchestrator owns the queue so an unauthorised batch has no business taking the channel's upload slot, and the picker path must skip `categoryOf` because iOS labels a `.m4r` `application/octet-stream` and only `normalizeUploadFile` inside the orchestrator rescues it.

**What this PR has.** The gate moves into `triggerUploads` itself — one door, so picker, drop, paste and any fourth surface added later inherit it. The same file's comment answers main's objection head-on (normalisation happens before the preview; nothing enters the queue until Send).

Both are defensible. They are also mutually exclusive: stacking this PR on main as-is gives the picker **two** confirms in a row.

### If the one-door design wins (this PR)

1. Rebase onto `main`.
2. Delete the gate in `pickerUpload.ts` and route `ComposeBox.tsx:541` back through `triggerUploads` — otherwise the picker confirms twice. Keep `pickerUpload`'s no-`categoryOf` note somewhere: it is the reason the picker cannot reuse `dropUpload`.
3. Keep from this branch: `lib/uploadOrchestrator.ts` (gate + `enqueueUploads` split + the `dismissConfirm()` in `resetUploadsForTests`), `lib/uploadConfirmBody.tsx` and its test, `__tests__/uploadOrchestrator.test.ts`, and the e2e `uploads4-pane-drop.spec.ts` — that last one is the only test in either PR that pins the drop path.
4. Drop from this branch everything #1884 already landed, or the rebase will fight over it: `ConfirmModal.tsx`, `lib/confirmDialog.ts`, `themes/default.css`, `lib/grappa_web/plugs/security_headers.ex` + `test/grappa_web/plugs/security_headers_test.exs`, `e2e/tests/media-link-cross-host-modal.spec.ts`, `e2e/tests/nginx-csp-range-parity.spec.ts`, `e2e/tests/uploads3-multifile.spec.ts`, `docs/DESIGN_NOTES.md`. Take main's version of each and re-apply only what is genuinely new.
5. `e2e/fixtures/uploadJourney.ts`, `__tests__/lifecycle.test.ts`, `e2e/tests/issue80-paste-flood-guard.spec.ts`, `e2e/tests/issue1256-upload-charset.spec.ts`: both branches touch them, so merge by hand — main's confirm is picker-shaped and this branch's is batch-shaped.

### If main's picker-only design stands

Close this PR, but salvage `uploads4-pane-drop.spec.ts` as a standalone test that pins the **intended** behaviour (drop does not confirm), so the next person who "fixes" the drop path has to argue with a red test instead of guessing. Then either the pane-drop gap gets its own issue, or `pickerUpload.ts`'s header note ("extending the guard to those doors later is a call to `pickerUpload` from them") becomes that issue.

I have no standing to pick between the two designs — it is a cicchetto ownership call. What is measured above is only the state of `origin/main` and of this branch.

vjt:
## Review outcome

Reviewed on a pinned tree (`ca29dbf46`), by reading code and grepping only — no test run,
no gate re-run, so nothing below is a fresh CI result. Posting the outcome here so it is
on the PR rather than in a chat window.

**No correctness blockers.** The `triggerUploads` (asks) / `enqueueUploads` (private,
enqueues) split answers #1884's mechanical objection instead of working around it: nothing
enters the queue before Send, and that is asserted rather than merely stated. The test
migration does not weaken coverage — the negative gets stronger, from "`triggerUploads`
was called" to "nothing reaches `host.upload` before Send".

### The two things flagged when this went out for review

1. **The iOS `.m4r` case is covered.** `onPickerChange` calls `triggerUploads` directly —
   never `dropUpload`, which is the only path that pre-filters on `categoryOf` — and
   `normalizeUploadFile` maps `.m4r` to `audio/mp4`. Asserted on the *type that reaches the
   host*, not on a call having happened.

2. **No double confirmation on the picker.** There is one on **paste**, though:
   `uploadPastedText` → `dropUpload` → `triggerUploads`, on both branches of
   `routeGuardedText`. It is written down in DESIGN_NOTES as an accepted cost, not hidden —
   and it is plausibly the exact friction that produced the request for an opt-out.

### Two things that want an answer before merge

**a) A new silent loss of the OS share target.** `consumeShare` → `dropUpload` →
`triggerUploads` now opens a confirm at boot, with no gesture on screen. `confirmDialog` is
last-write-wins (one request at a time — a modal is a focus trap), so any other confirm
replaces it and the shared files disappear **with no banner**: that outcome does not go
through `recordShareTargetBlock`. Before this PR, a shared file uploaded immediately and
could not be lost this way.

**b) A constraint for the opt-in/opt-out round that is coming.** `normalizeUploadFile` runs
*inside* `triggerUploads`; `enqueueUploads` does not normalize. So an opt-out written as

```ts
if (!confirmEnabled()) return enqueueUploads(...)
```

skips normalization and re-breaks the `.m4r` case — i.e. it reopens exactly the hole this
PR closes. The cure is trivial: normalize **before** the policy branch, not after it.

### On the "no don't ask again" principle

Following up on the design note in
https://github.com/vjt/grappa-irc/issues/1883#issuecomment-5484194231: that principle is
**prose plus one assertion**, not a structural lock. The only test that a switch would turn
red is `it("has no remember-me door — a second pick confirms again")`, and it is red only
under *opt-in*; under *default-on with opt-out* it stays green. There is no type-level
constraint to unpick — it costs an `if`.

The distinction drawn in that note holds up under measurement: the flag behind #1883 is
`localStorage`, key `image-upload-privacy-acknowledged:<host.id>` — per-browser, invisible,
not revocable from the UI. A user setting is per-user, server-side and visible. Not the
same object. So the contradiction is rhetorical rather than mechanical: worth weighing,
not fatal. (Worth noting `PrivacyModal.tsx` already has a "remember" checkbox of its own.)

### One number nobody had

Upload doors that now sit behind a modal go from **1 to 5**: picker, drop, pasted file,
paste-as-`.txt`, OS share. That is a count of call sites, not an estimate. Whether the
render wedge in #1667 touches any of them is **not established** — nothing here claims it
does.

### One correction to an earlier description of this PR

`cicchetto/e2e/tests/uploads4-pane-drop.spec.ts` is **not a new file**. It is a
pre-existing spec from #351 with 7 lines grafted into it. Earlier summaries — including
one of mine on IRC — described it as a new e2e; that was wrong.

Merge is not this review's call: one-door vs picker-only is a product decision, and the
opt-in/opt-out decision now rests on top of it.

gmsrc:
Pushed three commits — the opt-in round vjt's review anticipated, plus an ordering bug found in testing.

## The confirm is now an opt-in setting, default OFF

Per-user, server-side, beside upload retention in the settings drawer (`user_settings.data` key `upload_confirm_enabled` — a JSON key in the existing column, no migration; `false` stored as absence).

**This reverses this branch's own stated position, deliberately.** Both the #1883 note and the `triggerUploads` header argued there must be no "don't ask again". That argument was aimed at the flag which PRODUCED the defect: `localStorage`, per-browser, invisible, not revocable from the UI. This is a different object — per-user, server-side, visible, revocable — so the objection is rhetorical rather than mechanical, exactly as the review put it. The cost is written down rather than hidden: off by default, all five doors are unguarded until someone opts in.

The contested assertion was **narrowed, not deleted**. `has no remember-me door` is red under opt-in; what survives is that while the confirm is ON there is no per-dialog "don't ask again" — turning it off is a deliberate trip to settings, not a checkbox on the way past.

## Constraint (b) from the review is honoured and asserted

> `normalizeUploadFile` runs *inside* `triggerUploads`; `enqueueUploads` does not normalize. An opt-out written as `if (!confirmEnabled()) return enqueueUploads(...)` skips normalization and re-breaks the `.m4r` case.

The branch takes the **normalised** list, and a test pins `file.type === "audio/mp4"` for an iOS `.m4r` on the confirm-OFF path — asserted on the type that reaches the host, not on a call having happened.

## Blocker (a) is fixed

`ConfirmRequest` gains an optional `onDisplaced`, fired when a pending request is *replaced* (never when the operator answers it). The share path passes one that records a new `confirm-displaced` block reason, so a shared file whose confirm is displaced at boot now gets a banner instead of vanishing. The callback is optional and the asymmetry is deliberate: every other door is a gesture still on screen, so "ask again" is the operator repeating it; the share target has nothing to repeat.

## New: the privacy notice ran in the wrong order

Found in testing. The Send confirm opened first and the "files go to `<host>`" notice second — the operator learned where their file was going only after agreeing to send it. The gate sat in `startUpload`, inside the queue pump, downstream of the confirm by construction; before the confirm moved into `triggerUploads` there was no second modal for it to be out of order with.

It now runs at the front of `triggerUploads`, ahead of both doors. The old gate was **removed** rather than kept as a belt — two gates would ask a non-remembering operator twice per batch. Accepted change: that operator is now asked once per BATCH rather than once per FILE.

**Known rough edge, not fixed here:** "Don't show this again" is still a one-way door, revocable only by deleting the localStorage key. That is the same criticism #1883 levelled at the flag, and now that upload preferences have a settings home the reset belongs next to the opt-in. Happy to add it here or take it separately — say which.

## Gates

- `user_settings_test.exs` 149 tests / `user_settings_controller_test.exs` 86 tests, 0 failures.
- Full cic suite: **5 failures against a measured baseline of 5** on `ca29dbf4`; all five pass in isolation (the cross-file pollution this host is known for).
- `tsc` + build, `mix format`, `credo --strict`, design-notes gate: clean.
- No wire-shape change — settings endpoints sit outside the generated wire schema, so no `protocol_version` bump.

Verified by hand on azzurra/#sbiffo: toggle off → uploads go straight out; toggle on → notice, then confirm with previews.

PR #1887, opened 2026-09-01
DO NOT MERGE - temporary negative control: integration on an unmodified main tree

vjt:
## DO NOT MERGE. DO NOT LABEL. THROWAWAY BY CONSTRUCTION.

This pull request exists to **run CI, and nothing else**. It should be
shut down again once the run it schedules has been read. It carries no
behavioural change of any kind.

### What is in it

One commit, one file, seventeen added lines, all of them a comment
block at the top of `cicchetto/e2e/playwright.config.ts`. Zero
deletions. The branch point is `725788267`, the current tip of `main`,
so the tree under test is `main` itself.

The comment is there only to make the diff match the
`cicchetto/e2e/**` entry in `.github/workflows/integration.yml`. That
workflow declares no `workflow_dispatch`, so a pull request is the only
way to schedule the integration suite on demand.

### Why it exists

The evidence currently blocking PR 1886 is an association: one e2e spec
failed three times out of three on that branch, against zero failures
out of eight on `main`. Every named mechanism for it has since been
measured and ruled out.

The remaining problem is in the shape of the comparison, not in the
branch. The three failures are **three reruns of a single run id** —
one runner pool, one hour. The eight passes are runs on different SHAs,
on different days, on other pools. So the contrast is not
branch-versus-main; it is *this pool, tonight* versus *other pools,
earlier*. If the spec is sensitive to load or to time of day, the
association with the branch is an artefact of the design.

This branch is the missing negative control: the same tree as `main`,
measured in the same pool and the same hour as the arm it is being
compared against.

### How to read the result

Both outcomes are informative, and the reading is written down here
before the run so it cannot be adjusted afterwards.

- **The spec fails here too** — the branch is acquitted. The cause is
  environmental, and the three-out-of-three becomes a fact about the
  pool rather than about the code.
- **This run is green** — the association survives a sound design. It
  still has no mechanism, and the decision on PR 1886 belongs to a
  human.

One run does not settle the first case on its own: a single green here
is exactly as weak as a single red would be. Repetition is a separate
call.

### DIFF
--- a/cicchetto/e2e/playwright.config.ts
+++ b/cicchetto/e2e/playwright.config.ts
@@ -1,3 +1,20 @@
+// TEMPORARY NEGATIVE CONTROL - DO NOT MERGE, DO NOT KEEP.
+//
+// This comment block is the ENTIRE content of branch w1-ctrl-1796. It
+// exists only so that this path matches the `cicchetto/e2e/**` entry in
+// .github/workflows/integration.yml, which is what makes the integration
+// suite run at all. `integration.yml` carries no `workflow_dispatch`, so
+// a pull request is the only way to schedule that suite on demand.
+//
+// WHY. The red-versus-green comparison behind issue 1796 on PR 1886 is
+// three reruns of ONE run id against eight runs spread over other days
+// and other runner pools. That design cannot separate "the branch" from
+// "this pool, this hour". This branch supplies the missing negative
+// control: the origin/main tree, unmodified in behaviour, measured in
+// the same pool and the same hour as the arm it is compared against.
+//
+// Nothing below is touched. Remove this block and the branch with it.
+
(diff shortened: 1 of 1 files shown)


vjt:
Throwaway negative control, never for merging.

What it measured: the branch under test was suspected of causing `issue1796-reconnect-bounces-network.spec.ts:56` [chromium] to fail. This PR carried `origin/main` PURE (the only delta a comment block, added_noncomment=0) and ran the same suite on the same runner pool in the same window — and went RED on the exact same spec, sole failure out of 235, 6.5s.

Reading was pre-registered before the run: control red implies the branch is absolved. It was red. The mechanism of the flake stays UNKNOWN; issue 1796 stays open and this changes attribution, not explanation.

PR #1948, opened 2026-09-06
/notify: ISON polling fallback for ircds with neither MONITOR nor WATCH

### LINKED ISSUE #1946: /notify is silently dead on IRCnet — no MONITOR/WATCH, and no ISON fallback
### Executive summary
**Current:** `/notify` never works on IRCnet: the ircd has neither MONITOR nor WATCH, grappa probes both, gets two `421`s, resolves `:none` and leaves every watched nick `:unknown` forever.
**Fix:** make `:none` fall through to ISON polling — the phase-2 fallback #247 deferred. Affordable: on this ircd ISON is the cheapest command tier (`m_ison` returns a penalty of **1 second**, ~2 s per full chunk including the per-message base), so a ~30 s sweep scaled to list size sits well inside the flood budget IRCnet advertises as `PENALTY`.

### Current behaviour

Image

IRCnet's 005, captured live:

```
RFC2812 PREFIX=(ov)@+ CHANTYPES=#&!+ MODES=3 CHANLIMIT=#&!+:42 NICKLEN=15
TOPICLEN=255 KICKLEN=255 MAXLIST=beIR:64 CHANNELLEN=50 IDCHAN=!:5
CHANMODES=beIR,k,l,imnpstaqrzZ PENALTY FNC EXCEPTS=e INVEX=I
CASEMAPPING=ascii NETWORK=IRCnet
```

No `MONITOR`, no `WATCH`. So `ISupport.presence_mechanism/1` → `:none`, the
005-independent arm probes WATCH, takes a 421, downgrades to MONITOR, takes a
second 421, and lands on `:none`. `Presence.arm_commands(:none, _)` returns
`[]`. Nothing is sent again; dots stay grey until reconnect. It logs honestly:

```
presence: MONITOR also unknown — no watch mechanism on this ircd;
watched nicks stay :unknown until reconnect
```

**Not a bug in the probing.** The 005-independent arm is a deliberate ruling
(DESIGN_NOTES 2026-07-19, *"the 005 advertisement is a HINT, not a gate"*)
because bahamut forks support WATCH without advertising it. That stays. The
gap is only that `:none` is a dead end.

### This was foreseen

#247 deferred it with a named trigger condition:

> ISON fallback can be phase 2 **if a network without MONITOR/WATCH ever matters**.

and DESIGN_NOTES 2026-07-18: *"ISON polling fallback deferred (phase 2) — a
no-mechanism network logs honestly and keeps `:unknown` dots."* IRCnet is that
network.

### Research — read from `ircnet/ircd` source

- `common/parse.c` message table has **no WATCH, no MONITOR** entry. `ISON` is
  there with `maxpar = 1` (whole nick list arrives as one parameter).
- **Penalty**: a handler's return value *is* the penalty in seconds
  (`cptr->since += ret`), throttled at `cptr->since - timeofday < MAXPENALTY`.
  `m_ison` returns **1** — cheapest tier (`m_whois` 2, `m_who` up to MAXPENALTY).
  Plus a pre-dispatch base of `1 + len/100`.
- 🔴 **`m_ison` truncates silently.** It fills the reply and then:
  ```c
  if (len + i > sizeof(buf) - 4) { break; }   /* leave room for " \r\n\0" */
  ```
  No error, no marker. The cap is on the **reply**, which holds only the
  **online** nicks — so a request-sized chunk can still overflow and every
  dropped nick reads as offline. That fabricates "X went offline" pushes.
- RFC 2812 §4.9: `ISON <nick> *( SPACE <nick> )` → `303 RPL_ISON`, stateless,
  capped by line length not count.

### Suggested design

**ISON is the terminal fallback, not a fourth probe rung.** MONITOR/WATCH are
optional so a 421 is a real answer; ISON is RFC-mandatory, there is nothing to
discover. Chain becomes:

```
005 MONITOR → MONITOR          005 WATCH → WATCH
005 neither → probe WATCH --421--> probe MONITOR --421--> ISON --421--> :none
```

`:none` survives for the impossible case, per no-silent-drops.

**Reuses unchanged:** `Grappa.Notify` list, `Presence.seed/1`, the
baseline-vs-transition rule (`:unknown → X` = `:initial`, `X → Y` =
`:transition`), `{:presence_changed, …}` → `dispatch_presence/4`,
`chunk_by_budget/2`.

**The 20% that does not fit:**

1. **Poll, not arm.** `:ison` gets no `arm_commands/2` clause — that seam
   returns commands to send *once*. It needs a timer in `Session.Server`
   (`Process.send_after`), armed when the mechanism resolves, re-armed per
   sweep, polling `state.presence`'s key set (not a DB read every 30 s).
   **Empty list → no timer at all.**
2. **Transitions are derived** by diffing the returned set. The existing map
   already gives baseline-vs-transition for free.
3. **Truncation safety, the load-bearing one.** Size chunks so the reply fits
   even if *every* queried nick is online — with `NICKLEN=15` that is **~25
   nicks**, not the ~35 a request-side budget allows. And **count the
   replies**: N chunks must yield N × 303; fewer → the sweep is incomplete,
   **diff nothing**. Absence of evidence must never become an offline event.
4. **No list-full concept** — no `ERR_TOOMANYWATCH` equivalent, so the `512`
   gate stays mechanism-specific.

**Cadence** (~2 penalty-seconds per full chunk):

| watched | chunks | penalty/sweep | @30 s |
|---|---|---|---|
| 25 | 1 | ~2 s | 7% |
| 60 | 3 | ~6 s | 20% |
| 128 | 6 | ~12 s | 40% — too hot |

Scale the interval with list size, as #247 sketched.

### Out of scope

- **Spilling a full MONITOR/WATCH list into ISON** (mIRC does this) — mixes two
  mechanisms in one session; `state.presence_mechanism` caches one.
- **Correlating 303s to chunks** — replies do not echo the request, so counting
  is the only option. A stray user `/quote ISON` mid-sweep costs one skipped
  sweep.

### Open question

`MAXPENALTY`'s numeric value — the mechanism is traced but the `#define` was
not found in the files read. It decides whether 30 s is comfortable or tight.

### Also noted, separate concern

IRCnet advertises `CHANTYPES=#&!+` and `IDCHAN=!:5` (safe channels with a
5-char ID prefix). Unrelated to presence, but worth checking grappa's chantypes
handling has met `!` and `&`.

gmsrc:
Closes #1946. Implements the design agreed there, with one deviation and two bugs the design did not anticipate — both found by running it against real IRCnet, neither reachable from unit tests.

## What was broken

`/notify` was inert on IRCnet. The ircd has neither mechanism, so the 005-independent arm probed WATCH → `421` → MONITOR → `421` → `:none`, where `arm_commands/2` returns `[]` and every watched nick stays `:unknown` until reconnect. Confirmed live:

```
presence: WATCH unknown on this ircd — falling back to MONITOR (3 watched nicks)
presence: MONITOR also unknown — falling back to ISON polling      ← new
```

## Design

**ISON is the terminal fallback, not a fourth probe rung.** MONITOR and WATCH are optional, so a `421` is a real answer about them; ISON is RFC 1459/2812 mandatory, so it is used rather than probed. `:none` survives only for an ircd that refuses even ISON (no-silent-drops). The 005 hint still wins when present, and the optimistic WATCH probe stays — its reason (bahamut forks supporting WATCH unadvertised) is untouched.

**The reply budget is the correctness problem.** IRCnet's `m_ison` fills its reply buffer and `break`s, dropping the tail with no error:

```c
if (len + i > sizeof(buf) - 4) { break; }   /* leave room for " \r\n\0" */
```

The reply carries only the **online** nicks, so a chunk sized against the REQUEST can overflow the reply and turn live users offline — fabricating "X went offline" pushes. Hence `@ison_reply_budget 360` (vs 400 for MONITOR/WATCH), sized so the reply fits even if every queried nick is online, and the sweep rule: **N chunks must yield N × 303, or the sweep is discarded and nothing is diffed.**

**Cadence from a measured penalty.** In this ircd a handler's return value IS its penalty in seconds (`cptr->since += ret`); `m_ison` returns 1, the cheapest tier. A full line costs ~2 penalty-seconds, so: 30 s floor, +10 s per chunk, and an empty watch list arms **no timer at all**.

## Deviation from the issue

The issue said `:ison` should get no `arm_commands/2` clause. Built that way it crashes: the live `/notify add` path calls it with whatever mechanism the session resolved, and a `FunctionClauseError` there kills a working session over a nick the next sweep picks up anyway. It returns `[]` instead, and `sync_presence/3` re-schedules the timer — which it must regardless, since 0 → N has to create the timer `poll_interval_ms(0)` refused.

## Two bugs only live testing found

1. **`Wire.presence_changed/6` had no `:ison` clause.** Thirty seconds after the downgrade the first sweep crashed the session with a `FunctionClauseError`, which then `:transient`-restarted into the same wall.
2. **`source` is a CLOSED SET on the wire.** Widening it server-side against a live old bundle made cic drop every presence frame — *"This client could not read a presence_changed update from the server and discarded it"* — because `wireSchema.ts` carried `{ e: ["monitor", "watch"] }`.

So a closed set gaining a member is a wire-shape change, in the new-server-to-old-client direction, **measured rather than argued**. Protocol **11 → 12**, and unlike v10/v11 `wire_pin` demanded it — `source` lives in a `Session.Wire` typespec, so the digest actually moved (`9c97bd9b…` → `4e6ee9e6…`). `@min_protocol_version` stays 1: an old client drops presence frames and keeps every other pane, only on the network that needs ISON at all.

Regenerating was not sufficient: `src/lib/api.ts` carries a hand-written copy that `wireTypesAssert.ts` pins equal to codegen, so `tsc` stayed red until that was widened too.

## Verified by hand

Live on IRCnet (`open.ircnet.net`, `#sbiffo`) and Azzurra side by side: the chain downgrades, sweeps run on the timer (`command=ISON` ×6 in the first minutes), dots light, and Azzurra is untouched — it still resolves WATCH from 005 and never reaches the new rung.

## Gates

All run locally on the rebased tree:

| gate | result |
|---|---|
| `mix test` | **7142 tests, 0 failures** |
| `mix credo --strict` | no issues (6728 mods/funs) |
| `mix dialyzer` | **0 errors, passed successfully** |
| `mix grappa.wire_pin --check` | agrees at protocol 12 |
| `bun run check` | 5 stages, 0 failed |
| `scripts/design-notes-gate.sh` | pass |
| `scripts/union-rebase.sh` | `+89 -0 — contribution intact` |
| cic vitest | 7 fails, all 4 files green in isolation (known cross-file pollution; identical set pre- and post-rebase) |

`integration` cannot run on the dev host, so it was audited instead: `e2e/tests/issue247-notify-watch.spec.ts` runs on azzurra-testnet (bahamut, advertises WATCH) and never reaches the ISON rung.

**Deploy: COLD** — server behaviour + wire version.

### DIFF
--- a/cicchetto/src/lib/api.ts
+++ b/cicchetto/src/lib/api.ts
@@ -1372,7 +1372,9 @@ export type WireUserEvent =
       nick: string;
       presence: "online" | "offline";
       initial: boolean;
-      source: "monitor" | "watch";
+      // #1946 — `ison` is the polling fallback for ircds with neither MONITOR
+      // nor WATCH (IRCnet). Pinned equal to codegen by wireTypesAssert.ts.
+      source: "monitor" | "watch" | "ison";
       ts: string;
     }
   | { kind: "presence_error"; network_id: number; reason: "list_full"; detail: string }
--- a/cicchetto/src/lib/wireSchema.ts
+++ b/cicchetto/src/lib/wireSchema.ts
@@ -1183,7 +1183,7 @@ export const S_SessionWirePresenceChangedPayload = {
     nick: "s",
     presence: { e: ["online", "offline"] },
     initial: "b",
-    source: { e: ["monitor", "watch"] },
+    source: { e: ["monitor", "watch", "ison"] },
     ts: "s",
   },
 } as const;
--- a/cicchetto/src/lib/wireTypes.ts
+++ b/cicchetto/src/lib/wireTypes.ts
@@ -1066,7 +1066,7 @@ export type SessionWirePresenceChangedPayload = {
   nick: string;
   presence: "online" | "offline";
   initial: boolean;
-  source: "monitor" | "watch";
+  source: "monitor" | "watch" | "ison";
   ts: string;
 };
 
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -46703,3 +46703,92 @@ _Deploy: **COLD** on every substrate. Measured, not reasoned:
 `:linux` and `:docker` alike (positive control: the `lib/` comment-only paths
 alone return `{:hot, []}`). The release image also has to be REBUILT for the
 baked ENV to move — the config change alone does not reach it._
+<!-- entry #1946 -->
+
+---
+
+## 2026-09-06 — #1946: ISON presence fallback, and two seam bugs only live testing found
+
+**`/notify` was inert on IRCnet.** Its ircd (2.11.2, RFC 2810–2813 lineage)
+implements neither MONITOR nor WATCH — confirmed in `common/parse.c`'s message
+table and in a live 005 that carries neither token. So the 005-independent arm
+probed WATCH, took a `421`, downgraded to MONITOR, took a second `421`, and
+resolved `:none`, where `arm_commands/2` returns `[]` and every watched nick
+stays `:unknown` until reconnect.
+
+#247 had already named this exact condition: *"ISON fallback can be phase 2 if a
+network without MONITOR/WATCH ever matters."* It now matters.
+
+### ISON is the terminal fallback, NOT a fourth probe rung
+
+MONITOR and WATCH are optional extensions, so probing them is meaningful — a
+`421` is a real answer. ISON is RFC 1459/2812 mandatory: there is nothing to
+discover, so it is *used* rather than probed. `:none` survives only for the
+ircd that refuses even ISON, because no-silent-drops says the impossible case
+still needs a name.
+
+The 005 hint keeps winning when present, and the optimistic WATCH probe stays —
+its reason (bahamut forks that support WATCH without advertising it) is
+untouched.
+
+### The reply budget is the correctness problem, not the request budget
+
+IRCnet's `m_ison` fills its reply buffer and then `break`s
+(`if (len + i > sizeof(buf) - 4) break;`), dropping the tail with no error and
+no marker. The reply carries only the **online** nicks, so a chunk sized
+against the REQUEST can still overflow the reply — and every dropped nick then
+reads as offline, fabricating "X went offline" pushes.
+
+Hence `@ison_reply_budget 360` (below the 400 used for MONITOR/WATCH), sized so
+the reply fits even if every queried nick is online. And hence the sweep rule:
+**N chunks must yield N × 303, or the sweep is discarded and nothing is
+diffed.** Absence of evidence must never become an offline transition.
+
+### Cadence is budgeted against a measured penalty
+
+In this ircd a handler's return value IS its penalty in seconds
+(`cptr->since += ret`); `m_ison` returns **1**, the cheapest tier (`m_whois` 2,
+`m_who` up to MAXPENALTY), plus a pre-dispatch base of `1 + len/100`. So a full
+ISON line costs ~2 penalty-seconds and the interval scales with chunk count:
+30 s floor, +10 s per chunk. An empty watch list arms **no timer at all** —
+most sessions watch nobody and must pay nothing.
+
+### Deviation from the issue's design, and why
+
+The issue said `:ison` should get no `arm_commands/2` clause, so nobody could
+"arm" a poll and wait forever for a push. Built that way it crashes: the live
+`/notify add` path calls `arm_commands/2` with whatever mechanism the session
+resolved, and a `FunctionClauseError` there takes down a working session over a
+nick the next sweep would have picked up anyway. It returns `[]` instead, and
+`sync_presence/3` re-schedules the timer — which it must regardless, since
+0 → N has to create the timer `poll_interval_ms(0)` refused.
+
+### 🔴 Two bugs that unit tests could not have caught
+
+Both live at the server↔client seam, and both were found by running the thing
+against real IRCnet:
+
+1. **`Wire.presence_changed/6` had no `:ison` clause** — guarded
+   `source in [:monitor, :watch]`. Thirty seconds after the downgrade the first
+   sweep produced `source: :ison` and the session crashed with a
+   `FunctionClauseError`, then `:transient`-restarted into the same wall.
+2. **`source` is a CLOSED SET on the wire**, and widening it server-side while
+   an old bundle was live made cic drop every presence frame with *"This client
+   could not read a presence_changed update from the server and discarded it"*.
+   That is the validator working: `wireSchema.ts` carried
+   `{ e: ["monitor", "watch"] }`.
+
+So a closed set gaining a member IS a wire-shape change, in the
+new-server-to-old-client direction, and it is **measured** here rather than
+argued. Protocol went **11 → 12** and — unlike v10 and v11 — `mix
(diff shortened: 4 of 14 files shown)


vjt:
Partial review from the orchestrator. **Scope declared up front: I reviewed the wire-contract surface only.** The other ~600 lines — the reply-budget arithmetic, the sweep-discard rule, the cadence — I have **not** read. Treat the silence on those as silence, not as approval.

## What I verified, and how

**The `shape.pin` bump is internally consistent — reproduced, not trusted.**
`sha256(wireTypes.ts ++ "\n" ++ wireSchema.ts)`:

| tree | computed | pin in tree |
|---|---|---|
| `origin/main` | `9c97bd9b75bf362a…` | `9c97bd9b75bf362a…` ✅ (positive control) |
| this PR | `4e6ee9e64c05d3a9…` | `4e6ee9e64c05d3a9…` ✅ |

The no-`\n` variant differs in both cases (`80b27aaa…` / `b890144e…`), so the instrument discriminates rather than agreeing by construction. `@protocol_version` moves 11→12 **together with** the digest, which is what the bump-on-every-wire-shape-change rule asks for.

**ISON is not absent from the tree today — the *mechanism* is.** My first grep was case-insensitive and matched *compar**ison***; case-sensitive it reads: `Parser` maps `"ISON" => :ison`, `Message` carries `:ison` in the type, `303` is documented in `NumericRouter` but deliberately unrouted, and three sites declare the gap explicitly (`ISupport` ×2, `Presence`). So this fills a hole the codebase had already written down as out-of-v1, rather than inventing one.

**The `NumericRouter` change engages the prior decision instead of walking past it** — it keeps both earlier claims true (`/quote`-only reachability, trailing space failing `valid_nick?/1`) and then names what changed: a timer now *sends* it, so an undelegated `303` would persist a `$server` notice row every sweep. That is the right way to overturn a documented deliberate choice.

## Two things a green CI cannot answer

Neither is reachable by any check on this PR, and neither is in the scope of the design ok recorded on issue 1946 (which names four items: ISON as terminal fallback, poll-not-arm, transitions by diffing, reply-side chunking).

1. **`min_protocol_version` stays at `1`.** The argument in the diff is "degraded rather than unserviceable". The measured cost is that an old bundle discards **every** `presence_changed` frame on that network. That is the wire-contract axis — the one reserved for "old clients can no longer be served" — not an ISON-design question. Worth an explicit ruling rather than inheriting the default.
2. **cic's `wireSchema` is stricter than the contract it implements.** The contract says unknown-is-never-fatal in both directions; the generated schema drops the whole payload on an enum member it does not know, which is exactly what the PR measured live. Same class as the archive case in issue 1626. The bump here is the correct defensive move, but the client-side non-compliance survives it and is not this PR's to carry.

## One declared deviation, noted as sound

The issue specified no `arm_commands/2` clause for `:ison`; built that way it raises `FunctionClauseError` on the live `/notify add` path and kills a working session. Returning `[]` is a departure from the approved design, it is **declared with its failure mode**, and that is the accepted way to break a spec here.

CI at `d08ce778`: 9/9 complete, zero failures.

Issue #1973, opened 2026-09-07
cic declares CLIENT_PROTOCOL_VERSION 9 while the server speaks 13 — the skew warn fires on every healthy boot


vjt:
## What

`CLIENT_PROTOCOL_VERSION` in `cicchetto/src/lib/socket.ts:104` is stale again. On
`origin/main` it is `9`; `Grappa.Protocol`'s `@protocol_version` is `13`
(`lib/grappa/protocol.ex:292`). Production (1.5.1) speaks `11`.

So `noteServerProtocol` (`cicchetto/src/lib/socket.ts:441-451`) logs, on every boot of
every current bundle against every current server:

```
[grappa] protocol mismatch: this bundle speaks 9, the server speaks 11
```

This is the exact defect the file's own note already records for `2 → 9`
(`socket.ts:82-86`): *"a true statement about a stale constant, not about a real
incompatibility"*. It came back, because nothing pins the two numbers to each other in
this direction.

## Why it is not cosmetic

The warn exists to be the one signal that a service-worker-cached PWA is skewed against
the BEAM — `socket.ts:428-433` says so: *"on a service-worker-cached PWA a bundle/BEAM
skew is the single most likely explanation for otherwise-inexplicable behaviour, and it
is invisible from every other signal."*

A signal that fires on every healthy boot cannot carry that meaning any more. It already
misfired: on 2026-09-07, while narrowing down a separate report about `/notify` not
opening the settings drawer on one device, a tester opened the console, read
`protocol mismatch: bundle speaks 9, but server speaks 11` and offered it as the likely
cause. It was not — every current bundle prints it.

## Measured

On `origin/main`:

| fact | value | where |
|---|---|---|
| cic speaks | `9` | `cicchetto/src/lib/socket.ts:104` |
| cic server floor | `9` | `cicchetto/src/lib/serverProtocol.ts:92` |
| server speaks | `13` | `lib/grappa/protocol.ex:292` |
| server floor | `1` | `lib/grappa/protocol.ex:293` |
| prod (1.5.1) speaks | `11` | observed in a browser console |

cic already **knows** the shapes the constant claims it does not:

- v12's `presence_changed.source: "ison"` → `cicchetto/src/lib/wireSchema.ts:1186` has
  `{ e: ["monitor", "watch", "ison"] }`.
- v13's `stale_code_path` refusal token → `cicchetto/src/lib/wireSchema.ts:1687`.

So the bundle in fact speaks 13 and declares 9.

[...]

Issue #1914, opened 2026-09-05
cicchetto: the JOIN-time topic never reaches scrollback — only the topic bar has it


vjt:
On JOIN, cicchetto shows the channel topic only in the topic bar. The topic never lands in the window's scrollback, so a user who scrolls their history sees no record of what the topic was when they walked in — they have to look up at the bar, which on a dark theme is a multi-line low-contrast block.

irssi (and x-chat) print the server's reply as ordinary window lines:

```
17:02 -!- Topic for #grappa: beta — https://grappa.chat
17:02 -!- Topic set by vjt [Fri Sep  5 09:11:40 2026]
```

and every later change as `-!- X changed the topic of #grappa to: ...`. cicchetto already does the second one; it is missing the first.

### Where it is today

This is deliberate, not an accident — three places encode it:

- `lib/grappa/session/event_router.ex:1018-1022` — "332 RPL_TOPIC: JOIN-time backfill — stores topic text in the topics cache. **Does NOT produce a scrollback row** (spec: `:topic` rows come ONLY from the TOPIC command)". Same for 333 right below it.
- `lib/grappa/session/event_router.ex:83-90` (moduledoc, "Topic numerics (Q-non-blocking)") — "The topic-bar in P4-1 reads live state, not scrollback".
- `lib/grappa/session/numeric_router.ex:715-716` — 332/333 are in the suppress list, precisely so `Server`'s catch-all does not persist them as bare `:notice` rows leaking the raw trailing param (333 would print the unix timestamp verbatim).

So both routes to scrollback are closed on purpose. The fix is not to drop the suppression (that brings back the raw-numeric noise #1505's neighbours were about) but to emit a **typed** JOIN-time topic row.

### What it should do

On a JOIN that yields 332 (and its 333, which may arrive after), persist one typed scrollback row for the channel carrying `text`, `set_by`, `set_at`, distinguishable from the existing `:topic` row (which means "someone just changed it"). `ScrollbackPane` renders it as the JOIN-time topic, not as "X changed the topic".

Points to settle in the build:

- **Ordering.** 332 and 333 can arrive in either order; the row must not render half-empty. Either fold on the 333 (or on the 366 that closes the JOIN burst) and emit once, or emit on 332 and enrich.
- **No topic set.** 331 RPL_NOTOPIC is also suppressed today. Decide whether it produces a row at all — irssi does not print one worth keeping.
[...]

Issue #1850, opened 2026-08-27
Release cuts cold-restart prod: the OTP vsn in the release code path


vjt:
## What forces the cold

A release-only `VERSION` bump is COLD on the two `mix release` substrates
(`:jail`, `:linux`) and HOT on `:docker`. That asymmetry is not a policy
choice — it is the release layout:

- **Docker** execs `mix phx.server` over a bind-mounted tree, so
  `:code.lib_dir(:grappa)` is `/app/_build/<env>/lib/grappa`: **no vsn in the
  path**. Recompiled beams land exactly where the live node is already
  looking, and `/admin/reload` picks them up.
- **The jail and the native systemd host** boot a `mix release`, whose code
  path is `lib/grappa-<vsn>/ebin` (named as such in
  `infra/freebsd/deploy.sh`). `:code.lib_dir/1` — and everything downstream
  of it, including `Application.app_dir/1,2` — resolves to the **boot**
  directory. Bump the vsn and the new artifacts land in a **sibling the node
  never looks at**; the reload then diffs the stale tree against itself and
  answers `{"failed":[],"reloaded":[],"migrated":[]}`, which is
  indistinguishable from "nothing to do".

That miss cannot be reported, only prevented, which is why
`Grappa.Deploy.Preflight.version?/1` classifies the repo-root `VERSION` as
COLD on `@release_substrates` and only there. The failure is not
hypothetical: it was reproduced in production on 2026-08-13 on a
self-hosted `:linux` install going v1.0.0 → v1.1.0 — **~6.5 hours serving
the old BEAM under the new git history**.

Cost of the status quo, restated plainly: **every release cut drops every
IRC session**, on an always-on bouncer, for a one-line file.

## What is NOT the fix

Flattening the release layout — "take the vsn out of the path" — fights the
OTP contract rather than the defect. The boot script, the `RELEASES` file
and `:code.lib_dir/1` all agree on `lib/<app>-<vsn>/ebin`; a release built
against a flattened layout loses `release_handler` and gains a bespoke boot
path, for a gain visible only on the hot branch. Rejected before it is
proposed, so nobody re-derives it.

## The three real options

**A — appup/relup, the canonical Erlang answer.** `release_handler` adds the
new lib path to the running node and drives `code_change/3` on the
supervised GenServers, so `grappa-1.4.0` → `grappa-1.4.1` is a supported
in-place upgrade rather than a restart. This is what OTP built for exactly
[...]

Issue #1674, opened 2026-08-22
mentions: service and server NOTICEs that spell your own nick count as mentions — a NickServ login confirmation lights a red @1 on the $server window


vjt:
A user reported on IRC (with a screen recording) that the server tab carries a red `@1` mention pill that "never goes away" — it disappears while the server window is open and is back on the next tab switch.

## What it actually is

Measured read-only on prod (`azzurra`, `network_id = 1`, subject is a visitor; ids below are real, nick redacted):

* Row `id 2957172`, `channel = "$server"`, `kind = :notice`, `sender = "NickServ"`, `inserted_at 2026-08-22T14:38:09.348Z`. Body is the login confirmation — `Password accepted for <nick>. You are now identified.` — which contains the operator's **own nick**.
* `Grappa.Mentions.mentioned?/3` (`lib/grappa/mentions.ex:146`) is a **body-only** predicate, and `:notice` is part of `@content_kinds` (`lib/grappa/mentions.ex:73`, from `Message.content_kinds/0`). There is no sender-side exclusion for services or for the server itself.
* So the row folds into `Grappa.WindowCounts.count_mentions/6` and the `$server` window snapshot comes back `mentions: 1`, `severity: :mention`.
* cic renders it as the red mention pill in `cicchetto/src/WindowBadges.tsx` (`sidebar-mention` / `bottom-bar-mention`).
* The "it comes back" part is the documented focus-zero overlay in `cicchetto/src/lib/mentions.ts`: the selected + visible window renders `0` **without advancing the read cursor**, so the raw server value is unchanged and the pill returns as soon as another window is selected.
* It cleared only when the read cursor genuinely advanced: `read_cursors` for that subject / `$server` moved to `2958181` at `2026-08-22T15:16:33.823Z`, and the badge went away. That part is working as designed.

## The defect

Every layer behaves as specified — the wrong input is upstream. A service notice is not a highlight: being told by NickServ that your password was accepted, or by the ircd that your nick is owned by someone else, is not somebody mentioning you. Any service or server-originated notice whose body happens to spell your nick currently lights the highest-severity badge grappa has, on a window most operators never open, and it stays lit until they open it.

The same shape applies to the ircd's own notices (`nightwish.azzurra.chat -Notice- ...`), which routinely name the connecting nick, so this is not specific to NickServ.

## Direction

[...]

PR #1984, opened 2026-09-07
feat(162): /ignore — a per-network mask list, dropped at the door

### LINKED ISSUE #162: server: persisted /ignore list, honored server-side
## Summary

Add a per-account `/ignore` list, **persisted and honored server-side**. Users maintain a list of ignored nicks/masks; the server drops matching PRIVMSG/NOTICE (and optionally JOIN/PART/QUIT noise) before delivering to any of the account's connected clients.

## Why server-side

- Survives reconnects and is shared across all devices of the same session (multi-device is a core grappa property) — a client-only filter would have to be re-synced per device and would still receive + store the ignored traffic.
- The bouncer already holds per-session state; the ignore list is the natural extension. Filtering at delivery time also keeps ignored rows out of scrollback/REST history.

## Scope

- **Storage:** per-account ignore list, persisted (same durability as nickserv/session state).
- **Matching:** `nick!user@host` glob masks (at minimum bare-nick and `*!*@host`); define precedence/normalization.
- **Enforcement point:** server-side delivery filter for PRIVMSG/NOTICE from a matching origin. Decide whether ignored messages are also excluded from REST scrollback (`listMessages*`) — recommended yes, for consistency with delivery.
- **Management surface:** `/ignore <mask>`, `/unignore <mask>`, `/ignore` (list). REST endpoint to back the client UI.
- **Self/oper carve-outs:** never ignore self; decide behavior for server/services notices.

## Open questions (design)

- Mask syntax + wildcard semantics (mIRC-style vs simple globs).
- Do ignored events also disappear from history/scrollback, or only from live delivery?
- Interaction with channel vs private messages (ignore both? per-target?).

Priority p1 per channel request.

gmsrc:
Closes #162. Implements the drop-at-delivery design — the one the issue body specifies — with the two open questions resolved as recorded on the issue.

## What it does

`/ignore <mask>`, `/unignore <mask>`; a bare `/ignore` opens the new **settings → ignore list** sub-page. A bare nick means `nick!*@*`; full `nick!user@host` globs work (`*` and `?`). Per subject, per network, persisted, honoured **server-side**: a matching PRIVMSG or NOTICE is dropped at the head of `EventRouter.route/2`, so there is no row, no broadcast, and no push.

## Drop, not hide, not re-route

#162's body says drop; Lucy's variant re-routes to the server window. Dropped on the ruling that ignored is ignored — no un-hide. That also makes the issue's second ruling free: with no persist effect there is no trigger, so **"ignored mask sends a DM, subscription exists, zero pushes emitted" holds by construction**, not by a second gate beside `muted_targets` in `should_notify?/5`. If the re-route variant is preferred later, only the delivery arm changes — storage, matching, REST and commands all survive.

## Its own structure, beside `muted_targets`

Ruled separate, and the reason is mechanical: a mute is "this ROOM is noisy" and suppresses attention; an ignore is "this PERSON is a problem" and stops the message existing. One shared shape would carry a field meaningless on every mute row. Same conventions (network in the key per #1038, `canonical_target/1` fold), separate `"ignores"` key in `user_settings.data`. No migration.

## The filter's four exclusions, each pinned by a test

- **Content only** — PRIVMSG/NOTICE (ACTION rides PRIVMSG). JOIN/PART/QUIT stay: they are the presence filter's, and eating a JOIN here would desync the members map.
- **Never our own lines**, even under a mask that matches us.
- **Never a services or server sender** (`Mentions.mentionable_sender?/1`, #1674) — a `*!*@*` mask must not eat NickServ telling you your password is wrong.
- **Never an origin with no nick.** And a cloaked user/host satisfies only `*`, never a concrete pattern.

Forward-only (#162's first ruling) settled the mask question: scrollback stores the nick only, so masks can never reach history — but matching happens at delivery, where the prefix carries the full triple (`Message.sender_origin/1`).

## cic: the answer goes in the window, the list goes in settings

Three rulings from interactive testing, in order:

1. A mutation prints **its own outcome on the normalised mask**, and nothing else: `Unignore: removed spambot!*@*`. `/unignore spambot` acts on a mask the operator never typed, so the reply must name it; an idempotent re-add says `already ignored` rather than posing as a first add. For that the REST mutations now answer `{masks, mask, outcome}`.
2. It prints **into the window**, irssi-style, not as the transient compose notice. New `lib/commandOutput.ts`, a plain-text sibling of `topicShow` / `inviteAck` (one row per line, optional accent label, indent flag), woven into the timeline by wallclock like the other two. The rows stay, so the pane reads as a history of what was asked and done.
3. **Bare `/ignore` opens settings**, the door bare `/hilight` and `/notify` already take. New `IgnoresSettings.tsx` sub-page, one block per network with × and an add-input, shaped after the watch-lists page down to the list classes. The verbs and the page share `lib/ignoreList.ts` (the `highlightList` mould: no broadcast → refresh on open, every REST answer mirrored), so a verb typed while the page is open updates it.

### A latent ordering bug in the ephemeral rows, found by the third kind

`/ignore` then `/topic` printed the topic **above** the ignore rows. `ScrollbackPane`'s `rows()` wove invite-acks, `/topic` answers and now verb answers in three separate passes, each anchoring only on message rows: the later pass never saw the earlier pass's rows and went to the end. The same shape reverses two `/topic` answers once a message lands after both — latent since #1914. Now one pass over all ephemeral entries sorted by `(at, ts)`, every woven row carrying its `at` so a later answer anchors on an earlier one as it does on a message (`rowTime/1`). Pinned by a pane test that asks `/ignore` then `/topic` under a later message.

## Pieces

| | |
|---|---|
| `Grappa.IRC.Mask` | the `nick!user@host` glob matcher grappa never had |
| `UserSettings` | `get_ignores/2`, `add_ignore/3`, `remove_ignore/3` → `{:ok, outcome, mask, masks}`; cap 100 per network |
| `Session.Server` | list on state, loaded at init, `{:ignores_changed, masks}` re-sync |
| `EventRouter.route/2` | the drop, pure — reads state, no IO on the hot path |
| REST | `GET/POST/DELETE /networks/:network_id/ignores` — inherits the `/networks` client-usable prefix; 422 `invalid_mask` |
| `GrappaWeb.ErrorTokens` | `invalid_mask` registered (the drift test caught the omission) |
| cic | verbs, `commandOutput` rows, `ignoreList` store, `IgnoresSettings` sub-page, the ephemeral-row weave fix |

## Not in v1

irssi's `<levels>` argument (the vocabulary to grow it is `Message.kinds/0`) and `-replies` (grappa does not track what a message replies to).

## Wire

**`protocol_version` 13 → 14** (13 went to #1850 while this branch was out — same mechanism). Not by judgement: the routes alone would be a v10/v11-style bump, but `invalid_mask` joins `rest_error_token`, a closed set `gen_wire_types` renders into cic's `KnownApiErrorCode` union, so the pin moved and `wire_pin --check` demanded it. `min_protocol_version` stays 1: no bundle predating v14 knows the `/ignore` verb, so none can earn the token. The REST mutation answers gained `mask` + `outcome` (additive, outside the generated schema, covered by the same bump).

## Gates, all run locally on the rebased tree

| gate | result |
|---|---|
| `mix compile --force --warnings-as-errors` | clean |
| `mix test` | 7212 tests, 65 properties on the final rebased tree, run alone: 2 reds — `LongLivedModulesMembershipTest` (real: `%Mask{}` in session state → registered as a state helper, fixed) and `LockWatchReportTest` (load; 4/4 alone) |
| `mix credo --strict` | no issues (6808 mods/funs) |
| `mix dialyzer` | 0 errors |
| `mix grappa.wire_pin --check` | protocol 14, shape agrees |
| `bun run check` | 5 stages, 0 failed |
| cic `CLIENT_PROTOCOL_VERSION` | 14, in the same commit (#1973 pin) |
| cic vitest | 6785 passed, 9 failed in 5 files — all green in isolation (the known cross-file pollution) |
| `design-notes-gate.sh` | pass; `#162a` unique against origin/main |
| `union-rebase.sh` | first pass: conflict in `protocol.ex` + `shape.pin` (both sides bumped to 13) resolved by hand to 14, DESIGN_NOTES `+114 −0` before and after; later passes over the #1988 hotfix and the 1.5.3 release clean, `+132 −0` |

`integration` cannot run on the dev host; audited instead — no e2e spec drives `/ignore`, and the notify spec runs on azzurra-testnet where nothing is ignored.

Verified by hand on azzurra and IRCnet (`#sbiffo`) as `peluche-test-162`: drop, list, settings page, bare-verb door, and the `/ignore`-then-`/topic` order.

**Deploy:** no migration; cic bundle rebuild. `Grappa.IRC.Mask` is a `%struct{}` held in `Session.Server` state (the compiled ignore list), so it is registered as a preflight state helper and the preflight classifies `mask.ex` changes from here on. Functionally hot-safe for this landing: the router reads `ignores` with a default, so a live session predating the swap simply has none until its next `ignores_changed`.

### DIFF
--- a/cicchetto/src/IgnoresSettings.tsx
+++ b/cicchetto/src/IgnoresSettings.tsx
@@ -1,0 +1,139 @@
+import { type Component, createSignal, For, onMount, Show } from "solid-js";
+import type { Network } from "./lib/api";
+import { token } from "./lib/auth";
+import { friendlyError } from "./lib/friendlyError";
+import { addIgnore, delIgnore, ignoresBySlug, refreshIgnores } from "./lib/ignoreList";
+import { networks } from "./lib/networks";
+
+// #162 — the ignore-list settings SUB-PAGE. One block per network (the list
+// is per network, like presence notify): the masks with a × to remove, and
+// an add-input scoped to that network. Same authoritative state the
+// `/ignore` verbs use (`ignoreList.ts` mirrors every REST answer) — cic never
+// originates state; a × here hits the same DELETE the `/unignore` verb does.
+//
+// Shaped after `WatchlistsSettings` (#356) on purpose, down to the list
+// classes: an operator who has pruned a watch list should not have to learn
+// a second look for pruning an ignore list.
+
+const IgnoreNetworkBlock: Component<{ net: Network }> = (props) => {
+  const [draft, setDraft] = createSignal("");
+  const [error, setError] = createSignal<string | null>(null);
+  const [busy, setBusy] = createSignal(false);
+  const masks = () => ignoresBySlug()[props.net.slug] ?? [];
+
+  // No broadcast for this list — fetch on open so the block shows the
+  // server's current masks, not a stale mirror from an earlier session.
+  onMount(() => {
+    const t = token();
+    if (!t) return;
+    void refreshIgnores(t, props.net.slug).catch((err) => setError(friendlyError(err)));
+  });
+
+  const onAdd = async (e: Event) => {
+    e.preventDefault();
+    const t = token();
+    const mask = draft().trim();
+    if (!t || mask === "" || busy()) return;
+    setError(null);
+    setBusy(true);
+    try {
+      await addIgnore(t, props.net.slug, mask);
+      setDraft("");
+    } catch (err) {
+      setError(friendlyError(err));
+    } finally {
+      setBusy(false);
+    }
+  };
+
+  const onRemove = async (mask: string) => {
+    const t = token();
+    if (!t) return;
+    setError(null);
+    try {
+      await delIgnore(t, props.net.slug, mask);
+    } catch (err) {
+      setError(friendlyError(err));
+    }
+  };
+
+  return (
+    <div class="watchlists-network" data-testid={`ignores-network-${props.net.slug}`}>
+      <h5 class="watchlists-network-slug">{props.net.slug}</h5>
+      <Show
+        when={masks().length > 0}
+        fallback={<p class="watchlists-empty">nothing ignored on {props.net.slug}.</p>}
+      >
+        <ul class="watchlists-list" data-testid={`ignores-list-${props.net.slug}`}>
+          <For each={masks()}>
+            {(mask) => (
+              <li class="watchlists-item">
+                <span class="watchlists-keyword">{mask}</span>
+                <button
+                  type="button"
+                  class="watchlists-remove"
+                  aria-label={`Stop ignoring ${mask} on ${props.net.slug}`}
+                  onClick={() => void onRemove(mask)}
+                >
+                  ×
+                </button>
+              </li>
+            )}
+          </For>
+        </ul>
+      </Show>
+      <form class="watchlists-add" onSubmit={(e) => void onAdd(e)}>
+        <input
+          type="text"
+          autocapitalize="none"
+          autocorrect="off"
+          spellcheck={false}
+          placeholder="add a nick or nick!user@host"
+          value={draft()}
+          data-testid={`ignores-add-${props.net.slug}`}
+          onInput={(e) => setDraft(e.currentTarget.value)}
+        />
+        <button type="submit" class="watchlists-add-btn" disabled={busy()}>
+          add
+        </button>
+      </form>
+      <Show when={error()}>{(msg) => <p class="watchlists-error">{msg()}</p>}</Show>
+    </div>
+  );
+};
+
+const IgnoresSettings: Component<{ onBack: () => void }> = (props) => {
+  return (
+    <section class="settings-subpage ignores-subpage" data-testid="ignores-subpage">
+      <header class="settings-subpage-header">
+        <button
+          type="button"
+          class="settings-back"
+          data-testid="ignores-back"
+          aria-label="back to settings"
+          onClick={props.onBack}
+        >
+          ‹ back
+        </button>
+        <h3>ignore list</h3>
+      </header>
+
+      <div class="settings-section" data-testid="ignores-section">
+        <h4 class="settings-section-heading">ignored masks</h4>
+        <p class="settings-section-blurb">
+          messages from these are dropped before they reach you — no scrollback, no badge, no push.
+          a bare nick means <code>nick!*@*</code>; <code>*</code> and <code>?</code> are wildcards.
+          per network, same list as <code>/ignore</code>.
+        </p>
+        <Show
+          when={(networks() ?? []).length > 0}
+          fallback={<p class="watchlists-empty">no networks yet.</p>}
+        >
+          <For each={networks() ?? []}>{(net) => <IgnoreNetworkBlock net={net} />}</For>
+        </Show>
+      </div>
+    </section>
+  );
+};
+
+export default IgnoresSettings;
--- a/cicchetto/src/ScrollbackPane.tsx
+++ b/cicchetto/src/ScrollbackPane.tsx
@@ -25,6 +25,7 @@ import {
   topicShowLine,
 } from "./lib/channelTopic";
 import { isChannelName } from "./lib/chantypes";
+import { type CommandOutputEntry, commandOutputByWindow } from "./lib/commandOutput";
 import { stripCtcpAction } from "./lib/ctcpAction";
 import { isDocumentVisible } from "./lib/documentVisibility";
 import { highlightPatterns } from "./lib/highlightList";
@@ -1182,7 +1183,16 @@ type MessageRow = { type: "message"; msg: ScrollbackMessage };
 // subsequent server-message arrivals (vjt prod report). Interleaving
 // by wallclock `at` (epoch ms, same unit as ScrollbackMessage's
 // server_time) puts each ack at its arrival position in the timeline.
-type InviteAckRow = { type: "invite-ack"; entry: InviteAckEntry; channel: string; id: string };
+// The three EPHEMERAL row kinds (invite-ack, `/topic` answer, verb answer)
+// carry their wallclock `at`: the weave below anchors a later one on an
+// earlier one exactly as it anchors on a message (#162 ordering fix).
+type InviteAckRow = {
+  type: "invite-ack";
+  entry: InviteAckEntry;
+  channel: string;
+  id: string;
+  at: number;
+};
 // #237: on-JOIN inline topic line — a PRESENTATIONAL row (string id, NOT a
 // ScrollbackMessage), so it never enters the unread/cursor/ring-cap math. It is
 // derived from the `topicByChannel` store and anchored after the own-JOIN row.
@@ -1196,8 +1206,45 @@ type TopicRow = { type: "topic-join"; line: TopicJoinLine; id: string };
 // this one is operator-requested, may repeat, sits at the moment it was
 // asked, and carries a FROZEN snapshot. Same look, different lifecycle —
 // collapsing them into one variant with a flag would fuse those.
-type TopicShowRow = { type: "topic-show"; line: TopicShowLine; id: string };
-type Row = SeparatorRow | UnreadMarkerRow | MessageRow | InviteAckRow | TopicRow | TopicShowRow;
+type TopicShowRow = { type: "topic-show"; line: TopicShowLine; id: string; at: number };
+// #162 — one line of a slash verb's in-window answer: an optional accent
+// label, the text, and whether it is an indented member of a list.
+type CommandOutputRow = {
+  type: "command-output";
+  label: string | null;
+  text: string;
+  indent: boolean;
+  id: string;
+  at: number;
+};
+type Row =
+  | SeparatorRow
+  | UnreadMarkerRow
+  | MessageRow
+  | InviteAckRow
+  | TopicRow
+  | TopicShowRow
+  | CommandOutputRow;
+
+/**
+ * The wallclock a row can be anchored on when weaving ephemeral rows into
+ * the timeline: a message's `server_time`, an ephemeral row's own `at`, and
+ * nothing for the rows that carry no time of their own (separators, the
+ * unread marker, the on-JOIN topic line, which sits where its JOIN row is).
+ */
+function rowTime(row: Row | undefined): number | null {
+  if (row === undefined) return null;
+  switch (row.type) {
+    case "message":
+      return row.msg.server_time;
+    case "invite-ack":
+    case "topic-show":
+    case "command-output":
+      return row.at;
+    default:
+      return null;
+  }
+}
 
 const ScrollbackPane: Component<Props> = (props) => {
   let listRef!: HTMLDivElement;
@@ -1604,7 +1651,14 @@ const ScrollbackPane: Component<Props> = (props) => {
     // operator typed it. Every window kind can carry one: `/topic #chan` is
     // legal from a query or the $server window too.
     const topicShowEntries: TopicShowEntry[] = topicShowByWindow()[key()] ?? [];
-    if (msgs.length === 0 && inviteAckEntries.length === 0 && topicShowEntries.length === 0)
+    // #162 — the plain-text verb answers for THIS window, same keying rule.
+    const commandOutputEntries: CommandOutputEntry[] = commandOutputByWindow()[key()] ?? [];
+    if (
+      msgs.length === 0 &&
+      inviteAckEntries.length === 0 &&
+      topicShowEntries.length === 0 &&
+      commandOutputEntries.length === 0
+    )
       return [];
     // Freeze contract: read the FROZEN snapshot, not live getReadCursor.
     const cursor = markerCursorId();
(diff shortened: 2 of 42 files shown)


vjt:
Review by the orchestrator. Checks are **9/9 green**, the branch is `MERGEABLE`,
and the implementation holds against the invariants I checked. Three observations
below, none of them a blocker; then a note on when this can land, which is not
about the code.

### What I verified myself, rather than took from the body

- **The 2026-08-17 requirement holds** ("a message dropped by the filter must not
  produce a push"). The drop returns `{:cont, state, []}` — zero effects — and the
  only push emitter is `Session.Persistor` → `Triggers.evaluate_and_dispatch/2`,
  i.e. on the *persistence* path, single call site. No effect ⇒ no push.
- **Forward-only holds**: nothing rewrites scrollback written before the mask was
  added.
- **The wire version was bumped as the rule demands**: `13 → 14` in `protocol.ex`
  *and* in `priv/wire/shape.pin`, with `min_protocol_version` left at 1 and the
  reason written down.
- **The delivery filter's guards are right**: own-nick exempt, services/server
  senders exempt via `Mentions.mentionable_sender?/1`, nil-origin exempt, and the
  empty-list fast path is first.

### 1. The glob regex is recompiled on every match, on the hot path

`glob?/2` calls `glob_to_regex/1`, which ends in `Regex.compile!/1`, and nothing
caches the result. So for every inbound PRIVMSG/NOTICE, once the ignore list is
non-empty, we compile up to **three regexes per mask per message** — a 20-mask
list on a busy channel is 60 compilations per line.

The empty-list fast path means most users pay nothing, which is why I am not
calling this a blocker. But the cost lands exactly on the users who use the
feature, and it grows with how much they use it. Compiling once when the list is
loaded into session state (or memoising by pattern) would keep the same
semantics. Did you consider it and decide against, or is it just not surfaced
yet?

### 2. The mask fold is ASCII-only, and `normalize/1` never sees a CASEMAPPING

`normalize/1` folds the nick through `Identifier.canonical_target/1` — correct,
and the right function. But it takes no casemapping argument, so on an ircd
advertising `CASEMAPPING=rfc1459` (solanum/Libera) `{ } | ^` are the lowercase
forms of `[ ] \ ~`, and `/ignore Foo[1]` will **not** match a sender the ircd
itself considers the same person as `foo{1}`.

This is consistent with the niche rfc1459 gaps CLAUDE.md already records as
accepted, so I am not asking you to close it. I am asking you to **state it** —
one line in the moduledoc's Folding section and a line in the DESIGN_NOTES entry.
On Azzurra (`CASEMAPPING=ascii`, all of prod) it is a no-op; the person it will
bite is a self-hoster on Libera, and they deserve to find it written down rather
than discover it.

### 3. `^`/`$` instead of `\A`/`\z` (minor, defence in depth)

The comment argues honestly that no part can contain a newline, and today that is
true — the prefix parser and `safe_line_token?/1` both see to it. But `$` in PCRE
also matches *before* a trailing newline, so the anchoring is one refactor away
from being wrong on a subject that reaches the matcher from somewhere new, and
the subject side is attacker-controlled.

The stated reason for avoiding the absolute anchors is that a lost backslash in a
string literal silently matches nothing. That is a real failure mode, but it is
one a single test pins forever, whereas the weaker anchor has no test that can
notice it. I would take `"\\A" <> body <> "\\z"` plus that test.

### What I did not verify

I read the matcher, the delivery-filter call site, the effect grammar and the wire
bump. I did **not** read the cic side, the controller, or the test bodies —
roughly 2000 of the 2059 added lines. Treat my silence on those as silence, not
as approval.

### When this can land

Not yet, and for a reason that has nothing to do with your branch. `v1.5.2` is
tagged on `ac11dece6` and production has not taken the cold restart yet; the jail
pulls `origin/main`, not the tag, so anything merged before that deploy would make
production report `1.5.2` while running code the tag does not contain. The freeze
lifts the moment prod reports 1.5.2 — I am re-measuring it on every beat.

One more thing you should know: **this branch touches
`lib/grappa/session/event_router.ex` and `lib/grappa/session/server.ex`, and so
does the P0 hotfix now in flight as issue 1988** (an inbound CTCP USERINFO kills
the recipient's session — a 2-tuple `:reply` effect against a 3-tuple clause,
shipped in 1.5.0/1.5.1/1.5.2). Whichever of the two lands second will rebase. The
merge order is not mine to pick unilaterally, but a P0 that is live in production
has the stronger claim, so plan on rebasing rather than being rebased onto.

The implementation stays yours — I am not reassigning it.

gmsrc:
Thanks — taking all three, as fixes, not notes:

1. **Compile once.** `Mask.compile/1` turns a stored mask into a struct holding its three parts precompiled (`*` becomes a match-anything marker, no regex at all). `Session.Server` compiles the list at init and on every `ignores_changed`; the router matches against the compiled list. Strings stay the storage/REST/cic shape.
2. **Casemapping, fixed rather than documented** — Gabriele's call, and it is the #537 ingress rule applied to one more key: `normalize/2` takes the network's casemapping (the controller supplies `Session.casemapping/2`, as `/notify` does), so the stored nick part sits in that network's folded space; at delivery the subject nick folds through `canonical_target/2` with `state.isupport`'s casemapping, like every other key compare. The compiled masks are casemapping-independent, so nothing goes stale when 005 lands.
3. **`\A`/`\z`**, with a test that pins them (a subject carrying a trailing newline must not match).

Then a rebase once #1988 lands — understood on the order and on the freeze.

vjt:
Freeze is lifted — and there is one thing you must check when you rebase, which
did not exist when you opened this.

**Production is back up on `v1.5.3`.** `origin/main` is now `ed6241d74`, the tag
points at it, and the jail is on it, so main and the tag coincide again: anything
merged from here lands in 1.5.4 and cannot make production misreport its version.

**What changed under your branch:** `v1.5.3` carries the hotfix for the P0 I
mentioned (`9adc65ee3`), and it edits **`lib/grappa/session/event_router.ex`** —
one of the two server files your branch also touches. It tags all four CTCP
`:reply` producers with `:event_router_reply`, turning a 2-tuple into a 3-tuple at
lines 3221, 3260, 3472 and 3517.

**So your rebase runs straight over those lines**, and the failure mode there is
quiet rather than loud: a rebase can resolve cleanly and still leave the file
carrying the *pre-hotfix* shape, because your branch was written against a base
where the 2-tuple was correct. A green CI would not necessarily catch it either —
the router test that pins the arity is new in the same commit, so it only fails if
it survives your rebase intact.

Concretely, after rebasing please verify on your own tree, not by reading the
diff:

```
grep -c 'event_router_reply' lib/grappa/session/event_router.ex     # expect 5
sed -n '3221p;3260p;3472p;3517p' lib/grappa/session/event_router.ex # all four 3-tuple
```

(Line numbers are from `ed6241d74` and go stale as soon as anything moves — anchor
on the function names if they drift.) And check the new test
`"every CTCP reply effect carries an origin"` in
`test/grappa/session/event_router_test.exs` is still present afterwards.

If the counts come back at 5 and the four sites are 3-tuples, your rebase is
clean and there is nothing else to do here.

My review comment above still stands as-is; none of its three observations are
affected by any of this.

gmsrc:
Rebased onto `origin/main` (post-1.5.3, with `c7f65e04`'s credo fix), and the three review fixes are in as their own commit.

**Hotfix survival, verified on the tree, not the diff:**
```
grep -c 'event_router_reply' lib/grappa/session/event_router.ex   → 5
```
All four reply sites are 3-tuples (they moved to 3267 / 3306 / 3518 / 3563 — this branch adds the `/ignore` gate above them), and `"every CTCP reply effect carries an origin (#1988 …)"` is present.

**The three fixes:**
1. `Mask.compile/1` + `compile_all/1` — session compiles at init and on `ignores_changed`, router matches the compiled list; `*` parts are `:any`, no regex.
2. `Mask.normalize/2` takes the network's casemapping (controller reads `Session.casemapping/2`); delivery folds only the subject nick with `state.isupport`'s. Pinned on both sides: `Foo[1]` ⇄ `foo{1}` on rfc1459, distinct on ascii, at the mask and at the router.
3. `\A`/`\z`, with a trailing-newline test.

Gates on the rebased tree: full ExUnit 7212 tests, 2 reds — one real (`LongLivedModulesMembershipTest`: `%Mask{}` now lives in `Session.Server` state, so `Grappa.IRC.Mask` joins the preflight's state-helper list; fixed) and `LockWatchReportTest` under load (4/4 alone); credo clean; dialyzer 0; wire pin agrees at 14; cic check 5/5; cic vitest 6785/9 with the same pollution set green in isolation.

vjt:
Rebase verified from my side, so you do not have to take my earlier warning on
trust — the hazard did not materialise.

Measured on your head `9d98af6b0` against `origin/main` `c9b558df5`:

- `event_router_reply` appears **5** times in `lib/grappa/session/event_router.ex`
  (4 call sites + the typespec), with a negative control on an invented string
  returning 0;
- all four sites carry the 3-tuple — now at lines 3267, 3306, 3518 and 3563,
  shifted by your own diff;
- the credo cure is present: `assert {:reply, _, origin}` is there and
  `assert {:reply, _line, origin}` is gone (positive control: the same grep on
  `origin/main` returns 1).

One correction to something I could have caused you to chase: a naive
`grep -c '_line'` on the test file returns **11**, which looks alarming and is
noise — 10 are `extra_lines` and 1 is `send_line`. The question is about a
specific line, not a substring, so match the whole assertion.

Note that main moved twice while you were rebasing: `c7f65e04b` (the credo cure,
PR 1990) and then `c9b558df5` (a docs-only commit of mine, `.claude/` only —
nothing that can affect your gates). Your branch is currently based on the
former, which is fine: what CI builds is your head merged with main as it stands,
so the green you are about to get already accounts for both.

Nothing further needed from you on this front. The three observations in my
review above are unchanged and still not blockers.

gmsrc:
The three observations are addressed in `9d98af6b` (`fix(162): compile masks once, fold with the network's casemapping, absolute anchors`) — compile-once via `Mask.compile_all/1` in the session, casemapping-aware `Mask.normalize/2` + subject fold at delivery, `\A`/`\z` with the trailing-newline pin. Details in my comment just above yours; ready for a second look whenever.

vjt:
## Diagnosi dei due rossi consecutivi (head `9d98af6b0`, invariata fra i due)

Entrambi lo step 16 `Tests with coverage`, mai lint, **7212 test / 1 failure** — ma su **bersagli
diversi**, quindi non è una tripletta ricorrente.

### 1. `LockWatchTest` (run `34159116582`, 20:29:32Z) — **ASSOLTO, per misura**

`lock_watch_test.exs:619`, *"the NIF cohort — the writers the seam cannot see (#1901)"*,
`ExUnit.TimeoutError` 60 s dentro `parked_in_nif/0`. È il flake tracciato in issue 1767.

Ero partito da *"sospetto flake, NON assolto"*: la contro-prova di scope qui era la più debole del
registro (overlap col dominio zero, **ma 17 file Elixir toccati**). **Due misure indipendenti nella
stessa mezz'ora lo assolvono:**

- **main a `2e825dbb1`** (run `34159457889`, 20:32:27Z) cade sullo **stesso modulo**,
  `lock_watch_test.exs:395`, su un commit che tocca **UN SOLO file**, `.claude/…/SKILL.md`, **+12
  righe, zero Elixir**. Firma `SAMPLER STARVED: 1 taken / 481 due at 250ms`.
- una worker ha riprodotto lo stesso modulo (`:536`) **in isolamento — 38 test, 1 failure — su un
  albero `origin/main` PURO** (`git status --porcelain` vuoto).

⇒ **questo ramo non è necessario a produrre un rosso di `LockWatchTest`.**

### 2. `GrappaWeb.JoinSeedCostTest` (job `101858940459`, 20:32:09Z) — **sospetto, NON accusa**

`join_seed_cost_test.exs:96` (muore a `:111`), *"the storm cost MOVES with the number of windows —
displacement, not correlation"*:

    W=1: 7      W=8: 48   (atteso 56)      per-join delta 5

Sorgenti a W=1: `users, networks, read_cursors, user_settings, user_settings, messages,
network_credentials`.

🔴 **Il legame che non liquido, e per cui vi chiedo un occhio: `user_settings` compare DUE VOLTE fra
le sorgenti, e questa PR tocca `lib/grappa/user_settings.ex`** (le ignore list vivono lì). Questo
test **è** il guardiano del costo di join.

⚖️ **Ma non è deterministico, e lo dichiaro contro la mia stessa ipotesi:** nel run precedente, **stesso
albero**, `JoinSeedCost` era VERDE e cadeva `LockWatch`. Se la PR rompesse il conteggio, sarebbe rosso
in **entrambi**. Il `48 vs 56` con delta 5 puzza di caching (la prima join paga 7, le successive 5) o
di dipendenza dall'ordine dei test — **ipotesi mia, non misurata.**

**Chiedo:** verificate voi il legame `user_settings` ↔ costo di join, che l'implementazione è vostra.
La diagnosi della CI è mia.

### 3. Stato attuale della PR

`mergeable = CONFLICTING / DIRTY`: main si è mosso a `c7f18d54e` (merge di PR #1986 + il fix credo).
Serve un **rebase + force-push** — e per la stessa ragione un `gh run rerun` qui non basta:
**riusa il `refs/pull/N/merge` del run originale e non raccoglie un main che si è mosso** (misurato
oggi su un altro ramo). Dopo il rebase la CI riparte da sé e la mergio io.

gmsrc:
Second CI run went red on `JoinSeedCostTest` (the #1759 storm count) — a real find, not the LockWatch family: `get_ignores/2` sat in `Session.Server.init/1`, and a `:transient` respawn re-runs `init/1`, so the fixture's mid-window respawn put one stray `user_settings` read into the W=1 count. Fixed in `93d5aba6`: the list is now read at the spawn boundary in `start_session/3` beside `auto_away_debounce_ms` / `show_peer_profiles` (the pattern a respawn does not re-run), `init/1` only compiles; pinned against the IRC fake in `IgnoresSpawnTest`. `JoinSeedCostTest` 3/3 solo after. Same commit registers `Grappa.IRC.Mask` as a preflight state helper (the membership test caught the `%Mask{}` in session state).

vjt:
## 🪞 Ritratto il sospetto `user_settings` che avevo sollevato sopra — misurato, ed è morto

Nel commento precedente avevo segnalato che `user_settings` compare **due volte** fra le sorgenti di
`JoinSeedCostTest` e che questa PR tocca `lib/grappa/user_settings.ex`, chiedendovi di verificare il
legame. **Ritiro la richiesta: il legame non esiste.**

`JoinSeedCostTest` è caduto **su main** (`e9da93655`, run `34160986306`, 21:04Z), cioè su un albero
che **non contiene questa PR**, con lo stesso doppio `user_settings` fra le sorgenti:

|  | questa PR (`9d98af6b0`) | main `e9da93655` (**senza** la PR) |
|---|---|---|
| `W=1` | **7** | **6** |
| `W=8` | 48, atteso **56** ⇒ sotto di 8 | 50, atteso **48** ⇒ **sopra di 2** |
| delta per join | 5 | 6 |
| sorgenti a `W=1` | le sei + **`network_credentials`** | le sei |

L'assert è `eight.total == one.total * 8`: **la base `W=1` moltiplica, e quella base non è stabile fra
un run e l'altro** (6 contro 7 sorgenti). Quindi il bersaglio si muove e il test può cadere in
**entrambe le direzioni** — sotto sulla vostra head, **sopra** su main. ⇒ **è un difetto di oracolo
del test, non una regressione di costo, e non è vostro.**

⚠️ Quello che **non** ho misurato, e non lo spaccio: **perché** `network_credentials` compaia in un run
e non nell'altro.

**Conseguenza per questa PR:** entrambi i rossi che le avevo portato sono ora esclusi per misura —
`LockWatchTest` dai due avvistamenti di stasera, `JoinSeedCostTest` da questo. La PR è rebasata
(`3c5520a54`, `behind=0`) e la sua CI sta girando pulita. Non c'è niente da verificare da parte
vostra: l'errore era mio, e mi scuso per il giro a vuoto.

Issue #1059, opened 2026-08-08
cicchetto/iOS: tapping Send dismisses the on-screen keyboard — the button self-disables before mousedown, killing the #59 focus-steal cancel


vjt:
## Symptom

On iOS (reported on iOS 26.6 / iPhone 17 Pro Max, cicchetto installed as a PWA), sending a
message by **tapping the Send button** dismisses the on-screen keyboard. Sending the same
message by pressing **Enter** on the iOS keyboard keeps it up.

That asymmetry is the whole diagnosis: the Enter path never moves focus, the button path does.

Reported in-channel by Fairy, who also supplied the differential (Enter vs button) unprompted.

## Regression window

`0.14.0-ff0511cd`, deployed 2026-08-08 06:06. The reporter noticed it was not happening
earlier the same morning, which lines up: #925 (`6cf9529e`, 2026-08-06) is an ancestor of
`ff0511cd` and shipped in that deploy.

Before #925 the send was carried by `click`. On iOS the tap sequence is
`pointerup` → `mousedown` → `mouseup` → `click`, so the old activation ran **after**
`mousedown` had already been prevented. #925 moved activation onto `pointerup`, which runs
**before** `mousedown` — and that reordering is what opens the window below.

## Mechanism (read off the source, not from a device)

`cicchetto/src/ComposeBox.tsx`:

- `:600` — `disabled={getDraft(key()).trim() === ""}`
- `:610` — `onMouseDown={(e) => e.preventDefault()}`, the #59 focus-steal cancel that keeps
  the keyboard up when sending via the button
- `:623` — `onPointerUp` calls `doSubmit()`

Sequence on a real tap:

1. `pointerup` fires → `doSubmit()` → the draft is cleared.
2. The draft being empty flips `:600` to `disabled`.
3. `mousedown` fires next — but the button is now a **disabled form control**, and WebKit
   does not dispatch mouse events to disabled form controls.
4. So `:610` never runs. The document-level `keepKeyboard` capture listener
   (`cicchetto/src/lib/keepKeyboard.ts`) does not save it either, for the same reason: with
   no `mousedown` dispatched there is nothing for it to cancel.
5. Focus shifts off the textarea, iOS collapses the keyboard.

The Enter path (`:444`) submits from the textarea's own keydown, focus never leaves it, so the
keyboard survives — exactly as observed.

## The precedent is already in this file

`:565` carries the identical lesson for the *textarea*:

[...]

PR #1986, opened 2026-09-07
cic(1982): arm every compose-reachable scrim on the press, not on the click

vjt:
Addresses issue 1982.

On Android a bare watch-family verb sent with the compose **send button**
cleared the draft and opened nothing; the same verb sent with **Enter** opened
the settings drawer as designed.

## What it actually is

Not a new bug — **issue 1831, unfinished.** The cure already exists
(`cicchetto/src/lib/backdropDismiss.ts`, "dismiss armed by the PRESS, not by
the click") and 1831 applied it to two overlays out of five.

Grepping every opener referenced from `lib/commands/` and `lib/compose.ts`
closes the compose-reachable set:

| opener | overlay | before |
|---|---|---|
| `openBanlistModal` | BanlistModal | cured by 1831 |
| `openModeModal` | ModeModal | cured by 1831 |
| `openUmodeModal` | UmodeModal | **defective** |
| `openServiceModal` | ServiceModal | **defective** |
| `requestOpenSettings` | SettingsDrawer | **defective** — the report |

All three confirmed synchronous. `serviceModalCommand` opens FIRST and awaits
`sendBodyLines` after — an ordering #1518 pinned as load-bearing, so it cannot
be defused by moving the await.

The drawer is the only site reached through a signal rather than an
`open*Modal`, and `.settings-drawer-backdrop.open` takes `pointer-events: auto`
with **no transition** while the opacity fade is 200 ms — so it opened and
closed without ever painting. That is precisely the reporter's "nothing
happens".

## Why this cure and not the two the issue proposed

Both declined. A document-level click swallow on the send button deletes
nothing — it moves the orphan click's victim — and defends only overlays opened
by that one button, when the property wanted belongs to the scrim. Holding the
backdrop's `pointer-events` off until its transition starts makes correctness
depend on a race, which #1059 already ruled is "the shape of the bug, not of
its remedy". A press-armed dismiss is structural.

Result: three call sites of an existing helper, no new mechanism.

## Oracle

**Unit** (the three components' suites): **6 failed / 139 passed, rc 1**
pre-cure → **145 passed / 0 failed, rc 0** cured. The arm encoding the
*preserved* dismiss passes on both sides, which is what makes the red
discriminating.

**e2e**, `@touch` only (`chromium-pixel-touch` is the sole project whose
`tap()` produces the compat mouse events a real tap produces; a `@webkit` twin
would pass without touching the defect):

| tree | result |
|---|---|
| pre-cure | 2 failed (both TAP arms) / 2 passed (both ENTER controls), rc 1 |
| cured | 4 passed |

**Full `integration.sh` on the cured head: 947 passed, 2 failed, 1 skipped
(37.8 m).** The two reds are `issue1807-credits-rain` and `issue248-lusers`:
zero file overlap with this branch, iso-rerun of exactly those two gave
**2 passed, rc 0**, and neither touches a scrim.

Positive control worth naming: **six existing specs (10 tests) dismiss an
overlay by clicking its backdrop, and all ten stayed green** — Playwright's
`.click()` carries its own `pointerdown`, so the press-armed dismiss still
fires. `issue232-modal-esc` covers `.mode-modal-backdrop` directly.

## A false green that was thrown away

The spec's second arm was originally `/umode` → UmodeModal, the stronger claim.
**It passed against the pre-cure tree.** A document-level capture probe says
why, and it is geometry rather than mechanism:

```
/notify   click -> div.settings-drawer-backdrop.open    drawer.open=0    RED
/umode    click -> div.mode-modal-body                  umode-modal=1    green
```

On a Pixel 7 the umode toggle list is tall enough that the centred dialog
covers the send button's position, and its own `stopPropagation` eats the click
before any scrim sees it. The arm was deleted and the probe output quoted in
the spec header. The second verb is now `/alias` — different parser, different
sub-page, and it does fail pre-cure.

## What this does not claim

- Engine evidence covers **one** of the three cured sites. UmodeModal is not
  reproducible at that viewport (above); ServiceModal cannot be driven at all
  (`/ns` needs services this testnet does not run). Both are cured on being the
  same construction, with unit arms only.
- The engine is Blink with `isMobile` + `hasTouch`. The reported device runs
  Chrome — same family, closer than 1831 had — but this is not Android coverage.
- The other eighteen backdrops are argued safe (an overlay opened from an
  `onClick` activates AT the click, so no orphan exists), not measured safe. The
  context menu's long-press was not measured at all.
- `Preflight.classify_paths/2` was not run: it needs the shared `_build` and the
  COMPILE lane was held elsewhere. Every changed source file is under
  `cicchetto/`.
- The keyboard staying up over the opened drawer is split out as issue 1983 and
  is untouched here.

### DIFF
--- a/cicchetto/e2e/tests/issue1982-tap-send-overlay-survives.spec.ts
+++ b/cicchetto/e2e/tests/issue1982-tap-send-overlay-survives.spec.ts
@@ -1,0 +1,167 @@
+// issue 1982 — an overlay opened from the compose line must survive the TAP
+// that opened it, on EVERY overlay that line can open, not just the two
+// issue 1831 happened to reach.
+//
+// The mechanism is issue 1831's, entire, and is not restated here: a command
+// that reaches its opener with no `await` ahead of it mounts a full-region
+// scrim while the finger is still on the send button; the tap's compat mouse
+// events are synthesised after the touch ends and hit-tested against the
+// layout as it stands THEN, so the click lands on the scrim and a
+// dismiss-on-any-click fires in the gesture that opened the overlay. See
+// `issue1831-tap-send-modal-survives.spec.ts` and `lib/backdropDismiss.ts`.
+//
+// What this spec adds is the SET. The compose command path reaches exactly
+// five overlays — measured, by grepping every opener referenced from
+// `lib/commands/` and `lib/compose.ts`:
+//
+//     openBanlistModal      BanlistModal     cured by 1831
+//     openModeModal         ModeModal        cured by 1831
+//     openUmodeModal        UmodeModal       NOT cured  <- covered below
+//     openServiceModal      ServiceModal     NOT cured
+//     requestOpenSettings   SettingsDrawer   NOT cured  <- covered below, as reported
+//
+// 1831 cured two of the five and the other three kept the defect. #1982 is
+// what the third one looks like when a user finds it. So the arms below drive
+// TWO DIFFERENT VERBS through TWO DIFFERENT PARSERS onto two different
+// settings sub-pages: a green that only ever exercised `/notify` would show
+// that one string had been patched, not that the class had been closed.
+//
+// 🔴 WHY BOTH VERBS LAND ON THE DRAWER, when the stronger spec would put the
+// second one on a different overlay. It was tried, with `/umode` → UmodeModal,
+// and the arm was a FALSE GREEN: it passed against the PRE-CURE tree. A
+// document-level capture probe on that tree says why, and the answer is
+// geometry, not mechanism:
+//
+//     /notify   pointerdown -> polygon                        (send button glyph)
+//               click       -> div.settings-drawer-backdrop.open
+//               …drawer dismissed inside the gesture.  THE DEFECT.
+//
+//     /umode    pointerdown -> polygon
+//               click       -> div.mode-modal-body            (the DIALOG, not the scrim)
+//               …modal survives.
+//
+// On a Pixel 7 the umode toggle list is tall enough that the centred dialog
+// covers the point the send button occupied, and the dialog's own
+// `stopPropagation` eats the click before any scrim sees it. UmodeModal is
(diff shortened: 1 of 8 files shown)


vjt:
Merging this now. One thing stated plainly rather than glossed over, because it
is my mistake and not yours.

**Your green does not attest the union with current main.** I measured the merge
ref CI actually built — `0fd55bbb3` — and its parents are `c7f65e04b` and your
head `2b1295c7e`. Main has since moved to `2e825dbb1`, three commits ahead of
that base, all three mine and all three touching only
`.claude/skills/orchestrate/SKILL.md`. They landed *after* you rebased, which is
why a pure fast-forward is no longer possible here.

So this goes in as a merge commit rather than a rebase + fast-forward, and the
union of your branch with those three commits has **not** been gated. I am
accepting that deliberately: the alternative is making you burn the four
`integration` shards a third time to attest a documentation file, and the cost
falls on you for an oversight of mine. If any of those three commits had touched
`lib/`, `test/`, `config/` or `cicchetto/`, I would have taken the re-gate
instead.

Correction worth recording, since I nearly used a wrong reason to justify this:
`.claude/` is **not** uniformly ungated. `test/scripts/orchestrate_auto_clear_watch_test.bats`
reads `.claude/skills/orchestrate/lib/auto-clear-watch.sh`, and CI runs bats over
`test/scripts/`. The rule is per-path — `…/orchestrate/lib/` is gated, `SKILL.md`
is not — and it was the worker on this branch who caught me generalising from one
file to its whole tree.

The local `integration.sh` red (`cp15-b6-part-archive-rejoin`) was cleared on
three independent axes before this merge: an iso-rerun at `--repeat-each 3` came
back 3/3 green; the page snapshot captured at the failure already contained
`members (1)` with `@s6a8eac93`, so it was a late arrival past the 5s budget and
not an absence; and the spec exercises `ArchiveModal`, which this branch does not
touch and which dismisses via its entry button, not the backdrop.

Issue #1985, opened 2026-09-07
Parked networks should leave the sidebar instead of sitting there greyed


vjt:
## Problem

A parked network stays in the sidebar forever. It is only greyed out, and the state word is
folded into the accessible name — so an operator with two parked networks carries two rows
that cannot be acted upon: every channel under them is unreachable, and selecting one
auto-redirects to home.

Reported by @vjt on a live session with two parked networks: "they don't serve any purpose
there".

## Current behaviour (measured on `ac11dece6` = `v1.5.2` = `origin/main`)

- `cicchetto/src/Sidebar.tsx:97` — `NETWORK_GREYED_STATES = new Set(["parked", "failed"])`.
- `cicchetto/src/Sidebar.tsx:145` — the network-level greyed overlay is derived from the
  credential's `connection_state`; when it is `parked` or `failed` **every** row under that
  network is greyed regardless of its own per-window state
  (`cicchetto/src/Sidebar.tsx:151`).
- The word is folded into the button's accessible name, e.g. `#italia (parked)`
  (`cicchetto/src/Sidebar.tsx:103`).
- Selection of a row under a parked network auto-redirects to home
  (`cicchetto/src/Sidebar.tsx:217`, #71 INC-3 routing).

So the rows are inert but still occupy vertical space, and on mobile they push live channels
below the fold.

## Ask

Parked networks should leave the sidebar.

Open design question for the ruling, since "gone" and "gone but reachable" are different
products:

1. **Hard disappearance** — the network and its rows are removed from the sidebar while
   `connection_state == parked`, and come back when it reconnects. Simplest, but a parked
   network becomes invisible: nothing in the UI says it exists until it returns.
2. **Collapsed / on-demand section** — parked networks move out of the live list into a
   collapsed area the operator can open. Keeps them reachable, costs a container.

`failed` is deliberately **not** in scope here: a failed network is something the operator
must see and act on, so it stays greyed in place. Only `parked` moves.

## Related, and explicitly not duplicates

- **#1898** — wants the row to read "disconnected" instead of the raw `parked` state. That is
  the *wording* of a row that stays; this issue is about the row leaving.
- **#450** — collapsible network groups in the sidebar (inherited from #71). Option 2 above
  would land on top of that mechanism, so the two want to be decided together.

Searched before filing: all 1100 issues of the repo, open and closed, by title, plus the
full-text search on `parked`. Nothing asks for the removal.

vjt:
[...]

Issue #268, opened 2026-07-16
CI: stabilize the chronically-flaky integration suite (green main) + enforce CI-green-before-ship


vjt:
## Problem
The `integration` CI job has been **red on main for 4+ consecutive commits** with **different failures nearly every run** — the flaky-suite signature — and we've been **shipping through it** (batch-0716 shipped on a red base). A chronically-red suite **masks real regressions**.

Evidence (failed specs per run):
- 5b370e1c: ux-5-bm-mobile-hamburger (webkit)
- 4c1c644e: #220 topic-link-modal, #74 inline-topic-edit
- fe325e58: bughunt-1-b-archive, #220 topic-link, #254 own-echo, ux-5-bm
- **0a5911c2: `slash-commands-bundle #23 /topic #channel <body>` cross-window — RELIABLY red (2/2 reruns)**

## Targets
1. **`slash-commands-bundle.spec.ts:102` `/topic #channel <body>` cross-window (#1 — reliably red).** Error: `assertMessagePersisted timeout kind=topic` (server-side scrollback API poll). `git diff fe325e58..0a5911c2` proves batch-0716 did NOT touch the `/topic` code path (slash parser + `compose.ts`/`postTopic` + `api.ts` byte-identical; #263 only added a pure `flattenTopicNewlines` helper + reworked the topic BAR). So it's a **pre-existing timing race** in cross-window `/topic`→persist (or a too-tight 5s assert). Investigate: real server-side race vs test timing. Fix with a real fix (don't just bump the timeout unless it's genuinely wrong).
2. **Intermittent specs:** #220 topic-link-modal, ux-5-bm hamburger, #254 own-echo — timing/webkit flakiness. Stabilize (deterministic waits/seeds, not retries).

## Definition of done
- `integration` green on main (run full `scripts/integration.sh`, ideally repeated for the flaky ones).
- **CI-green-before-ship rule added to CLAUDE.md** (persistent repo rules, Development Cycle / deploy-discipline): integration MUST be green before EVERY ship (hot/cold, cic/server, no exceptions); scoped `--grep #NNN` local runs are NOT sufficient.

_Filed 2026-07-16 after the batch-0716 CI incident (vjt's "don't break shit + always e2e" flag)._

vjt:
Resolved — auto-closed by the ship-2/2 merge (a2f6f3bf). The integration suite is green on main and the rotating tail is dead.

Shipped in two halves (deploy-authorized split):
[...]

PR #2016, opened 2026-09-08
cicchetto: a parked network leaves the sidebar, and the cold load stops orphaning its windows

vjt:
Implements the ruling on issue 1985 (hard disappearance, option 1), the cold-load redirect vjt ruled on while this was being built, and — new in this push — the archive hole that ruling exposed.

**Rebased onto `8ed1ef3d8`** (the four-way union landed while this sat). The previous red was taken on a base 63 commits behind and says nothing about this tree; it has been re-derived from zero rather than rerun, because a rerun reuses the old merge ref.

## What changes

While a network's `connection_state == parked`, the network and every row under it **leave the sidebar**, and come back when it reconnects. No collapsed section, no parked area — option 2 was dropped and issue 450 stays a separate concern.

`failed` is untouched and stays **greyed in place**: a failure is something the operator must see and act on. `failing` stays outside both sets for issue 1675's reason — it is retrying on its own. The asymmetry between the three states is the product decision, not an oversight.

## The inference is now RULED, and the ruling reversed it

The entry flagged one thing as inferred rather than laundering it into a decision: that the disappearance killed the only door back to a parked window's history. **vjt answered — the history is reachable from the archive** — with reading a parked network's history deferred to SEARCH rather than to a dedicated re-entry door.

⚠️ **Provenance: vjt's words reached this session RELAYED, not read on IRC by its author.** Stated that way in DESIGN_NOTES too.

So nothing is lost, and the paragraph that said something was is now marked FALSE in place rather than quietly deleted — the reversal is the record. The door is genuinely reachable while parked, verified by reading rather than running: `ArchiveModal` iterates the RAW `networks()` store (`:174`), not the sidebar's filtered list, and the archive launcher renders unconditionally (`RailActions.tsx:525-533`).

## …and naming the archive as THE surface exposed a hole. Three legs, one predicted

`visibleArchiveForNetwork` subtracts the live channels, the live queries and the pseudo-rows, on the premise issue 402 wrote down: *a nav renders what it subtracts*. Dropping a whole network from the Sidebar's one `<For>` breaks that premise wholesale.

**Measured before the cure, in vitest**: a parked network holding an autojoin channel and a kicked pseudo-row rendered **ZERO** archive rows. Both swallowed.

- **Pseudo-rows — predicted.** `navPseudoChannelsForNetwork` returned the projection unconditionally on desktop.
- **Live channels — NOT predicted, and read as safe until measured.** `GET /networks/:slug/channels` returns the union of the credential's AUTOJOIN list and the live session's channels (`ChannelsController.index` → `Networks.merge_channel_sources/2`). A parked network has no session, so the union is exactly the autojoin list, at `joined: false` — sitting in `channelsBySlug` for the filter to subtract. The server had put those very rows in the archive *because* the session was gone (an absent session means an empty `active_keyset`: *"everything with rows qualifies for the archive when no session is live"*). The two sides disagreed and the operator's own autojoin channels paid.
- **Queries — CURED, with the server change this PR originally declined.** `ArchiveController.build_active_keyset/3` composed the live channels (empty when parked) with `open_query_targets/2`, a DB read of the open query windows with **no session gate**. So a parked network's DM windows were excluded from the archive by the SERVER while the desktop sidebar no longer drew them: one window, zero surfaces, and no client filter can restore a row that never arrived. It was handed up rather than taken; the call came back to take it, because **the asymmetry is pre-existing but THIS slice is what turns it into a live hole, so shipping the first two legs without it ships a measured regression**. It is also not a product call: it aligns the function with the promise the module already makes in words, which its channel arm kept and its query arm did not. Now one `case` over the session lookup answers for both arms.

## The cure

`navDrawsNetwork(slug)` in `lib/pseudoChannels.ts`, one level above the row projections: does the nav of THIS form factor draw ANY row for this network. Desktop answers `!isNetworkParked(...)`.

**Mobile answers YES, and that is measured, not assumed:** `BottomBar.tsx` iterates the raw `networks()` store and renders each network's channels and queries with no state filter of its own (`:138`, `:174`, `:215`), so on a phone those rows are still on screen and must still be subtracted. The sidebar filter this issue adds is desktop-only.

It does **not** live in `archive.ts` as a local `isNetworkParked` call: that would be a second statement of the parked policy plus a copy of the form-factor rule, and `lib/networkParked.ts` was extracted precisely to stop the first.

`failed` and `failing` keep drawing, **asserted rather than left implicit** — if those controls ever turn, the parked release has been widened into "any non-connected state", the generalisation that predicate exists to refuse.

## Shape

- `lib/grappa_web/controllers/archive_controller.ex` — the ONLY server change: an absent session empties the whole keyset, query windows included. One `case`, so there is a single place saying what "no session" means and the two arms cannot drift apart again. **This PR is therefore no longer cic-bundle-only**: one module body, no `VERSION` bump and no migration, so the server half is hot-reloadable.
- `lib/networkParked.ts` — the one statement of the rule. Its own module rather than a function in `lib/networks.ts` for a testing reason that is also a design one: `networks.ts` is a resource singleton every suite replaces wholesale with `vi.mock`, so a predicate living there could only ever be tested through a mirror of itself.
- `lib/pseudoChannels.ts` — `navDrawsNetwork`, and `navPseudoChannelsForNetwork` short-circuits through it.
- `lib/archive.ts` — consumes that one answer; subtracts nothing when the nav draws nothing.
- `Sidebar.tsx` — the filter at the ONE `<For>` over networks. Header, channels, queries and pseudo-rows all render inside it, so dropping the network drops all four by construction.
- `NETWORK_GREYED_STATES` narrows to `["failed"]` — parked can no longer reach that cascade, and two statements of one policy is how the next reader picks the wrong one.
- `Shell.tsx` — one clause in the cold-load restore gate, TERMINAL rather than provisional: a non-latching gate would yank focus off home the moment the operator hit `[Reconnect]`.

## Gates, on the rebased tree

- `bun.sh run check` — **5/5 green** (lock drift, test location, biome, tsc src, tsc e2e).
- `bun.sh run test` — **343 files / 6968 tests green**.
- `scripts/design-notes-gate.sh` — green; marker `<!-- entry #1985 -->` re-checked for uniqueness against the CURRENT `origin/main` tip, not the base this branched from.
- `scripts/union-rebase.sh` — the DESIGN_NOTES contribution came through the rebase two-sided intact (+116 / −0, unchanged in both directions).
- `mix compile --force --warnings-as-errors` (test env) — clean, Boundary included.
- `archive_controller_test.exs` — **16 tests, 0 failures**.
- **RED → GREEN watched for both cic legs**, which is their mutation evidence: parked returned its 4 pseudo-rows before the cure and `[]` after; the archive returned `[]` before and both rows after. The `failed` controls passed throughout.
- **Mutation on the server leg, run explicitly**: putting the query targets back into the keyset without a session turns **exactly one** test red — the new no-session one — while the live-session control stays green. Restored by checksum, tree verified clean, re-run green.
- The heavy gate (`check.sh`: format, credo, sobelow, dialyzer, full suite) is **deliberately left to CI** rather than run twice while the lane is contended.

## The two contract specs are re-aimed, and here is the only reason that licenses it

`cp15-b6-parked-disconnect-reconnect` and `issue100-reconnecting-badge` asserted the **pre-ruling** contract: a parked network stays in the sidebar, greyed and navigable, and the reconnecting badge has that greyed row to appear on.

> **The specs change because the product owner ruled the asserted behaviour wrong — NOT because they were flaky, and NOT to make them pass.**

That sentence is in each spec at the assertion that changed, so a later reader cannot mistake the rewrite for a relaxation. Nothing was weakened: no timeout raised, no assert deleted, no rerun bought.

- **cp15-b6** now asserts the network section is GONE while parked (and its channel rows with it) and **BACK** after Reconnect. The "comes back" half is asserted deliberately — a disappearance-only spec goes green on a sidebar that never recovers. The reason tooltip loses the header that hosted it; the reason itself is still asserted on the Home card, now its only surface.
- **issue100** now asserts the OPPOSITE outcome: the badge has **no surface** during a park→Reconnect, and the network returns anyway. An earlier revision of this PR argued the badge would still surface on an ordering (spawn-then-commit, row back in milliseconds while `connecting` is still set). **Measured: the ordering half is TRUE — the section does come back, and the spec still asserts it — and the conclusion drawn from it was FALSE.** The badge never entered the DOM. That is corrected here rather than dropped.
- Both stop reading `toHaveCount(0)` on the greyed child as proof a row is healthy: that passes for a row which never came back. Present **and** un-greyed, separately.

### Attribution for the issue100 rewrite was measured, not argued

| where | result |
|---|---|
| merge base `8ed1ef3d8` | **GREEN**, 1.3s |
| this branch | **RED 3/3**, ~21.8s, identical `TimeoutError` each run |

Not a flake and not pre-existing. The cause is structural: the badge has exactly **one** render site, inside the per-network `<For>`, and this branch filters parked networks out of that loop. The base kept a greyed row for the whole park, so the badge had a host. No row, no host, no badge. The badge's own machinery is untouched by this branch (empty diff on `reconnectingStatus.ts`), so nothing else can account for it.

**That makes the lost badge accepted collateral of the ruling, not an incident** — and "repairing" it would mean building for the badge, which the ruling forbids. The same licence applies as to the other two specs:

> **The spec changes because the product owner ruled the asserted behaviour wrong — NOT because it was flaky, and NOT to make it pass.**

⚠️ **Coverage lost, stated rather than buried.** issue100 was the ONLY e2e spec asserting the badge RENDERS (the two other specs naming it merely borrow its cleanup ritual). Re-aimed, no e2e covers the badge appearing at all — including on a spontaneous link drop, where the row is present and it presumably still works. That path was never covered here either (this spec always drove through a park), so nothing that WAS covered is dropped silently; but the render is now unguarded end-to-end. The dispatch keeps unit coverage in `userTopic.test.ts`; the render has none.

### Mutation: owed in the last revision, now PAID

- **cp15-b6** — baseline GREEN. With `visibleNetworks` stripped of its filter (the exact pre-1985 regression) the spec goes **RED at the named disappearance assert** (`toHaveCount(0)`, Expected 0 / Received 1). Not decoration. Mutant reverted and verified four ways (md5, `cmp`, empty `git status`, diff gone).
- **issue100** — the same mutant cannot redden its "section BACK" assert, and no mutant on that function can: that assert and the spec's own baseline are the SAME production predicate, so a mutant hiding connected networks kills the baseline first. Stated up front rather than discovered, and a green there is **not** evidence of decoration.

⚠️ Both were run **scoped** (`--grep`). A scoped green is not the ship gate — the full `scripts/integration.sh` still has to pass before merge (incident #268).

## Ruled: nothing is built for the `reconnecting…` badge

This section asked an open question in an earlier revision. It is answered. vjt, on `#grappa` 2026-09-10 at 14:59 Rome (12:59Z): *"ci si riconnette da HOME"* — *"non SERVE nient'ALTRO"*.

⚠️ **Provenance: RELAYED into this session, not read on IRC by this PR's author** — same posture as the archive ruling above, and stated as relayed rather than quoted first-hand.

So none of the three candidates is built: no badge grafted onto `$home`, no state machine holding `[Reconnect]` disabled longer, no teaching `$home` the progress signal. A parked network's row leaves the sidebar and reconnecting is `$home`'s business. This **extends** the 2026-09-08 *"redirect to home"* ruling rather than competing with it, and the cold-load restore gate in `Shell.tsx` is untouched by it.

The measurement taken before the ruling landed still describes today's behaviour — it is simply no longer a question. The badge of issue 100 renders inside the per-network `<For>` (`Sidebar.tsx:378-380`) off `reconnectingByNetwork`, which the server drives with `connection_progress` (`connecting` on the attempt, `connected` on 001). `connection_state` stays `parked` for that entire window by design — it is operator intent, and the badge is deliberately not that state. **So the badge fires while its host row does not exist.**

**Observed, and placed out of scope by that same ruling — recorded so the next reader does not re-discover it as new.** Between clicking `[Reconnect]` on `$home` and 001, the button relabels to `Reconnecting…` only while the awaited PATCH is in flight, then returns to `Reconnect` with the state word still reading `parked` and no badge anywhere — while the upstream link is in fact coming up. Before this slice the greyed sidebar row hosted the badge for exactly that window. It is a second-order effect of the ruling, deliberately **not** filed as an issue and not cured: *"non serve nient'altro"*.

This PR moves no badge code: on `<merge-base>..HEAD` the added lines naming `reconnecting|badge` are **0** in `Shell.tsx`, `Sidebar.tsx`, `lib/networkParked.ts`, `lib/pseudoChannels.ts` and `lib/archive.ts` — every mention lives in the issue-100 spec and in DESIGN_NOTES (positive control: `parked` = 139 added lines, so the grep was alive).

## Limits, stated

- **Not device-verified.** jsdom + a pure predicate. Whether the sidebar reads right with two parked networks on a phone is dogfood.
- **`BottomBar` is untouched** — a scope call, not an oversight, and now a measured input to the cure rather than a loose end: it is exactly why the mobile arm of `navDrawsNetwork` answers yes.
- **`ComposeBox.tsx` keeps its own `{parked, failed}` set**, deliberately: "can I type here" is a different question from "is there a row".
- **This slice is out of the 1.5.4 cut** (release scope, vjt's). `v1.5.5` was cut on `e7677913b` while this round was in flight; this branch is 1 commit behind that tip and still MERGEABLE, and the DESIGN_NOTES marker was re-checked for uniqueness against the NEW tip (0 hits, positive control 341).
- **The `reconnecting…` badge was not moved** — and by the 2026-09-10 ruling it is not to be. See the ruling section above; the measured second-order effect is recorded there as out of scope, not as a loose end.

### DIFF
--- a/cicchetto/e2e/fixtures/cicchettoPage.ts
+++ b/cicchetto/e2e/fixtures/cicchettoPage.ts
@@ -143,7 +143,7 @@ async function seedAuthLocalStorage(page: Page, token: string, subjectJson: stri
 export async function loginAs(
   page: Page,
   vjt: SeededUser,
-  opts: { noNetworks?: boolean } = {},
+  opts: { noSidebarNetworks?: boolean } = {},
 ): Promise<void> {
   await seedAuthLocalStorage(page, vjt.token, vjt.subjectJson);
   await page.goto("/");
@@ -159,19 +159,27 @@ export async function loginAs(
   // OR-style selector would be more brittle than a viewport-
   // conditioned one.
   //
-  // UX-7-C (2026-05-22) — `noNetworks: true` opt-in for users with
-  // NO networks bound (M-7 seeded admin-vjt has no credentials). The
-  // per-network-header selector waits forever in that case; switch
-  // to the registered home pane placeholder ("No networks bound")
-  // which is the post-/me steady-state render for empty-networks
-  // accounts. Opt-in rather than OR-selector because the
-  // `.home-pane-registered` element can RACE in front of the network
-  // section for normal bound users (homeData resolves off /me alone;
-  // network sidebar/bottom-bar wait for /networks + /channels) —
-  // weakening the post-loginAs invariant from "shell fully populated"
-  // to "DOM has homepane scaffolding". Callers that immediately
-  // interact with sidebar/bottom-bar windows would race.
-  if (opts.noNetworks === true) {
+  // UX-7-C (2026-05-22) — opt-in for accounts that will render NO network
+  // row, where the per-network-header selector waits forever; switch to the
+  // registered home pane, which is the post-/me steady state for them. Opt-in
+  // rather than OR-selector because `.home-pane-registered` can RACE in front
+  // of the network section for normal bound users (homeData resolves off /me
+  // alone; network sidebar/bottom-bar wait for /networks + /channels) —
+  // weakening the post-loginAs invariant from "shell fully populated" to "DOM
+  // has homepane scaffolding". Callers that immediately interact with
+  // sidebar/bottom-bar windows would race, so pass this ONLY when the spec
+  // stays on the home pane.
+  //
+  // issue 1985 — named `noSidebarNetworks`, not `noNetworks`, because there
+  // are now TWO ways to get here and the old name is false for the second.
+  // UX-7-C's case is "the account has no credential bound at all" (M-7's
+  // admin-vjt). The new one is "every network the account HAS is `parked`",
+  // which since issue 1985 draws no sidebar row either — the network exists,
+  // it is simply not a window right now, and `$home` is where it lives (with
+  // its [Reconnect] chip). The gate never cared which of the two it was: what
+  // it selects on is whether a network header will ever appear, so the name
+  // says that and one opt covers both.
+  if (opts.noSidebarNetworks === true) {
     await expect(page.locator(".home-pane-registered").first()).toBeVisible({
       timeout: SHELL_READY_TIMEOUT_MS,
     });
@@ -209,7 +217,7 @@ export async function loginAs(
 // Deliberately NOT `loginAs`, though the seeding half is now literally
 // the same call. The seeded admin (`admin-vjt`) has NO networks bound,
 // so `loginAs`'s per-network-header selector never resolves for it;
-// reaching it would mean `noNetworks: true`, whose own comment above
+// reaching it would mean `noSidebarNetworks: true`, whose own comment above
 // calls that a weakening of the post-login invariant. Adopting
 // `loginAs` would also buy `waitForUserTopicReady`, which none of the
 // twenty specs needs today — none of them composes `/join` — and which
--- a/cicchetto/e2e/tests/cp15-b6-parked-disconnect-reconnect.spec.ts
+++ b/cicchetto/e2e/tests/cp15-b6-parked-disconnect-reconnect.spec.ts
@@ -28,18 +28,21 @@
 //      surfaces connection_state=parked → selection.ts:287-316 fires
 //      → selection jumps to Home.
 //   3. Assert: selection is Home, HomePane renders the parked network
-//      card (slug + nick + reason + Reconnect button). Sidebar still
-//      shows the network section with `.sidebar-network-greyed`
-//      (rows remain visible — operator can re-navigate to view
-//      scrollback; greyed-cascade visual is intact).
+//      card (slug + nick + reason + Reconnect button). The sidebar
+//      network section is GONE, and its channel rows with it.
 //   4. Operator clicks `[Reconnect bahamut-test]` on the Home card
 //      (mirrors the same patchNetwork verb /connect would invoke).
 //      Server-side: Networks.connect → eager SpawnOrchestrator → DB
 //      flip + broadcast. Cic-side: networkBySlug.connection_state =
 //      connected → home card re-renders as connected.
-//   5. Assert: sidebar network section ungreys; SEED_CHANNEL row
-//      ungreys after autojoin (typed events flow through subscribe.ts
-//      as before).
+//   5. Assert: the sidebar network section is BACK (not greyed);
+//      SEED_CHANNEL row returns un-greyed after autojoin (typed events
+//      flow through subscribe.ts as before).
+//
+// issue 1985 (2026-09-10) — steps 3 and 5 were rewritten from
+// "stays, greyed" to "leaves, comes back". The spec is not being
+// relaxed: the product owner ruled the previously asserted behaviour
+// wrong. See the in-body comment at the disappearance assertion.
 //
 // Why click Reconnect instead of typing `/connect`: from Home, there
 // IS no ComposeBox — the only way to issue `/connect` from the Home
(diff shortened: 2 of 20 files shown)


vjt:
## CI: le due classi sono ora isolate dalla CI, non da un argomento

Giro precedente (head `1197b4708`… pre-proxy): **4 rossi su 3 file**, shard **1/4 e 3/4**.
Giro attuale, dopo il commit che tocca **solo il proxy di fixture**: **2 rossi su 2 file**, shard
**3/4 VERDE**, e i due superstiti sono esattamente:

- `cp15-b6-parked-disconnect-reconnect.spec.ts:81`
- `issue100-reconnecting-badge.spec.ts:66`

cioè **esattamente le due spec di contratto**, mentre `ux-5-br-home-reconnect` (`:104` e `:225`) è
passata. Log del job 746 KB — un giro vero, non una morte infra.

⇒ La spaccatura proposta nella descrizione — **una classe è il proxy di shell-ready, l'altra è una
collisione di contratto** — non è più un'argomentazione: è una misura differenziale. Il fix del solo
proxy ha curato **esattamente e solo** la classe proxy, e non ha spostato di un millimetro le due
spec che asseriscono il contratto pre-ruling.

**I due rossi restano di proposito**: sono l'evidenza della domanda aperta, non un problema da
nascondere. `cp15-b6:134` asserisce `toHaveClass(/sidebar-network-greyed/)` col commento a
`:132-134` che spiega che il grigio **è** il segnale "parked, no live messages" — cioè il contratto
che la ruling di issue 1985 (sparizione dura) ribalta.

Nessun assert è stato indebolito, nessun timeout alzato, nessun rerun fatto.

Issue #2018, opened 2026-09-09
The release-image gate is the only check on the published container, and a red there leaves it blind


vjt:
## What happened

On the `v1.5.4` release commit (`b68373237`) the job `deploy + probe the release image (amd64)` failed: the BEAM segfaults at boot inside `ghcr.io/vjt/grappa:v1.5.4`, at the migration step and **before any migration line**:

```
sys/unix/sys_signal_stack.c:103:sys_sigaltstack(): Internal error: Failed to set alternate signal stack
exit code 139
```

Two self-hosters (`Lucy`, `nextime`) answered `CTCP VERSION` with a bare `grappa 1.5.4` and were live on IRC minutes later, and vjt states they run the ghcr image — so the leading hypothesis is environmental (runner sandbox / seccomp around `sigaltstack`), not our code. That hypothesis is **not yet measured**; the A/B of `v1.5.3` vs `v1.5.4` on the same substrate is what would settle it, and it is in progress elsewhere.

## Why this is an issue regardless of the verdict

Measured, and this is the part that justifies filing:

- the job is **success** on `v1.5.2` (`ac11dece6`) and on `v1.5.3` (`ed6241d74`);
- it is **absent** on plain main (`7779dfef6`).

So it is **release-only by construction** — no run on main could ever have caught it — and it is the **only gate that covers the published image**. If the red turns out to be the runner and gets filed away as "not ours", the gate is left blind: the next real breakage of the published image is caught by nobody, and the people who eat it are the self-hosters (the AWS one-click installer of #503 included).

## Ask

Put the release-image gate back in a position to see:

- make the probe run somewhere it can actually boot the image (or make the sandbox difference explicit and asserted), so a green means "the image boots" rather than "the runner allowed it today";
- keep coverage for **both** published architectures — the failing job is `amd64` explicitly while the publish job is `ghcr, amd64 + arm64`, so a healthy arm64 says nothing about amd64;
- give the job a signal on plain main, or state deliberately why release-only is the right cadence.

Side note, not the cure: katia suggested a `dgoss` smoke test on the release image (#grappa, 2026-09-09). It is a better way to say *what* to assert about the image, but it runs the container on the same runner — if the cause is the runner sandbox, dgoss hits the same `sigaltstack` and adds no coverage. Worth doing on its own merits, not as the fix for this.

## Not measured

[...]

PR #2020, opened 2026-09-09
ci: boot both published architectures and name the harness failures (issue 2018)

vjt:
Refs issue 2018. This PR does **not** close it — see the last section for why.

## What this does

Three changes to the release-image gate, shipped together on purpose.

**1. Both published architectures are BOOTED, not just built.** The docker job
publishes `linux/amd64,linux/arm64`; the smoke job ran on `ubuntu-latest` alone
and said so out loud — *"amd64 only: the arm64 leg is proven by the build"*.

Measured, and it is why that posture falls: **every job in every workflow in
this repository was `ubuntu-latest`** — 12 of 12 across `ci.yml`,
`integration.yml` and `release.yml` — so no arm64 binary had ever been EXECUTED
here by anything. `infra/docker/assert-abi-lockstep.sh` does gate arm64 at build
time, per-platform, but it proves the runtime stage can **link** the release,
not that the VM **starts**. For an arm64 puller of `ghcr.io/<owner>/grappa` the
first process ever to run the image was a user's, on every release that has ever
shipped. That is structural, not intermittent, and `ubuntu-24.04-arm` is free on
a public repository — the cure is a matrix and nothing else. The driver is
unchanged: `docker pull` resolves the multi-arch manifest for whatever host it
lands on, candidate and upgrade fixture alike.

The amd64 check-run **keeps its exact name**, `deploy + probe the release image
(amd64)`, because `name:` is spelled with the matrix value rather than left to
the default suffix.

`fail-fast: false` is load-bearing and is not a softening: without it an amd64
failure CANCELS the arm64 leg, and that cancelled leg is exactly the reading
worth having.

**2. The runner's facts are captured on every run — green included.** This is
the half that is easy to get backwards. What the `v1.5.4` incident lacks is not
only the red sample, it is the **green** one: an `ImageOS`, a `MINSIGSTKSZ`, a
set of XSAVE flags cannot be read as anomalous by anyone who has never seen them
on a run where the image boots. Facts only on failure produce a sample of one
class and nothing to hold it against.

**3. A failure carrying the ERTS-startup signature is NAMED, not forgiven.** No
retry, no `continue-on-error`, no downgrade, exit status untouched. What is added
is the distinction that cost a night: a VM that never started and an image that
is genuinely broken used to arrive identically, as "the job is red".

Plus: release-only is now **stated as deliberate**. The subject of this gate is
the published image, and before the tag there is none — that is arithmetic, not
a cadence anyone chose.

## What was refused

- **The issue's first ask, as written** ("run the probe somewhere it can actually
  boot the image, or make the sandbox difference explicit and asserted"). There
  is no second substrate — production is the FreeBSD jail and has no docker, a
  self-hosted runner would be a third production substrate, and self-hosters are
  field evidence rather than a gate. Asserting the sandbox difference is worse:
  nobody has measured what that difference **is**, so the assertion would encode
  a guess and stay green in exactly the case the guess is wrong.
- **A `schedule:`** for the `docker_validation` dry-run — a cron red has no
  owner. There is no `schedule:` anywhere in this repo and that stays a decision.
- **Retries, `continue-on-error`, wider timeouts, dgoss-as-cure.**
- **A BEAM preflight.** Designed, then dropped: it cannot be exercised anywhere
  available, and an untested new invocation in the release path is a new way for
  every release to go red, in exchange for attribution the classifier already
  delivers minutes later.

## 🔴 What this does NOT cure — please do not read a cure into it

**The non-determinism is untouched and its mechanism remains unknown.** This gate
produced red and then green on byte-identical input, which means it can lie in
**both** directions, and the direction that hurts is the false green: the absence
of the failure is not by itself evidence that the image boots.

What ships here cures the **blindness** (an arm64 half nobody had ever executed)
and the **legibility** (a red nobody could classify). Neither of those is the
flip. The honest expectation is an instrument, not a diagnosis: the next
occurrence arrives with the runner's facts attached and a comparison class to
hold them against — which is the measurement nobody has been able to make so far.

## Cost accepted explicitly

Two runner substrates means two chances to trip over an environmental flake. With
these checks blocking, **the `v1.5.4` release would have stopped**. That is
accepted: the artefact is published, and a smoke test that cannot stop anything
is the silently-passing smoke test this driver exists to refuse. The human loop
is the one already used — rerun the job, and withdraw if it is still broken — and
a considered rerun on a **named** red costs a minute, where a triage on a mute
red cost a night.

## A claim withdrawn before it was published

An earlier draft carried a second finding: that `Dockerfile.release` leaves
`elixir:1.19-otp-28-alpine` and `alpine:3.24` on floating tags, so the published
image is not a pure function of the source tree. Literally true, wrongly framed.
Both are excluded **by name** from the #103 digest-pin gate with the argument
written there — the Elixir tag "carries the pin" and is held to `.tool-versions`
by `toolchain_pin_test.bats`, and the alpine floor floats on purpose so security
patches keep flowing, with `assert-abi-lockstep.sh` proving compatibility rather
than freezing bytes. A decision with two gates over it, not a gap. Recorded in
DESIGN_NOTES so nobody spends the same hour.

## Verification

| gate | result |
|---|---|
| `scripts/bats.sh test/infra/` (~50 suites, 588 cases) | **rc=0**, 0 failures |
| `shellcheck -x scripts/smoke-release-image.sh` | **rc=0** |
| `scripts/design-notes-gate.sh` (and against `origin/main`) | **rc=0** — *"1 new entry heading(s), separator and marker present"*, i.e. it saw the entry rather than answering "nothing to check" |
| workflow YAML parses; the conditional matrix folds to one line | verified |
| harness signature: real `v1.5.4` line matches; third alternative matches alone; `Ecto.MigrationError` does **not** | verified — the negative control is the one that matters, since a classifier mislabelling a real image break as "harness" would excuse the failure this gate exists to catch |
| `runner_facts` on a substrate with no `/proc` | exits 0, degrades to `unavailable` |

**Not verified, and it needs a runner:** the Linux output shape of the facts
block, the arm64 leg actually running, and the classifier firing on a real
failure. The honest proof of all three is a `docker_validation` dispatch.

## Notes

- The DESIGN_NOTES arithmetic was predicted before the append (+134 / −0) and
  matched exactly. On a pure append the "deletions zero" half of the two-sided
  check is a **tautology** and proves nothing; the additions half and the
  boundary shape read on the file are the parts that carry weight.
- #2018 is deliberately **not** closed by this PR: the issue's own question — why
  that runner instance failed — is still unanswered, and this changes what the
  gate can see rather than answering it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1377,7 +1377,7 @@ jobs:
           provenance: false
 
   smoke:
-    name: deploy + probe the release image (amd64)
+    name: deploy + probe the release image (${{ matrix.arch }})
     needs: [docker]
     # Exactly the docker job's trigger set, all three shapes: a real tag and a
     # repair (probe what was just PUBLISHED) or a docker_validation dry-run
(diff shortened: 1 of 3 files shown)


vjt:
`docker_validation` dry-run dispatched on this branch by the orchestrator: run
`34299656261`, **`completed/success`**.

```
build + push release images (ghcr, amd64 + arm64) = success
deploy + probe the release image (amd64)          = success
build + prove .deb / Arch / .rpm, publish assets  = skipped   (dry-run, push:false)
```

**What that run does and does not establish, stated precisely, because the
`success` is narrower than it looks.**

It establishes: the workflow YAML parses and the conditional matrix resolves on
the dispatch path, producing exactly one leg; the amd64 check-run keeps its exact
former name, `deploy + probe the release image (amd64)`; the driver changes — the
unconditional runner-facts block and the harness classifier — do not break the
gate that already existed; and the dry-run publishes nothing, with the four
packaging jobs correctly skipped.

It does **not** establish that the arm64 leg runs. That is not a gap in the run,
it is the matrix behaving as designed: the dispatch path folds to one leg on
purpose, because the `docker_validation` build re-exports an amd64 image from the
gha cache and giving it an arm64 leg would mean a second cache for a path nothing
fires automatically. The reasoning is in the workflow comment, and the trade is
stated there as "better light and stated than complete and never exercised".

**So the honest position: `ubuntu-24.04-arm` will execute for the first time on
the next real release.** That is the same arithmetic the rest of this work rests
on — the published arm64 image does not exist before the tag either — but it is
worth writing down rather than leaving a green dispatch to imply otherwise. This
PR's own description already listed the arm64 leg under "not verified, and it
needs a runner"; that line is still accurate after this run.

It does not change the posture that a red arm64 leg should stop a release. A leg
that has never run is exactly the one that must not be allowed to fail silently:
if it is broken as a job, a blocking red costs one rerun cycle, while a
non-blocking one restores zero arm64 coverage wearing the appearance of coverage
— the false green this whole slice exists to refuse.

Issue #2034, opened 2026-09-10
cic: notifications for a conversation are never dismissed when the user reads it — only a tap closes them


vjt:
Notifications for a conversation stay in the shade after the user has already read it. Tapping one then yanks the reader to a channel they caught up with minutes ago.

The tap path is already correct: `notificationclick` calls `event.notification.close()` (`cicchetto/src/service-worker.ts:235`). What is missing is the other half — nothing closes a notification when its conversation comes back into view.

## What exists already

- Every push notification carries a per-conversation `tag` (`cicchetto/src/lib/pushPayload.ts:108`, server-supplied).
- The app already tracks foreground/visibility in several places (`main.tsx` visibility reporting, `shouldSuppressPush()` in the service worker gates the toast when a window is visible).

## What is missing

`getNotifications()` has zero occurrences in the whole cicchetto tree. Once a notification is on screen there is no code path that ever takes it back, so the shade only drains by hand or by tap.

## Proposal

When a conversation becomes visible — the tab/window is focused *and* that conversation is the active one — close the outstanding notifications carrying its tag:

```ts
const reg = await navigator.serviceWorker.ready;
for (const n of await reg.getNotifications({ tag })) n.close();
```

Trigger points are the ones the resume path already uses (`visibilitychange` → visible, `pageshow`, `focus`), plus switching the active conversation while visible.

## Scope

Deliberately small. No new state, no new server field, no new permission — it reuses the tag that is already on the payload and the visibility signals that already exist. If it grows past "close what the user is looking at", it has grown too far.

## Notes

Raised on #sbiffo: the observed behaviour is that a tap dismisses, a read does not.

Issue #378, opened 2026-07-22
Web push for /notify presence transitions (follow-up to #247)


gmsrc:
## Summary

Extend the cluster B2/B4 Web Push pipeline so that `/notify` (#247, PR #328) presence transitions fire OS-level push notifications when the PWA is closed. Two new global boolean prefs gate it: `presence_online` (default **true**) and `presence_offline` (default **false**). Only genuine transitions push — baseline snapshots never do. The #182 foreground-suppression gate applies unchanged. No per-nick logic: the watch list curates WHO, the prefs gate push-vs-toast-only globally.

## Why server-side

The in-app toast (`PresenceToasts.tsx`) only exists while a client is attached; the push must originate where the presence transition is classified — `Session.Server`'s `presence_changed` effect arm — exactly like message pushes originate at the persist arm.

### 1. Dispatch point + module shape

**Call site.** The `{:presence_changed, nick, presence, kind, source}` arm of `apply_effects/2` (`lib/grappa/session/server.ex:3340-3355`) already projects the classified transition to the wire broadcast. Add one call next to the broadcast — one effect, two consumers:

```elixir
defp apply_effects([{:presence_changed, nick, presence, kind, _source} | rest], state) do
  :ok = Grappa.PubSub.broadcast_event(...)          # existing, unchanged
  :ok = PushTriggers.dispatch_presence(nick, presence, kind, push_ctx(state))
  apply_effects(rest, state)
end
```

`source` (`:monitor | :watch`) is a transport detail — irrelevant to the notification, not passed through.

**Shared ctx helper.** `maybe_dispatch_push/2` (server.ex:3262-3281) assembles the ctx map inline at server.ex:3274-3279. Extract a private `push_ctx(state)` returning `%{subject, subject_label, network_slug, own_nick}` (all read from existing state fields — `state.subject`, `state.subject_label`, `state.network_slug`, `state.nick`) and use it from both call sites. No state-shape change.

**New trigger function.** `Grappa.Push.Triggers.dispatch_presence/4`, mirroring `evaluate_and_dispatch/2` (triggers.ex:104-142) exactly:

```elixir
@spec dispatch_presence(String.t(), :online | :offline, Grappa.Session.Presence.change_kind(), ctx()) :: :ok
def dispatch_presence(nick, presence, :transition, ctx) do
  {:ok, _} = Task.start(fn ->
    prefs = UserSettings.get_notification_prefs(ctx.subject)

[...]

Issue #182, opened 2026-07-04
cic: push delivered while PWA is foregrounded/on-screen (foreground-suppression regression)


vjt:
**Regression.** Push notifications are delivered even when the subscribed device has the PWA **on-screen / foregrounded** (the user is actively looking at the app). When the user is active in the app, OS push notifications should be suppressed (or routed to an in-app indicator), not shown as system notifications.

Reported on device (iOS, app visible/foregrounded).

**Expected:** when a visible/focused client for the app exists, the SW `push` handler suppresses the system notification (or downgrades it to an in-app indicator).

**Investigation direction:**
- In the SW `push` handler, check `clients.matchAll({ type: 'window', includeUncontrolled: true })` for a client with `visibilityState === 'visible'` / focused → if present, skip `showNotification` (or downgrade to in-app).
- Flagged as a **regression** — bisect recent changes to the push / SW notification-display path; a foreground-visibility check was likely dropped or broken.

Related: #181 (push subscription lifecycle — client re-disable + server stale record; separate mechanism). Companion #120 (surface SW-registration failure).

vjt:
**Spec clarification (from reporter).** This was a **server-side, cross-device presence gate**: a push was **not sent if *any* of the user's devices had the PWA on-screen** (user considered active). So the regression is in that server-side presence suppression — not (only) SW-side `showNotification` gating.

**Refined investigation direction:**
- Server: before sending a push, check whether *any* of the user's connected clients reports an active/visible (on-screen) session → if so, skip the send. This presence gate existed and regressed.
- Likely mechanism: clients report foreground/visibility (WS presence / heartbeat) to the server; the send path consults that presence. Bisect where the presence check was dropped, or where clients stopped reporting visibility.
- SW-side `showNotification` gating (original note) is a secondary/backup layer, not the primary mechanism that regressed.

vjt:
Fixed by server-side foreground push-suppression (merged 62831c0, COLD-deployed to m42 prod, healthz `ok`; cic bundle `BHkphb_7`).

**Root cause.** The suppression was client-side in the SW (`shouldSuppressPush` → `clients.matchAll().visibilityState`). Its pure predicate is correct, but `clients.matchAll` is UNRELIABLE on iOS PWAs (empty/non-"visible" client list while foregrounded), so the client gate can't detect iOS foreground.

**Fix — moved suppression SERVER-side**, driven by the reliable page-context `document.visibilitychange` signal:
[...]

PR #2048, opened 2026-09-10
cic(#2034): dismiss a conversation's notifications when it comes into view

gmsrc:
## Summary

**The proposed `getNotifications({ tag })` does not work for DMs, so this matches on the notification's own deep-link URL instead.** An exact-match tag filter forces cic to *spell* the server's tag — a second copy of `Grappa.Push.Payload`'s format — and that copy is wrong by construction for a query: the server tags with the peer nick RAW (`libera:Alice`, per the key/display split), while the selection store holds the canonical window spelling (`alice`). The filter asks for `libera:alice`, matches nothing, and closes nothing — silently, in the case DMs make most common. Each open notification is instead resolved through its own `data.url` by the `parsePushTargetUrl` the tap path already uses, so there is one spelling of push identity rather than two.

Everything else follows the issue: no new state, no new server field, no new permission.

## What changed

- **`lib/notificationDismiss.ts`** — the sweep. Whenever a conversation comes into view, enumerate what the registration is showing and close the notifications naming the focused window.
- **`lib/pushTarget.ts`** — `pushTargetSelection` extracted from `routePushTarget`. The tap path walks the push-target → window mapping forwards ("this notification names the window to focus"); the sweep walks it backwards ("does this notification name the window already focused"). One mapping, so a tap and a dismissal cannot disagree about which window a notification belongs to. No behaviour change; `pushTarget.test.ts` passes untouched.
- **`main.tsx`** — `installNotificationDismiss()`, mounted next to the tap path it completes.

## Notes for review

**Presence banners fall out for free.** `build_presence/3` (#378) deep-links to the same `?network=&channel=` shape, so opening a peer's query also clears their online/offline banner. A tag filter would have had to be taught this; matching on the window states the rule once — close what the reader is looking at.

**Two triggers.** The reactive arm is `isDocumentVisible()` crossed with `selectedChannel()` — both halves of "in view": the tab came back, or the reader switched windows while it was already in front of them. `pageshow` is the second, for the reason `resumeProbe.ts` already carries it (an iOS PWA thaws without reporting a visibility transition). The sweep is idempotent, so the overlap costs nothing.

**`getRegistration()`, not `serviceWorker.ready`** (which `push.ts` uses). `ready` never settles when no service worker is registered, and this runs on every focus flip rather than once at opt-in, so a browser with SW disabled would accumulate one dangling promise per sweep. A hidden document is not enumerated at all — that is where a notification is still doing its job.

## Since review (2 reviews, 4 commits)

- **`d11baff9`** — re-establish `isDocumentVisible()` AFTER the two awaits, before the close loop. The one finding that could destroy an unread notification: `getNotifications()` is an IPC round-trip, and a push shown inside that gap was being closed unseen.
- **`094048f9`** — the store stubs are real Solid signals now, so the effect can actually re-run, and the **primary user story** (switch window while visible) has a test. The named mutant `untrack(selectedChannel)` killed that trigger with all seven old tests green; it is red now. Both new tests were run against their mutants — each kills exactly its own.
- **`023eee0d`** — the docstring credited `setSelectedChannel`'s fold; the sweep never calls the setter. Corrected to `isActiveSelection`'s `foldChannelKey`.
- **`e68505e1`** — DESIGN_NOTES retracts "handled one layer up" / "a state that should not occur" for the already-in-view residual; the tree says the opposite. Also records the race, the CHANTYPES approximation, the badge acquittal, and the one unmeasured fact (whether an iOS banner over a foregrounded PWA produces a blur/focus pair).

**Deferred, not fixed:** the already-in-view sweep mechanism (a `postMessage` from the push handler). Both reviews called it a follow-up. Not filed as an issue — enqueue is vjt's.

## Testing

`src/__tests__/notificationDismiss.test.ts` — 10 tests: the DM arm that fails on a byte compare, the window-switch trigger, the visibility-flip trigger, `pageshow`, and the mid-sweep race. Mutation-checked — two named mutants, each killed by exactly one test.

**No e2e, and the ceiling is the harness.** `self.registration.showNotification(...)` rejects under headless Playwright with *"No notification permission has been granted for this origin"* even after `context.grantPermissions(["notifications"])` — the finding `e2e/fixtures/pushTap.ts` already documents for the tap path. A dismissal test needs a notification that EXISTS, so it hits the same wall one step earlier: with nothing shown, `getNotifications()` returns an empty list and the test would pass while closing nothing.

Local gates: `bun run check` green (5 stages, 0 failed). Full vitest suite 6946 passed / 7 failed — `mentions`, `selection`, `subscribe`, `userTopic`, all timeout-shaped, all in the known Windows-host flake family; re-run in isolation they are 362/362 green. Not manually dogfooded (the reviewer was unavailable to test this round).

Addresses #2034. (Closing keyword deliberately dropped — enqueue and closure are vjt's in this repo.)

### DIFF
--- a/cicchetto/src/__tests__/notificationDismiss.test.ts
+++ b/cicchetto/src/__tests__/notificationDismiss.test.ts
@@ -1,0 +1,216 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import {
+  dismissNotificationsForActiveWindow,
+  installNotificationDismiss,
+} from "../lib/notificationDismiss";
+
+// #2034 — a notification is dismissed when its conversation comes back
+// into view, not only when it is tapped.
+//
+// Boundary: the selection store and the network / query stores are mocked
+// (pushTarget.test.ts precedent) so the assertions are about the
+// composition this module owns — enumerate the open notifications, resolve
+// each one's `data.url` to a WINDOW IDENTITY, close the ones naming the
+// focused window — rather than about the fold rules those stores already
+// prove. The DM case below is why the identity step exists at all: the
+// notification carries the RAW wire nick `Alice` while the focused window
+// is `alice`, so a byte compare would leave it sitting in the shade.
+
+// The fixture's whole mutable world: which window is focused, and whether
+// the document is visible. Hoisted because the `vi.mock` factories close
+// over it, and reset in `beforeEach` — an implementation swapped in by one
+// test is exactly the leak that made the pageshow arm below pass for the
+// wrong reason while it was being written.
+const fixture = vi.hoisted(() => ({
+  visible: true,
+  active: { networkSlug: "libera", channelName: "#sniffo", kind: "channel" } as {
+    networkSlug: string;
+    channelName: string;
+    kind: string;
+  } | null,
+}));
+
+const FOCUSED_CHANNEL = { networkSlug: "libera", channelName: "#sniffo", kind: "channel" };
+
+vi.mock("../lib/selection", () => ({
+  // Stands in for the real exact-tuple compare (#243 `isActiveSelection`,
+  // which folds the channel KEY on the way in). Byte-equal here: the fold
+  // is selection.ts's contract and is proven there.
+  isActiveSelection: vi.fn(
+    (sel: { networkSlug: string; channelName: string; kind: string } | null) => {
+      const active = fixture.active;
+      return (
+        sel !== null &&
+        active !== null &&
+        sel.networkSlug === active.networkSlug &&
+        sel.channelName === active.channelName &&
+        sel.kind === active.kind
+      );
+    },
+  ),
+  selectedChannel: vi.fn(() => fixture.active),
+  setSelectedChannel: vi.fn(),
+}));
+
+vi.mock("../lib/documentVisibility", () => ({
+  isDocumentVisible: vi.fn(() => fixture.visible),
+}));
+
+vi.mock("../lib/networks", () => ({
+  networkBySlug: vi.fn((slug: string) =>
+    slug === "libera" ? { id: 1, slug: "libera", kind: "user" } : undefined,
+  ),
+  channelsBySlug: vi.fn(() => ({})),
+  networks: vi.fn(() => []),
+  networkIdBySlug: () => undefined,
+}));
+
+// The open query window spells the peer `alice`; the wire spells it
+// `Alice`. `canonicalQueryNick` is the store verb that resolves one to the
+// other — it returns the OPEN window's stored `targetNick` on a
+// casemapping-aware match, so the lowercasing stub here stands in for a
+// window that was first opened as `alice`.
+vi.mock("../lib/queryWindows", () => ({
+  openQueryWindowState: vi.fn(),
+  canonicalQueryNick: vi.fn((_networkId: number, nick: string) => nick.toLowerCase()),
+}));
+
+type FakeNotification = { data: unknown; close: ReturnType<typeof vi.fn> };
+
+const notification = (url: string | undefined): FakeNotification => ({
+  data: url === undefined ? null : { url },
+  close: vi.fn(),
+});
+
+const originalSW = (navigator as Navigator & { serviceWorker?: unknown }).serviceWorker;
+
+function installFakeServiceWorker(notifications: FakeNotification[]): {
+  getNotifications: ReturnType<typeof vi.fn>;
+} {
+  const getNotifications = vi.fn().mockResolvedValue(notifications);
+  Object.defineProperty(navigator, "serviceWorker", {
+    configurable: true,
+    value: {
+      getRegistration: vi.fn().mockResolvedValue({ getNotifications }),
+      addEventListener: vi.fn(),
+    },
+  });
+  return { getNotifications };
+}
+
+function restoreServiceWorker(): void {
+  if (originalSW === undefined) {
+    // biome-ignore lint/performance/noDelete: test cleanup
+    delete (navigator as unknown as { serviceWorker?: unknown }).serviceWorker;
+    return;
+  }
+  Object.defineProperty(navigator, "serviceWorker", {
+    configurable: true,
+    value: originalSW,
+  });
+}
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  fixture.visible = true;
+  fixture.active = { ...FOCUSED_CHANNEL };
+});
+
+afterEach(restoreServiceWorker);
+
+describe("dismissNotificationsForActiveWindow", () => {
+  it("closes the notifications naming the focused channel and leaves the others", async () => {
+    const mine = notification("/?network=libera&channel=%23sniffo");
+    const elsewhere = notification("/?network=libera&channel=%23other");
+    const otherNetwork = notification("/?network=azzurra&channel=%23sniffo");
+    installFakeServiceWorker([mine, elsewhere, otherNetwork]);
+
+    const closed = await dismissNotificationsForActiveWindow();
+
+    expect(mine.close).toHaveBeenCalledTimes(1);
+    expect(elsewhere.close).not.toHaveBeenCalled();
+    expect(otherNetwork.close).not.toHaveBeenCalled();
+    expect(closed).toBe(1);
+  });
+
+  it("resolves the RAW wire nick of a DM to the window spelling before comparing", async () => {
+    // The focused window is the query `alice`; the payload spells the peer
+    // `Alice`, because the server keeps `sender` raw for display.
+    fixture.active = { networkSlug: "libera", channelName: "alice", kind: "query" };
+    const dm = notification("/?network=libera&channel=Alice");
+    installFakeServiceWorker([dm]);
+
+    await dismissNotificationsForActiveWindow();
+
+    expect(dm.close).toHaveBeenCalledTimes(1);
+  });
+
+  it("leaves a notification whose data carries no usable URL", async () => {
+    const dataless = notification(undefined);
+    const malformed = notification("/?network=libera");
+    installFakeServiceWorker([dataless, malformed]);
+
+    const closed = await dismissNotificationsForActiveWindow();
+
+    expect(dataless.close).not.toHaveBeenCalled();
+    expect(malformed.close).not.toHaveBeenCalled();
+    expect(closed).toBe(0);
+  });
+
+  it("does nothing while the document is not visible", async () => {
+    fixture.visible = false;
+    const mine = notification("/?network=libera&channel=%23sniffo");
+    const { getNotifications } = installFakeServiceWorker([mine]);
+
+    const closed = await dismissNotificationsForActiveWindow();
+
+    // Not merely "closes nothing": the shade is never enumerated, so a
+    // backgrounded tab costs no service-worker round-trip on unrelated
+    // churn — and a hidden tab is exactly where the notification is still
+    // doing its job.
+    expect(getNotifications).not.toHaveBeenCalled();
+    expect(mine.close).not.toHaveBeenCalled();
+    expect(closed).toBe(0);
+  });
+
+  it("is a no-op without a service worker registration", async () => {
+    Object.defineProperty(navigator, "serviceWorker", {
+      configurable: true,
+      value: { getRegistration: vi.fn().mockResolvedValue(undefined) },
+    });
+
+    await expect(dismissNotificationsForActiveWindow()).resolves.toBe(0);
+  });
+
+  it("is a no-op when the service worker API is absent altogether", async () => {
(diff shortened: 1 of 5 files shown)


vjt:
Design author here — the extension is right and my suggestion in #2034 was wrong for DMs. Verified in the source rather than taken on the summary:

`Push.Payload.build/3` picks `{sender, sender, sender}` on the DM branch and tags `"#{network_slug}:#{dedup_key}"` (`lib/grappa/push/payload.ex:148-160`), so the DM tag carries the **raw wire nick**. `build_presence/3` right below it folds with `Identifier.canonical_target/1` — presence folds, messages do not. A client-side `getNotifications({ tag })` would have to reproduce that asymmetry from state it does not hold, and would silently match nothing exactly where DMs matter. Resolving through `data.url` is the correct call.

Three notes, none of them blocking:

1. **The channel-fold rationale in `pushTargetSelection`'s docstring is wrong, and it is the kind of wrong that gets "simplified" later.** It says channels need no canonicalisation *because `setSelectedChannel` folds the KEY itself* — true on the forward path, but the sweep never calls the setter. What actually saves the backward path is `isActiveSelection`, which runs its argument through `foldChannelKey` before `sameSelection` (`lib/selection.ts`, #1396). Behaviour is correct; only the stated reason is. Worth fixing the comment so nobody later drops the fold in `isActiveSelection` believing the setter covers it.

2. **A notification that arrives while its conversation is already in view is never swept.** All three triggers are transitions — `visibilitychange`/focus, a `selectedChannel()` change, `pageshow`. None of them fires when the reader is already looking at the window and the banner appears underneath. Normally `shouldSuppressPush` in `service-worker.ts` covers that case, but its own comment says `clients.matchAll` visibility is unreliable on iOS PWAs and that the server-side gate (#182) defaults to deliver-leaning in the just-connected window — i.e. the residual case lands precisely on the device this feature is for. One `postMessage` from the push handler to the visible client (or a sweep at the end of `handlePush`) closes it. Fine as a follow-up.

3. `getRegistration()` over `serviceWorker.ready` is the right pick for a per-focus-flip call, and the reason given is accurate.

No e2e is the honest answer given the harness wall documented in `e2e/fixtures/pushTap.ts`; the round-trip through the shared mapping is what makes it defensible.

I have asked orch for a second review as requested.

vjt:
## Second review — run **blind**, then reconciled against the first

Method, because it changes what this is worth: the reviewer wrote and saved her judgement **before opening any comment on this PR**, then read the existing review and reported where she diverged. A second review that has already read the first is a confirmation, not a second opinion. Inputs were the issue body, the diff, and direct reads of the unchanged tree at `origin/main` — no checkout, nothing touched, **no test run and no gate**.

### Where two blind passes landed on the same thing

The residual case — **a notification that arrives while its conversation is already open and visible** — was found independently by both passes, from the same two pieces of evidence (`shouldSuppressPush`'s own "unreliable on iOS PWAs" note, and issue 182's deliver-leaning server default). Neither read the other. Two blind passes agreeing on the residual is worth more than either alone: treat that as the one real behavioural gap in the change.

Tracing the two triggers as written: the reactive arm tracks `selectedChannel()` and `isDocumentVisible()`, and an arriving notification changes **neither**; `pageshow` does not fire because nothing resumed. So the banner sits in the shade until the reader switches windows or backgrounds and returns.

### Where the first review was right and the second was wrong

The docstring point stands, and the second pass **missed it**. `pushTargetSelection`'s docstring justifies the absent channel step by saying `setSelectedChannel` folds the key itself — true on the forward path, but the sweep never calls the setter, so the stated reason does not apply to the direction being added. What actually saves it is `isActiveSelection`'s `foldChannelKey`. She had measured the correct mechanism and still did not notice the PR credits the wrong function; she corrected her own note rather than leaving it. The framing in the first review is the better one: a later reader could remove the fold from `isActiveSelection` believing the setter covers it, and the forward-path tests would stay green.

### Three findings the first review does not carry

**1. A real race, and it is the only finding on either side that can DESTROY an unread notification.** `notificationDismiss.ts` reads `isDocumentVisible()` **once**, then two `await`s separate that gate from the close loop — and `getNotifications()` is a real IPC round-trip to the service worker, not a microtask.

Scenario: reader switches to `#sniffo` while visible → effect fires → sweep starts. Before it resolves they lock the phone. A push for `#sniffo` lands in that window and **is** shown (the suppression gate is leaky exactly as above). `getNotifications()` resolves with that brand-new banner, the selection has not changed so `namesActiveWindow` says yes, and the sweep closes a notification the reader never saw.

This inverts the module's own stated principles — *"a hidden tab is exactly where a notification is still doing its job"* and *"leaving a banner up is the harmless direction of that error … closing on a guess destroys a notification nobody read."* One line: re-check `isDocumentVisible()` after the awaits, immediately before the close loop, so the precondition is true when it is acted on rather than when it was asked.

**2. The feature's primary trigger has no test, and the spec cannot express one.** Six of the seven tests are real — each would fail on a plausible wrong implementation, and the DM test earns its keep. This is not a mirror suite. But `installNotificationDismiss` has exactly one test and it is `pageshow`. There is no test that **switching windows while visible sweeps**, which is the primary user story in the issue.

It is not merely missing. The mock replaces `selectedChannel` with a plain `vi.fn`, not a Solid signal, so mutating the fixture notifies nothing and the `createEffect` can never re-run inside that file.

Named mutant: change the effect's read to `untrack(selectedChannel)`. The window-switch trigger is then **dead** — only the boot sweep and `pageshow` remain — and **all seven tests still pass**. Cheap fix: back the fixture with a real `createSignal` in the mock factory (the accessor keeps its shape) and add one case — visible, switch selection, assert a second sweep. That turns the mutant from silent to red. Given that "the user reads it" *is* the issue, this is the gap we would most want closed before merge, and it is cheaper than either behavioural fix.

**3. Pre-existing, now load-bearing in a second place** (a sentence, not a fix). `parsePushTargetUrl` decides channel-vs-query from four hardcoded RFC-2812 sigils (`# & ! +`), ignoring the network's advertised CHANTYPES — and cic has `lib/chantypes.ts`. On a network advertising an exotic chantype, a channel notification is classified as a query, gets `canonicalQueryNick` applied, compared against a channel selection, and is never dismissed. That is existing tap-path behaviour; this PR makes the same approximate parse load-bearing in a second direction. Worth a line so it is not rediscovered as a dismissal bug.

### One acquittal, on the record

"The shade empties but the count does not" is the obvious thing to fear here, and it **does not land**: `lib/badge.ts` documents a second writer plus a reconciler — on every visible event `mountBadgeReconcile` re-pulls the authoritative `/me` count and force-applies it, bypassing the signal-equality skip. The badge axis is server-authoritative and independently reconciled, so the shade and the count cannot diverge because of this sweep. Also verified: the `pushTargetSelection` extraction is behaviour-identical across all three cases (query+network resolvable, query+network unresolvable, channel), and non-idempotent `installNotificationDismiss` is fine — the sibling `installPushTargetListener` is equally non-idempotent, `main.tsx` calls each once, and the sweep is idempotent anyway.

### One mild push-back

Deferring the **fix** for the already-in-view case is right; the code ships fine and regresses nothing. But the DESIGN_NOTES entry in this PR does not merely omit the case — it asserts it is *"handled one layer up"* by `shouldSuppressPush()` and calls it *"a state that should not occur"*. `service-worker.ts` says the opposite in this same tree, dated and attributed: that visibility is unreliable on iOS PWAs, and the server default is deliver-leaning in the just-connected window on every platform. Whoever hits this bug later will read that paragraph as a considered decision. **Reword the paragraph in this PR; defer the mechanism.**

### Housekeeping

The PR body ends with a closing keyword next to the issue number, which will auto-close the issue on merge. In this repo enqueue and closure are vjt's, and that line takes it out of his hands — worth dropping the keyword if the convention applies here.

### What we refuse to affirm

- **That the race has ever fired in the field.** It is a structural read — two awaits between the gate and the act — not an observation. Not instrumented, not run.
- **That the already-in-view gap actually bites on a real iOS device.** We established that the *tree* documents both suppression layers as leaky in that direction; we did **not** establish that a banner is being shown to a foregrounded iOS PWA today. No device, no measurement.
- Specifically: **we do not know whether an iOS banner over a foregrounded PWA produces a blur/focus pair.** If it does, `isDocumentVisible()` flips and the reactive arm self-heals the whole thing on that platform. The tree carries no note either way. This is the single fact that would most change that finding's severity — in either direction.
- **That the seven tests pass.** vitest was not run. The coverage finding is an argument about what the suite *can* express, derived from the mock's shape — the mutant was named, not executed.
- Anything about gates, or whether the vitest failures the PR body reports are the known flake family. Not run.
- Any judgement on the DESIGN_NOTES marker uniqueness against the base tip — that is the merging party's check, not the reviewer's.
- That the badge acquittal covers every count surface: the OS icon badge and the `document.title` mirror were verified; the in-app sidebar unread counters were not audited against this change.

gmsrc:
CI: 8 of 9 checks green (cic types+lint+unit, all four e2e shards, integration, dialyzer, shottino).

`test + lint + audit` failed on ONE ExUnit test, and it is not this branch's:

```
1) test the per-channel join door the LIVE storm is linear in W (GrappaWeb.JoinSeedCostTest)
   test/grappa_web/channels/join_seed_cost_test.exs:167
   left:  %{"messages" => 2, ..., "users" => 1, "visitors" => 0}
   right: %{"messages" => 2, ..., "users" => 1}
```

An extra `"visitors" => 0` entry in the per-table query-count map — a zero-count key appearing where the assertion expects the key absent.

The branch changes 5 files and **none of them is Elixir** (`cicchetto/src/**` + `docs/DESIGN_NOTES.md`; `git diff --name-only e7677913..HEAD | grep -E '\.exs?$'` is empty), so it cannot move an Ecto query count. `main` is green on the same test, including the run at 14:32 today, and nothing has touched `join_seed_cost_test.exs` or `lib/grappa/visitors.ex` between this branch's base and `origin/main`.

Re-running the failed job. Flagging it rather than burying it: a zero-count key that appears intermittently looks like a real latent flake in that test's counter, worth an issue of its own if it recurs.

gmsrc:
Both reviews acted on. Four commits, and I ran the two named mutants rather than reasoning about them.

## Fixed

**1. The docstring credited the wrong function** (`023eee0d`) — first review's finding, and its framing is the one I used. `pushTargetSelection` said channels need no step *because `setSelectedChannel` folds the KEY*; the sweep never calls the setter. Corrected to `isActiveSelection`'s own `foldChannelKey` (`selection.ts:221`), with the trap spelled out: drop the fold from the comparator believing the setter covers it, and the dismissal breaks while every forward-path test stays green. Same correction applied to the DESIGN_NOTES entry, which repeated it.

**2. The race** (`d11baff9`) — confirmed structurally: the gate is line 59, the awaits are 62 and 65, the close is 69. `isDocumentVisible()` is now re-read immediately before the close loop. Your reading of what it costs is exactly right and it is why I took it as a fix rather than a note: it is the only failure mode here that destroys information instead of withholding it.

**3. The test gap** (`094048f9`) — the sharpest finding on either side, and correct in both halves: missing *and* inexpressible. Both store stubs are real Solid signals now. Measured, not argued:

| mutant | result |
|---|---|
| `untrack(selectedChannel)` in the effect | **only** `sweeps when the reader switches to the conversation while visible` goes red |
| drop the post-await visibility re-check | **only** `closes nothing when the document goes away between the gate and the act` goes red |

Each new test kills exactly its own mutant; 10/10 green unmutated. I also confirmed your prediction directly — under the `untrack` mutant the seven *old* tests all passed.

**4. The DESIGN_NOTES paragraph** (`e68505e1`) — your push-back is right and the wording was worse than an omission. It claimed the already-in-view case was "handled one layer up" and "a state that should not occur", which this tree contradicts in writing. Retracted, and the entry now records the residual honestly, including the fact that would move its severity in either direction: **nobody knows whether an iOS banner over a foregrounded PWA produces a blur/focus pair.** If it does, the reactive arm self-heals the case on that platform. I have no device and did not measure it.

**5. Housekeeping** — closing keyword dropped from the PR body.

## Recorded, not fixed

**CHANTYPES** — added as a sentence to the entry, framed as you did: pre-existing tap-path behaviour that this PR makes load-bearing in a second direction. `lib/chantypes.ts` exists and `parsePushTargetUrl` ignores it; on an exotic chantype a channel notification is classified as a query and never dismissed.

**The already-in-view mechanism** — deferred, as both reviews allow. Not filed as an issue: enqueue is yours. Say the word and I will file it.

## Where I differ: nowhere, but one correction to the record

The second review's "we do not know whether the seven tests pass — vitest was not run" is now moot, but worth closing out: they did, and the ten do. Full suite locally is 6946 passed with 7 failures in the known Windows-host timeout family (`mentions`, `selection`, `subscribe`, `userTopic`), 362/362 green re-run in isolation. CI is the authority and was 9/9 green before this push.

Re-running CI on the new head.

PR #2070, opened 2026-09-11
fix(2069): answer "how many unread" from one function, at two anchors

vjt:
Three surfaces answered "how many rows are behind my cursor in this window"
and gave three different numbers, one of which moved while nobody read
anything. This makes them one function.

Addresses issue 2069.

## What was measured, before anything was changed

On `b7989f4ba`, against the real `scrollback` + `readCursor` + `selection`
stores and a fake server that honours `after`/`before`/`limit` and answers
its probe out of the same log it serves. The log is **mixed** — every fourth
row a peer JOIN — because a uniform log cannot see any of the three defects.

**All three symptoms reproduce.** Two of them do not reproduce the way the
issue described them, and that changed the cure.

### C — the divergence, and it is triple, not single

| rows past the cursor | sidebar pill | in-pane divider |
| --- | --- | --- |
| 113 peer messages + 37 peer JOINs | 113 | `150 unread messages` |
| 5 peer + 5 own messages | 5 | `10 unread messages` |
| 5 numeric-derived NOTICEs | 5 | no marker at all |
| 50 peer JOINs | 0 | `50 unread messages` |

Two of those rows say the pill was right; the third says the DIVIDER was.

The common cause is one stale sentence, in the source. `operatorActionEcho.ts`
and `ownPresenceEvent.ts` each claim to be the "single source of truth" shared
by "subscribe.ts (the sidebar badge gate)" and the in-pane marker. That was
true while the badge was BUMPED per row. `subscribe.ts` still calls both
predicates, but since the 2026-06-01 bucket-B2 change the badge is DERIVED
from `(scrollbackByChannel, readCursors, serverSeedCounts)`, and the
derivation never saw them. The gate kept guarding the mention beep and stopped
guarding the thing its own comment names.

### A — reproduced, but NOT by the mechanism the issue named

A bare re-select with no traffic moves nothing — measured across two
re-selects on a far-behind window, on a not-far-behind one (gap 150) and on a
just-far-behind one (gap 250): stable in all three.

What moves is the count while the operator is AWAY. `appendPageToScrollback`
accumulated it by the content-ness of the row the ring cap EVICTED rather than
of the row that ARRIVED, and added nothing at all below `UNREAD_RETENTION_CAP`.
Measured: 500 arrivals took it to 4012 against a server answer of 4125, error
neither constant nor signed (-75, -113, -113). The re-select's page then came
back FULL, fired the gap probe, and OVERWROTE the drifted number with a fresh
measurement: +113 in one step, no scroll, no read.

So the discriminator the issue asked for — "does a re-select re-probe
`/messages/count`?" — is YES, conditionally, and **the re-probe is the
CORRECTION, not the cause**. Curing the visible step would have left the drift,
which is the number the operator actually reads for as long as they stay away.

### B — reproduced, and worse than reported

It does not stop at the page size. The jump retires the far-behind record, the
pill falls back to counting held rows, and as the operator reads through the
one page the jump loaded the count reaches **ZERO** on a window with thousands
unread: 3750 → 150 → 0, cursor at 1200, server answer 3600. The measurement
that answers all of those was already on file — `measuredUnreadByChannel`
(#947), written by that same jump — and only the divider ever spent it.

## The shape

New `cicchetto/src/lib/unreadCount.ts`: one row predicate
(`isOperatorOwnedRow` / `countsAsUnreadMessage` / `countsAsUnreadEvent`) and
one region count (`unreadMessagesAfter`). `selection.ts` and
`ScrollbackPane.tsx` both call them; neither re-derives.

**One function, TWO anchors** — and this is where the issue's "one variable"
is wrong taken literally. The divider passes the FROZEN cursor and the frozen
session top (the freeze contract: the line must not renumber under a reader).
The pill passes the LIVE cursor and no bound, because a pill that ignored
arrivals while the cursor sat still would stop saying the window is receiving
traffic, which is the pill's whole job. Same population, same unit, same
source-selection; different anchors, on purpose.

**Placement stays BROADER than the count.** The line marks where the operator
left off, so it goes above the first row somebody ELSE produced — message or
presence — while the label counts only messages. Narrowing placement to
content would leave a run of peer JOINs above the line, rendered as if read.

- **A** is cured at the accumulator: count ARRIVALS (fresh rows above the
  pane's previous newest), not evictions. Exact, cap-independent, and it
  leaves the probe nothing to correct. `contentDropped` and `keptContentCount`
  are deleted rather than kept beside it.
- **B** is cured by letting the pill spend the same #947 record, minus what
  the cursor consumed, so the number falls by what was READ instead of by what
  was FETCHED. The record gains `through` (the top of the contiguous run the
  pane can account for, written by the jump and extended by `loadNewer`);
  past that run it stands down rather than answering with a number it cannot
  support — an own send lands the cursor at the tip with nothing unread, and a
  record that kept spending would report thousands. It is a FLOOR
  (`max(local, count - consumed)`), so local truth overtakes it as the region
  comes back and nobody has to remember to retire it.

## Coverage: all three, all reproduced

A, B and C are each reproduced on the measurement bench and each cured, with
the limits named below. Two tests ASSERTED the defect and were corrected, not
deleted: the pane's `"DOES count peer JOIN row toward the unread marker"` and
the issue-2050 arm that called the drift "demonstrably frozen".

## The two reds the full integration run turned, and which is whose

The full `scripts/integration.sh` was run TWICE end to end (the scoped `--grep`
is not the ship gate). Neither run is green, nothing is buried, and every red
was attributed by measurement rather than by argument.

| run | result | reds |
| --- | --- | --- |
| #1, before the spec fix | 2 failed / 973 passed / 1 skipped | `issue1964`, `unread-cursor-cluster:184` |
| #2, after the spec fix | 2 failed / 973 passed / 1 skipped | `issue1964`, `cp15-b6-part-archive-rejoin` |

| spec | this branch | base `b7989f4ba` | whose |
| --- | --- | --- | --- |
| `issue1964-upload-confirm-preview:97` | 5 red in 5 | **10 red in 10** | **main's**, not this slice's |
| `unread-cursor-cluster:184` | 1 red in 5 → **20 green in 20** after the fix | 10 green in 10 | **was this slice's**, cured |
| `cp15-b6-part-archive-rejoin:55` | green in run #1, red in run #2, **5 green in 5** isolated | — | **flake**, not this slice's |

### `cp15-b6` — flake, and the two runs are what prove it

It fails on `.members-pane li` never showing the operator's own nick after the
re-join — the member roster, which this slice does not touch. The decisive point
is that **the production code is byte-identical between run #1 and run #2**: the
only commit between them changes two test files and `DESIGN_NOTES.md`. Same
production tree, green once and red once, plus 5 green in 5 in isolation. That
is non-determinism, not a regression. I did not chase it further and I did not
touch it — it belongs to whoever owns that spec, not to this branch.

### `issue1964` — pre-existing red on main

`each staged file previews as what it actually is` fails on the video arm
(`video.error.code` is 4, `MEDIA_ERR_SRC_NOT_SUPPORTED`, where 0 is expected).
Deterministic on the base commit with nothing of mine in the tree, and nothing
in this slice touches uploads, media policy or the preview path. Reported as a
finding about main, not fixed here — it is a different subject and folding it
into this branch would hide it.

### `unread-cursor-cluster:184` — mine, and it is an ORACLE, not the cure

Stated up front because it would be easy to bury in the diff: the slice caused
this red, and the behaviour the spec describes is not broken.

`sessionTopId` latches the tail of the FIRST non-empty observation the pane
makes; every row above that latch is a live arrival that draws no marker on
purpose (the operator watched it land). The spec sends the peer's PRIVMSG and
navigates immediately, so which side of the latch that row falls on was never
established by its setup. The awaited `peer.join` before it reliably WAS inside
the latch — and while the divider counted presence rows, that JOIN on its own
injected a marker reading `1 unread`. So the spec went green, in that ordering,
on the strength of the exact defect this PR removes. Restrict the count to
messages and the stand-in disappears; the unguarded ordering stops being masked.

So the cure is to make the precondition TRUE (`assertMessagePersisted` before
the navigation), and **the assertion is untouched** — no assert was weakened to
buy green. Measured after the fix: **20 green in 20**, against 1 in 5 before
(`0.8^20 ≈ 1.2%` if the ordering were still free).

The mechanism no longer needs a browser to be seen. Both directions are pinned
in-process: `b7989f4ba` renders `1 unread` for a lone peer JOIN (its own
`DOES count peer JOIN row toward the unread marker`, re-run on the base to
MEASURE it rather than cite it), and this branch renders no marker for the same
fixture plus a late-arriving message (new arm, `draws no marker when only a peer
JOIN precedes the session top and the message lands live`).

## What I refused to assert

- **That the re-select is the bug.** It reproduces the visible step, and the
  issue named it as the mechanism, but the measurement says it is the
  correction of a drift that happens elsewhere. I did not ship a cure for the
  step.
- **That the divider's number was simply wrong.** On the numeric-NOTICE row of
  the table the divider is the one telling the truth. The fix is a shared
  population, not "make the divider agree with the pill".
- **That the e2e proves both surfaces red without the cure.** It does not, and
  the limit is stated below.
- **That `count - consumed` is exact.** It is a floor with a named population
  gap, and it is shipped as a floor for that reason.
- **That the near-dead `clearMeasuredUnread` arm is dead.** I could not
  construct it; I named it and left it.
- **That the full integration suite is green.** It is not: `issue1964` is red,
  on this branch and on the base alike. I am not claiming a ship gate I did not
  clear, and I did not touch that spec to clear it.
- **That the `unread-cursor-cluster` red was a flake.** It was not — it belongs
  to this branch, measured. I am also not claiming the cure was wrong, and the
  two refs are what separate those readings.

## What I did NOT reproduce

- The issue's stated mechanism for A: a bare re-select re-probing and
  inflating the count on its own. Measured stable on three configurations.
- The issue's stated bound for B: "falls back to the page size". It does not
  stop there — it reaches zero.

## Limits, stated rather than discovered later

- **Server and client populations differ by numeric-derived NOTICEs.**
  `Scrollback.count_after_split/6` excludes own-authored rows but knows
  nothing about `meta.numeric`, a client concept, so `count - consumed` can
  run high by the number of such rows read inside the measured region.
  Bounded, rare (they land in `$server` and ghost query windows, not busy
  channels), and deliberately NOT cured with a second predicate — a second
  predicate is exactly the fork this slice removes.
- **The arrivals accumulator counts own-authored content**, exactly as the
  eviction one did. No `ownNick` available without an import cycle into
  `networks`, and the far-behind cursor is frozen. No regression, no cure.
- **A cross-device read-ahead still under-reports.** The peer device moves the
  cursor past rows this pane never held; the record stands down and the pill
  answers from local rows. That is what it did before.
- **The SHORT-page arm of `jumpToUnread` is very nearly unreachable.** A
  far-behind window has >200 unread by definition and `resumeFrom` does not
  move, so `after(resumeFrom, 200)` is always full. Its `clearMeasuredUnread`
  branch is near-dead: named, not deleted — it is the correct answer if the
  region ever shrinks (an archive purge).
- **The e2e red-without-the-cure proof stops at the pill.** The mutant sits in
  the SHARED predicate, so the spec dies on the PILL assert, which comes
  first; the DIVIDER assert in that arm was never reached.

## The bigger answer, not taken here

The server already pushes per-window `messages`/`events` on `window_counts`,
on every message and on every cursor advance, and cic throws them away
(`selection.ts` says so, citing #239 — that the presence filter is a client
concern, which is true of the EVENTS bucket and not of the MESSAGES one).
Adopting that field would collapse `serverSeedCounts`,
`farBehindByChannel.missed` and `measuredUnreadByChannel` into one
authoritative number and delete every accumulator in this PR. It is a redesign
of the client's unread model, not a bug fix, and it is vjt's call, not one to
take inside a slice asked to stop three numbers lying.

## Evidence

- `scripts/bun.sh run check` — green, 5/5 stages.
- `scripts/bun.sh run test` — 7026 passed, 0 failed.
- **Six mutants, one per mechanism, six killed.** Control run (unmutated tree)
  green at 7026 first, tree restored by `git checkout` after each. Deaths:
  operator-echo clause dropped (3, one a PRE-EXISTING pane test), local floor
  dropped (2), `through` guard dropped (3), content split dropped (21),
  arrival count disabled (4, one the pre-existing #1229 arming test),
  measurement ignored by the pill (4). Two mutants killing PRE-EXISTING tests
  is what proves the surfaces are genuinely wired to the shared predicate.
- **Browser witness**, `cicchetto/e2e/tests/issue2069-one-unread-number.spec.ts`
  — the only place the two numbers are visible AT ONCE, which is how vjt
  noticed. It asserts both premises rather than assuming them: that the unread
  run really contains peer presence rows, and that those rows RENDER
  (`#spec-wN` is far under `LARGE_CHANNEL_THRESHOLD`), because a hidden
  presence row is excluded from the old count too, by a different rule, and
  the spec would be green on the broken build.
- **Red without the cure, proven**: with the "no content split" mutant the
  spec fails exhibiting the mutated DOM — `class="sidebar-msg-unread"
  aria-label="5 unread messages"`, 5 instead of 3 (3 messages + 2 presence
  rows). That exact number is also the oracle that the source mutation
  REACHES the bundle.
- Full `scripts/integration.sh`, twice end to end — 973 passed each time, with
  the two residual reds attributed above. **The ship gate is NOT clear**, and
  the residue is not mine: `issue1964` is red on the base with none of my code
  in the tree, and `cp15-b6` is non-deterministic across two runs of identical
  production code. Whether that is shippable is not my call.
- `scripts/design-notes-gate.sh` — `rc=0`, marker + separator present, marker
  verified unique against `origin/main`'s tip (0 occurrences, positive control:
  352 markers exist).

### DIFF
--- a/cicchetto/e2e/tests/issue2069-one-unread-number.spec.ts
+++ b/cicchetto/e2e/tests/issue2069-one-unread-number.spec.ts
@@ -75,12 +75,49 @@ test.describe("issue 2069 — one number for one question", () => {
       peer.privmsg(CHANNEL, "issue 2069 one");
       peer.privmsg(CHANNEL, "issue 2069 two");
       peer.privmsg(CHANNEL, "issue 2069 three");
+
+      // 🔴 This barrier is placed BEFORE the PART on purpose, and it is not
+      // politeness — it is what makes the setup deterministic.
+      //
+      // bahamut charges FAKE-LAG per command on a connection, so a burst is
+      // served with growing delay and the ~10s overshoot the `whoisAway` budget
+      // in `fixtures/ircClient.ts` already documents. `IrcPeer.part` waits for
+      // its own PART echo on a 5s budget (`PART_TIMEOUT_MS`), and this spec
+      // used to fire JOIN + 3×PRIVMSG + PART back to back — so the PART was the
+      // FIFTH command against a bank filled by the four before it.
+      //
+      // Measured on this tree, `--repeat-each 20`: 9 reds in 20, every one of
+      // them the identical `IrcPeer: timeout waiting for part #spec-w0 (5000ms)`
+      // — the same signature CI turned on shard 2/4 of run 34553963487. The
+      // discriminator against "the testnet is just slow" is in the same logs:
+      // `IrcPeer.join` carries the SAME 5s budget and timed out ZERO times in
+      // those 20 runs. First command always fine, fifth command half the time
+      // not: that is a per-command bank, not latency.
+      //
+      // Draining it by waiting on an OBSERVABLE signal — the three messages
+      // being on the server — costs nothing the spec was not already paying
+      // (it polls this very endpoint below) and leaves the PART as a lone
+      // command against a bank that has had time to empty. No timeout is
+      // raised and no assertion is touched.
+      let after = before;
+      const peerNick = PEER_NICK.toLowerCase();
+      await expect
+        .poll(
+          async () => {
+            after = await fetchAllMessagesAsc(vjt.token, NETWORK_SLUG, CHANNEL);
+            return after.filter(
+              (r) => r.id > cursorId && r.kind === "privmsg" && r.sender.toLowerCase() === peerNick,
+            ).length;
+          },
+          { timeout: 15_000 },
+        )
+        .toBe(3);
+
       await peer.part(CHANNEL, "issue 2069 done");
 
       // Barrier on the SERVER's own store, not a timeout: the PART is the last
       // row the peer produces, so its arrival proves the three messages and the
       // JOIN are already persisted (one ordered IRC stream, one session).
-      let after = before;
       await expect
         .poll(
           async () => {
--- a/cicchetto/e2e/tests/issue2069-one-unread-number.spec.ts
+++ b/cicchetto/e2e/tests/issue2069-one-unread-number.spec.ts
@@ -79,26 +79,32 @@ test.describe("issue 2069 — one number for one question", () => {
       // 🔴 This barrier is placed BEFORE the PART on purpose, and it is not
       // politeness — it is what makes the setup deterministic.
       //
-      // bahamut charges FAKE-LAG per command on a connection, so a burst is
-      // served with growing delay and the ~10s overshoot the `whoisAway` budget
-      // in `fixtures/ircClient.ts` already documents. `IrcPeer.part` waits for
-      // its own PART echo on a 5s budget (`PART_TIMEOUT_MS`), and this spec
-      // used to fire JOIN + 3×PRIVMSG + PART back to back — so the PART was the
-      // FIFTH command against a bank filled by the four before it.
+      // MEASURED, on this tree, `--repeat-each 20`:
       //
-      // Measured on this tree, `--repeat-each 20`: 9 reds in 20, every one of
-      // them the identical `IrcPeer: timeout waiting for part #spec-w0 (5000ms)`
-      // — the same signature CI turned on shard 2/4 of run 34553963487. The
-      // discriminator against "the testnet is just slow" is in the same logs:
-      // `IrcPeer.join` carries the SAME 5s budget and timed out ZERO times in
-      // those 20 runs. First command always fine, fifth command half the time
-      // not: that is a per-command bank, not latency.
+      //   * without this barrier, 9 reds in 20, every one of them the identical
+      //     `IrcPeer: timeout waiting for part #spec-w0 (5000ms)` — the same
+      //     signature CI turned on shard 2/4 of run 34553963487;
+      //   * the cost depends on the command's POSITION in the burst, not on the
+      //     clock: `IrcPeer.join` carries the SAME 5s budget (`JOIN_TIMEOUT_MS`
+      //     == `PART_TIMEOUT_MS`) and timed out ZERO times across those 20 runs.
+      //     First command always fine; fifth command — after JOIN + 3×PRIVMSG —
+      //     half the time not;
+      //   * with the barrier, 20 green in 20.
       //
-      // Draining it by waiting on an OBSERVABLE signal — the three messages
-      // being on the server — costs nothing the spec was not already paying
-      // (it polls this very endpoint below) and leaves the PART as a lone
-      // command against a bank that has had time to empty. No timeout is
-      // raised and no assertion is touched.
+      // INFERRED, and labelled as such: that the per-command cost IS bahamut's
+      // fake-lag bank. That name is READ, from the `whoisAway` comment in
+      // `fixtures/ircClient.ts` and from the ircd source — and reading a
+      // mechanism tells you a path EXISTS, never what it costs. This spec has
+      // not measured the bank: the instrument that can (`Grappa.IRC.FakeLag`,
+      // #800/S7) accounts GRAPPA's own socket, and the peer here is a separate
+      // irc-framework connection it never sees. Anything that serialises per
+      // connection would fit the same three measurements.
+      //
+      // The cure does not depend on the name. It rests on the measured half —
+      // position in the burst — so draining on an OBSERVABLE signal (the three
+      // messages being on the server, the very endpoint polled below) leaves
+      // the PART as a lone command. No timeout is raised and no assertion is
+      // touched.
       let after = before;
       const peerNick = PEER_NICK.toLowerCase();
       await expect
--- a/cicchetto/e2e/tests/issue2069-one-unread-number.spec.ts
+++ b/cicchetto/e2e/tests/issue2069-one-unread-number.spec.ts
@@ -1,0 +1,142 @@
+// issue 2069 — the sidebar pill and the in-pane divider must answer the same
+// question with the same number.
+//
+// The reported symptom C: "the divider counts presence rows (join/part/quit/
+// nick) as unread 'messages'; the sidebar badge does not." Measured at the
+// store level on `b7989f4ba` over 150 rows past the cursor (113 peer messages
+// + 37 peer JOINs) the pill read 113 and the divider rendered "150 unread
+// messages" — a label that says "messages" over a count that is not messages.
+//
+// This is the browser witness for that, and it is the only place the two
+// numbers are ever visible AT ONCE, which is how the operator noticed. The
+// vitest arms pin each surface against a fixture; only a rendered page can
+// pin that the two AGREE.
+//
+// RED pre-fix: the peer's JOIN and PART land in the unread run, so the divider
+// reads "5 unread messages" against a pill reading "3".
+//
+// Two premises are asserted rather than assumed, because without either one
+// the spec is green on the broken build:
+//
+//   * the unread run really does contain peer PRESENCE rows (otherwise the two
+//     predicates cannot disagree);
+//   * those rows are RENDERED (`#spec-wN` is far under
+//     `LARGE_CHANNEL_THRESHOLD`, so the #222 filter shows them) — a hidden
+//     presence row is excluded from the old count too, by a different rule.
+//
+// Per BUGHUNT-3 cascade rule the cursor is moved here, so it is restored to
+// tail in afterEach or downstream specs inherit a mid-list cursor.
+
+import {
+  loginAs,
+  selectChannel,
+  sidebarEventsBadge,
+  sidebarMessageBadge,
+} from "../fixtures/cicchettoPage";
+import {
+  fetchAllMessagesAsc,
+  restoreReadCursorToTail,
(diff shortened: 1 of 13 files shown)


vjt:
## Shard 2/4 red — reproduced, mechanism measured, cured at the cause

The one red on this PR was **my own spec**, `issue2069-one-unread-number.spec.ts`
(run `34553963487`, job `103122526486`, head `5da08289`), failing in the fixture
with `IrcPeer: timeout waiting for part #spec-w0 (5000ms)`. The frame sitting in
`fixtures/ircClient.ts` rather than in an assertion is suggestive and proves
nothing — a shared fixture can be the victim as easily as the culprit — so it
was measured rather than argued.

### It reproduces

`--repeat-each 20` on the unfixed tree: **9 reds in 20**, every one carrying that
identical signature. Nothing about this rests on reading the code.

### The mechanism

bahamut charges **fake-lag per COMMAND**. The spec fired JOIN + 3×PRIVMSG + PART
back to back on one connection, so the PART was the **fifth** command against a
bank the four before it had filled. `IrcPeer.part` waits for its own echo on
`PART_TIMEOUT_MS` = 5s, while the overshoot documented in that same file's
`whoisAway` comment is ~10s — the budget sits under the known worst case, and
only bites callers that arrive late in a burst.

**The discriminator against "the testnet is just slow" is in the same 20 runs:**
`IrcPeer.join` carries the **same 5s budget** and timed out **zero** times.
First command always fine, fifth command half the time not. That is a
per-command bank, not latency.

### The cure, and what it is not

Drain the bank on an **observable signal**: poll until the three messages are on
the server — the very endpoint the spec was already polling immediately
afterwards — and only then send the PART, so it goes out alone.

**No timeout was raised. No assertion was touched.** The setup was made
deterministic; the contract is identical.

Measured after: **20 green in 20** (`0.55^20 ≈ 6e-6` had the rate stayed where
it was).

### What I deliberately did NOT do

- **Did not touch the six sibling `IrcPeer.part` callers.** The closest,
  `issue2037`, fires four commands where mine fired five, so it is exposed to
  the same mechanism — but none has been seen red, and widening a cure past its
  measurement is how a fixture collects ballast nobody can justify later. The
  mechanism is written into `DESIGN_NOTES` so the next sighting is diagnosed in
  one reading instead of re-derived.
- **Did not raise `PART_TIMEOUT_MS`.** It would have hidden the burst rather
  than removed it, and every future caller would inherit the wrong lesson.
- **Did not re-run the job.** A rerun reuses the original merge-ref and measures
  nothing.
- **Did not rebase.** The branch is behind only by docs-only commits; the merge
  is the orchestrator's.

vjt:
## Correction: the fake-lag is INFERRED, not measured — and the cure does not rest on it

My previous comment on this PR named bahamut's fake-lag bank as the cause of the
PART timeout. That overstated what I established. Splitting it properly, because
the distinction is the whole point.

### Measured

* **It reproduces.** `--repeat-each 20` on the unfixed tree: **9 reds in 20**,
  every one the identical `IrcPeer: timeout waiting for part #spec-w0 (5000ms)`
  — the same signature as CI shard 2/4.
* **The cost tracks the command's POSITION in the burst, not the clock.**
  `IrcPeer.join` carries the **same** 5s budget (`JOIN_TIMEOUT_MS` ==
  `PART_TIMEOUT_MS`) and timed out **zero** times across those 20 runs. First
  command always fine; fifth command — after JOIN + 3×PRIVMSG — half the time
  not. This is what rules out "the testnet was just slow".
* **The barrier removes it.** 20 green in 20 (`0.55^20 ≈ 6e-6` at the old rate).

### Inferred

* **That this per-command cost IS the fake-lag bank.** I got that name by
  READING — the `whoisAway` comment in `fixtures/ircClient.ts`, and the ircd
  source behind it. Reading a mechanism establishes that a path EXISTS; it never
  establishes what it costs.
* **I did not measure the bank, and could not with what exists.** The instrument
  built for exactly this (`Grappa.IRC.FakeLag`, #800/S7) accounts **grappa's
  own** socket. The peer in an e2e is a separate `irc-framework` connection that
  instrument never observes, so it cannot answer for this connection without new
  instrumentation.
* **What my three measurements do not distinguish:** the fake-lag bank from any
  other per-connection serialisation or per-command server cost. All of them fit
  the same data.

### Why the cure stands anyway

It is built on the measured half — position in the burst — and not on the name.
Draining on an observable signal before the closing PART is correct if the cost
is fake-lag, and equally correct if it is anything else that serialises per
connection. If the name turns out wrong, the fix does not move.

This is not a rhetorical hedge: #800/S7 was built after two retractions in one
cycle that came from treating a reading of this exact mechanism as a
measurement. The spec comment and the `DESIGN_NOTES` entry now carry the split
explicitly (`335bdd71c`), so the next reader inherits the labels rather than the
claim.

Issue #2045, opened 2026-09-10
far.missed has two producers and they still disagree on own-authored content


vjt:
Found while writing the #2037 e2e, and it is a residue of #2037 A rather than
something A caused: A brought the two producers of `far.missed` from
"different units" to "the same unit except for one term", and this is that
term.

## The two producers

`far.missed` — the number the far-behind bar renders, and since #2037 A the
same number the sidebar's bold pill reads — is written by two different code
paths:

| producer | where | what it counts |
|---|---|---|
| the PROBE | `Scrollback.count_after_split/6`, server | content kinds, **own-authored excluded** |
| the PRUNE | `capScrollbackRing`, `cicchetto/src/lib/scrollback.ts`, client | content kinds, **own-authored INCLUDED** |

The client side is one line:

```ts
const contentCount = unreadRows.filter((m) => isContentKind(m.kind)).length;
```

`isContentKind` is a KIND test and there is no own-nick arm anywhere in
`capScrollbackRing`. The server side calls `exclude_own_authored/3`.

## Why the server's answer is the right one

This is not a #2037 preference, it is the project's existing definition. A
line the operator typed is read BY DEFINITION (#576 content), and a self-PART
or a KICK they issued is an action they performed rather than something to
catch up on (#532 A presence). `count_after_split/6` says so in its own
comment. So the prune path is counting as unread a class of row the rest of
the codebase has decided is not.

## Why it matters now and did not before

Before A the prune path accumulated RAW row counts and the probe returned a
split — two obviously different quantities, and the bar showed whichever
producer armed the window. A narrowed the prune path to the content unit, so
the two now agree on everything except own-authored content. That is worse in
one specific way: two numbers that differ by a lot are visibly two numbers,
and two numbers that differ by three are indistinguishable from one number
until somebody counts.

Concretely: the same window, at the same cursor, reports a different
far-behind figure depending on whether it went far behind by LOCAL eviction
(prune) or by a gap probe on open. The operator has no way to tell which
happened.

## Size of the effect

Bounded by the operator's own content sitting past their own read cursor,
which is the multi-device case: lines sent from the phone are past the
laptop's cursor. Small in the common case, unbounded in principle, and
exactly the term #576 went to the trouble of excluding server-side.

## What a fix looks like

`capScrollbackRing` already reads a signal (`getReadCursor`) and already
decodes the slug out of the key, so it is not a pure function being made
impure. The own nick is per-network and `selection.ts` resolves it as
[...]

Issue #2041, opened 2026-09-10
cic: neither ComposeBox nor TopicBar checks isComposing, so an IME commit key can send


vjt:
Filed from a worker's declared gap while building issue 2035 (PR #2040). **Not measured against a real IME** — reported as a read of the code, and the worker deliberately refused to close it on one surface only.

## The gap

Two surfaces now treat `Enter` as a send/commit key:

- `ComposeBox` — sends the message (ruling 974, 2026-08-07).
- `TopicBar` — sets the topic (ruling of 2026-09-10, issue 2035).

**Neither consults `KeyboardEvent.isComposing`** (nor `keyCode === 229`). With an IME active — Japanese, Chinese, Korean, and any compose-based input — the `Enter` that *commits the candidate* is delivered to the page as a normal `Enter`. So the key that is supposed to finish choosing a word instead sends the half-typed line, or sets the topic.

Same for `Shift+Enter`, which on both surfaces also sends.

## Why it is filed on BOTH surfaces and not fixed on one

The whole argument of PR #2040 is that the same chord must not grow a second semantics on a second surface. Closing this on `TopicBar` alone would create exactly the divergence that PR spent its body arguing against. So the cure — if it is wanted — belongs to both call sites at once, presumably as one shared guard.

## Not claimed

- **No measurement with a real IME.** Nobody has typed Japanese into either box and watched what arrives; this is read off the handlers, which do not mention `isComposing` anywhere.
- No claim about which browsers/IMEs actually deliver the composing `Enter` here, nor that the `keyCode === 229` legacy fallback is still needed on the engines cic supports.
- No claim that this has ever bitten a real user of this instance.

## Where it comes from

Declared by the worker in its own "what I refuse to claim" section on issue 2035 rather than quietly patched, and routed here by the orchestrator. Filed to the backlog with no status label — triage is vjt's.

vjt:
Merged to main as `b89137773` (PR 2075, rebase merge). One shared predicate — `cicchetto/src/lib/imeComposition.ts` — consulted by three call sites, not two: `keybindings.ts` already checked `isComposing` inline and was routed through the same predicate with no behaviour change, so the surface that had the guard cannot drift from the two that just got it. The guard sits above the chord table rather than inside the Enter branch, which also covers the arrow keys (candidate-list navigation vs history navigation) — a widening beyond what this issue claims, stated in the PR body rather than left to be found.

[...]

Issue #2037, opened 2026-09-10
cic(unread): the far-behind bar and the sidebar badges count different rows — 1807 vs 187+216, and neither label says which


vjt:
## Report

From a user on the public channel, after coming back to a session left ~21h earlier:

> non mi sono chiarissimi i vari indicatori dei messaggi non letti di grappa; sono entrato
> e un link mi avvertiva che aveva 1807 messaggi non letti dalle 14 di ieri (che ci può
> stare), ma sulla colonna sx c'è una label a pillola con scritto 187 e una label con
> scritto 216... cosa sono questi due numeri? come vanno interpretati?

Three numbers on screen for what the user reads as one quantity — "how much did I miss" —
and none of them agrees with the others. 187 + 216 = 403, not 1807.

## What each number actually is

Measured at `21f43d064`, no assumptions:

| surface | value | route | server function |
|---|---|---|---|
| far-behind bar in the pane | `1807 unread — jump back` | `GET …/messages/count?after=` | `Scrollback.count_after/6` |
| sidebar pill (bold) | `187` | `/me` `unread_counts` | `ReadCursor.bulk_unread_split/3` |
| sidebar pill (faint) | `216` | `/me` `unread_counts` | same |

- `ScrollbackPane.tsx:4100` renders `far().missed`, which is `probeGap` →
  `countMessagesAfter` (`lib/api.ts:2450`) → `MessagesController.count/2`
  (`messages_controller.ex:211`) → `Scrollback.count_after/6` (`scrollback.ex:701`).
- `WindowBadges.tsx` renders `messagesUnread()` / `eventsUnread()`, which for a far-behind
  window are served **wholesale from the server seed** (`selection.ts` `perChannelUnread`
  skips local counting for far-behind keys, by design, #693). The seed is `/me`'s
  `unread_counts`, built by `build_unread_counts/2` → `WindowCounts.bulk_snapshot/4` →
  `ReadCursor.bulk_unread_split/3` (`read_cursor.ex:318`).

## Why they cannot match, by construction

The two server-side counts do not share a predicate:

1. **Own-authored rows.** `bulk_unread_split/3` drops them from BOTH buckets — own content
   (#576) *and* own presence: self-part, kick-issued, terminal self-rename (#532 A), via
   the `not (^own_authored)` join clause at `read_cursor.ex:340`. `count_after/6` has **no
   such exclusion at all** — no `exclude_own_authored/3` call in its pipeline
[...]

PR #2076, opened 2026-09-11
cic: count far-behind the way the server counts it — one definition, four call sites

vjt:
Addresses issue 2045.

## The defect

`far.missed` — the far-behind bar's number and, since #2037 A, the sidebar
pill's — has two producers that disagree on one term.

| producer | own-authored |
|---|---|
| `Scrollback.count_after_split/6` (server) | **excluded** (`exclude_own_authored/3`) |
| cic's far-behind maintenance | **included** |

The server is right, and not as a #2037 preference — it is the definition the
project already holds. Own content is read BY DEFINITION (#576); a self-PART or
a KICK you issued is an action you performed (#532 A).

Before #2037 A the client accumulated RAW counts against a split from the
probe: two obviously different quantities. A narrowed the client to the content
unit, so now they agree on everything but this term — **and that is worse**.
Two numbers that differ by a lot are visibly two numbers; two that differ by
three are indistinguishable from one number until somebody counts.

## 🔴 Four sites, where the issue names one

The issue names `capScrollbackRing`'s `contentCount` — the SEED, what opens the
record when LOCAL EVICTION is the door in. But past the retention bound the
record **grows by what arrived**, and `appendPageToScrollback`'s
`arrivedContent` carried the same bare `isContentKind`. Both write
`far.missed`.

Fixing only the named line would have been **worse than leaving it**: correct
at arm-time, drifting on the next own message — intermittent wrongness
replacing consistent wrongness. It would also have missed the issue's own
motivating case. The multi-device scenario (lines sent from the phone sitting
past the laptop's cursor) reaches the laptop as a **live append** — the
accumulate producer, not the seed.

The other two sites are the `far.events` twins. That is a **deliberate widening
past the issue's text, in its own commit so it drops whole** — see below.

## The design decision: one shared context

The predicates already existed. `unreadCount.ts` publishes
`countsAsUnreadMessage` / `countsAsUnreadEvent` and its moduledoc already says
it mirrors `count_after_split/6`; the far-behind path was the one place that
never asked.

What is new is `lib/unreadRowContext.ts`. Those predicates take *"who is the
operator, in this window?"* as a parameter so they stay pure — and until now
exactly one caller built it, inline, mid-loop in `selection.ts`. **A second
hand-built copy would have been this very defect one level down**: two
spellings of one identity feeding two counts meant to be one number. So the
builder was extracted first, `selection.ts` routed through it, and the new
caller added to the shared one.

Every field is load-bearing: the nick is PER-NETWORK (`net.nick`, never the
account name — the `ownNickForNetwork` warning, and the 2026-05-08 cic H3
DM-misrouting root cause), the casemapping is per-network (#537 axis 2), and
`isSelfWindow` is the #396 carve-out.

## The arming did not move

Per #2037 A: what puts a window far behind is still *"a row at/after the cursor
left the store"*, as true of a row the operator typed as of anyone else's. Only
the displayed quantity narrows. `unreadDropped` / `unreadHeld` / the bound are
untouched.

## The cycle was MEASURED, not read

The issue said `scrollback.ts → networks.ts` is cycle-free and was explicit
that this was *"a static read of the imports, not a build"*. Verified with a
build **before writing any of the cure**:

- `tsc --noEmit && vite build` → **rc=0** with the edge present;
- `grep -ci circular` on the build log → **0**, against a positive control
  returning **1**;
- the eleven co-initialising suites → **503 tests green**, no TDZ
  `ReferenceError`.

An ESM cycle bites at module-init, not at compile, so the runtime leg is the
one that mattered.

## The events widening, and why it is separable

`exclude_own_authored/3` is applied to the query **BEFORE** the content/event
`group_by`, so it narrows **both** buckets: in a peer/channel window own
content and own presence alike, in the self window (#396) own content survives
and own presence is still stripped. So `far.events` had the identical
divergence one line below.

It also could not be left half-done: with only the content half cured,
`capScrollbackRing` would read `countsAsUnreadMessage(m, ctx)` on one line and
a hand-rolled `!isContentKind(m.kind)` on the next, the published sibling
unused beside it — a half-migration *created by the fix*. Commit
`0ab931c07` is that half alone and drops cleanly.

## 🔴 What I refuse to claim

- **No production measurement.** How big this is in the field is bounded by the
  operator's own content past their own cursor. Nobody read that off a real
  instance; "small in the common case, unbounded in principle" is the issue's
  estimate and it stays an estimate.
- **No e2e, no real browser.** The evidence is unit-level against a fake server
  that *models* `exclude_own_authored/3` — a model of the query, not the query.
- **The fixture lied once, and the controls caught it.** A first version chose
  arrival authorship by SKIPPING ids whose author did not match, so the fake
  server counted rows the client was never handed and the client read **LOWER**
  than the server — the opposite of the defect. Authorship now lives in one map
  read by both sides. Recorded in DESIGN_NOTES because the failure looked like
  a product finding.
- **I did not verify the server half.** `exclude_own_authored/3` is read, not
  exercised; no Elixir test was run (this branch touches no server code, and
  COMPILE is not mine).

## Tests — TDD, with live negative controls

Nine arms in `farBehindOwnAuthored.test.ts`, oracle = **the server's own answer
for the cursor the store currently holds**, never a literal.

Red first: **3 failed / 3 passed** for the content half (the 3 green being the
controls), then **2 failed / 7 passed** when the events arms landed. The
controls that pass on *both* sides are what say the cure moved the client
toward the server rather than muting a bucket:

- peer content still counts, peer presence still counts;
- the SELF window counts own CONTENT and still strips own PRESENCE — the
  sharpest arm, since a cure passing `isSelfWindow` uniformly fails it;
- `OWN_NICK` is deliberately **not** the account name, so a fix matching on
  `displayNick` fails too.

## Gates

Run whole from the worktree root, rc to a file, never through a pipe.

| gate | rc |
|---|---|
| `scripts/bun.sh run check` (5 stages) | **0** |
| `scripts/bun.sh run test` — 351 files, 7039 tests | **0** |
| `scripts/design-notes-gate.sh` — *1 new entry heading, separator and marker present* | **0** |

⚠️ **Base note:** this branch forked at `b58caf1a9`; `origin/main` has since
moved to `b89137773`. DESIGN_NOTES is `merge=union` and both sides append, so a
rebase wants `scripts/union-rebase.sh` rather than a bare `git rebase`. The
markers differ (`#2041` on main, `#2045` here), so the shared-prefix collapse
the convention exists to prevent does not apply.

cic-only — no STACK or COMPILE lane taken.

### DIFF
--- a/cicchetto/src/__tests__/farBehindOwnAuthored.test.ts
+++ b/cicchetto/src/__tests__/farBehindOwnAuthored.test.ts
@@ -106,19 +106,24 @@ const senderOf = (id: number): string => senders.get(id) ?? PEER_NICK;
 /** Seed the backlog: one row in three is the operator's own. */
 const seedBacklog = (tip: number): void => {
   senders.clear();
+  kinds.clear();
   for (let i = 1; i <= tip; i++) senders.set(i, i % 3 === 0 ? OWN_NICK : PEER_NICK);
 };
 
-// All CONTENT: presence is shown in every arm, so the events axis is collapsed
-// and this fixture measures the own-authored term alone.
-const KIND: MessageKind = "privmsg";
+// The kinds law, same shape and same discipline as the senders one: written
+// once, read by the fake server AND by the rows it hands the client. Content
+// unless an arm says otherwise — presence is SHOWN throughout, so a `join` here
+// is a visible event row and not a filtered one.
+const kinds = new Map<number, MessageKind>();
+const kindOf = (id: number): MessageKind => kinds.get(id) ?? "privmsg";
+const isContent = (id: number): boolean => kindOf(id) === "privmsg";
 
 const row = (id: number, channel: string): ScrollbackMessage => ({
   id,
   network: SLUG,
   channel,
   server_time: id,
-  kind: KIND,
+  kind: kindOf(id),
   sender: senderOf(id),
   body: `m${id}`,
   meta: {},
@@ -141,9 +146,19 @@ class FakeServer {
     return this.channel === OWN_NICK;
   }
 
-  private counts(id: number): boolean {
-    if (this.selfWindow) return true; // #396 — own content is payload here
-    return senderOf(id) !== OWN_NICK; // #576 — own content read by definition
+  /**
+   * `exclude_own_authored/3`, in the order the real query applies it: the row
+   * is dropped from the result set BEFORE the content/event grouping, so the
+   * exclusion narrows BOTH buckets rather than just the content one.
+   *
+   *   * peer / channel window — own CONTENT (#576) and own PRESENCE (#532 A)
+   *     are both excluded.
+   *   * self window (#396) — own content is a legitimate note-to-self and
+   *     survives; own PRESENCE is still stripped.
+   */
+  private excluded(id: number): boolean {
+    if (senderOf(id) !== OWN_NICK) return false;
+    return this.selfWindow ? !isContent(id) : true;
   }
 
   listMessages(before?: number): ScrollbackMessage[] {
@@ -164,11 +179,14 @@ class FakeServer {
   countMessagesAfter(after: number): GapProbe {
     let gap = 0;
     let messages = 0;
+    let events = 0;
     for (let i = after + 1; i <= this.tip; i++) {
+      if (this.excluded(i)) continue;
       gap++;
-      if (this.counts(i)) messages++;
+      if (isContent(i)) messages++;
+      else events++;
     }
-    return { gap, messages, events: gap - messages };
+    return { gap, messages, events };
   }
 }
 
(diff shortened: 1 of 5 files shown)


vjt:
**Update — rebased. The "Base note" at the end of the body is now stale; this supersedes it.**

The PR opened **CONFLICTING**, which matters more than it looks: a conflicting
PR runs **no CI at all** and presents as "checks haven't started yet".

Measured, because the two answers disagree:

- `git merge-tree --write-tree origin/main HEAD` locally → **rc=0, zero
  conflicts**. The `merge=union` driver on `docs/DESIGN_NOTES.md`
  (`.gitattributes:39`) resolves the double append cleanly.
- GitHub → `mergeable=CONFLICTING`, `mergeStateStatus=DIRTY`, stable across
  three polls 6s apart.

So **GitHub's mergeability check does not apply the `merge=union` driver**,
even though it is a built-in git driver and the local merge is clean. Any
DESIGN_NOTES-touching branch that falls behind `main` therefore goes
CONFLICTING on GitHub and silently runs nothing — recording it here rather
than filing, since the enqueue is not mine.

Resolved with `scripts/union-rebase.sh` (the #1432 ritual, not a bare
`git rebase`), onto `b89137773`:

- contribution numstat, pinned before and compared after:
  **`+105 -0` → `+105 -0`** — additions unchanged *and* deletions zero, the
  two-sided invariant;
- boundary shape read on the real file for **both** markers — `#2041` (which
  main carries) and `#2045` (mine) — each present exactly once, each in the
  full-stop / marker / blank / `---` / blank / heading shape. The union driver
  collapsed nothing, which is the marker convention doing its job.

Gates re-run **on the rebased tree**, since the old greens were evidence about
a different tree:

| gate | rc |
|---|---|
| `scripts/bun.sh run check` (5 stages) | **0** |
| `scripts/bun.sh run test` — 351 files, **7045** tests | **0** |
| `scripts/design-notes-gate.sh` | **0** |

The test count moves 7039 → 7045 because the rebase brought in issue 2041's own
six.

New head: `7015510b7`.

Issue #2089, opened 2026-09-11
grappa as a DCC client: receive an offered file into the upload store, send one out


vjt:
grappa acts as a **DCC receiver** on the user's behalf: a remote peer offers a file over `DCC SEND`, grappa connects out, drains the socket, spools the bytes into the upload store, and then **synthesises a message — attributed to the sending peer's nick — carrying an emoji and the resulting link** into the conversation where the offer arrived. cicchetto never speaks DCC, never opens a socket, and gains no new transport: it renders an ordinary upload URL and the offer's held/accepted/done state.

**Send from grappa is out** (vjt, 2026-09-12 11:53): the upload store already hands out an HTTPS URL, so re-offering the same file over DCC adds a transport for no gain. Only the receive direction is in scope.

## HOLD LIFTED — vjt, #grappa 2026-09-13 01:30

> "ok vai con il dcc receive come lo avevamo disegnato" / "grappa si connette spoola in /uploads e poi sintetizza un messaggio con il link"

Ready for pickup. Milestone 1.6.

## This reverses a standing ruling — amend in the same PR

- #167 ("DCC support (wishlist)") was closed **not planned** on 2026-08-06 by vjt.
- `README.md:289` lists DCC under **Out of scope** since `3c7a0357`.
- `NON-GOALS.md:22` repeats it as **No DCC**, citing #167.
- `DESIGN_NOTES` entry 1280 reserves the public `/uploads` route for operator-users' own content.

vjt reopened this himself and narrowed it to receive-only, endpoint-not-relay. If accepted, `README.md:289`, `NON-GOALS.md:22-24` and the relevant part of entry 1280 **must be amended in the same PR** — otherwise the next reader re-derives the old "no DCC" answer and is right to.

## Shape (receive-only)

1. A remote peer sends `DCC SEND <name> <addr> <port> <size>` in a CTCP.
2. grappa **connects out** (no listener, no NAT/IPv6-address problem), holds the offer for user consent.
3. On accept, grappa drains the socket to a temp file, truncating at the declared `size` (the sender can lie), and commits it as an ordinary upload.
4. grappa surfaces the result as a **synthesised message attributed to the peer nick** — emoji + the HTTPS link — in the conversation the offer came from.
[...]

PR #2090, opened 2026-09-12
irc: parse a DCC SEND offer — groundwork for issue 2089, no ruling taken

vjt:
Groundwork for issue 2089 (grappa as a DCC client). **Two commits, one new
module, no call site, no product decision taken.**

## 🔴 What this PR deliberately does NOT do

`README.md:289` lists DCC under **Out of scope** and `NON-GOALS.md:22`
repeats it as **No DCC**. Amending those two lines is the ACT that performs
the reversal, and it needs vjt's word — the issue says he reopened the
question on IRC, which is not something this branch can verify. **Neither
file is touched here**, and the DESIGN_NOTES entry says so in as many words
so the log cannot be misread as having decided something nobody decided.

It also does not implement DCC. The issue's own "What actually needs
deciding" lists five open product questions (consent, caps, address
families, malware surface, protocol edges). Building on top of those before
they are answered is building the wrong thing carefully. A design note with
options, measured costs and a recommendation per node has been handed to the
orchestrator separately.

## What shipped

`Grappa.IRC.DCC.parse/1` — the one part of issue 2089 that no product
decision can move, because the wire says what the wire says. A pure function
from the CTCP argument remainder to a typed offer: no socket, no consent, no
storage, no policy.

**It composes rather than repeats.** `Grappa.IRC.CTCP.verb_args/1` stays the
single CTCP framing parser and `parse/1` takes the `args` half it already
returns, so this file never touches `\x01`. The address decodes all the way
to an `:inet.ip_address()` tuple, which is exactly what
`Grappa.Net.Ssrf.safe_public_ip?/1` already takes — a peer-supplied address
is textbook SSRF input, so the eventual dial site reuses the guard hardened
for #75 instead of growing a second one.

**Three refusals, each a safety property rather than a shortcut.**

- *The sizeless historical form.* Without a declared size a receiver cannot
  check a byte cap BEFORE dialling, so the cap degrades from a precondition
  to a hope; and the right-to-left split that recovers an unquoted spaced
  filename stops being decidable.
- *Port 0 with no trailing token.* A passive offer missing its token can be
  neither dialled nor answered, so calling it well-formed would hand the
  caller an unusable struct.
- *Non-strict address literals.* `:inet.parse_strict_address/1` only,
  matching `Grappa.Net.Ssrf`'s posture: a short or octal form must never be
  quietly decoded into a loopback address behind the guard's back.

**The filename comes back verbatim**, traversal shapes included. Nothing
downstream may use a peer filename as a path — `Uploads.storage_path/2`
derives every on-disk name from a base32 slug and raises otherwise — so
sanitising here would build a second, weaker guard beside the real one.

`CHAT`, `RESUME` and `ACCEPT` return `{:unsupported_verb, verb}` rather than
`:malformed`. Whether grappa answers a resume is an open product question,
and a parser is the wrong place to settle it; naming the verb is truthful
whichever way it lands.

## Two measurements worth reading before the ruling

Both are in the DESIGN_NOTES entry, because they would otherwise be
re-derived wrongly in opposite directions.

**1. #1280's stated reason forbids the INBOUND leg, and receiving is
outbound.** Entry #1280 rules DCC out "permanently", but the reason it
records is *"accepting **inbound** P2P connections from arbitrary IRC
nicks"*. In `DCC SEND` the offerer listens and the receiver dials out — so
grappa **receiving** a file sits outside that rationale entirely, while
grappa **sending** one sits squarely inside it. Receive-only is therefore
not a reversal of #1280. The issue does not cite #1280 at all, so this was
not visible from the ticket. Rider: **passive/reverse DCC inverts the
roles** (the receiver listens), so "receive-only" does not imply "no
listener" unless passive offers are refused outright — an edge the issue
does not list.

**2. The existing upload store cannot be the destination.** The issue's
shape says the bytes land in it "as if the user had uploaded it". Measured:
`UploadsController`'s `@mime_categories` is a **closed** MIME allowlist and
a DCC offer carries **no MIME at all**; `GET /uploads/:slug` is **public and
unauthenticated** by design; and #1280's last paragraph already reserves
that route for *"content the operator's own users chose to publish — never a
proxy for arbitrary third-party URLs."* `Grappa.Avatars` is the precedent
and went the other way for the same reason — own context, own cap, own TTL,
own reaper — because a shared data model with a type flag across two trust
domains is a boundary violation, not reuse. The HTTP read path in the issue
survives; the store does not.

## Gates

`scripts/check.sh` green at branch HEAD: Credo strict `found no issues`,
Sobelow, Dialyzer `Total errors: 0`, ExUnit `8 doctests, 68 properties,
7329 tests, 0 failures`, bats. `scripts/design-notes-gate.sh` reports
`1 new entry heading(s), separator and marker present`.

New tests: 30 examples + 3 properties in `test/grappa/irc/dcc_test.exs`.
They are not a mirror — one of them caught a real bug in the parser during
development (`SEND f 0 1 1`, an active offer from 0.0.0.0, was being read as
a passive one because its own filename landed in the address slot).

## Not claimed

- **No interop evidence.** No real `DCC SEND` from a real client was ever
  observed for this work. The parser is validated against the protocol as
  read, not against a live sender.
- `scripts/integration.sh` was not run: this branch adds no runtime, no HTTP
  surface and no e2e path.
- No `protocol_version` change — there is no wire-shape change here. The
  first DCC event will need one.
- No mutation testing.

### DIFF
--- a/docs/DESIGN_NOTES.md
+++ b/docs/DESIGN_NOTES.md
@@ -54137,3 +54137,97 @@ quote precisely because it is one. Colour only: muted, never hidden.
 - Out of scope, as the issue says: turning the reply into a structured field
   with its own wire representation. This is a colour on a region that was
   already identified.
+<!-- entry #2089 -->
+
+---
+
+## 2026-09-12 — #2089: the DCC wire parser, and the two measurements that must outlive the ruling
+
+vjt reopened DCC on #sbiffo. #167 is closed not-planned, `README.md:289`
+lists DCC under **Out of scope** and `NON-GOALS.md:22` repeats it as **No
+DCC** — so the reversal is an ACT, it needs vjt's word, **it has not been
+made, and neither of those two files is touched here.** What landed is the
+one piece no product decision can move (the wire shape) and the two
+measurements a later reader would otherwise get wrong in both directions.
+
+### #1280's stated reason forbids the INBOUND leg — and receiving is OUTBOUND
+
+Entry #1280 says *"DCC is out of scope, permanently"*, and the reason it
+gives is: *"accepting **inbound** P2P connections from arbitrary IRC nicks
+has no place in that architecture"*. In `DCC SEND` the OFFERER listens and
+the RECEIVER dials out. So the two halves are not symmetric under that
+sentence: **grappa receiving a file is an outbound connect and is outside
+the recorded rationale; grappa sending one requires a listener and is
+squarely inside it.** Receive-only is therefore not a reversal of #1280 —
+it is work the recorded reason never reached. That asymmetry is the whole
+argument for shipping the receive half first, and it is not visible from
+the issue, which does not cite #1280 at all.
+
+Two riders. #1280 also says the question *"was never seriously
+considered"*, so the `permanently` is broader than the analysis behind it.
+And **passive/reverse DCC inverts the roles**: on `DCC SEND <name> <addr> 0
+<size> <token>` it is the RECEIVER that must listen, so "receive-only" does
+not imply "no listener" unless passive offers are refused outright.
+
+### The existing upload store cannot be the destination, and this is measured
+
+The issue's shape says the bytes *"land in the existing upload store as if
+the user had uploaded it"*. Three measurements say otherwise:
+
+- `UploadsController`'s `@mime_categories` is a **closed** MIME allowlist
+  (image / video / document / audio; unknown → 415). A `DCC SEND` carries
+  **no MIME at all** — a name and a size. Either everything outside the
+  allowlist is refused, which makes DCC useless for the archives that are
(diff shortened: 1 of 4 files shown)


vjt:
Closing on vjt's order (2026-09-12 11:51): the DCC **send** direction is dropped from #2089 — grappa already hands out an HTTPS upload URL, so serving the same file back over DCC adds a transport for no gain. This PR parsed the `DCC SEND` offer as groundwork for exactly that side.

#2089 continues as **receive-only** (connect out, drain into the upload store, surface the URL) and stays on hold until vjt lifts it. A fresh PR will start from the receive path; nothing here is reused as-is, so this one is archived rather than reworked.

Issue #1964, opened 2026-09-06
cicchetto: pasted-text upload confirm has no preview and a constant filename


vjt:
Reported by @gmsrc on IRC (2026-09-06 23:09-23:11), filed after reading the code. He will pick up the work.

## What happens

Pasting a block large enough to trip the flood guard and then choosing "upload it as a text file" opens the **upload confirm** dialog — the one designed around an image thumbnail — with **no preview at all** and with a **constant filename**. The dialog therefore says nothing about what is being uploaded.

## Where it comes from (measured, not guessed)

* `cicchetto/src/lib/pasteRoute.ts:78` — `PASTE_UPLOAD_FILENAME = "paste.txt"`.
* `cicchetto/src/lib/pasteRoute.ts:89-94` — `uploadPastedText` mints `new File([text], PASTE_UPLOAD_FILENAME, { type: "text/plain; charset=utf-8" })` and hands it to `dropUpload`, i.e. the same door as a picked or dropped file.
* `cicchetto/src/lib/uploadOrchestrator.ts:705-711` — `toAttachment` sets `thumbnail: categoryOf(staged.file.type) === "image" ? staged.file : null`, so `text/plain` gets `null`.
* `cicchetto/src/ConfirmModal.tsx:42-43` — `blob === null` ⇒ no object URL is minted and the `confirm-modal-attachment-thumb` node is not rendered.

The `null` is deliberate: the comment above `toAttachment` argues that a picture is the only preview worth showing because *"the name is what distinguishes `contract-final.pdf` from `contract-draft.pdf`"*.

## Why that argument does not hold on this path

That reasoning assumes a **file the operator named**. On the paste path the name is the constant `paste.txt` for every paste in every window, so the one signal the design relies on carries zero bits here. What survives is `detail: formatBytes(size)` — a byte count. The operator is asked to authorise an upload whose content they cannot see and whose name never varies.

Two directions, both open (whoever takes it decides):

1. Render the first N lines of a `text/*` attachment as a textual preview in the same slot the thumbnail occupies.
2. Keep the slot empty but make `detail` carry something discriminating for `text/*` (line count alongside size), and/or derive a filename from the paste rather than using a constant.

## Scope note

The confirm is opt-in and OFF by default (#1883, server-side `upload_confirm_enabled`), so this is only visible to operators who turned it on — @gmsrc is one.

PR #2095, opened 2026-09-12
uploads: choose the TTL in the pre-send confirm (issue 2094)

### LINKED ISSUE #2094: uploads: choose the TTL in the pre-send preview modal, not only in settings
Today the upload TTL is a per-user preference and nothing else: `SettingsDrawer` writes
`user_settings.data["upload_ttl_seconds"]`, `uploadOrchestrator` caches it, and at dispatch time
`pickHostTokenFromSeconds(host, uploadTtlSeconds()) ?? host.defaultTtl` turns it into the host token
(`cicchetto/src/lib/uploadOrchestrator.ts:478`). Server side, `POST /api/uploads` already accepts a
per-request `expire` from a closed ladder — `@allowed_ttl_seconds [3600, 43_200, 86_400, 259_200]`,
default `86_400` (`lib/grappa_web/controllers/uploads_controller.ex:145`).

So the per-upload knob exists on the wire and is simply not reachable from the UI. The consequence is
that retention is decided once, in a drawer, for every future file — and the moment where the operator
actually knows how long a file should live (they are looking at the thing they are about to post) is
exactly the moment they cannot say so.

## Ask

Put a TTL control in the pre-send preview/confirm modal (#1964), alongside the staged rows, so the
choice is made per batch at the point of decision.

- Ladder = the server's `@allowed_ttl_seconds` (1h / 12h / 24h / 72h), not a free-form duration —
  the closed set is already the contract and the host token ladder maps off it.
- Default = the persisted `upload_ttl_seconds` preference, falling back to `host.defaultTtl`. The
  preference stays the default; the modal overrides it for this batch only.
- Scope = per batch, not per row, unless there is a reason to go finer. The confirm dialog asks one
  question about one set of files.
- Whether an override should write back to `upload_ttl_seconds` ("remember this") is a separate
  decision — my instinct is no, a one-off should stay one-off.

## Design question worth settling first

The confirm modal is **opt-in** (#1883, `upload_confirm_enabled`, default `false`). If the TTL picker
lives only there, then the operators who never enabled the confirm can never choose a TTL per upload,
and the feature is invisible to the default configuration. Options, roughly:

1. TTL picker only inside the confirm modal — cheapest, but gated behind an unrelated opt-in.
2. Enabling a per-upload TTL implies showing the confirm — couples two settings that are today
   independent.
3. A separate lightweight affordance in the compose/attachment area for the no-confirm path — more
   surface, but keeps the two opt-ins orthogonal.

@vjt — which of the three, before anything gets built? (1) is what the title asks for literally; (3)
is what makes it usable by default.

## Notes

- No server change appears to be needed: `parse_ttl/1` already takes `expire` per request and 400s
  anything off the ladder. Worth confirming the litterbox host path carries the same ladder.
- Wire contract: nothing new is emitted, so no `protocol_version` implication.

gmsrc:
Closes #2094. vjt's ruling in the issue was **option 1**: the TTL picker lives inside the pre-send confirm (#1964) and nowhere else.

## What changed

The per-request `expire` has been on the wire since the embedded host landed — `UploadsController.parse_ttl/1` takes it from `@allowed_ttl_seconds [3600, 43_200, 86_400, 259_200]` and 400s anything else. Nothing in cic could reach it. The only way to say how long an upload lives was `upload_ttl_seconds`, a preference set once in the settings drawer, so retention was decided in advance for every future file by an operator with no file in front of them.

**No server change.** The ladder cic offers for the embedded host IS `@allowed_ttl_seconds`, spelled in seconds; the wire shape is unmoved, so there is no `protocol_version` implication.

1. **The store learned a generic CHOICE, not a TTL.** `ConfirmRequest` gains `choice: ConfirmChoice | null` beside `alternative` and `attachments`, on identical terms — a label, pre-formatted options, a reactive `value()`, an `onSelect` closure — and `confirmDialog.ts` knows nothing about uploads. Ten call sites spell `choice: null` explicitly, the same contract the two neighbours carry.
2. **The answer rides the queue.** `QueuedUpload` and `lastAttempt` carry `ttlSeconds`; the selection is a signal created per REQUEST, so a cancelled dialog takes its half-made choice with it. Three layers at dispatch: the batch's answer, then the stored preference, then the host default — the opt-out path enqueues `null`, which is exactly the old behaviour.
3. **Seconds → host token happens at DISPATCH**, never in the dialog: `activeHost()` is a reactive read of an admin setting, so a token picked at drop time could be a token for the wrong host. A retry re-sends on the terms that were chosen.
4. **The seed is validated against the active host's ladder** — `upload_ttl_seconds` is a bare integer with no host attached, and a preference this host cannot serve would seed the dropdown with an option that is not in it.

## Known cost, accepted rather than worked around

The confirm is **opt-in** (#1883, `upload_confirm_enabled`, default `false`), so an operator who never switched it on cannot choose per upload and keeps exactly today's behaviour. That is option 1 as ruled; it is the same shape as the limitation #1883 already accepted when it put the confirm toggle inside the TTL fieldset.

Also deliberately not built: no write-back to `upload_ttl_seconds` (a one-off stays one-off) and no "remember this" checkbox (a third control on a dialog that just gained one).

## Tests

- `uploadOrchestrator.test.ts` — 10 new: the ladder and its seeding (preference / host default / off-ladder preference), a host with no ladder asks nothing, the chosen duration overrides the preference without being written back, the whole batch gets it, a retry keeps it, a cancelled batch drops it.
- `ConfirmModal.test.tsx` — 5 new: absent when `choice` is null, options + selection rendered, named visibly and only once, a pick reported back, and the caller's value followed without the request being replaced (so focus is not re-stolen).
- `e2e/tests/issue2094-upload-confirm-ttl.spec.ts` — drives the picker to "1 hour" and reads `expire` back out of the real multipart POST body. The unit tests prove a function call; only this proves the choice reached the server.

## Gates

- `scripts/bun.sh run check` — **5 stages ran, 0 failed**.
- `scripts/bun.sh run test` — 7132 passed, 6 failed across 4 files (`compose`, `e2eConstantMirrors`, `moduleRootGuard`, `userTopic`). **All four are timeouts, and all four pass in isolation** (517 passed, 4 files) — the known Windows-host flake set, not this branch.
- e2e cannot run on this host; every spec that opens a confirm was audited file-by-file against the extra element first — all testid- or body-scoped, no child-count, snapshot, tab-order or geometry assertion touches `.confirm-modal`.
- `scripts/design-notes-gate.sh` — 1 new entry heading, separator and marker present.

### DIFF
--- a/cicchetto/e2e/tests/issue2094-upload-confirm-ttl.spec.ts
+++ b/cicchetto/e2e/tests/issue2094-upload-confirm-ttl.spec.ts
@@ -63,7 +63,7 @@ test("2094 — the chosen duration is the one the server is asked for", async ({
   // The control is named in words. A bare dropdown reading "24 hours" says
   // nothing about what happens then, which is the whole reason it carries a
   // visible label rather than only an accessible one.
-  await expect(page.getByTestId("confirm-modal-choice")).toContainText("Expires after");
+  await expect(page.getByTestId("confirm-modal-choice")).toContainText("Delete after");
 
   const select = page.getByTestId("confirm-modal-choice-select");
   // The embedded host's ladder, which is the server's own `@allowed_ttl_seconds`
--- a/cicchetto/e2e/tests/issue2094-upload-confirm-ttl.spec.ts
+++ b/cicchetto/e2e/tests/issue2094-upload-confirm-ttl.spec.ts
@@ -32,11 +32,17 @@ const png = (name: string) => ({
 // Collect the multipart bodies of real POSTs to the embedded host. Same
 // same-origin constraint as #1883's counter — a `page.route()` stub would
 // block cic's own bootstrap — so the request is observed, never intercepted.
+//
+// `postDataBuffer()`, not `postData()`: the latter returns the body decoded as
+// UTF-8 and answers NULL when it is not valid UTF-8, which a multipart body
+// carrying PNG bytes never is. Measured — the first run of this spec read `""`
+// and failed on an empty haystack. latin1 is the lossless byte-per-char
+// decode, and the field being read is ASCII.
 function collectUploadBodies(page: Page): () => string[] {
   const bodies: string[] = [];
   page.on("request", (req) => {
     if (req.method() === "POST" && req.url().endsWith("/api/uploads")) {
-      bodies.push(req.postData() ?? "");
+      bodies.push(req.postDataBuffer()?.toString("latin1") ?? "");
     }
   });
   return () => bodies;
@@ -84,7 +90,11 @@ test("2094 — the chosen duration is the one the server is asked for", async ({
   // default anyway would be green everywhere else and red here. Read the
   // field's own segment rather than the whole body — the PNG bytes are in
   // there too, and "contains 3600 somewhere" is not evidence.
-  const expireField = bodies()[0]?.split('name="expire"')[1]?.slice(0, 120) ?? "";
+  const body = bodies()[0] ?? "";
+  // Fail on the READ before failing on the value: an unreadable body makes
+  // every assertion below vacuously about an empty string.
+  expect(body).toContain('name="expire"');
+  const expireField = body.split('name="expire"')[1]?.slice(0, 120) ?? "";
   expect(expireField).toContain("3600");
   expect(expireField).not.toContain("86400");
 });
--- a/cicchetto/e2e/tests/issue2094-upload-confirm-ttl.spec.ts
+++ b/cicchetto/e2e/tests/issue2094-upload-confirm-ttl.spec.ts
@@ -84,16 +84,32 @@ test("2094 — the chosen duration is the one the server is asked for", async ({
   await sendPickedFiles(page);
 
   await expect(scrollbackLine(page, "privmsg", "📸").first()).toBeVisible({ timeout: 15_000 });
-  await expect.poll(() => bodies().length, { timeout: 15_000 }).toBe(1);
+
   // The wire, not the intent: the multipart body carries the `expire` field
   // the controller parses. A UI that changed its own label and posted the
-  // default anyway would be green everywhere else and red here. Read the
-  // field's own segment rather than the whole body — the PNG bytes are in
-  // there too, and "contains 3600 somewhere" is not evidence.
+  // default anyway would be green everywhere else and red here.
+  //
+  // FOUR stages, each naming its own cause (vjt's review of this spec's first
+  // red). A single `?? ""` collapsed worlds with opposite causes into one empty
+  // string: no POST captured at all, a POST whose body could not be read, and a
+  // body carrying no `expire` part all reported identically, and the reader was
+  // handed `Received string: ""` for any of them.
+  await expect
+    .poll(() => bodies().length, { message: "no POST to /api/uploads was seen", timeout: 15_000 })
+    .toBe(1);
+
   const body = bodies()[0] ?? "";
-  // Fail on the READ before failing on the value: an unreadable body makes
-  // every assertion below vacuously about an empty string.
-  expect(body).toContain('name="expire"');
+  // Stage 2 — the body was READABLE. This is the one that actually fired:
+  // `postData()` returns null on a body that is not valid UTF-8, which a
+  // multipart carrying PNG bytes never is, so the collector pushed "" while
+  // the upload itself had gone through (the scrollback link above proves it).
+  expect(body.length, "the upload POST body was captured but could not be read").toBeGreaterThan(0);
+  // Stage 3 — the request builder put an `expire` part in it at all.
+  expect(body, "the upload POST carried no expire field").toContain('name="expire"');
+
+  // Stage 4 — and its value is the one the operator picked. Read the field's
+  // own segment rather than the whole body: the PNG bytes are in there too, and
+  // "contains 3600 somewhere" is not evidence.
   const expireField = body.split('name="expire"')[1]?.slice(0, 120) ?? "";
   expect(expireField).toContain("3600");
   expect(expireField).not.toContain("86400");
--- a/cicchetto/e2e/tests/issue2094-upload-confirm-ttl.spec.ts
+++ b/cicchetto/e2e/tests/issue2094-upload-confirm-ttl.spec.ts
@@ -1,0 +1,90 @@
+// 2094 — the operator chooses how long the upload lives, in the dialog that
+// shows them what it is.
+//
+// The defect: the TTL ladder was reachable only from the settings drawer, so
+// retention was decided once, in advance, for every future file — while the
+// moment an operator actually knows what a file is worth is the moment they
+// are looking at it. The server has taken a per-request `expire` since the
+// embedded host landed (`UploadsController.parse_ttl/1`); nothing in the UI
+// could reach it.
+//
+// Why a real browser and not jsdom: the unit tests prove the orchestrator
+// hands the chosen seconds to the host, which is a statement about a function
+// call. What has to be true is that the CHOICE reaches the SERVER, and the
+// only honest witness to that is the multipart body of the real POST.
+
+import type { Page } from "@playwright/test";
+import { TINY_PNG_HEX } from "../fixtures/bytes";
+import { loginAs, scrollbackLine, selectChannel } from "../fixtures/cicchettoPage";
+import { setUploadConfirmEnabled } from "../fixtures/grappaApi";
+import { AUTOJOIN_CHANNELS, NETWORK_SLUG } from "../fixtures/seedData";
+import { expect, specNick, specUser, test } from "../fixtures/test";
+import { sendPickedFiles } from "../fixtures/uploadJourney";
+
+const CHANNEL = AUTOJOIN_CHANNELS[0];
+
+const png = (name: string) => ({
+  name,
+  mimeType: "image/png",
+  buffer: Buffer.from(TINY_PNG_HEX, "hex"),
+});
+
+// Collect the multipart bodies of real POSTs to the embedded host. Same
+// same-origin constraint as #1883's counter — a `page.route()` stub would
+// block cic's own bootstrap — so the request is observed, never intercepted.
+function collectUploadBodies(page: Page): () => string[] {
+  const bodies: string[] = [];
+  page.on("request", (req) => {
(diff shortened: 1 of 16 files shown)


vjt:
Ruling confirmed by vjt (option 1), so the design is settled and this is purely about the red.

**Not merging: `integration` settled RED, and the only failure is this branch's own new spec.**

- Run [34704734592](https://github.com/vjt/grappa-irc/actions/runs/34704734592), shard **2/4**: `1 failed, 246 passed (11.9m)`, **0 retries, 0 flaky**. Shards 1/4, 3/4, 4/4 green; the whole `ci` workflow green (dialyzer, cicchetto types+lint+unit, shottino, test+lint+audit).
- The single red is `tests/issue2094-upload-confirm-ttl.spec.ts:45 › 2094 — the chosen duration is the one the server is asked for`, at line 88:

```
Expected substring: "3600"
Received string:    ""
```

This cannot be a known flake **by construction** — the spec name is introduced by this branch, so it appears in no flake register, and it is precisely the test the PR description nominates as the only one that proves the choice reaches the server. The unit tests prove a function call; this one is the end-to-end claim, and it says no.

**One request before the next push, because the oracle cannot currently tell you which half is broken.**

```js
const expireField = bodies()[0]?.split('name="expire"')[1]?.slice(0, 120) ?? "";
expect(expireField).toContain("3600");
```

The `?? ""` collapses two different worlds into the same empty string:

1. `bodies()` is empty — no multipart POST was captured at all (the dispatch never happened, or the route interception missed it), or
2. a body *was* captured and simply carries no `name="expire"` part — the picker's answer never reached the request builder.

Those have opposite causes and the assertion reports them identically. Splitting it — assert `bodies().length` first, then that the body contains `name="expire"`, then the value — makes the next red name its own cause instead of handing the reader an empty string.

The run kept the artefacts that answer it directly: `error-context.md`, `video.webm` and `trace.zip` under `test-results/issue2094-upload-confirm-t-34a0f-.../`. The DOM snapshot in `error-context.md` is usually enough to say whether the dialog ever got as far as dispatching.

**Declared limits of what I measured:** one run only, so I have not established whether this is deterministic — and I have not read the trace, so I am not claiming to know which of the two branches above it is. What is established is that the red belongs to this branch and not to the base: the mechanical gates are clean (0 commits behind `origin/main`; DESIGN_NOTES marker `<!-- entry #2094 -->` unique against `origin/main` as it stands now, with positive and negative controls; boundary shape exact; no stray closing keyword in the commit messages).

Ball is yours — ping me when it is pushed and I will re-gate.

gmsrc:
Both asks done, and the artefacts settle the question you deliberately left open.

**Which of the two worlds it was: the READ, not the dispatch.** I pulled `playwright-traces-shard-2` from run 34704734592. `error-context.md` carries the uploaded link in scrollback — `https://nginx-test/uploads/gg7tn6q24djczeb4okoozqpyiq.png` — so the POST happened and the server answered 201. Combined with `expect.poll(() => bodies().length).toBe(1)` having passed before the red, the body WAS captured and the empty string came from reading it.

Cause: `req.postData()` decodes the body as UTF-8 and returns `null` when that fails, which a multipart body carrying PNG bytes guarantees. My collector's `?? ""` then pushed an empty string. The assertion never saw a TTL — it failed on an empty haystack, exactly as you read it.

Fixed in `53f21486`: `req.postDataBuffer()?.toString("latin1")` — the raw bytes, losslessly decoded byte-per-char, and the field being read is ASCII.

**The oracle split, in `e1d0bbda`.** Four stages, each with its own failure message:

```js
await expect.poll(() => bodies().length, { message: "no POST to /api/uploads was seen" }).toBe(1);
const body = bodies()[0] ?? "";
expect(body.length, "the upload POST body was captured but could not be read").toBeGreaterThan(0);
expect(body, "the upload POST carried no expire field").toContain('name="expire"');
const expireField = body.split('name="expire"')[1]?.slice(0, 120) ?? "";
expect(expireField).toContain("3600");
expect(expireField).not.toContain("86400");
```

Stage 2 is the one that fired last time, and it is the one your review showed had nowhere to report itself.

**Limits of what I am claiming.** I read `error-context.md`, not `trace.zip`, so "the body was captured" rests on the poll having passed rather than on seeing the bytes; and I cannot run e2e on this host, so the fix is verified by CI and not locally. `scripts/bun.sh run check` is 5 stages, 0 failed on both commits.

Also on the branch since your gate: `26ea13f6` renames the control from "Expires after" to "Delete after" (Gabriele's call) — one string, three places, all in sync.

Pushed. Ready for the re-gate.

vjt:
Re-gated `e1d0bbda`. **Still red, same shard, same spec — and the split oracle earned its keep immediately: it names the stage.**

Run [34704734833](https://github.com/vjt/grappa-irc/actions/runs/34704734833) — sorry, the shard job is `103585843019` — `1 failed, 246 passed (12.6m)`, shards 1/3/4 green, `ci` workflow green throughout.

```
Error: the upload POST body was captured but could not be read
  expect(body.length, ...).toBeGreaterThan(0)
  Expected: > 0
  Received:   0
  at tests/issue2094-upload-confirm-ttl.spec.ts:106
```

**Stage 1 passed.** A POST to `/api/uploads` was seen — `bodies().length === 1`. Stage 2 failed: the body is empty.

**So the UTF-8 diagnosis is falsified, not confirmed.** The cure swapped `postData()` for `postDataBuffer()?.toString("latin1")` and **the symptom did not move**: empty before, empty after. That is the displacement test this project holds everyone to, and it came back negative — a correct-sounding mechanism that turned out not to be the one operating. Worth saying plainly because the reasoning was good and the artefact citation was real; it just was not load-bearing.

**My lead, and I am labelling it UNMEASURED because I have not run it.** Chromium does not hand the Network domain a request body built from a file blob — `postDataBuffer()` is documented to return null when the body is not available, and an upload assembled from a `File` is the canonical case. If that is what is happening here, then *no* amount of decoding fixes it, because there is nothing to decode: the approach of reading the multipart bytes off the intercepted request cannot work for a file upload in this browser. I have not verified that against this Playwright version, and you are better placed to than I am.

**A falsification that costs one run:** log `req.postDataBuffer() === null` versus `.length === 0` in the collector. `null` says the browser never exposed the body and the approach is structurally dead; a zero-length buffer says something else is going on and the lead is wrong.

**If it is structural, the way out is to stop asserting on the request and assert on the consequence.** The server already parses `expire` from a closed ladder and 400s anything off it — so the TTL that was actually received is observable *server-side*, and an oracle that reads it there proves the same claim without depending on what Chromium is willing to show the test. That keeps the end-to-end assertion the feature needs (the unit tests prove a function call; they cannot stand in for this).

**Where that leaves the merge:** held. The mechanical gates are clean on this head too — 0 behind `origin/main`, marker `<!-- entry #2094 -->` still unique against `origin/main` as it stands, no closing keyword in the commit messages, and the only file touched since my last gate is the spec, so the product code has not moved. The design is settled and the implementation may well be correct. What is missing is the proof, and it is missing twice now.

**Declared: two runs, both red, same failure, so I will call it deterministic rather than flaky — but two is what I have, and the first red was a different assertion reading the same empty string, so it is really one symptom observed twice, not two independent facts.**

gmsrc:
**You are right and I was wrong: the UTF-8 diagnosis is retracted.** The displacement test is the whole argument — empty before, empty after — and I should have applied it myself instead of reading a plausible mechanism into a symptom that never moved. Thank you for saying it plainly.

**Your lead is now MEASURED, and it holds.** Standalone Playwright 1.59.1 (the version `e2e/` pins) driving Chrome 152 on this host via `executablePath`, three POSTs through ONE collector so the answer is comparative:

| body | `postData()` | `postDataBuffer()` |
|---|---|---|
| multipart **with a `File`** | `null` | `null` |
| multipart, text parts only | `string(244)` | `buffer(244)` |
| plain `expire=3600` | `string(11)` | `buffer(11)` |

The two controls read fine through the same listener, so the collector is not the problem and the `File` is the variable. Exactly your falsification, and it is `null`, not zero-length: **the browser never exposes the body, so the approach is structurally dead** and no decoding was ever going to cure it.

Declared limits: this is the host's Chrome, not CI's bundled Chromium build, and one Playwright version. What it establishes is the mechanism, and CI's two reds are the same symptom.

**Which is your third point, and it was already pushed before I read this: `2d995807` moves the oracle to the consequence.** The spec no longer touches the request. It reads the 201's `expires_at` and asserts the lifetime is an hour — a window (30 min .. 2 h), not an equality, because the timestamp carries the server's clock and is read against the runner's, and the thing it must be distinguished from is the 24-hour default. That is the stronger claim: not that cic spelled a form field, but that the file really will be deleted when the operator said.

Your split stages are kept and are why the second red could be read at all: a 201 arrived / its body parsed / it carries `expires_at` / the lifetime is an hour, each with its own message.

`3b9fcca8` records the measurement in DESIGN_NOTES as a general rule for the next spec — an upload request body is not observable, assert on the response or on server state — with the retraction and the displacement test that produced it.

Two commits since your gate, both spec/docs; the product code has not moved since `26ea13f6`. CI is running on `3b9fcca8`. Ping when you want the re-gate.

gmsrc:
**Green on `3b9fcca8`, all nine checks, including the shard that carried the two reds.**

Run [34707257103](https://github.com/vjt/grappa-irc/actions/runs/34707257103) — head SHA `3b9fcca83498b24463d7ae5f5f64914c7630b2c2`, matching the branch tip:

- `cicchetto + grappa + azzurra-testnet` shards 1/4 … 4/4 — **pass** (shard 2/4, the one that was red twice, 15m18s)
- `integration (all shards)` — **pass**
- `cicchetto (types + lint + unit)`, `dialyzer`, `shottino (build + tests)`, `test + lint + audit` — **pass**

**What I am NOT claiming.** One green run. The spec that turned green is the one I rewrote, so this is the first observation of the new oracle passing, not evidence that it is stable — and it now depends on a clock window (30 min .. 2 h) that a pathologically skewed runner could in principle fall outside of. The 24-hour default is 24× outside the upper edge, so a wrong TTL still reds; a broken clock would be the false red, and I have not measured skew on the runners.

**What is established:** the previously-red assertion is gone, the claim it was trying to make is now made against the server's own `expires_at`, and nothing in the product code moved to achieve it — the last non-test commit is still `26ea13f6`.

Ball's yours for the re-gate.

Issue #1040, opened 2026-08-08
cicchetto: on the home window the rail actions must be expanded, with the buttons visible


vjt:
On the home window the rail actions must be **expanded, with the buttons visible** — not
collapsed behind the launcher the way they are on a channel.

vjt, 2026-08-08 on #grappa: *"in home il menu actions deve essere espanso"*, *"e i bottoni
visibili"*.

## Why home is the window where the collapse buys nothing

`#500` collapsed every rail affordance behind one launcher pinned to the bottom of the rail,
and the moduledoc at `cicchetto/src/RailActions.tsx:37` states exactly what problem that
solved:

> *"Before #500 they were an always-expanded column: on a big channel the nick list overflowed
> and pushed the column below the fold (desktop: unreachable) / squeezed the list (mobile)."*

That cost is a **nick list** cost. The home window has no nick list: `RailContext`
(`cicchetto/src/RailContext.tsx:32`) says it *"renders NOTHING for kinds with no context
content (channel already has the MembersPane above; home/admin/list/mentions have none)"*.
So on home the rail is an empty column with a single launcher pinned at its bottom, and the
operator pays a tap to reach buttons that had nowhere to be pushed to in the first place.

Home is also the window an operator lands on with nothing selected — the one place where the
actions ARE the content.

## Current behaviour

- `RailActions.tsx:162` — `const [open, setOpen] = createSignal(false)`; the menu starts closed
  on every window kind and nothing varies it by kind.
- `RailActions.tsx:514` — `.rail-actions-launcher` is the only thing in the flex flow;
  `RailActions.tsx:241` — `.rail-actions-menu` is the overlay it opens.
- `cicchetto/src/themes/default.css:2340`-ish — `.rail-actions` is the `position: relative`
  containing block with `margin-top: auto` flooring the launcher at the bottom; the menu is
  absolute with `bottom: 100%` so it expands UPWARD over the nick list.

## What needs deciding while implementing

The expanded home state is not simply "open the overlay by default" — the overlay geometry is
built for a menu that floats over a list:

- The menu is absolutely positioned against the launcher and opens upward (`bottom: 100%`),
  with a `max-height` derived from `spaceAbove` (`lib/menuPosition.ts`, `RailActions.tsx:213`)
  because a menu opening upward only has the space above the launcher. On home, with the whole
  empty column available, laying the buttons out in normal flow from the top reads better than
  a popover pinned to the bottom edge — but that is a layout decision, not a default-open flag.
[...]

PR #2099, opened 2026-09-12
cic: unread rollup badge on the archive launcher (2096)

### LINKED ISSUE #2096: cicchetto: unread rollup badge on the archive launcher in the RailActions drawer
Reported by Fairy on IRC; ordered by vjt (#grappa, 2026-09-12 20:50): *"aggiungiamo un badge sull'archive nel drawer actions se ci sono messaggi non letti li"*.

## What happens now

#532 B put per-row unread badges **inside** `ArchiveModal` (`cicchetto/src/ArchiveModal.tsx`, the `archive-modal-unread` block, keyed by `unreadKey(slug, target)` off the same `messagesUnread` / `eventsUnread` seed the sidebar reads).

The **launcher** carries nothing: the archive button in the RailActions drawer (`cicchetto/src/RailActions.tsx`, `rail-action-archive` / `data-testid="mobile-panel-archive"`) is a bare icon. Since #473 that button is the single archive surface on both form factors, so an archived window with unread messages is invisible until you open the modal and expand the right network group. Fairy hit exactly that.

## What we want

A rollup badge on the archive launcher when any archived window has unread. Same visual language as the sidebar/bottom-bar badges (`sidebar-msg-unread` / `sidebar-events-unread`), so messages and events stay distinguishable rather than collapsing into one dot.

## The design knot to settle first

`ArchiveModal` loads archive rows **lazily, per network group** — `loadArchive(slug)` fires on `<details onToggle>`, and the comment in the file says the list can be O(hundreds) per network. A rollup on the launcher needs the archived target set *before* anything is expanded. Two directions:

1. **Client-side eager:** load every network's archive list at startup (or on first drawer render) and roll the counts up client-side. Simple, but pays the O(hundreds)×networks cost the lazy design deliberately avoided.
2. **Server-side aggregate:** have the server report an archived-unread rollup (count, or just a boolean per kind) alongside the existing `unread_counts` seed. No list load, but it is a wire change — protocol version bump territory, so it needs a wire ruling.

Worker should measure both before picking; do not implement direction 1 by default just because it touches less.

## Notes

- Mute interaction: `lib/activeWindows.ts` already encodes "the mute always wins" for the on-screen affordances while the counts stay intact. Whatever the badge does must follow that rule, not invent a second one.
- Both form factors — desktop rail and mobile drawer render the same `RailActions` component.

vjt:
Closes #2096.

Fairy's report: an archived window holding unread was invisible until you opened `ArchiveModal` **and** expanded the right network group. #532 B put per-row badges inside the modal; the launcher that opens it carried nothing.

## The wire shape: NONE. No `protocol_version` bump.

That is a **change from the first ruling on the issue**, and it came from a measurement this branch took before building anything.

The issue framed two directions — eager-load every network's archive list, or ship a server-side aggregate beside the `/me` `unread_counts` seed — and vjt ruled for the second. Both premises were wrong in the same place: **the seed already carries archived windows.** `ReadCursor.bulk_unread_split/3` is driven purely from `read_cursors ⋈ networks ⋈ messages`, with no active-window filter anywhere in it.

That was a code read, so it was taken to the far end of the tube. A throwaway `ConnCase` with a **live session** against the fake ircd (without one `build_active_keyset` falls on `{:error, :no_session}`, the keyset is empty, every target reads as archived and the control collapses into the subject) read two decoded HTTP bodies in one run — `GET /networks/:slug/archive` as the **oracle** for "is this archived", `GET /me` as the **subject**:

```
oracle  GET /archive          : ["#m2096-archived", "#m2096-nocursor"]
subject GET /me unread_counts : {"#m2096-archived": 1, "m2096-peer": 1}
```

Three controls, all asserted before any number printed:

| | window | oracle says | in `unread_counts` |
|---|---|---|---|
| **POSITIVE** | `m2096-peer` — open query window, live session | ACTIVE | ✅ yes — the reader is in the right place |
| **NEGATIVE A** | `#m2096-nocursor` — rows, no cursor | ARCHIVED | ❌ no — the envelope is **not** a mirror of the listing |
| **NEGATIVE B** | `#m2096-nonesuch` — invented | absent | ❌ no |
| **SUBJECT** | `#m2096-archived` — rows + cursor | ARCHIVED | ✅ **yes** |

Neither set is a subset of the other. vjt re-ruled on that measurement: **derive client-side, no wire change, no bump, `loadArchive` stays lazy per group.**

Two further reasons the aggregate was dropped. It would have put a synchronous `GenServer.call` per network back on the `/me` cold-load path (`Session.list_channels/2` is `call_session`) — the work #498 removed. And it would have been a **boot-time snapshot**: nothing re-emits it when the operator opens the archived window, so the badge could not fall without a second push. The derivation is live for free.

## What changed

- **`lib/archive.ts`** — the subtraction comes out of `visibleArchiveForNetwork` into `archiveSuppressionForNetwork/2` + `archiveTargetSuppressed/3`. Behaviour-identical; the archive now has two consumers asking one question of different inputs, and two statements of it would drift.
- **`lib/archiveRollup.ts`** (new) — pure `rollupArchivedUnread` + a thin memo, the `orderUnreadWindows`/`activeWindows` split. Skips `$server` (`list_archive/3` excludes it) and any slug cic renders no group for (GH #105). It cannot over-count by construction otherwise: a key with unread has rows, and every non-active target with rows is in the listing.
- **`RailActions.tsx` + `default.css`** — the badge, on the same component both form factors mount. Two pills, `sidebar-msg-unread` / `sidebar-events-unread`, clustered like `.archive-modal-unread`.
- **`lib/selection.ts` + `lib/userTopic.ts`** — one real bug the badge exposed: a destructive archive DELETE left its seed count standing until the next cold load. Invisible while nothing summed seed keys with no window behind them. `archive_purged` now also calls `clearServerSeedCount`. The read cursor there still does not clear, for the reasons already written at that site.

## Mute

Untouched. "The mute always wins" governs the **affordance** and leaves the **counts** intact — the sidebar and the modal's own rows render a muted window's numbers. A count badge inherits that by doing nothing; subtracting mutes here would be a second rule no other badge obeys.

## Gates

| gate | result |
|---|---|
| `bun.sh run test` | **7159 passed / 353 files**, rc=0 |
| `bun.sh run check` | 5 stages, 0 failed (lock drift · test location · biome · tsc src · tsc e2e) |
| `design-notes-gate.sh` | `1 new entry heading(s), separator and marker present` |
| `scripts/check.sh` | see the comment below — it was still running when this was opened |
| **e2e** | **NOT RUN HERE.** Needs the exclusive runner lane; CI gives the verdict. |

## Limits stated, not discovered later

- **The e2e is not green on this host.** It is written and it asserts the visible outcome on both form factors — baseline, then a peer message into a channel the operator is **still in** (the control: the badge must not move), then PART (appears, +1), then JOIN (disappears) — all in one continuously-open rail menu. But it has never executed. CI owns that verdict.
- **The measurement's substrate is a fake ircd and `Phoenix.ConnTest`**, not production. The instrument was temporary and is not in this branch; the tree is clean.
- **No mention pill.** A mention is always also a content row, so `mentions > 0` implies `messages > 0`: omitting it cannot hide the existence of unread, only a severity signal.
- **No dimming when every contributor is muted.** "All of them are muted" is a different predicate from #1077's per-window "this one is", and inventing it is the same second rule.
- **The badge lives one tap inside the collapsed drawer.** `RailActions` renders its menu under `<Show when={expanded() || open()}>` and `expanded()` is true only on home (#1040). Better than "open the modal and expand the group", which is what the issue asked for; a dot on `rail-actions-launcher` itself is a separate decision, deliberately not taken here.

### DIFF
--- a/cicchetto/e2e/tests/issue2096-archive-launcher-unread-badge.spec.ts
+++ b/cicchetto/e2e/tests/issue2096-archive-launcher-unread-badge.spec.ts
@@ -1,0 +1,152 @@
+// issue 2096 — the unread rollup badge on the archive launcher.
+//
+// Reported by Fairy: an archived window holding unread was invisible until you
+// opened the ArchiveModal AND expanded the right network group. #532 B put
+// per-row badges INSIDE the modal; the launcher that opens it carried nothing.
+//
+// What this pins, in ONE continuously-open rail menu, so every transition is
+// observed LIVE rather than across reloads:
+//
+//   1. baseline — whatever the shared backend already holds, read off the
+//      badge itself (the suite shares one account across specs, so an
+//      absolute `0` would be a hostage to whatever ran before);
+//   2. THE CONTROL — a peer message lands in a channel the operator is still
+//      IN. The unread is real (the +1 in step 3 proves cic counted it), and
+//      the badge must NOT move: an ACTIVE window has its own surface and is
+//      not the archive's business. Without this reading, step 3 would only
+//      say "a number appeared", not "the subtraction works";
+//   3. APPEARS — PART the channel. Same unread, no new message, the window
+//      simply stops having a nav surface, and the rollup gains exactly 1;
+//   4. DISAPPEARS — JOIN it back. Still the same unread, the surface returns,
+//      the rollup falls back to the baseline.
+//
+// Steps 3 and 4 are what makes this spec a real one rather than a green mirror
+// of the code: the badge is DERIVED from `/me`'s seed minus what the nav
+// draws (`lib/archiveRollup.ts`), so a rollup wired to anything else — the
+// raw seed, a boot-time snapshot, the modal's lazily-loaded list — fails at
+// 2, at 4, or at both.
+//
+// Both form factors, because `RailActions` is ONE component mounted by both
+// branches of Shell's `isMobile()` split (#473). The desktop entry runs on the
+// `chromium` project by subtraction; the `@touch` twin runs on
+// `chromium-pixel-touch`. NOT `@webkit`: the claim here is "the rail menu on a
+// phone-shaped viewport shows the badge", and WebKit tap semantics are a
+// different axis (issue 1831) that this badge does not touch.
+
+import type { Page } from "@playwright/test";
+import { loginAs, openRailMenu, selectChannel } from "../fixtures/cicchettoPage";
+import {
+  assertMessagePersisted,
+  awaitPartEcho,
+  joinChannel,
+  partChannel,
+  restoreReadCursorToTail,
+} from "../fixtures/grappaApi";
+import { IrcPeer } from "../fixtures/ircClient";
+import { AUTOJOIN_CHANNELS, NETWORK_SLUG } from "../fixtures/seedData";
+import { expect, specLiveNick, specNick, specUser, test } from "../fixtures/test";
+
+const CHANNEL = AUTOJOIN_CHANNELS[0];
+const WITNESS = "#2096 — the line that must show up behind the archive door";
+
+test.afterEach(async () => {
+  const vjt = specUser();
+  // Restore the seed-time joined state and a tail cursor: this spec PARTs the
+  // shared autojoin channel and leaves an unread behind, and neither may
+  // poison a sibling spec (or this one under --repeat-each). Both idempotent,
+  // both guarded so a mid-test failure cannot cascade.
+  await joinChannel(vjt.token, NETWORK_SLUG, CHANNEL).catch(() => {});
+  await restoreReadCursorToTail(vjt.token, NETWORK_SLUG, CHANNEL).catch(() => {});
+});
+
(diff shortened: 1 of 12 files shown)


vjt:
`scripts/check.sh` finished after the PR was opened: **rc=0**.

```
ExUnit : 8 doctests, 65 properties, 7301 tests, 0 failures (38 excluded)
bats   : 645 ok, 0 not ok
```

Nothing Elixir changed on this branch, so that is a baseline rather than coverage of the diff — the diff is entirely `cicchetto/` plus the DESIGN_NOTES entry.

Still outstanding and unchanged: **the e2e has never executed here** (it needs the exclusive runner lane). `issue2096-archive-launcher-unread-badge.spec.ts` is the only gate that actually exercises the badge in a browser, on both form factors, and CI owns its verdict.

Issue #1911, opened 2026-09-04
auth: OIDC login for grappa/cicchetto, tested against Kanidm


vjt:
Requested by katia on `#grappa` (2026-09-05 00:49 CEST), approved by vjt for target 1.6.

## What exists today

Client-facing auth is entirely local:

- `POST /api/auth/login` → `GrappaWeb.AuthController.login/2` (`lib/grappa_web/router.ex:253`, `lib/grappa_web/controllers/auth_controller.ex:124`), backed by `Grappa.Accounts.Login.authenticate/1` (`lib/grappa/accounts/login.ex:79`) — Argon2 over `users.password_hash`.
- Second factor: TOTP (`POST /api/auth/totp/verify`) and passkeys/WebAuthn via `wax_` (`router.ex:254-258`), shipped in #442.
- The `users` schema (`lib/grappa/accounts/user.ex:47-55`) has `name`, `password_hash`, `totp_*`, `passkey_mode`, `is_admin` — **no external-identity columns**.
- `mix.exs` carries **no OIDC/OAuth2 client dependency** (`assent`/`oidcc`/`ueberauth` are all absent).
- Frontend login lives in `cicchetto/src/Login.tsx`.

So this is greenfield: an OIDC authorization-code + PKCE flow, a provider config surface, an identity-link table, and a button in cicchetto.

## Scope

1. **Server: authorization-code flow with PKCE.** Discovery via the provider's `/.well-known/openid-configuration`, JWKS fetch + cache + rotation, `id_token` signature/`iss`/`aud`/`nonce`/`exp` validation, `state` binding to the browser. Ends in the *same* opaque grappa session the local login issues — everything downstream (`:authn`, `RequireFullSession`, the request budget) must stay untouched.
2. **Identity linking.** New table keyed on `(issuer, subject)` → `users.id`. The `sub` claim is the identity, never the email or the preferred username — both are mutable at the IdP. Decide (open question below) whether a first unseen `sub` provisions a user or is refused.
3. **Config surface.** Provider(s) configured by the operator (issuer URL, client id, client secret, scopes, redirect URI), plus a switch for whether local password login stays enabled alongside.
4. **cicchetto.** "Sign in with …" on `Login.tsx`, callback handling, and a sane error path when the IdP round-trip fails.
[...]

PR #2102, opened 2026-09-13
#1911 - OIDC support

iakat:
glm-5.3-flash generated code
----
oidc wannabe


vjt:
Thanks for this — it is a large, coherent contribution, and the server-side `state`/`nonce`/PKCE machinery is correct. Because it is new **authentication** code we gave it a full security review rather than a skim, and we are sending everything in one delivery: CI status, the rebase that is owed, and 20 review findings.

Two notes on how to read this. Nothing below is in production — this branch is not merged, so none of it is a live vulnerability; it is ordinary review of unmerged code. And the failure mode we looked for is *plausible-and-wrong* rather than anything malicious: **8 of the 20 findings are a comment or a moduledoc asserting a check, an enforcement, or a test suite that is not there.** That is the characteristic residue of agent-authored code, and ours has the same habit — it is why we read it closely rather than a judgement about the work.

---

## 1. CI is red, 9 of 10 jobs

We approved and ran the workflow. The baseline was established first: `main` at `abfd6e01d` is green on every one of those jobs, so **none of this red is inherited**, and our rebase touched only `protocol.ex`, `socket.ts` and `shape.pin` — none of the failing files. Three causes, all small:

- **`mix format --check-formatted`** fails on `lib/grappa/auth/oidc.ex`, a file this branch adds.
- **`tsc` TS2614**: `cicchetto/e2e/tests/issue1911-oidc-login.spec.ts` imports `OIDC1911_USER` from `../fixtures/seedData`, and that symbol does not exist — not on `main`, and not on this branch either, where it appears only as an import and a use, never as an export. `seedData.ts` is not among the 52 files in the diff.
- **biome**: 2 errors.

## 2. A rebase is owed, and one value has to move

The branch is now 7 commits behind `main`. More importantly: this branch sets `@protocol_version 21`, and **21 was taken this morning** by another change that has since landed on `main`. On the rebase it has to become **22**, in three places that must stay in lockstep — `@protocol_version` in `lib/grappa/protocol.ex`, `CLIENT_PROTOCOL_VERSION` in `cicchetto/src/lib/socket.ts`, and the digest in `priv/wire/shape.pin`. `mix grappa.wire_pin --check` is the gate that catches a mismatch.

---

## 3. HIGH

### H1 — a crafted URL fragment logs any visitor into an attacker's account

`cicchetto/src/lib/oidc.ts` → `readOidcLanding`; `cicchetto/src/Login.tsx` → `onMount` / `consumeOidcLanding`; `cicchetto/src/lib/auth.ts` → `installSharedSession`.

`onMount` reads `window.location.hash` and, on a landing of `kind: "session"`, calls `installSharedSession(landing.token, landing.subject)` then `navigate("/")`. The only thing between the fragment and the credential store is `isSessionLanding`, which checks the payload's **shape**. Nothing checks its **provenance**: no server round trip, no one-shot token to redeem, no value the client minted at `beginOidcLogin()` and requires echoed back.

So an attacker with an ordinary account and its bearer `T` can build `{"kind":"session","token":"T","subject":{…their id…}}`, base64url it, and send a victim `https://<host>/login#oidc=<encoded>`. The victim's browser installs `T`; `clearOidcLanding()` then scrubs the address bar, so they see a normal logged-in grappa. Every network they bind and every message they send lands in the **attacker's** account. Login CSRF / session fixation, and it needs only that the victim follow a link.

The module comment calls this "the same install the share-link route uses". True of the install, false of the part that matters: `POST /auth/share/consume` hands the server a signed one-shot token and the **server** mints the session. Here the credential arrives pre-made in attacker-controlled input, and the `state`/`nonce`/PKCE machinery never sees this path because this path never talks to the server.

### H2 — `:issuer` is never required to be HTTPS

`lib/grappa/auth/oidc/config.ex` → `build/1`, `required/2`; `lib/grappa/auth/oidc/discovery.ex` → `refresh/1`, `build/2`, `fetch_https/2`.

`Discovery.build/2` puts `authorization_endpoint`, `token_endpoint` and `jwks_uri` through `fetch_https/2`, which rejects a non-`https` scheme. `issuer` goes through `fetch_string/2` instead, and `Config.build/1` validates it only as a non-empty binary. `refresh/1` then concatenates the discovery URL off `config.issuer` and hands it to `Req.get/2`, which speaks cleartext without complaint.

An operator setting `GRAPPA_OIDC_ISSUER=http://idm.internal` — plausible on a LAN-only IdM, and nothing refuses it — lets anyone on that path answer the plaintext discovery request with a document whose `issuer` matches (so `check_issuer/2` passes) and whose three endpoints are `https://attacker.example/…` (so `fetch_https/2` passes; they hold a valid certificate for their own domain). grappa then POSTs the authorization code **and the HTTP Basic client credentials** to the attacker's token endpoint and verifies the `id_token` against the attacker's JWKS.

What makes this a finding rather than operator error is that the code states the opposite in three places, including `Discovery.build/2`'s own docstring: *"An HTTPS endpoint is a hard requirement: a bearer code and a client secret must not be handed to a cleartext hop."* It is enforced for the three fields the provider supplies and skipped for the one the operator supplies — the one that decides whether the conversation is protected at all. The cure is one guard clause in `Config.build/1`.

### H3 — ~1950 lines of new authentication code with zero ExUnit tests

`lib/grappa/auth/oidc/` and `lib/grappa_web/controllers/oidc_controller.ex`.

Measured: `git ls-files | grep -i oidc` returns 25 files on the branch and **none under `test/`**; `git grep -c 'oidc' -- test/` exits 1 with no output, while the same call for `passkey` exits 0 with 17 files — so both the grep and the convention work. The added automated coverage is `issue1911-oidc-login.spec.ts` with exactly **two** Playwright tests (happy path, `not_linked` refusal), plus `scripts/oidc-e2e.sh`.

Three comments claim otherwise: `scripts/oidc-e2e.sh` (*"the Bypass-driven controller tests prove grappa's half"*), the `oidc-acceptance` job comment in `integration.yml` (*"the unit suite proves grappa's half over Bypass"*), and `verify_claims/5`'s docstring. Three test seams were built and are dead: `Jwks.put_test_keys/2`, `Discovery.put_test_discovery/2`, `Config.put_test_config/1`.

Why it matters concretely: both the script and the spec drive a *real, correct* provider, and a correct provider cannot be made to emit `alg: none`, a wrong `aud`, a mismatched `nonce`, an expired `exp`, a duplicate `kid`, or an `enc`-purpose key. The alg allowlist, the issuer compare, the audience arms, the nonce compare and the key-selection refusals — the whole forgery-resistance surface — have no test that can fail, so a future edit weakening any of them goes green. M7 and L4 below are both defects inside exactly that untested region, which is the evidence that the gap is not theoretical.

---

## 4. MEDIUM

- **M1 — the temporary CI trigger is still in the workflow.** `.github/workflows/integration.yml`, `on:` block: `push: branches:` goes `[main]` → `[main, oidc]`, and your own comment says *"temporary lane trigger (fork-only, drop before the upstream PR)"*. It must come out before merge, or our `integration` fires on pushes to a branch named `oidc`. This is the one finding whose cost is incurred by **merging** rather than by running.
- **M2 — no step-up re-authentication.** `oidc_controller.ex` → `start_link/2`, `unlink/2`: `POST /me/oidc/link` and `DELETE /me/oidc` change the account's credentials with only the ambient session.
- **M3 — the callback 500s on two expected error returns.** `oidc_controller.ex` → `admit/2`, `sync_admin_flag/3`: reproduces the bug class the surrounding code says it fixed.
- **M4 — on Docker the two group-gate variables never reach the container.** `compose.yaml` `services.grappa.environment` vs `.env.example` and `config/runtime.exs` ⇒ admin **demotion** silently never happens.
- **M5 — unbounded unauthenticated growth in `Transaction`.** `transaction.ex` → `put/2`, `handle_call({:take, …})`, `drop_expired/2`, and the client mints one entry per login-page view.
- **M6 — the two new unauthenticated routes are missing from `RouterScopeTest`'s allowlist**, which the router's own comment instructs. We predicted `router_scope_test.exs` RED before CI ran; it is two lines in an allowlist.
- **M7 — a multi-audience `id_token` with no `azp` is accepted**, while the comment says `azp` pins the client. `id_token.ex` → `check_audience/2` list arm, `check_azp/2`.

## 5. LOW (titles only — ask and we will paste the detail)

L1 `:persistent_term` as a 60s TTL cache on two hot paths · L2 the real transport/TLS failure discarded at three provider boundaries · L3 provider-controlled text logged verbatim into a newline-delimited log · L4 the moduledoc claims an `iat` check that does not exist, and `check_time/3` is a trap · L5 `scripts/oidc/flowdbg.sh` is a stale pre-fix copy of `flow.sh` · L6 `busybox` pulled unpinned in a CI-executed script, beside digest-pinned siblings · L7 predictable fixed `/tmp` filenames for TLS generation · L8 the acceptance stack binds `0.0.0.0` on the host network · L9 a trust-anchor private key written world-readable · L10 migration hygiene: hand-typed stamp, out-of-order version, unbalanced `writable_schema`.

---

## 6. Suggested order

Not a plan, just what the dependencies imply. **H1** is the only finding a stranger can exploit against a default deployment with nothing but a link, and its cure is client-side and small. **M1** wants settling before merge regardless of the rest. **H2** is one guard clause. **M6** is two lines. **H3** is the largest by effort and the reason the others were findable at all.

## 7. What we did not review

Everything above is static reading: we ran no test suite, no `scripts/oidc-e2e.sh`, no Playwright, no Sobelow/Credo/Dialyzer, and stood up no provider — M6 in particular was a prediction about a gate, and is labelled as one. Not read line by line: `docs/DESIGN_NOTES.md` and `docs/oidc-kanidm.md` prose against the code (only grepped for insecure operator guidance — one hit, and it *forbids* a skip-verification switch), `themes/default.css`, `cicchetto/e2e/oidc-op/server.mjs` (scanned for hard-coded keys, weak secret comparison and privilege; it reaches no production path), and the lab scripts. No timing-channel analysis of `find_identity/2`.

---

We would like these addressed on the branch rather than taking the work over — it is yours, and you know the design. H1, H2, M1 and M6 are the ones that gate a merge for us; the rest we are happy to discuss. Rebase onto current `main` and move the protocol number to 22 while you are in there, and CI will re-run on the new head.

Happy to expand any finding, and to be argued out of any of them with a measurement — three of the twenty came from reading a comment that turned out to be right about intent and wrong about the code, and we would rather be corrected than be politely agreed with.

iakat:
meh  @vjt acconcia no?

vjt:
Sono out of tokens e non ho modo di testare perché non ho questa infra in casa.

Puoi farlo tu, per favore, dato che puoi fare un test end to end che vale di più dei test sintetici, e poi metter su tutta la CI?

:-)

Issue #2118, opened 2026-09-13
Architecture review 2026-09-13 - 10 findings: 0 CRITICAL, 4 HIGH, 5 MEDIUM, 1 LOW


gmsrc:
**Base:** `main` @ `c030399a` (1.5.6). **Previous review:** 2026-08-15 @ `575e203a` (29 days).
**Method:** the six architecture concerns from `docs/reviewing.md`, run by one reviewer, read-only, no production instance. The reviewing host has no toolchain, so no `mix`, no Credo, no Dialyzer, no `boundary.find_violations` — every number below is a counted grep or a raw newline count.

**Total: 0 CRITICAL, 4 HIGH, 5 MEDIUM, 1 LOW — 10 findings.**

## Headline

The buckets filed 2026-08-17/18 (#1390–#1403, #1406) are 15 closed, 1 open, and the closures are mostly real work: #1393 codegen went **3/162 → 64/192** schemas consumed, #1402 hoisted the window-state set into `windowStateSets.ts`, #1396 took `compose.ts` from 2,345 → 1,501 lines, #1400 took casts 83 → 67, #1401 was ruled "not planned" with the measurement recorded.

**Two numbers regressed, and they are the ones that matter:**

- **#1390 was closed and its subject kept growing.** `Session.Server` **7,231 → 8,297 lines (+14.7%)** in 29 days — now 81 state keys, 370 defs, **72 `handle_call` + 55 `handle_info` + 1 `handle_cast`** on a process that exists one per `(user, network)`.
- `EventRouter` **4,401 → 5,450 lines (+23.8%)**.

## Findings

| # | Sev | Concern | Finding |
|---|---|---|---|
| A1 | HIGH | Abstraction | `EventRouter`'s state contract is `optional(any()) => any()`; it reads 24 fields; #1390's drift pin covers `Session.Server` only |
| A2 | HIGH | Cohesion | `Session.Server` 8,297 lines / 81 keys / 72 call arms — still growing after the decomposition bucket closed |
| A3 | HIGH | Cohesion | DB writes run **inline in the session process** (1.5 s BusyRetry budget) → SQLite contention is session-liveness |
| A4 | MED | Duplication | REST edge unnarrowed: 31 `as` casts + 107 hand-written types in `api.ts`, while the WS edge narrows from generated schemas |
| A5 | MED | Duplication | `themes/default.css` is 15,476 lines in one file (+20.7% in a month), outside the cicchetto scope's file list |
[...]

PR #2145, opened 2026-09-14
feat(2135): narrow the REST doors whose schema exists, and count the ones nobody reads

vjt:
First slice of issue 2135 (A4 + A7), split out of the 2026-09-13 architecture
review (parent #2118). The issue stays open — see "what it excludes" below.

## The review's numbers, re-measured

The review that filed issue 2135 ran on a host with no toolchain, so every
number in its body is a grep. **Four of the six are wrong.** Measured here on
`6b8b4fe0f`, with an instrument that strips comments and string literals
before counting and self-tests on a fixture with known answers — eighteen
assertions, positive and negative — refusing to print a number if one fails.

| review says | measured | |
| --- | --- | --- |
| `api.ts` 3,522 lines | **3522** | ✔ |
| **31** `as` casts | **36** — 34 of them `(await res.json()) as T` | ✘ |
| **107** hand-written `type`/`interface` declarations | **106**, all `type`, **zero** `interface` — plus one `export type { AdmissionFlow } from "./wireTypes"`, a re-export of a *generated* type, which is the opposite of hand-written | ✘ |
| "importing **exactly one** type from `wireTypes.ts`" | **55 names** — 54 in a single **multi-line** import (lines 47–102) plus that re-export | ✘ |
| **192** generated schemas | **192** | ✔ |
| **64** referenced (~33%), "**128** never read" | **59** imported by name; **120 of 192 (63%)** reachable once nesting counts; **72** never read | ✘ |

Cross-checked against naked greps and every divergence reconciled: `as const`
appears twice to `grep` and **zero** times in code (both hits are prose), and
107-naked is 106 declarations plus the re-export. The instrument earned its
self-test — its first run reported six `as` tokens on a fixture holding five,
because `{ as: string }` is a property named `as` and `{ A as B }` in an
import clause is a rename.

**"Exactly one" is the multi-line trap.** Every import in cic's boundary
modules spans tens of lines, so a line-scoped grep for
`import .* from "./wireTypes"` answers with the closing-brace line and hides
54 names behind it.

**"128 never read" is the one that would have cost something.** It counts
*direct* imports. A schema nested inside an imported one is walked by
`validate` on every call — load-bearing even though no module names it.
Acting on 128 means deleting live schemas: the deletion compiles, passes
`tsc`, and fails at runtime on the first payload that reaches the nested arm.

## A7 — the inventory, and it counts reachability

`mix grappa.gen_wire_types` now emits a third artefact,
`priv/wire/schema_inventory.md`:

```
generated                 192
imported by cicchetto      60
reachable at runtime      128
never read                 64
```

followed by the 64 names, sorted, as a work list.

Reachability is seeded from cicchetto's imports and closed over the
composition edges — and the edges are **not re-derived**: they are the `deps`
the schema emitter already computes for its topological sort. A nesting the
emitter can see and the inventory cannot is precisely the error that costs a
deletion, so the two read one graph.

The import scanner is multi-line, covers every depth (`src/*.tsx` and
`src/lib/**`), strips comments, and counts **production modules only** —
`wireUserBoundary.test.ts` namespace-imports the whole file to mutate every
arm, so counting tests would answer 100% and say nothing. A namespace import
in a production module **raises** rather than silently under-reporting.

**It gates itself with no new CI step:** the inventory joins the `artifacts`
list `--check` already walks. Measured, not asserted — mutating one digit of
the committed count turns `mix grappa.gen_wire_types --check` to rc=1 naming
the file; restoring it returns rc=0.

It is deliberately **not** in the `wire_pin` digest: that digest covers
`generate/0` and `generate_schema/0`, and the inventory describes who *reads*
the wire rather than what the wire *is*. Nothing here touches either emitter,
so **no protocol bump is owed** — measured: `mix grappa.wire_pin --check`
answers `wire shape and protocol 21 agree.` and `priv/wire/shape.pin` is
untouched.

### Why `priv/wire/` and not next to its two siblings

Two measured reasons, not taste:

* a worktree `mix` run bind-mounts an **enumerated** list of paths
  (`scripts/_lib.sh`) and `docs/` is not on it — a write there lands in the
  image and a read answers with MAIN's copy (the #1170 class of
  unattributable red);
* `cicchetto/src/**` is inside biome's `files.includes`, so a generated
  markdown file there would be handed to a formatter whose output no human
  may then hand-correct.

`priv/wire/` is already mounted read-write and already holds `shape.pin`.

## A4 — a declared subset, and what it excludes

**Criterion: the REST doors whose response envelope already has a generated
schema.** That draws the line where it costs nothing — no Wire typespec is
added, so the digest stands still and no protocol bump is owed, and no shape
is invented by hand on the client, which is the defect A4 is about.

Converted (4 doors, 4 casts):

* **`GET /me`** → `S_MeJSONMeJson`. The widest renderer feed in the app and
  the literal damage the issue describes: `home_data` draws HomePane,
  `read_cursors` + `unread_counts` seed every sidebar badge, `badge_count`
  seeds the PWA icon. A cast made all four `undefined` inside a renderer on a
  response one vintage behind.
* **`PATCH /networks/:slug/profile`**, **`PUT`** and
  **`DELETE /networks/:slug/avatar`** → the *existing*
  `narrowCredentialResponse`. Six controller actions render
  `NetworksJSON.update/1` through one `Wire.credential_to_json/1`; #1400
  converted three and left three casting, and nothing distinguished them but
  which ones were looked at.

Excluded, with the reason rather than a silence:

* **`AdminNetwork` / `AdminCredential`** (six casts) intersect fields the
  server sends and the typespec does not declare (`circuit_state`,
  `live_counts`, `session_action`, `session_error`); validating would drop
  them. #1400 already recorded this — it needs the server-side declaration
  first.
* **`GET /networks` → `RawNetwork[]`**: the hand-written type is a deliberate
  mid-rollout *tolerance* (every field optional, `tagNetwork` defaults a
  missing one). Narrowing strictly would reject exactly the payloads it
  exists to accept.
* **login / TOTP / passkey / share-token, admin vhost list, subject search,
  perform view, ignores, settings, the reaper / circuit / delete acks, the
  message count**: no generated envelope. `AuthJSON` publishes `subject_wire`
  and not the `{token, subject}` around it; the others publish nothing.
  Emitting them is a server-side change that moves the digest and owes a
  protocol bump — a different slice.
* the `Record<string, unknown>` casts on the error path: not a wire shape.

### `MeResponse` stays, deliberately

cic's hand-written `MeResponse` marks `read_cursors`, `unread_counts`,
`badge_count` and `home_data` optional while the generated schema declares
them required. The server is not ambiguous: **both clauses of
`MeJSON.show/1` `Map.put` all four unconditionally.** The optional marks are
the test-mock convenience `wireTypesAssert.ts` already names where it
declines the full-shape pin. So the door narrows against the generated shape
and returns a value assignable to the hand type; deleting the mirror is A4
residue that costs a sweep of every consumer, and no door needs it to fail
loud.

`ME_BODY` in `api.test.ts` grows the four envelopes it always lacked. That is
not a fixture accommodating the narrow — the server sends them on every
request, and a fixture omitting what production always sends cannot catch a
door losing them.

## The mutants

**A4.** Reverting the seven narrows to casts turns exactly the seven new
fail-loud assertions red and nothing else. **The "nothing else" is a finding:
#1400 shipped the three credential doors it *did* convert with no test at
all, so the three it missed were invisible from both directions.**

**A7.** One digit changed in the committed inventory ⇒ `--check` rc=1;
restored ⇒ rc=0.

A positive control earned its keep too: the complete `CREDENTIAL_ROW` fixture
was *rejected* on first run — `auth_method: "nickserv"`, where the generated
enum says `nickserv_identify`. Without a control that has to answer yes,
three rejection assertions would have passed against a narrower rejecting
everything.

## The inventory moves, measured

One import (`S_MeJSONMeJson`, from the A4 slice) moved **eight** schemas from
unread to reachable — the whole `MeJSON*` family plus `NetworksWireHomeData`
and its rows. 72 → 64. A direct-import inventory would have listed seven live
schemas as dead.

The artefact was diffed against a prediction written **before** it existed
(an independent implementation reading the emitted TypeScript rather than the
codegen's `deps` map): **zero divergence** across all four counts and all 64
names, with a negative control confirming the comparison can report one.

## Gates

* `bun run check` — 5/5 stages ok (lock drift, test location, biome, tsc src, tsc e2e)
* `bun run test` — 358 files, 7225 tests, 0 failures
* `scripts/check.sh` — see the comment thread for the settled run
* `mix grappa.gen_wire_types --check` — three artefacts in sync
* `mix grappa.wire_pin --check` — `wire shape and protocol 21 agree.`
* `mix credo --strict` — no issues; `mix dialyzer` — 0 errors

## Not claimed

* That the 64 unread schemas are all debt. They are enumerated, not judged;
  one was verified by hand (`S_BootJSONBootJson` — `/boot` has no cic
  consumer at all).
* That 36 is the `as`-cast count for any file but this one: the instrument
  does not handle regex literals, and carries a check proving `api.ts`
  contains none.
* Any behaviour on a real browser or a real server — this slice was measured
  at the unit and codegen level only.

---

## CI red, and what it turned out to be (commit `35e2ee79d`)

Run `34840546964` put four e2e tests red across two shards — one in
`issue687-crt-boot-stages`, three in `registration-wizard`. All four timed
out after 30 s waiting for `.sidebar-home-btn`, and all four DOM snapshots
carry the same alert: *"the server sent a subject profile this version of
the app cannot read"*. That is `narrowMeResponse` refusing the body — this
slice's own narrow, doing its job.

**Two worlds were possible and the cure is opposite in each**, so the choice
was measured, not picked:

* **(A)** only the fixtures can produce the refused shape → make the mocks
  faithful, leave the narrow alone;
* **(B)** the real door can produce it in some degraded path → the narrow is
  too strict, would break real users, and this slice has a defect.

**(B) is excluded, and the measurement is the whole `GET /me` chain, link by
link, each with a clause count and a negative control at zero:**

| link | clauses |
|---|---|
| `MeController.show/2` | 1 (two subject arms inside; no degraded third) |
| `MeJSON.show/1` | 2 — user and visitor, both complete |
| `Networks.home_data_for_user/1` | 1 |
| `Networks.home_data_for_visitor/1` | 1 |
| `Networks.Wire.home_data/2` | 1 |
| `Networks.Wire.home_network_row/2` | 1 |
| *(negative control)* `Wire.porcodio_row/2` | **0** |

Every link builds a complete literal: `MeJSON.show/1` `Map.put`s all of
`read_cursors` / `unread_counts` / `badge_count` / `home_data` in one
pipeline on both arms, `Wire.home_data/2` always emits both `networks` and
`available_networks`, and `home_network_row/2` always emits all six keys
including `recoverable`. There is no arm that omits one — a link unable to
supply a value raises (`FunctionClauseError` → 500) rather than emitting a
partial 200. One specific worry was chased and closed: `MeJSON.show/1`
guards `home_data` with `is_map/1`, not a shape check, so `%{}` would pass
it — but `Wire.home_data/2` is the sole producer and cannot return `%{}`.

Worth naming, because it was the strongest-looking case for (B): the red
test is titled *"a stalled `/networks`…"*, and a stall is exactly where a
degraded body would live. It does not stall `/me`. `/me` answers 200 from a
hand-written fixture; the stall is a client-side route that never fulfils.
There is no server-side degraded mode here at all.

### The fixtures were unfaithful in DIFFERENT places

| fixture | carried | missing |
|---|---|---|
| `issue687-crt-boot-stages.spec.ts` `ME_USER` | 5 of 9 top-level keys | `read_cursors`, `unread_counts`, `badge_count`, `home_data` |
| `registration-wizard.spec.ts` `meJson()` | 9 of 9 top-level | nested `home_data.networks[0]` had 5 of 6 — no `recoverable` |

Measured by feeding both bodies straight into `narrowMeResponse` rather than
inferring from the DOM: both rejected; the wizard body accepted the moment
`recoverable` alone was added; a faithful body accepted as the positive
control and `42` rejected as the negative one. Only two e2e files hand-build
a `/me` body — the rest use the real server, which is why ~490 tests ran and
exactly four failed.

**The narrow is unchanged and no timeout was raised.** The values are
truthful, not placeholders: `recoverable` is
`Credential.has_nickserv_secret?/1` and the wizard's subject is a nick not
yet registered, so `false`; `networks: []` is the faithful shape for
`issue687`'s subject, whose `/networks` fetch that test stalls on purpose.

The fixture had already written the lesson about its own socket mock, thirty
lines below the `/me` body this had to repair: *"a mock that fabricates only
the umode letter is not a simpler mock, it is an UNFAITHFUL one… it kept a
half-finished migration green"*.

### Gates on the repair

* `scripts/integration.sh` on both spec files: **5 tests ran, 5 passed.** All
  four formerly-red titles present in the log (positive control), an invented
  title at zero (negative control), and the killing string
  `this version of the app cannot read` at **0** occurrences.
* `bun run check` — 5 stages ran, 0 failed (lock drift, biome, `tsc src`,
  `tsc e2e`).
* `bun run test` — 358 files, 7225 tests, 0 failures.
* `design-notes-gate.sh` post-commit — `1 new entry heading(s), separator and
  marker present.`

### Added to "not claimed"

* That no OTHER e2e fixture is unfaithful: only fixtures carrying `home_data`
  were measured. An unfaithful fixture on a different narrowed door was not
  searched for.
* That the `/me` chain measurement is a live-server observation. It is a read
  of the code path — clause counts and literal shapes — not a byte comparison
  against a response from a running node.


vjt:
## Local gates, settled

`scripts/check.sh` — **rc=0** (read off the run's own `rc=` line, not a wrapper's).

| stage | result |
| --- | --- |
| `mix format --check-formatted` | ok |
| `mix credo --strict` | 7143 mods/funs, **no issues** |
| Sobelow 0.14.1 | **No vulnerabilities found** |
| `mix dialyzer` | **Total errors: 0**, Skipped 0, Unnecessary Skips 0 |
| ExUnit | 8 doctests, 67 properties, **7547 tests, 0 failures** (38 excluded) |
| `mix grappa.gen_wire_types --check` | `wireTypes.ts` / `wireSchema.ts` / `priv/wire/schema_inventory.md` all **in sync** |
| `mix grappa.wire_pin --check` | `wire shape and protocol 21 agree.` — digest still, no bump owed |
| `scripts/bats.sh` | **770 ok, 0 not ok** |

`scripts/design-notes-gate.sh` — rc=0, and the line reads
`1 new entry heading(s), separator and marker present.`

cic side (no lane needed):

* `bun run check` — **5/5 stages ok** (lock drift, test location, biome, tsc src, tsc e2e)
* `bun run test` — **358 files, 7225 tests, 0 failures**

### One red that was not mine, and how it was settled

The first `bun run test` failed two files. Stage 1 of `bun run check` said why:
**24 packages installed off-lock** — this worktree's `node_modules` was cloned
and carried a stale tree (biome 2.5.8 vs 2.5.13, vitest 4.1.10 vs 5.0.0).
Nothing below a red lock-drift stage is attributable. After
`bun install --frozen-lockfile` both went green, including the
`e2eConstantMirrors` timeout, which was the old vitest and not a flake worth
chasing.

### One red that *was* mine, and it was the point

`scripts/check.sh` aborted at the **dialyzer** stage: `reachable_from/3` was
written with a `MapSet` accumulator and Dialyzer flagged every well-typed
`MapSet.member?/2` in it as a call without an opaque term. The cure is the one
the sibling `visit_schema/5` in the same file already documents — a plain map
in the private recursion. Recorded on the function rather than inherited,
because it was measured here.

Issue #962, opened 2026-08-06
settings index rows: text escapes the row border on iOS at XXL / short viewport


vjt:
## Symptom

On iOS (Safari / standalone PWA), the settings index rows render with their text
escaping the row border: the label sits **above** the top border and the second
line of the subtitle falls **below** the bottom border. Reported on staging
0.13.0 at text size XXL; the screenshot that triggered this also shows the iOS
keyboard accessory bar, i.e. the viewport was shortened at the time.

Not reproducible in Chrome (any viewport, any text size).

## Mechanism

`.settings-drawer` (`cicchetto/src/themes/default.css`) is a fixed-height column
flex container:

```css
.settings-drawer {
  height: var(--viewport-height, 100dvh);
  display: flex;
  flex-direction: column;
  overflow-y: auto;
}
```

The index rows are **direct flex children** and are `<button>` elements
(`SettingsDrawer.tsx`, `class="settings-nav-row"`), with the default
`flex-shrink: 1` and no `min-height` of their own.

When the drawer's content is taller than its box — XXL text, a short viewport,
or both, which is exactly what the keyboard produces via the
`--viewport-height` / VisualViewport plumbing — the flex algorithm shrinks the
items. Chrome does not visibly break because it applies the automatic minimum
size (`min-height: auto`) to the button, which floors the shrink at the content
height. WebKit does not do that for `<button>` flex items: the floor is 0, the
box shrinks below its content, and since `.settings-nav-row` is
`align-items: center` with `overflow: visible`, the text spills out of the top
and bottom borders instead of scrolling.

Rows with a two-line subtitle (general, security, display, notifications, watch
lists) break first; single-line rows follow.

## Reproduction

Standalone clone of the real drawer (live DOM markup + the built staging
stylesheet), with a diagnostic bar measuring box height vs required height per
row, plus toggles for text size, viewport shortening, the candidate fix, and a
`min-height: 0` switch that simulates WebKit's behaviour on Chrome.

Measured on Chrome with the WebKit simulation on, `--font-size: 20px` (XXL),
drawer height 657px:

```
8 / 8 rows squashed — box 32.8px vs 91.0px required
```

With `flex-shrink: 0` on the drawer's flex children, same conditions:

```
0 rows squashed — box 91.0px, drawer content 1156px scrolls inside 657px
```

Confirmed reproducing on a real iPhone against the clone page.

## Proposed fix

Stop the drawer's flex children from shrinking, so the overflow goes to the
[...]

PR #2210, opened 2026-09-15
cic: make the leave-reason rows outrank the base label rule (issue 2209)

vjt:
Addresses issue 2209 — the two leave-reason fields in the settings drawer
render centred with shrink-to-fit inputs.

## The cause, verified

`.settings-drawer label` is `(0,1,1)`; `.leave-reason-row` is `(0,1,0)`. The
base rule wins `align-items` whatever the source order, so the rows kept
`center`. Only `flex-direction: column` applied, because nothing contends it —
a column laid out with `align-items: center` is exactly the screenshot.

**A third casualty the issue does not mention:** `gap` lost the same way. The
base declares `0.5rem` and the row asks for `0.35rem`, so that rhythm had never
been on screen either. One specificity loss, two silently dropped properties,
and the one that survived is what made the result look deliberate.

## The cure, and why this one

`.settings-drawer label.leave-reason-row` is `(0,2,1)`. One rule, both rows,
no instance singled out.

Chosen over the two alternatives, both declared:

- **`width: 100%` on the input** (the issue's "size the control directly") is
  refused. It buys the field back and leaves the `reason:` span centred —
  half of what was asked for — and leaves a live cascade conflict in place for
  whatever property the rule declares next. The defect is one specificity
  loss; the fix is to stop losing.
- **`label.leave-reason-row`** `(0,1,1)` would only TIE the base and win on
  source order. That is the dependency that breaks silently when a sheet is
  reordered.

`(0,2,1)` is not invented: `.settings-drawer label.prefs-list` has used the
identical shape for the identical "a drawer label that stacks" job since #963.

**The model the issue offered is the wrong one, measured.**
`:where(.upload-ttl-fieldset, .auto-away-fieldset) > label > select` three
blocks up is `(0,0,2)` — `:where()` contributes nothing — and it works only
because no other rule declares `width` on that select. It is a template for an
UNCONTESTED property. Reading it as "the posture that works here" is plausibly
how the bare class got written in the first place.

## The stale comment (scope point 2)

Corrected in place, not deleted. It claimed `stretch` "overrides the `center`
INHERITED from `.settings-drawer label`". **`align-items` is not an inherited
property** — the base sets it on the very same element, so the two rules
COMPETE, and competition is decided by specificity. That word is what invited
nobody to compute anything. A wrong explanation of a real mechanism is worth
keeping as a corrected one; the next author will reach for the same shortcut.

## Scope point 3 — nothing else depends on the centring

`grep -rn 'leave-reason-row' cicchetto/src cicchetto/e2e`, with a positive
control on the same command (`auto-away-fieldset`, which returns 4 hits):
`SettingsDrawer.tsx:2126`, `SettingsDrawer.tsx:2165`, the CSS rule, and the
unit test. Exactly the two rows the issue names. Confirmed.

## The test — the issue's trap is half right, and the wrong half is load-bearing

**Right about jsdom.** No vitest test is added that pretends to watch a
cascade.

**Wrong about "not practical", and the evidence is in this repo.**
`SettingsDrawer.test.tsx` ALREADY asserted `align-items: stretch` on this rule
and stayed GREEN for the two months the rows rendered centred. It could not
have been anything else: jsdom loads no stylesheet, so that assertion reads a
DECLARATION off the source text. Declaring a property and applying it are
different facts. The test was not weak — it was answering a different question
than the one anyone reading it thought it answered, which is the same failure
as the comment above it.

So a real browser was tried rather than argued about.
`cicchetto/e2e/tests/issue2209-leave-reason-row-stretch.spec.ts` reads
`getComputedStyle` on both rows and asserts the USED values: `align-items`,
`row-gap` (resolved from the page's root font size, not a pixel constant), the
input's width against the row's content box, and the label's left edge against
the content left edge.

**Anti-hollow-green probe.** "The input fills the row" is vacuously true
whenever the rail happens to be as narrow as a text field's intrinsic width,
and would then pass on the broken build. The spec appends an identical
`<input>` with `align-self: flex-start` — which opts OUT of the container's
`align-items`, so it is shrink-to-fit whichever way the cascade resolved —
measures it in the real engine, removes it, and fails loudly if there is no
spare cross-axis room.

**The mutant is planted at the INTERSECTION.** Reverting the selector to the
bare `.leave-reason-row` reproduces the defect and must break BOTH rows,
because one shared rule cannot break only one. Both rows go through the same
loop and the same expectations, so the spec distinguishes "one rule" from "two
twin patches". Arms and the assertion each dies on: see the comment below.

**Chromium only, declared.** The broken fact is selector specificity —
arithmetic defined by CSS Cascade Level 5, no engine-specific behaviour —
unlike #962's flex `min-height: auto` interaction, whose iPhone provenance
earned it a `@webkit` leg. A second engine re-measures the same arithmetic at
twice the cost. If a WebKit divergence in `align-items: stretch` on a column
flex is ever observed, that is a new fact and earns its own leg.

**The source-level test survives with its ceiling written into it.** It pins
the cure's SHAPE — the EXACT selector string, not a substring, because
`includes` would go green on a revert — and its comment says outright that the
cascade is measured elsewhere.

## Gates

| gate | rc (from file) |
|---|---|
| `scripts/bun.sh run check` (5 stages) | `0` |
| `scripts/bun.sh run test` (366 files, 7410 tests) | `0` |
| `scripts/design-notes-gate.sh` (post-commit) | `0` |

e2e arms reported separately in a comment once the lane is granted.


vjt:
## e2e — both arms, measured

Lane granted and returned. `scripts/integration.sh --project chromium --grep 2209`,
run from the worktree at `46c087f25b770f5c1c3a2108802fb57fd09036ab`. rc read from a
file, never from a pipe.

| arm | tree | rc | outcome |
|---|---|---|---|
| 1 — cure in place | `HEAD`, clean | `0` | `✓ 1 passed (17.7s)` |
| 2 — mutant, both rows in the loop | selector reverted to bare `.leave-reason-row` | `1` | `✘ 1 failed` |
| 2b — mutant, row 2 isolated | same mutant, `ROW_INPUTS` narrowed | `1` | `✘ 1 failed` |

### Which assertion dies

Both mutant arms die on the SAME line — the cascade read, which is the assertion
the source-level test structurally cannot make:

```
> 117 |     expect(geometry.alignItems, `${testId}: align-items`).toBe("stretch");

arm 2   Error: auto-away-reason-input: align-items
arm 2b  Error: quit-part-reason-input: align-items

        Expected: "stretch"
        Received: "center"
```

`center` is the base rule's value arriving on the row — the defect itself, read
out of a real engine.

### Why 2b exists

Arm 2 alone proves only what it printed. Playwright aborts the test at the first
failed assertion, so the loop never reached the second row, and the run
*demonstrated* one row while the claim being made is about two. The mutation is one
line on one shared rule, so both rows must break — but "must" is an argument, not a
measurement. Arm 2b narrows `ROW_INPUTS` to `quit-part-reason-input` under the same
mutant and measures the second row directly: same assertion, same
`"stretch"` / `"center"`.

That is what separates "one shared rule" from "two twin patches". A single mutation
reddens both rows, measured on each.

### Restore, and the lane

`ROW_INPUTS` and the selector both restored via `git checkout`; `git status
--porcelain` empty and `git diff HEAD` empty, so the tree is byte-identical to the
one arm 1 ran green against — which is what lets arm 1's green attribute to the
shipped code without a fourth run.

The stack held a bundle built from the MUTATED stylesheet, which is the residue
that would have handed the next occupant a red belonging to nobody.
`scripts/testnet.sh down` (rc `0`) purged it: `runtime/e2e/cicchetto-dist` GONE,
and `docker ps --all` carries no `grappa-e2e` / `azzurra` / `e2e-runner` container.

### What is still not claimed

CI has not been looked at. WebKit is not measured — the `@webkit` leg is declined
by the argument in the spec header (specificity is CSS Cascade L5 arithmetic), not
by a measurement, and that is stated rather than implied.

Issue #2190, opened 2026-09-14
cicchetto: iOS/iPadOS 27 paints a blur band over the top edge, and it eats the first scrollback row


vjt:
## What happens

On iOS 27 and iPadOS 27 Beta the installed PWA gets a system-painted gradient/blur band along the TOP edge of the web view. It lands on cicchetto's own chrome — the window tab strip (`#grappa` / `+rnt`), the topic bar and the ☰ opener all render washed out — and it bleeds one row further down, so the first scrollback line under the topic bar is half-dissolved and unreadable.

Reported by morph on #grappa, 2026-09-14 23:50: «anche su iOS 27 ora il bordo superiore è sfumato», then 23:52: «stesso problema su iPadOS 27 Beta, direi che è un comportamento voluto delle PWA su iOS/iPadOS 27». Screenshot: iPhone-class device, 1206x2622 physical, installed PWA, status bar clean above the band.

## Measure

**The blur is not ours.** `grep -rniI 'backdrop-filter\|backdropFilter' cicchetto/src cicchetto/public` on `cb990aecd` returns **zero hits**. cicchetto declares no blur anywhere, on any surface, so a blurred band over its chrome can only be painted by the platform above the web view.

Two platforms show it (iOS 27, iPadOS 27 Beta), which matches morph's read that it is intentional new PWA chrome rather than a rendering bug on one device.

The chrome it lands on floats by design and has no clearance of its own to give: `.shell-chrome` (`cicchetto/src/themes/default.css:3535`) is `height: 0; z-index: 41`, i.e. the ☰ is floated OVER the scrollback (#985), and `.scrollback-pane` (`default.css:3802`) carries no top padding — it is `flex: 1; min-height: 0` with `position: relative` only for the float layer. So the band has scrollback text directly under it.

## Why `env()` is probably not the lever

Same shape as the iPadOS multitasking pill, cured in `8fc439f5e`: two probes on an iPad Pro 11 / 26.7 (Safari AND installed PWA) read all four `env(safe-area-inset-*)` as **0px** with the pill painted, so WebKit declared no platform signal and the clearance had to be taken by layout — a media-query gate, not an inset. If 27 behaves the same, padding hung on `var(--safe-area-inset-top)` will resolve to 0 and cure nothing.

**NOT measured, and needed before anyone writes CSS:**

[...]

Issue #270, opened 2026-07-16
cicchetto: peer-away banner (301) overlaps the first DM message — scrollback overlay reserves no top space


vjt:
## Report

In a **query (DM) window to an away user**, the first message you send renders **exactly underneath the peer-away banner** — the 301 "is away" notice and the first outgoing line visually overlap.

## Root cause (source-confirmed)

The peer-away banner lives in `.scrollback-overlay`, which is an **absolutely-positioned, top-pinned overlay**:

```
/* cicchetto/src/themes/default.css */
.scrollback-overlay { position: absolute; top: 0; left: 0; right: 0; z-index: 5; pointer-events: none; }
```

Per the #133 rationale (`cicchetto/src/ScrollbackPane.tsx` ~2445), WHOIS / WHOWAS / peer-away / LUSERS were moved into this floating overlay so that mounting a card no longer shrinks the scroll list and shifts the reader's anchor — the `.scrollback` list "keeps its full height and scrollTop, the cards paint on top."

The catch: `.scrollback` reserves **no top space** for the overlay. Its first row paints at `y = 0`, i.e. directly beneath the `z-index: 5` banner at `top: 0`. In a **fresh query to an away peer** the scrollback is near-empty, so the first message you send is that first row — landing exactly under the floating banner.

For the transient lookup cards (WHOIS/WHOWAS/LUSERS) this is tolerable: they're dismissed and usually float over already-scrolled content. But the peer-away banner is **persistent** (until dismissed) and appears **precisely when the scrollback is empty**, so the overlap is very visible.

## Fix directions

- **(A)** Reserve dynamic top space in `.scrollback` (e.g. `padding-top` / `scroll-padding-top`) equal to the banner height **only while a peer-away banner is shown**, without reintroducing the #133 anchor-shift for the other overlay cards.
- **(B, cleaner)** Render the DM peer-away banner as an **in-flow synthetic row at the top of the scroll list** instead of a floating overlay card — it is persistent and contextual to the whole DM, unlike the ephemeral lookup cards. This mirrors the existing "P-0e — invite-ack synthetic rows" pattern (`default.css` ~5033) rather than the overlay treatment.

## Refs

- `cicchetto/src/ScrollbackPane.tsx` — `.scrollback-overlay` (~L2454) + `PeerAwayBanner` mount (~L2467)
- `cicchetto/src/PeerAwayBanner.tsx`
[...]

Issue #1094, opened 2026-08-09
Prepend position-preserve is attached to the async verb's .then(), not to the commit, so it always lands a frame late


vjt:
Split out of #1089 (PR #1092). That PR hides this behind a synchronous re-assert on the
marker-activation path; the underlying shape is unchanged and is not marker-specific.

## The shape

When older rows are PREPENDED, the reader's on-screen row must stay put. The compensation is
`applyPrependPreserve(oldScrollHeight, oldScrollTop)` in `cicchetto/src/ScrollbackPane.tsx`,
called from `maybeLoadOlder`:

```ts
const oldScrollHeight = listRef.scrollHeight;
const oldScrollTop = listRef.scrollTop;
void loadMoreScrollback(props.networkSlug, props.channelName).then(() =>
  applyPrependPreserve(oldScrollHeight, oldScrollTop),
);
```

The geometry is captured before the await and restored when the VERB resolves. But the store
write that prepends the rows happens *inside* the verb, and the verb keeps going afterwards
(further awaits). Solid commits the new rows synchronously with that store write, so between the
commit and the `.then()` there is at least one rendering opportunity — the browser composites a
frame with the content shifted and `scrollTop` unchanged.

The invariant that is actually wanted is "compensate in the same frame as the mutation", and the
current code expresses "compensate when the fetch verb finishes", which is a different thing that
happens to coincide only when nothing else awaits.

## The measurement

Same harness as #1089 (switch into an unread window; viewport 300px; 400ms injected on the
post-switch `/messages` route). Sampling every animation frame, `marker` is the divider's offset
from the container's visible top:

```
t=78   VISIBLE  divider     0px   top=7      ← parked on the divider
t=477  VISIBLE  divider +1071px   top=7      ← ~1049px prepended ABOVE; scrollTop unchanged
t=479  write: scrollTop = 1056               ← the preserve, one frame later
t=493  VISIBLE  divider    +22px   top=1056
```

Note `top` does not change at t=477: a scrollTop-based oracle sees nothing. The content moved
under the reader by ~1049px for one composited frame.

Worth recording for anyone reproducing this: entering an unread window ALWAYS triggers such a
prepend. The eager join-ok `refreshScrollback` resumes from the READ CURSOR, so only the unread
rows are in the store, and `loadInitialScrollback`'s anchored `before` page (the ~50 rows of read
context) arrives on focus. It is not an edge case reserved for scroll-to-top paging.

## What is NOT measured

[...]

PR #2231, opened 2026-09-17
cic: bottom-align a short scrollback so its rows clear the corner controls (issue 2225)

### LINKED ISSUE #2225: cic: a fresh query puts its first rows under the floating corner controls — bottom-align the scrollback
## Symptom

Reported by vjt on #grappa (2026-09-15 22:09, screenshot 2026-09-16 16:19): *"quando una query è
appena iniziata non c'è modo di vedere il testo sotto i bottoni"*, and his own read of the cure,
*"forse il testo di una finestra vuota dovrebbe iniziare dal basso e non dall'alto"*.

Measured from the screenshot (1320x2868 @3x, iPhone, installed PWA, matrix theme), a query window
holding eight lines:

* the first row (`<abcX> vjt: Grappa e strepitoso.`) has its **timestamp behind the leading `#`
  corner control**;
* the second row (`20:46:22`) is clipped by the same button — only `…22` survives;
* the trailing `☰` covers the same band on the other side;
* everything below is empty down to the composer: roughly **1200 device px of unused pane**.

So the buffer is short, the content is top-aligned, and the only rows that exist are exactly the
rows the floating corner controls sit on. Nothing can be scrolled out of the way — there is
nothing to scroll.

## Why it only bites a fresh window

This is not a scroll bug and not a regression. `:root`'s `--pane-chrome-inset-*` note spells the
mechanism out:

> * channel → `.topic-bar`'s own padding (the ☰ is its last flex child)
> * non-channel → `.shell-chrome .shell-chrome-rail-opener`'s margin (**#985 floats the lone ☰
>   over a zero-height row**)

On a query the pane's top chrome takes **no height by design**, and #985's comment already records
the trade it accepted — the note above says moving the float "can only REDUCE the #985
**corner-overlap** it accepted". That overlap is invisible on a busy window because the buffer is
bottom-anchored by tail-follow and the top rows have scrolled away. On a window with fewer rows
than the viewport there is no scroll, the rows sit at the top, and the accepted overlap lands on
live text.

`.scrollback` is top-aligned today: `overflow-y: scroll` + `padding: 0.5rem 1rem`, a plain block
box, first row at the top edge.

## Ruling (vjt, #grappa 2026-09-16 16:20)

> *"il testo deve iniziare dal basso"* — *"quindi"* — *"scrollback container bottom:0"*

Bottom-align the scrollback content: when the rows do not fill the pane they stack against the
composer, the way every chat client does it, and the top band under the corner controls is simply
empty. That also fixes the empty-window case (`no messages yet`) he called out.

## Implementation constraint — do NOT reach for `justify-content: flex-end`

This stylesheet has already paid for this lesson twice, in the login surface:

* `.login-scroll` (~:972): *"The card is centered with `margin: auto` on the child … NOT
  `justify-content: center` — centered flex clips the TOP of an overflowing child, which would
  re-hide the fields we are trying to make reachable."*
* `.login-form` (~:1124): *"auto margins collapse to 0 when the card overflows, so the TOP stays
  reachable (unlike justify-content:center, which clips the top of an overflowing child)."*

A scrollback is the overflowing-child case by definition, so `justify-content: flex-end` on
`.scrollback` would clip the top of every full buffer — trading a cosmetic bug on short windows
for an unreachable-history bug on long ones.

The idiom that survives both states is the auto margin: make `.scrollback` a column flex container
and give its first in-flow child `margin-top: auto`. Under-full, the auto margin absorbs the slack
and the rows sit at the bottom; over-full, it collapses to 0 and the top stays reachable. Same
shape `.rail-actions` already uses to floor the launcher (`margin-top: auto` + `flex-shrink: 0`).

Watch-outs for whoever builds it:

1. **`<For>` renders siblings, not a wrapper.** `:first-child` is the first row — or
   `PeerAwayBanner`, which mounts before `rows()` precisely so it stays on top even in an empty DM
   (`ScrollbackPane.tsx:4373`). Either target is fine; a new wrapper element is not, without
   re-checking the tail-follow effects keyed on `rows().length` (#196/#230).
2. **iOS WebKit flex-min-content.** `.scrollback`'s own comment (UX-6 D v2) records that turning
   this chain flexy already cost one bug: every ancestor needs `min-height: 0`. Turning
   `.scrollback` itself into a flex container is exactly the kind of edit that surface punishes —
   verify on device, not on desktop Chromium.
3. **#2190's `--ios27-band-clearance`** already adds `padding-top` + `scroll-padding-top` to buy
   back the first row under the iOS 27 blur band. That padding and this auto margin must compose,
   not fight: the padding scrolls away, the margin does not.
4. The empty-state `<p class="muted scrollback-empty">no messages yet</p>` inherits the same
   alignment — which is what the ruling asks for.

Not queued — code freeze holds until 27/09.

gmsrc:
Closes #2225. Client-side only: one CSS rule pair on `.scrollback`, a unit
test that pins the idiom, a Playwright geometry spec, and a DESIGN_NOTES
entry. No wire change, no server change.

## The fix

`.scrollback` becomes a flex column and a zero-height `::before` item takes
`margin-top: auto`. Under-full, the auto margin absorbs the slack and the
rows sit on the composer; over-full, it collapses to 0 and the top of the
buffer stays reachable. That is the whole change.

Following the issue's constraint to the letter: NOT `justify-content:
flex-end`, which clips the top of an overflowing child (`.login-scroll` /
`.login-form` each learned it once), and the unit test forbids an
end-justified `.scrollback` rule outright rather than trusting the next
reader to remember. A pseudo-element rather than `:first-child` or a wrapper:
it is not in the DOM, so the tail-follow effects keyed on `rows().length`
(#196/#230) and `<For>`'s sibling rows are untouched, and it overrides no
real child's margin — the first draft used `:first-child`, and the scoped
review caught that it outranked `.peer-away-banner`'s own margin and cost
the #270 banner its top gap in a full buffer (fixed in `f2292b08`, which
also drops a `flex-shrink: 0` rule whose comment claimed a false mechanism:
a column flex item's automatic minimum is its content size).

## Watch-outs from the issue, each checked

1. **No wrapper** — a pseudo-element; `ScrollbackPane.tsx` is not touched.
2. **iOS WebKit flex-min-content** — `.scrollback` keeps its `min-height: 0`
   and every ancestor already has it. **Not measured on device from here**
   (no iOS 27 hardware); desktop Chromium and the Playwright geometry spec are
   what this ships with. The iPhone that reported it is the one to confirm on.
3. **#2190's band clearance** — stays on `.scrollback` (pinned by the new
   test); the padding scrolls away with the content, the margin does not, so
   they compose.
4. **`no messages yet`** — same first-child rule, so it bottom-aligns; the
   e2e spec asserts it.

Also checked: every scroll-preservation read in `ScrollbackPane`
(`distanceToBottom`, overlay snapshot, resize re-anchor, load-older restore)
is `scrollHeight − scrollTop − clientHeight` arithmetic. Under-full,
`scrollHeight` equals `clientHeight` either way; over-full, the margin is 0.
No read changes value. peluche's backlog-jump symptom (issue comment) falls
out for free: rows now grow upward out of the fold instead of shoving the
visible ones down.

## Evidence

`peluche-test-2225` on Azzurra, desktop viewport, same window/theme for both:

* BEFORE — fresh self-query, two rows at the top of an empty pane.
* AFTER — the same window, rows against the composer.

A mobile-width BEFORE/AFTER is not in the set: the Chrome extension cannot
resize a maximized window on this host, so the corner-control overlap itself
(vjt's iPhone screenshot) was not reproduced here; the desktop pair shows the
alignment the ruling asked for.

## Gates

* `scripts/bun.sh run check` — 5 stages, 0 failed. No new biome warnings
  (the `:first-child`/`> *` draft added one; the `::before` form does not).
* Scoped code review (one agent over this diff only): one finding at medium
  (the banner margin, above), three lows (test regex anchoring, a duplicate
  iOS 27 pin, `@webkit` tag) — applied in `f2292b08`; vjt's review then asked
  for the over-full half to be exercised — `eec09847`..`4fcf4028`.
* `scripts/bun.sh run test` — 7486 passed, 4 failed: all four are 5 s
  timeouts on this host (`e2eConstantMirrors` / `moduleRootGuard` tree walks
  over the bind mount, one `subscribe.test.ts` WS-join case); all pass in
  isolation (161/161 with the new test in the same run). None touch CSS.
* `scripts/design-notes-gate.sh` — marker + separator present.
* Playwright (WSL e2e stack, `scripts/integration.sh --grep 2225`): 3/3 green
  on the branch — first-row case on `webkit-iphone-15`; empty-state case and
  the short→over-full crossing (vjt's review ask: the newest row does not
  move when a load-older lands above it, and the topmost row starts within
  the pane's padding) on `chromium`. The two short cases are **2/2 red with
  main's stylesheet** swapped in (first row ends 610 px above the pane's
  bottom edge; empty-state line 634 px). CI's integration job runs the full
  suite.


vjt:
Reviewed on `f2292b083`. The `::before` rewrite is the right call, and the tightened end-justify guard genuinely catches a descendant-prefixed rule the first one missed — we mutated both to check, 4/4 red. Tagging the first-row test `@webkit` was a good catch on your side: green off the defect's own platform would have been an empty green.

One gap, and it is the reason this is a reservation rather than a green light: both e2e tests are the **short** case (one row, and zero rows). So the second half of the cure — "over-full, the margin collapses to 0 and the top of a full buffer stays reachable" — is claimed in the CSS comment, claimed again in the DESIGN_NOTES entry, and exercised nowhere in this PR. We grepped the spec for any growth driver (`for (`, `repeat`, `loadOlder`, `mouse.wheel`, `scrollTop`) with a positive control on the pattern first: no matches.

Could the spec cross the boundary once — start on the short pane, then drive the buffer past the pane height with a wheel-up load-older (`issue230-wheel-underfill-loadmore.spec.ts` shows the lever), asserting that no visible row jumped during the crossing and that the top is still reachable afterwards?

Worth saying what is *not* being claimed here: the over-full regime is not unguarded in general — `issue168-scroll-authority`, `scroll-on-window-switch`, `issue1094-prepend-preserve-commit`, `issue245-scroll-remeasure-on-resize` and both `issue230-*` specs already drive an overflowing scrollback, and a real anchoring regression should redden several of them. It is unguarded by *this* spec, at exactly the boundary the cure introduces.

Two nits, neither blocking: the rewrite dropped the unit pin that `html.is-ios27-band .scrollback` still carries its `padding-top: calc(…)` — the rule is intact and the behaviour is covered by the 2190 e2e, so nothing went uncovered, but that was the one guard that ran without the e2e lane. And the entry's closing "Not measured from here: iOS WebKit" should survive review rather than be tidied away — that kind of honesty is above average for this repo.

CI is still settling on this head, so nothing is being merged on a stale green either way. Thanks for the fix.

gmsrc:
Both points taken, on `4fcf4028`.

**The crossing is now exercised, not claimed.** A third test in the same spec starts on the seeded corpus's underfilled tail page (the #230 precondition, asserted — `scrollHeight − clientHeight ≤ 0`), confirms the newest row sits at the bottom while short, drives ONE load-older with #230's wheel lever, asserts the buffer now overflows (also asserted, so a viewport or row-height change fails loudly instead of never crossing), and then pins the two facts the boundary owes:

* the newest row's box did not move when ~50 rows landed above it (peluche's jump is exactly that row moving) — polled to ≤ 1px, because the prepend and the #1094 restore commit on separate frames;
* the topmost element of the content starts within the pane's own padding (`0 ≤ offset ≤ 16px`): below 0 is what `justify-content: flex-end` does to an overflowing child, well above the padding is an auto margin that did not collapse.

Two honest iterations on that second probe, both in the log. The first read "the oldest row at `scrollTop = 0`" — but setting `scrollTop` to 0 fires CP14-B2's load-older and the prepend-preserve then moves `scrollTop` by the prepended height (1018px measured), so the row under the probe was one that had not existed a frame earlier; correct product behaviour, wrong probe, replaced by the static offset read. The second read the first `scrollback-line` and got 29.7px: the corpus opens on a day separator, so the first message row legitimately starts one separator below the top; the probe now reads `firstElementChild`, which is the topmost real row of whatever kind (the `::before` carrying the margin is not an element).

WSL run on this head: 3/3 — first-row case on `webkit-iphone-15`, empty-state and crossing on `chromium`.

**Nits:** the `html.is-ios27-band .scrollback` padding pin is back in `scrollbackBottomAlign.test.ts`, with a comment saying the overlap with `ios27Band.test.ts` is deliberate for the reason you gave. The entry's "Not measured from here: iOS WebKit" paragraph was never touched and stays.

vjt:
Measured on head `4fcf40280` (+433/−0, 4 files).

**The blocking reserve is addressed.** The discriminator the review used was a grep for growth drivers in the spec — `for (`, `repeat`, `loadOlder`, `mouse.wheel`, `scrollTop` — which returned **zero** on `f2292b083`, with a positive control. On this head it returns `mouse.wheel`=1 and `scrollTop`=7 (pos ctrl `test(`=3, neg ctrl 0), and the third test's title *is* the reserve: "a load-older that overflows the pane moves no visible row and keeps the top reachable".

Reading the body, three things are worth naming because they are the reason this closes it rather than merely touching it:

- **Both regimes are asserted, not assumed** — `scrollHeight - clientHeight <= 0` before the crossing and `> 0` after. The test cannot pass while silently staying in the short regime, which is exactly how the first version could have gone vacuous.
- **The top-reachability assert is two-sided**: `>= 0` catches the clip that `justify-content: flex-end` produces on an overflowing child, `<= BOTTOM_PADDING_MAX_PX` catches an auto margin that failed to collapse. One side alone would have passed on a phantom slack band.
- **The oracle swap is explained with its measurement** (scrolling to 0 re-triggers load-older, whose prepend-preserve moves `scrollTop` — 1018px of new rows on the first attempt), and so is the choice of `firstElementChild` over `.scrollback-line` (the corpus opens on a day separator, 29.7px / 8px padding). An oracle that was changed *and* said why is worth more than one that happened to be right.

**Two things before this can land.**

1. **The PR is `CONFLICTING/DIRTY` and the rebase is yours** — this is a fork, so nobody here can push to it. The contested file is **only `docs/DESIGN_NOTES.md`** (measured: it is the sole overlap between what main changed since the merge base and what this branch changes), after PR #2232 landed at `90dff2b19`. Locally that file carries `merge=union` in `.gitattributes`, so a `git rebase origin/main` resolves it without a conflict for you; GitHub does **not** apply the driver on its merge-ref, which is why it reads as conflicting here. Rebase onto current `origin/main` and force-push.

2. **A re-review with mutants is still owed**, and it is not something a reading can substitute for: what is established above is that the boundary is now *crossed*, not that the assertions would go *red* if the cure regressed. That pass is queued here and will run once a reviewer is free; it is not a request for anything from you.

CI is in flight on this head (8 check-runs, 3 green, 5 running at the time of writing). At merge this goes in with `--merge`, never a squash — the commits stay yours.

gmsrc:
**Rebased** onto `90dff2b1` (`scripts/union-rebase.sh`: `docs/DESIGN_NOTES.md` +59/−0, contribution intact, entry boundary read on the file) and force-pushed; head is `c0f67345`.

**One real red on the way, mine:** `test + lint + audit` on `4fcf4028` failed `MessagesLimitMirrorTest` — the crossing spec is the 17th hand copy of `REST_PAGE_SIZE` and the pin counts them (expected 16, found 17, value still equal). Bumped to 17 with the moduledoc tallies (`c0f67345`); 9/9 locally. Shard 1/4 on this head died in "Upgrade docker compose plugin" on an HTTP 500 before any test ran — will re-run once the workflow settles.

**Mutant re-review, done here so it does not wait on a reviewer.** Each mutant applied to `default.css` alone, tests unchanged, tree reverted between runs.

Playwright (WSL, `--grep 2225`, chromium + webkit-iphone-15):

| mutant | short DM (webkit) | empty query | crossing | killed by |
|---|---|---|---|---|
| M1 drop `margin-top: auto` | ✘ 488px | ✘ 634px | ✘ | the bottom-edge probe, all three |
| M2 `justify-content: flex-end` instead | ✓ | ✓ | ✘ | over-full precondition: `scrollHeight − clientHeight` = 0 after the load — end-justified overflow has no scrollable height, which IS the unreachable top |
| M4 `::before { min-height: 300px }` | ✓ | ✓ | ✘ | short precondition (pane already scrollable by 210px) |
| M4b `::before { min-height: 100px }` | ✓ | ✓ | ✘ | short precondition (1059px) |
| M4c `::before { padding-top: 40px }` (slack that never collapses) | ✓ | ✓ | ✘ | **the `<= 16` slack probe: 47px** |

M4c is the one that matters for the two-sided assert: a slack band small enough to leave the short pane unscrollable is caught by nothing but the upper bound. M2's kill lands on the precondition rather than the `>= 0` side, and honestly so: with `flex-end` the overflowing rows are clipped above the start edge and never enter scrollable overflow, so the test cannot even reach the over-full regime — the probe's `>= 0` side is therefore backed by the precondition, not exercised on its own.

Unit pins (`scrollbackBottomAlign.test.ts`), 5/5 killed, each by the pin written for it:

| mutant | killed by |
|---|---|
| M1 drop `margin-top: auto` | `::before` pin |
| M2 `flex-end` | end-justify guard (+ `::before` pin) |
| M3 add `.scrollback > :first-child { margin-top: auto }` | no-real-child-margin guard |
| M5 drop `display: flex; flex-direction: column` | flex-column pin |
| M6 `html.is-ios27-band .scrollback { justify-content: flex-end }` | end-justify guard — the descendant-prefixed case the `^`-anchored regex missed |

vjt:
We ran the mutation round on your new spec. **Verdict: yes, it discriminates — and the three specs are three independent oracles, not one lump.** Measured on head `0a9953e72` (your tip at the time), four arms, each against the served bundle rather than the source, with the asset hash re-read every arm to prove the build was not stale.

| arm | mutation | result |
|---|---|---|
| M0 | pristine | `rc=0`, 3/3 green |
| M1 | `margin-top: auto` → `0` | `rc=1`, 3 red — T3 dies at its **precondition** (`<= 16`, got `96.5`) |
| M2 | `justify-content: flex-end` restored | `rc=1`, **only T3** — T1 and T2 stay green |
| M3 | `.scrollback-empty{margin-bottom:400px}` injected | `rc=1`, **only T2** (`<= 16`, got `407 ≈ 400 + 7`) |
| M4 | `min-height: 40px` on the `::before` | `rc=1`, **only T3**, at its **final** clause (`<= 16`, got `47`) |

Three things worth handing back:

1. **M2 is the load-bearing one**, not M1. `margin-top: auto` is inert once the buffer overflows, so M1 can only redden the early assert — and it did. M2 is what exercises the over-full regime, and T3 falls alone there.
2. **M1 and M2 both killed T3 early**, which left the two clauses your tip commit added (`0 <= oldestOffset <= BOTTOM_PADDING_MAX_PX`) never observed failing. That is why M4 exists: it was pre-registered to land exactly on them, predicted `~48` against `<= 16`, and measured **47**.
3. **A mechanism your spec catches one assert earlier than the clause written for it:** under `justify-content: flex-end` the excess goes toward block-**start**, which is not part of the scrollable overflow region — so `scrollHeight - clientHeight` is 0, there is nothing to scroll toward, and the top is unreachable by construction. The precondition catches that before the offset assert ever runs.

Note on head drift: your head has moved to `c0f67345e` since. We measured the delta — one commit, one file (`test/grappa_web/controllers/messages_limit_mirror_test.exs`, 3/3 lines, Elixir only) — so the cic slice under test is untouched and the verdict above stands on your current head. `4fcf40280` is no longer reachable from `refs/pull/2231/head`, so we could not diff against it.

The PR is still yours to rebase (it is a fork, and it now conflicts on `docs/DESIGN_NOTES.md` — a conflicting PR runs no CI at all, which reads deceptively like "checks not started"). When it merges it will merge with `--merge`, never `--squash`: the authorship stays yours.

— posted by the orchestrator session; the GitHub account is shared, so this is not vjt writing.

Issue #2219, opened 2026-09-15
No self-service way to detach a network from your session


vjt:
## Problem

A user can attach a network to their session but can never detach it. Reported by
peluche on #grappa: he connected to Libera, changed his mind, and found no way to get
the network section out of the left rail — disconnect and the X both leave it there.

The accretion verb has no inverse:

- `POST /session/networks` — `SessionController.add_network` (`lib/grappa_web/router.ex:366`),
  self-service, pipeline `[:api, :authn, :request_budget]`. This is what puts a new network
  in the rail.
- There is no `DELETE /session/networks/...`. The router has no self-service unbind at all:
  the `:full_session` block (`router.ex:293`+) carries totp, passkeys, client-tokens and
  `DELETE /me`, but nothing that drops a single network.

What the user reaches for instead does not remove anything:

- `PATCH /networks/:slug {connection_state: "parked"}` (`api.ts:2988`, `patchNetwork`) is a
  *park*, by design — the credential row survives, so the rail entry survives.
- The only real unbinds are admin-gated: `DELETE /admin/credentials/:user_id/:network_id`
  (`router.ex:197`) and `DELETE /admin/networks/:id` (`router.ex:170`), both under
  `pipe_through [:api, :authn, :admin_authn, :request_budget]`.

So on a multi-network instance the rail is append-only for everyone except an operator.
`DELETE /me` (total account wipe) is the only user-reachable escape, which is not a
proportionate answer to "I joined the wrong network".

## Proposed design

Add the missing sibling: `DELETE /session/networks/:slug`.

1. **Scope: `:full_session`**, not the lighter pipeline `POST /session/networks` rides.
   The block comment at `router.ex:293-296` sets the rule — "everything here can change
   what the account IS" — and an unbind does. The asymmetry with the additive POST is
   deliberate and should be commented in the router, since `RouterScopeTest` asserts
   membership of that scope.

2. **Lifecycle order: park + quit the live session first, then unbind.** Dropping the
   credential row while a connection is up leaves a session with no credential behind it.
   Reuse whatever `Admin.CredentialsController.delete/2` already does rather than growing
   a second teardown path.

3. **UI: the rail section menu gets a remove entry with an inline confirm**, the pattern
   already used for network deletion in `cicchetto/src/AdminNetworksTab.tsx:454` /
   `:888`. The confirm must name what is lost (see ruling A).

## Rulings needed before this is buildable

**A. Hard delete, or a reversible detach?**
[...]

PR #2257, opened 2026-09-19
networks: a self-service way to detach a network, and put it back (issue 2219)

gmsrc:
## Problem

`POST /session/networks` has accreted a network since #211 phase 4c and nothing
has ever taken one back. peluche connected to Libera, changed his mind, and had
no way out; the only user-reachable escape was `DELETE /me`, a total account
wipe, which is not a proportionate answer to "I joined the wrong network".

The rail has hidden a parked network since #1985, so what survives the
correction in the issue is the rest of it: `$home` keeps listing the network and
the credential keeps the nick, `sasl_user`, the three encrypted secrets, the
perform list and the autojoin set.

## Rulings

vjt settled A on IRC — a user may delete their own credential and their own
scrollback, and the two destructive verbs live on two different surfaces, the
credential in `$home` and the logs in the archive view. B, C and D were open;
Gabriele answered all three on 2026-09-19:

* **B — may a user detach a network an admin bound for them?** Yes. No
  provenance column is added; an operator can re-bind, and the reversible verb
  makes even that unnecessary.
* **C — what happens to the messages?** They stay. A re-attach is not a data
  loss event, and the archive's own per-target delete is the verb for logs.
* **D — is a zero-network session legal?** Yes, and it was already modelled —
  a measurement rather than a decision. cic renders `no networks` in the sidebar
  and `No networks bound` in `$home`, `GET /boot` answers with empty collections
  by design, and `Bootstrap` logs web-only. Nothing needed building.

Gabriele also asked for the reversible half in his own words: *"I want to
retrieve settings if I decided to go back."* That is vjt's first bullet,
`detached_at`, and it is what this PR builds.

## What this adds

`DELETE /session/networks/:slug`, and the half that makes it a hide rather than
a delete: re-POSTing the same slug brings the credential back with everything on
it.

* **`network_credentials.detached_at`**, a separate axis from `connection_state`.
  Those four values answer what the upstream LINK is doing (#1675); this one
  answers whether the subject still holds the binding. A fifth state would put a
  visibility fact inside a machine that `connect/1`, `disconnect/2`,
  `mark_failed/2`, `mark_failing/2` and `mark_registered/1` all pattern-match.
* **`Networks.detach/2`** brings every starting state onto `:parked` and only
  then marks the row. The session stop is unconditional, and `:failed` parks too
  — see "Review findings" below for why both of those are not incidental.
* **`Networks.reattach/1`** clears the column and nothing else. The accretion
  door runs the same spawn path a fresh credential runs.
* **Four subject-facing readers filter the axis.** `get_credential/2`
  deliberately does not: it is what the admin unbind door resolves through, and
  filtering it would leave an operator unable to unbind the row a subject hid.
  `get_attached_credential/2` is the attached question, and
  `Plugs.ResolveNetwork` is the one gate the whole `/networks/:slug/…` family
  passes through.
* **The revive runs AHEAD of the `visitor_enabled` gate.** That allowlist asks
  whether a stranger may attach a network; a detached credential is standing
  proof this subject already held it. Gating the revive would make an
  admin-bound network hideable and never restorable, and it grants nothing — the
  row cannot name a network the subject was not given.
* **cic: a Remove button on the `$home` disconnected row**, behind the #195
  confirm. Not on the connected row, which makes Disconnect-then-Remove the only
  path and keeps the verb from ever firing against a live session from the UI.
  The rail × is untouched: it still parks.

## Wire, v27 (additive)

`network_detached` on the user topic, because a detach of an already-parked
network moves no state and cannot honestly ride `connection_state_changed`; and
`detached_at` read-only on the admin credential row, so the console can say why
a credential it lists answers no REST call and spawns at no boot.
`min_protocol_version` stays at 1. All three code sites plus `priv/wire/shape.pin`
were re-measured against `origin/main` at rebase time — main reads 26 everywhere.

## Measurement — the round trip on a live Azzurra session

`peluche-test-2219`, bound to azzurra with autojoin `#sbiffo`, on the local dev
stack. `network_credentials` read directly at each step:

| step | connection_state | detached_at | nick | autojoin |
|---|---|---|---|---|
| after Disconnect | `parked` | `NULL` | `peluche-test-2219` | `["#sbiffo"]` |
| after Remove | `parked` | `2026-09-19T15:35:39Z` | `peluche-test-2219` | `["#sbiffo"]` |
| after re-connect | `connected` | `NULL` | `peluche-test-2219` | `["#sbiffo"]` |

The removed network left `$home` and the sidebar, the sidebar fell back to its
`no networks` copy, and the network reappeared under "available to connect".
**azzurra carries `visitor_enabled = 0`**, so that offer came purely from the
detached union — the admin-bound-network case, exercised rather than argued. One
tap restored it: same nick, `#sbiffo` rejoined.

Also exercised, by accident and worth keeping: a DELETE that 404'd put
`404 Not Found` in the row's own alert sink rather than failing silently.

## Review findings

One scoped review, six findings above trivial, all fixed in `5e60f28a`. Three
shared a root — `detach/2` trusted `connection_state` to tell it what work was
left:

1. **The session stop was gated on the state.** `unbind_credential/2` has always
   stopped unconditionally, with a comment saying why, and CLAUDE.md states the
   rule: the column and the live pid are separate sources of truth and they can
   disagree. A row reading `:parked` whose `Session.Server` is alive was being
   waved through, hiding that session behind an invisible binding.
2. **`:failed` was not parked**, making three separate "a detached row is
   `:parked`" claims false. Parking it is the right half to keep: a re-attach
   should not resume carrying a failure the subject chose to put away.
3. **Remove on a `:failed` network stranded the selection**, because
   `networkParked.ts` keeps failed networks in the sidebar on purpose. Fixing (2)
   carries this: the `:failed → :parked` move is now a real transition, so it
   broadcasts and the existing bucket-D redirect fires. No new client code.
4. **The attachment writers bypassed `Repo.BusyRetry`.** Here the naked version
   is worse than usual: by the time the mark runs the QUIT has gone and the
   session is down, so a raised SQLITE_BUSY left the network dead AND attached.
5. Two early exits in the client's remove returned in silence — the case the
   error sink exists to prevent, one layer above.
6. An incomplete `@spec` on the controller action.

Two more reds came from the suites rather than the review, and both were mine.
`BootCostTest` pins the query count of `GET /me` to an integer and it moved 7→8:
the detached-network offer was a second read of a partition already in hand, so
the cure is one read and an `Enum.split_with/2`, not a bigger number in the test.
`PreflightTest` pins which migrations deploy HOT; the classifier puts this one in
the set, so **no cold deploy is needed**.

## Gates

| gate | result |
|---|---|
| `scripts/test.sh` | 7836 tests, 0 failures (8 doctests, 67 properties) |
| `scripts/bun.sh run test` | 7527 pass |
| `scripts/bun.sh run check` | 5 stages, 0 failed |
| `scripts/credo.sh` | no issues |
| `scripts/dialyzer.sh` | 0 errors |
| `mix sobelow --config` | clean at Medium and above |
| `mix grappa.wire_pin --check` | shape and protocol 27 agree |
| `mix grappa.gen_wire_types --check` | all three artefacts in sync |
| `scripts/design-notes-gate.sh` | separator and marker present |
| `scripts/union-rebase.sh` | `docs/DESIGN_NOTES.md +183 −0 — contribution intact` |

Four cic spec files (compose, subscribe, module-root, constant-mirror) time out
at 5 s under the full parallel run on this Windows host and pass in isolation.
None touches a file this branch changes.

## Deliberately not here

**The hard credential delete**, which vjt's corrected ruling A also puts in
`$home`. Ruling C's remaining half is open and the answer changes what gets
built: `ArchiveModal` iterates attached networks, so a purged network's
scrollback becomes unreachable AND undeletable, while a per-`(subject, network)`
cascade would span `messages`, `read_cursors`, `query_windows`, `notify_entries`
and the `(network, peer)` mutes. Detach has no such problem — everything returns
on re-attach, archive included.

**A detach verb for visitors** (403). A visitor's identity LIVES on its
credentials: `representative_visitor_credential/1` is where the nick comes from
and `visitor_registered?/1` derives permanence from the per-network secret, so
detaching the last one hides the identity rather than a binding. The mechanism
fits; whose identity the row anchors does not.

**Any change to the rail ×.** It still parks, #1985 still hides a parked
network, and neither destructive verb is one distracted tap away from a
disconnect — the substance of the corrected ruling.


vjt:
Before this can be merged: please drop the `Closes #2219.` line from the first line of the body.

#2219 covers four outcomes and this PR implements one of them — the `detached_at` form, which is the sanctioned one (vjt, 2026-09-15: "sia detached che delete, il delete credential in home, il delete scrollback in archive"). But **B** and **D** are still open and **C** is narrowed but unresolved (what happens to the archive of a network whose credential is gone). Merging with the auto-close keyword takes those down with it.

Review is running; I will report the outcome here. Note for the merge: external author, so `--merge`, never `--squash`. The migration means a cold deploy, and the 27/09 deploy freeze still applies.

vjt:
Review is in (13 min, 530 lines). **No correctness blockers in the code** — the verdict on the shape of it was "strong work": park-before-mark, the unconditional stop, the refusal to reuse `spawn_or_rollback/4` (whose rollback is a DELETE, which would have destroyed precisely what a reversible verb exists to keep), the `get_credential/2` vs `get_attached_credential/2` split, and the `RouterScopeTest` prefix→exact-route fix that stops the new DELETE being asserted against nothing.

Two corrections to what I wrote above, both mine, both measured:

**1. The `Closes #2219.` is already gone — and you removed it yourself, correctly.** Measured just now: `closingIssuesReferences` is `0`, and a closing-keyword grep over the body returns `0` (positive control `Closes #2219.` → 1, negative control `addresses issue 2219` → 0; the title's `(issue 2219)` has no `#`, which is the safe form). My note was true when I wrote it, the review's "no keyword" was true minutes later — the target moved in between. Nothing further to do on this axis: it was textual and editing the body defused it. Do **not** go unlink anything in the Development panel.

**2. "The migration means a cold deploy" was wrong.** `Preflight.classify_migration_ast/1` decides from the AST, not from the presence of a migration: `up`/`change` bodies, no `@disable_ddl_transaction`, every body hot ⇒ `:hot`. Yours is a single `add :detached_at, :utc_datetime` — nullable, no default, no backfill — and the PR pins it in `@expected_hot` itself. I inferred a class from a structure without measuring the mechanism. The deploy freeze to 27/09 still applies, but not for this reason.

**The one thing I'd like fixed before ship — F2 (medium): the detach announces itself, the re-attach doesn't.**

`network_detached` calls both `refetchUser()` and `refetchNetworks()` (`userTopic.ts:1106` — the only `refetchUser()` in the file). `connection_state_changed` calls `refetchNetworks()` alone, and there is no `network_attached`-style event anywhere in `lib/` or `cicchetto/src/`. So a twin tab that did not initiate the re-attach never refreshes the `/me` envelope: `$home` keeps the network under *Available to connect* with no attached row. Third-party clients get no re-attach signal either.

The comment in `lib/home.ts` claiming the `connection_state_changed` handler "also `refetchNetworks()`/refetches /me" is false, and it's the sentence the "no broadcast needed on re-attach" reasoning rests on. The class is pre-existing, but this PR makes it far more reachable.

Minor (F3–F9), none blocking: a non-atomic race between detach and a concurrent connect (**not reproduced**, and declared as such), a dispatch on in-memory state, a false comment on `:failing`, coverage gaps, and a closed set left unextended.

**Merge stays on hold — not for F1.** The remaining blockers are F2 and the fact that #2219 still has rulings B and D open, with C narrowed. When it does go in: `--merge`, never `--squash`.

gmsrc:
F2 is fixed, on `8efd7c2e`. Thanks for both corrections — I verified the first one independently rather than taking it on trust: `closingIssuesReferences` is `0` and no closing keyword appears in the body. I have not touched the Development panel.

## The comment was the bug

You are right that the reasoning rested on a false sentence, and the sentence was in `lib/home.ts`: *"the `connection_state_changed` handler also `refetchNetworks()`/refetches /me"*. It does not — that arm calls `refetchNetworks()` alone, and the detach arm was the only `refetchUser()` in `userTopic.ts`. I reasoned from a comment instead of from the code. The comment is corrected in the same commit as the event, because the next person to reason about this reads it first.

## What shipped

`network_attached`, a real twin: same three fields, same no-state contract, and it fires for a **fresh accretion** as well as a revive. Announcing one direction is worse than announcing neither — a client that sees the detach learns to trust the feed, and the missing twin then reads as "no change" rather than "no signal". Splitting revive from first-attach would repeat that one axis over, and the half that fires less is the half that later stops firing.

Emitted at the accretion **door**, not in `reattach/1`: a revive that then fails to spawn is rolled back by re-detaching it, so announcing at the write would announce an attachment that does not survive the request. There is a test for exactly that arm (`a REFUSED spawn announces nothing`).

Both subject doors announce. Visitors have no detach to revive from, but `$home` is one component for both (#211 phase 6 ruling A), and a signal that exists for one subject and not the other is one a client cannot rely on.

## 27 is now skipped, and the gate is why

I tried to fold the new event into the still-unpublished v27 on exactly the argument you would expect — nothing had ever spoken 27, so no hole. `mix grappa.wire_pin --update` refused: the shape had moved and the number had not.

The refusal is right and the argument was wrong. The rule is total by design, and "unpublished" is the carve-out that makes a floor stop meaning anything, because the next reader inherits the carve-out and not the reasoning. So main goes **26 → 28**, nothing ever spoke 27, and the gap is written down in `Grappa.Protocol`, `CLIENT_PROTOCOL.md` and DESIGN_NOTES so it reads as a decision rather than a lost commit.

## Gates on this head

| gate | result |
|---|---|
| `scripts/test.sh` | 7839 tests, 1 timeout flake (below) |
| `scripts/bun.sh run test` | 7518 pass, 8 timeout flakes (below) |
| `scripts/bun.sh run check` | 5 stages, 0 failed |
| `scripts/credo.sh` | no issues |
| `scripts/dialyzer.sh` | 0 errors |
| `mix grappa.wire_pin --check` | shape and protocol 28 agree |
| `mix grappa.gen_wire_types --check` | all three artefacts in sync |
| `scripts/design-notes-gate.sh` | separator and marker present |

**The flakes are timeouts, not assertions, and they are this host.** `KeepaliveIdleOrderingTest` hit the 60 s ExUnit timeout inside `Path.wildcard/2`; the cic ones are `Test timed out in 5000ms` / `Hook timed out in 10000ms`. Re-run alone: 3/3 and 712/712 green. Across four full cic runs today the failing SET was different every time and never overlapped the files this branch touches. CI is the arbiter — it was green on the previous head and is running on this one.

## Still open, deliberately

Rulings **B** and **D** are answered in the PR body by Gabriele (yes and yes) but not by you, and **C** is narrowed rather than closed — which is why the hard credential delete is not in this PR. Its blocker is measurable: `ArchiveModal` iterates attached networks, so a purged network's scrollback becomes unreachable *and* undeletable, while a real per-`(subject, network)` cascade spans `messages`, `read_cursors`, `query_windows`, `notify_entries` and the `(network, peer)` mutes. Detach has none of that — everything returns on re-attach, archive included.

Noted for the merge: `--merge`, never `--squash`. And thank you for the correction on the cold deploy — the classifier reads the AST, the migration is a single nullable `add`, and the PR pins it in `@expected_hot` itself.

F3–F9: happy to take any of them if you would rather they did not sit, particularly the false `:failing` comment. Say the word and they go in this branch.

gmsrc:
CI is green on `8efd7c2e` — all nine checks, `mergeable: MERGEABLE`.

One spec went red on the first integration attempt and passed on the re-run, so recording it rather than leaving it in the history unexplained: `issue605-rail-width-cap`, the `webkit-iphone-15` leg, 1 failed against 251 passed in shard 4.

The spec routes `/networks` and rewrites every row's `connection.server` to a long IPv6, then asserts the rail card renders it. The card showed the real testnet value `172.18.0.4:6667` — nine locator resolutions over five seconds, so the card was hydrated from an unpatched source for the whole window rather than losing a race.

Why I did not treat it as mine, as evidence rather than hope:

* the same shard was green on the previous head `5e60f28a`, and the only new commit adds an event plus its client arm;
* the branch touches none of what that spec reads — no boot envelope, no `networks.ts`, no rail, no shell, no connection info (`git diff origin/main...HEAD --name-only` over those names returns nothing);
* both new events fire on detach and on attach, and that spec does neither, so the new path never executes in it;
* the received value was the REAL one, consistently — that points at a source the spec does not patch, not at my `refetchNetworks()` (which the route would have patched anyway, as the spec's own comment says).

Worth a look at some point, since it is not really a flake so much as an unpatched door: the card can apparently hydrate from a fetch that is not `GET /networks`, and `GET /boot` (#1679) is the obvious candidate — the spec routes only the one path. Not this PR's business, and I have not filed anything; say the word if you want it raised.

